PR #31, opened 2026-03-27
chore: remove offline query lint validators

javiermolinar:
## What this PR does

Very opinionated PR. Adding Prometheus and Loki just to validate their queries feel too much. It bloats the binary size and add a ton of deps surface to maintain. I would drop them for now keeping into considartion that not even Grafana enforce query validations when importing dashboards. A better approach would be having something like this:

https://github.com/grafana/lezer-traceql

- removes the offline PromQL and LogQL lint validators from the linter bundle
- drops the `target-valid-promql` and `target-valid-logql` dashboard rules and their docs
- removes the Loki and Prometheus parser dependencies pulled in only for those validators
- updates architecture/reference docs to reflect the simpler linter builtin surface

This aligns gcx with Grafana's behavior, where dashboard create/update does not enforce datasource query syntax at write time, and trims the binary/dependency footprint.

## Validation
- `GRAFANACTL_AGENT_MODE=false make all`

## Notes
- `gcx dev lint` no longer performs offline PromQL/LogQL syntax validation for dashboard targets
- datasource query execution behavior is unchanged


dafydd-t:
Diff after rebasing vs today's main is 17MB (119MB v 136MB)

I don't think this is hugely important either way, at the moment.

PR #399, opened 2026-04-07
fix(logs): prefer log body field over id/tsNs in legacy Loki frames

dhftah:
In the legacy Grafana Loki data frame format (field names Time/Line/tsNs/id), convertLegacyFormat was assigning valueIdx to the last string-typed field, which is "id" (e.g. 1775564496989203231_94cf3bc) rather than "Line" (the actual log content). Loki queries therefore printed opaque identifiers instead of log lines.

Fix by preferring a field named "Line" or "body" when selecting the body field, only falling back to the first-seen string field if neither is present.

### DIFF
--- a/internal/query/loki/client.go
+++ b/internal/query/loki/client.go
@@ -421,7 +421,11 @@ func convertLegacyFormat(frame DataFrame, result *QueryResponse) {
 		case "time":
 			timeIdx = i
 		case "string", "number":
-			valueIdx = i
+			// Prefer named body fields ("Line", "body") over generic string fields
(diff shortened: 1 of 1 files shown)


radiohead:
Hey, thank you for your contribution and apologies for the late reply!

This has been superseded by #426, so I'm going to close this PR – I am sorry we haven't accepted your earlier fix, but we're currently iterating pretty quickly and we don't have time for reviewing 3rd party contributions.

I sincerely hope this does not discourage you from further contributions in the future!

PR #403, opened 2026-04-07
feat: add `gcx skills` subcommand and an initial set of skills

annanay25:
Initial groundwork for the skills subcomment, also porting some skills over.


radiohead:
Superseded by #450

Issue #430, opened 2026-04-09
support oncall commands through oauth


dafydd-t:
Currently, gcx oncall resources error when using oauth for auth, because the oauth proxying logic doesn't handle sending requests to other targets.

We need to support sending requests to backends external to grafana, aiming particularly at oncall.

- We should see no oncall related errors when we call `gcx resources list` (currently we get a load of 403 errors)
- oncall related commands should work: `gcx oncall schedules list` (currently returns 403)

PR #431, opened 2026-04-09
feat(oncall): route OnCall requests through OAuth proxy

dafydd-t:
## Summary

When using oauth, the OnCall client can't call the OnCall API directly (there is no SA token client-side). This routes requests through the assistant's new external provider proxy instead.

- Adds `IsOAuthProxy()` to `NamespacedRESTConfig` to detect OAuth proxy mode
- OnCall client uses the HTTP client (with `RefreshTransport` for gat_ token auth) in OAuth mode, instead of the standalone `ExternalHTTPClient()`, so it uses the oauth mechanisms and proxies requests to the oauth proxy.
- Handles pagination URL host mismatch: OnCall API returns pagination URLs pointing to the real OnCall host, which differs from the proxy base URL

## How it works

```
Direct mode:
  gcx OnCall client → oncall-prod.../oncall/api/v1/... (Authorization: {saToken})

OAuth mode (new):
  gcx OnCall client → {proxyEndpoint}/api/cli/v1/proxy/external/oncall/api/v1/...
                       (Authorization: Bearer gat_xxx via RefreshTransport)
       ↓
  Assistant proxy → discovers OnCall URL from Grafana plugin settings
       ↓
  Assistant proxy → oncall-prod.../oncall/api/v1/... (Authorization: {saToken})
```

## Context

Part of https://github.com/grafana/gcx/issues/430

Fixes 15 OnCall "Invalid token" failures when using `gcx resources get` with OAuth auth. The assistant-side proxy is in https://github.com/grafana/grafana-assistant-app/pull/5812.

## Test plan

- [x] `IsOAuthProxy()` returns true/false for OAuth vs token auth configs
- [x] Proxy mode client doesn't set its own Authorization header
- [x] Pagination works when next URL points to a different host than the proxy
- [x] All existing OnCall client tests continue to pass (direct mode unchanged)

For full e2e tests, we can wait until the backend is in dev & verify

### DIFF
--- a/internal/config/rest.go
+++ b/internal/config/rest.go
@@ -25,6 +25,11 @@ type NamespacedRESTConfig struct {
 	oauthTransport *auth.RefreshTransport
 }
 
+// IsOAuthProxy reports whether the config is using OAuth proxy mode.
+func (n *NamespacedRESTConfig) IsOAuthProxy() bool {
+	return n.oauthTransport != nil
+}
+
 // SetOnRefresh registers a callback that is invoked after a successful OAuth
 // token refresh. This allows the call site (which has access to the config
 // source) to persist refreshed tokens back to the config file.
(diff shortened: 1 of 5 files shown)


javiermolinar (approved):
LGTM

radiohead (approved):

PR #461, opened 2026-04-13
fix(slo): pass parent config loader to definitions commands

radiohead:
SLO definitions commands (list, get, push, pull, delete, status, timeline) created their own `ConfigLoader` internally, ignoring the `--context` flag bound on the parent `slo` command. This meant `gcx slo definitions list --context <name>` always used the default context.

**Fix**: Mirror the pattern already used by the `reports` subpackage — accept a `GrafanaConfigLoader` interface from the parent and thread it through to `NewTypedCRUD`. The lazy factory (used by resource adapters) still creates its own loader from context, so that code path is unaffected.

## Changes

- `internal/providers/slo/definitions/commands.go` — accept `GrafanaConfigLoader` parameter instead of constructing one internally
- `internal/providers/slo/definitions/resource_adapter.go` — update adapter construction to pass the loader
- `internal/providers/slo/definitions/status.go` — same threading for status command
- `internal/providers/slo/definitions/timeline.go` — same threading for timeline command
- `internal/providers/slo/provider.go` — pass parent loader down when registering definitions commands

### DIFF
--- a/internal/providers/slo/definitions/commands.go
+++ b/internal/providers/slo/definitions/commands.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/grafana/gcx/internal/config"
 	"github.com/grafana/gcx/internal/format"
 	cmdio "github.com/grafana/gcx/internal/output"
 	"github.com/grafana/gcx/internal/resources"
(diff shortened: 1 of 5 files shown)


javiermolinar (approved):
LGTM

PR #462, opened 2026-04-13
feat(irm): consolidate OnCall + Incidents under unified irm provider

### LINKED ISSUE #430: support oncall commands through oauth

### LINKED ISSUE #460: Unify oncall + incidents providers into single IRM provider using plugin proxy API
## Problem

The `oncall` and `incidents` providers reflect an old product split that no longer exists — OnCall and Incidents have been combined into **IRM** (Incident Response Management), shipping as one team, one repo ([grafana/irm](https://github.com/grafana/irm)), and one Grafana plugin (`grafana-irm-app`).

### Current architecture

```
gcx oncall     →  OnCall public API v1 (/api/v1/...)     →  SA token + external URL discovery
gcx incidents  →  IRM plugin proxy (/api/plugins/grafana-irm-app/resources/api)  →  Grafana auth
```

The `oncall` provider uses the **legacy public API v1** with its own auth model (service account tokens, `X-Grafana-Url` header, OnCall URL discovery via plugin settings). This causes:

1. **Auth failures with OAuth** — the public API v1 doesn't accept OAuth tokens, breaking 15 OnCall resource types (#430, #433)
2. **Broken endpoints** — `alerts get` calls `/api/v1/alerts/{id}/` which doesn't exist; the working endpoint is the plugin proxy
3. **Skim data** — the public API returns minimal fields (e.g., alerts: just `id`, `alert_group_id`, `created_at`), while the plugin proxy returns rich data (labels, annotations, status, firing instances) — the same data the Grafana UI shows
4. **User-token auth gaps** — personal-notification-rules requires user-token auth that the public API doesn't support (#261)
5. **Complex config** — OnCall URL discovery, SA token handling, `X-Grafana-Url` headers — none of which other providers need

Meanwhile, the `incidents` provider **already** goes through the plugin proxy correctly.

### Target architecture

```
gcx irm  →  IRM plugin proxy (/api/plugins/grafana-irm-app/resources/...)  →  Grafana auth
```

One provider, one auth model (same as SLO, incidents today), richer data, simpler config.

### IRM plugin API surface (from [grafana/irm](https://github.com/grafana/irm) source)

The IRM plugin proxy exposes:

- **OnCall domain (REST):** 135+ endpoints — alert groups, alerts, integrations, escalation chains/policies, schedules, shifts, routes, webhooks, users, teams, notification policies, slack/msteams/telegram channels, labels, resolution notes, shift swaps, heartbeats, direct paging
- **Incidents domain (RPC):** 22 service classes — IncidentsService, ActivityService, SeveritiesService, RolesService, TasksService, etc.
- **Plugin IDs:** `grafana-irm-app` (unified), `grafana-oncall-app` (legacy fallback), `grafana-incident-app` (standalone)

The UI dynamically selects `grafana-irm-app` when present, falling back to `grafana-oncall-app`.

### Issues this closes or advances

| Issue | How |
|-------|-----|
| #430 (OnCall via OAuth) | Plugin proxy uses Grafana auth — no SA token needed |
| #433 (OAuth: 15 OnCall resources 403) | Same root cause — resolved by plugin proxy |
| #261 (personal-notification-rules user-token auth) | Plugin proxy handles auth server-side |
| #207 (list filters) | Easier to add — plugin proxy already accepts these query parameters |

### In scope

- New `internal/providers/irm/` provider with sub-packages for oncall and incidents domains
- Client targeting plugin proxy (`/api/plugins/grafana-irm-app/resources/...`)
- Migrate all 16 oncall resource adapters + incidents adapter
- Migrate all CLI commands under `gcx irm` namespace
- Auth via Grafana REST config (same as SLO provider) — drop OnCall URL discovery + SA token handling
- Plugin ID detection: prefer `grafana-irm-app`, fall back to `grafana-oncall-app` (matching UI behavior)

### Out of scope (follow-up)

- Expanding to new IRM endpoints not currently supported
- Deprecation/removal of old `oncall`/`incidents` CLI commands (keep as aliases initially)

## CLI UX sketch

```
gcx irm alert-groups list [--status firing] [--integration-id X]
gcx irm alert-groups get <id>
gcx irm alert-groups get-alerts <id>        # rich alert data via plugin proxy
gcx irm schedules list
gcx irm escalation-chains list
gcx irm incidents list
gcx irm incidents get <id>
gcx irm incidents create --title "..."
# ... all current oncall + incidents commands under one namespace
```

`gcx oncall` and `gcx incidents` remain as aliases during transition.

## Acceptance Criteria

- [ ] `gcx irm` provider registered, discoverable via `gcx providers list`
- [ ] All current `oncall` resource adapters migrated (16 types)
- [ ] All current `incidents` resource adapters migrated (1 type)
- [ ] All requests route through plugin proxy — no direct OnCall API v1 calls
- [ ] OAuth auth works for all IRM resources (verifies #430, #433 fix)
- [ ] `gcx irm alert-groups get-alerts <id>` returns rich data (labels, annotations, status)
- [ ] `gcx oncall` and `gcx incidents` work as aliases
- [ ] `gcx resources list` shows IRM resources without 403 errors under OAuth
- [ ] Smoke test on dev and ops contexts with both token and OAuth auth

dafydd-t:
closes https://github.com/grafana/gcx/issues/460
closes https://github.com/grafana/gcx/issues/430

## Summary

Replaces the standalone `oncall` and `incidents` providers with a single `irm` provider that routes all OnCall requests through the IRM plugin's `/resources` proxy (`/api/plugins/grafana-irm-app/resources/`). This works for both token-based and OAuth auth modes. OnCall and incidents are both using the plugin proxy.

**Before:** `gcx oncall schedules list`, `gcx incidents list`
**After:** `gcx irm oncall schedules list`, `gcx irm incidents list`

## Why

The old oncall provider called the OnCall backend directly with a raw SA token, which fails in OAuth mode.. PR #431 and PR #5812 attempted to solve this with a custom external provider proxy in the assistant. This PR takes a different approach: route through the IRM plugin's existing proxy - the same pattern incidents already uses - eliminating the need for any assistant-side changes.

## Request flow

```
Token auth:   gcx → Grafana → IRM plugin → OnCall backend
OAuth auth:   gcx → assistant grafana proxy → Grafana → IRM plugin → OnCall backend
```

## Key changes

- New `internal/providers/irm/` package consolidating OnCall + Incidents under a single provider
- OnCall client (`oncall_client.go`) talks to the internal API (`api/internal/v1/`) via the IRM plugin proxy, with generic helpers (`iterResources`, `getResource`, `createResource`, etc.) for all 16 resource types
- Dual-client architecture: `OnCallAPI` interface with two implementations — plugin proxy client for OAuth, public API client (`oncallpublic/`) for SA token auth (temporary, until IRM plugin supports SA tokens)
- Shared types extracted into `oncalltypes/` sub-package so both clients reference identical types without import cycles
- Incidents commands reused as-is (exported constructors, attached under `gcx irm incidents`)
- Resource adapter registrations for `gcx resources get/push/pull` on all IRM resources
- Old `internal/providers/oncall/` and `internal/providers/incidents/provider.go` deleted (~5000 lines net deletion)
- Agent command annotations updated to `gcx irm oncall/incidents` paths

## How to review this PR

The PR is structured as 6 commits, each compilable and testable independently. Review commit-by-commit for easiest reading.

**Commit 1:** `feat(irm): add provider scaffolding, OnCall client, and internal API types`
Foundation commit. Adds the new `irm` provider with:
- `provider.go` — `IRMProvider` struct with `init()` registration (empty Commands/TypedRegistrations)
- `config.go` — `configLoader` wrapping `providers.ConfigLoader`, adds `LoadOnCallClient()` (no URL discovery needed — plugin proxy handles routing)
- `oncall_constants.go` — API group/version constants + shared `defaultStripFields`
- `oncall_types.go` — Go types matching the OnCall internal API response shapes (field names differ from the public API, e.g. `verbal_name` not `name`, `pk` not `id` for AlertGroup/User)
- `oncall_client.go` — HTTP client using `rest.HTTPClientFor()`, `doRequest`, pagination (extracts relative path from backend `next` URLs), and all CRUD methods for 16 resource types
- `oncall_client_test.go` — 9 tests: URL construction, auth delegation, page+cursor pagination, alert group actions, current user, direct paging input format
- Blank import added to `cmd/gcx/root/command.go` (old imports kept)

**Commit 2:** `feat(irm): add OnCall commands and table codecs`
Wires all oncall commands under `gcx irm oncall`:
- `oncall_commands.go` — 16 resource group commands (list/get) with `newListSubcommand`/`newGetSubcommand` pattern + all table codecs adapted for internal API field names
- `oncall_commands_extra.go` — alert-group actions (acknowledge/resolve/silence/etc.), `escalate` (using internal `direct_paging` endpoint with per-user importance), `users current`, `schedules final-shifts` (using `filter_events?type=final`)
- `provider.go` updated to wire oncall subcommands

**Commit 3:** `feat(irm): add incidents subcommand via existing incidents package`
Minimal change — reuses existing incidents code:
- Exports 7 command constructors in `incidents/commands.go` (`newListCommand` → `NewListCommand`, etc.)
- `incidents_commands.go` — thin wrapper attaching exported constructors under `gcx irm incidents`
- `incidents/provider.go` updated to use exported names (still functional as standalone provider at this point)

** Commit 4:** `feat(irm): add resource adapter registrations for oncall and incidents`
Enables `gcx resources get/push/pull` for all IRM resources:
- `oncall_adapter.go` — 16 `TypedCRUD` registrations using same GVKs as old oncall provider (`oncall.ext.grafana.app/v1alpha1`)
- `incidents_adapter.go` — delegates to existing `incidents.NewAdapterFactory`
- Exports `StaticDescriptor`, `IncidentSchema`, `IncidentExample` from incidents package
- `provider.go` updated to return registrations from `TypedRegistrations()`

**Commit 5:** `feat(irm): remove old oncall and incidents providers`
Cleanup — net deletion of ~5000 lines:
- Deletes `internal/providers/oncall/` entirely
- Deletes `internal/providers/incidents/provider.go` (keeps package as library)
- Removes incidents `init()` natural key registration (now in irm adapter)
- Updates `cmd/gcx/root/command.go` imports
- Updates `internal/agent/command_annotations.go` paths

**commit 6**: move the incidents provider internals into the irm provider dir

mostly moving files, fixing a couple of naming collisions

**commits 7/8:** Add oncallpublic client, which is a client for the oncall provider which uses the public endpoint. We need this temporarily while we sort out support for SA tokens in the IRM plugin proxy. Common types and interfaces for the two oncall clients are extracted into oncalltypes temporarily, so we could keep all the oncallpublic files in a subdir without having an import cycle.

later commits: linting

## Test plan

- [x] `go test ./internal/providers/irm/...` — 9 client tests pass
- [x] `go test ./internal/providers/incidents/...` — existing tests pass
- [x] `go test ./cmd/gcx/root/ -run TestConsistency` — annotation consistency passes
- [x] `go test ./...` — full suite passes
- [x] Manual e2e with token auth: `gcx irm oncall integrations list`
- [x] Manual e2e with OAuth auth: `gcx irm oncall integrations list`
- [x] Verify `gcx oncall` returns "unknown command"

## Context

Part of https://github.com/grafana/gcx/issues/430. See [analysis comment on PR #5812](https://github.com/grafana/grafana-assistant-app/pull/5812#issuecomment-4216949080) for why the `/resources` proxy approach was chosen over the assistant-side external provider proxy.


internal/providers/irm/oncall_commands.go
dafydd-t, line 1: think of this as recreating https://github.com/grafana/gcx/blob/main/internal/providers/oncall/commands.go

internal/providers/incidents/provider.go
```diff
@@ -36,13 +36,13 @@ func (p *IncidentsProvider) Commands() []*cobra.Command {
 	loader.BindFlags(incCmd.PersistentFlags())
 
 	incCmd.AddCommand(
```
dafydd-t: this file will get deleted later

dafydd-t: this will get deleted later

dafydd-t:
@claude review

wardbekker:
## Code review findings

Ran a code simplifier pass over the new `irm` provider code. Findings below, ordered by impact. The dual oncall client architecture (`oncall_client.go` + `oncallpublic/`) is understood as intentional and temporary — not flagged.

---

### 1. Bug: `ErrNotFound` not wrapping `adapter.ErrNotFound` — breaks push upsert

**`incidents_client.go:17`**

```go
var ErrNotFound = errors.New("incident not found")
```

This is a standalone sentinel. The adapter layer at `adapter/typed.go:356` checks `errors.Is(err, adapter.ErrNotFound)` to detect not-found and fall through to create during push upsert. Since this `ErrNotFound` doesn't wrap `adapter.ErrNotFound`, the adapter will never recognize it, breaking the upsert path.

Other providers do this correctly, e.g.: `ErrRuleNotFound = fmt.Errorf("rule: %w", adapter.ErrNotFound)`.

### 2. `ListFilterEvents` in oncallpublic ignores the `days` parameter

**`oncallpublic/client.go:468`**

```go
params.Set("end_date", startingDate) // will be computed properly by the command
```

`end_date` is set to the same value as `start_date`, ignoring the `days` parameter entirely. The command calculates `days` from start/end and passes it here, so the public API path always fetches shifts for a single day regardless of the requested range.

### 3. `CreateDirectPaging` in oncallpublic loses per-user importance granularity

**`oncallpublic/client.go:893-895`**

Only `input.Users[0].Important` is used to set a single `"important"` flag for the entire escalation. If multiple users are passed with different `Important` values, only the first user's importance is used, silently ignoring the rest. The internal API client correctly passes per-user `UserReference` objects with individual importance flags.

### 4. Organizations `get <id>` requires an ID argument but silently discards it

**`oncall_commands.go:420-423`**

```go
newGetSubcommand(loader, "Get organization info.",
    func(ctx context.Context, c OnCallAPI, _ string) (*Organization, error) {
        return c.GetOrganization(ctx)
    }),
```

This creates a `get <id>` subcommand (via `cobra.ExactArgs(1)`) that ignores the ID. A user running `gcx irm oncall organizations get ANYTHING` gets back the singleton organization regardless. This should either not use `newGetSubcommand` or take zero args.

### 5. Duplicate marshal-to-unstructured logic

**`oncall_commands.go:113-138`** vs **`oncall_commands_extra.go:474-505`**

`newListSubcommand` has inline marshal → map → delete-ID → wrap-as-unstructured logic that is identical to the `itemsToUnstructured` helper. The inline version in `newListSubcommand` could just call `itemsToUnstructured`.

### 6. Near-identical `TypedCRUD` construction duplicated

**`incidents_resource_adapter.go:131-163`** vs **`167-209`**

`newTypedAdapter()` and `NewTypedCRUD()` construct almost identical `TypedCRUD[Incident]` structs. The `GetFn`, `CreateFn`, `UpdateFn`, `DeleteFn`, `StripFields`, and `Descriptor` are copy-pasted. A shared builder accepting `client`, `namespace`, and `query` parameters would eliminate this duplication.

### 7. 16 identical `Decode` methods across oncall table codecs

**`oncall_commands.go`** — lines 573, 595, 637, 675, 714, 759, 800, 844, 882, 904, 927, 949, 975, 997, 1041, 1089

Every table codec has the same 3-line `Decode` returning `errors.New("table format does not support decoding")`. Consider embedding a shared `noDecodeCodec` struct.

### 8. `discoverOnCallURL` reloads config from disk

**`config.go:62`**

`LoadOnCallClient` calls `l.LoadGrafanaConfig(ctx)` at line 25. If it falls through to the SA-token path, `discoverOnCallURL` calls `l.LoadGrafanaConfig(ctx)` again at line 62. The `restCfg` from the first call could be passed as a parameter.

### 9. Inconsistent error prefixes between the two oncall clients

The internal client uses `"irm: list %s"` / `"irm: get %s"` while the public client uses `"oncall: list %s"` / `"oncall: get %s"`. Since both implement `OnCallAPI` and callers can't distinguish which is in use, inconsistent prefixes make debugging harder.

### 10. String-based `idField` parameter is fragile

**`oncall_commands.go`** — `newListSubcommand` takes an `idField string` parameter (`"id"` or `"pk"`) that must be manually kept in sync with the actual struct field name. Since the `ResourceNamer` constraint is already enforced (providing `GetResourceName()`), using that method directly would eliminate a class of copy-paste bugs.

radiohead (approved):
This comment has been generated by Claude Code

### New findings — recommended follow-ups

1. **High — Unvalidated `oncall-url` / discovered URL used with SA token** (`config.go:55-68`, `oncallpublic/client.go:58-68`)

Both the user-configurable `oncall-url` key and the auto-discovered URL from plugin settings are used as the target for SA-token-bearing requests without scheme/host validation. A misconfigured value could redirect the token to an unintended host. Recommend requiring `https://` scheme at minimum before use.

2. **Medium — Config key namespace mismatch** (`config.go:55`, `provider.go:64-67`)

Provider name is `"irm"`, `ConfigKeys()` declares `oncall-url` — suggesting `GRAFANA_PROVIDER_IRM_ONCALL_URL`. But `discoverOnCallURL` calls `LoadProviderConfig(ctx, "oncall")`, so the actual working env var is `GRAFANA_PROVIDER_ONCALL_ONCALL_URL`. The deleted `provider_test.go::TestDiscoverOnCallURL_EnvVar` would have caught this. Either align the namespace or document the legacy "oncall" namespace explicitly.

3. **Medium — Dropped OnCall adapter and provider tests** (`internal/providers/oncall/resource_adapter_test.go`, `provider_test.go`)

`resource_adapter_test.go` (round-trip tests for Integration, Schedule typed resources) and `provider_test.go` (env var discovery tests) were deleted without replacements. The incident adapter has equivalent round-trip tests in `incidents_resource_adapter_test.go`. Recommend adding OnCall equivalents.

4. **Low — Missing compile-time interface check for `OnCallClient`** (`oncall_client.go`)

`oncallpublic.Client` has `var _ oncalltypes.OnCallAPI = (*Client)(nil)` (line 926). `OnCallClient` — the 50+ method internal client — lacks one. One-liner fix: `var _ OnCallAPI = (*OnCallClient)(nil)`.

PR #466, opened 2026-04-13
fix(auth): persist rotated OAuth tokens across command paths

rif7t:
Fixes the issue where gcx could require `gcx auth login` again after the access token expired because some command paths refreshed OAuth credentials in memory without persisting the rotated refresh token back to config.

What changed:
- Added a centralized config helper to build REST configs and always wire OAuth token persistence.
- Updated config command paths to use the centralized helper.
- Updated provider config loader paths to use the centralized helper.
- Added regression tests covering token persistence for the new helper.

Why:
- OAuth refresh token rotation must be persisted after a successful refresh.
- Without persistence, a command can succeed once, leave the old refresh token on disk, and cause later commands to fail with refresh/auth errors.

Manual verification:
- Forced token expiry locally.
- Ran an authenticated resource command (`resources get folders`).
- Confirmed `oauth-token`, `oauth-refresh-token`, and `oauth-token-expires-at` were updated in the config file.


javiermolinar (commented):
Hi, I don't see an evidence in this PR why this change is needed. I see it centralizes the logic and it uses in the check command.

Could you provide a failing reproduction on current main and ideally a test that fails on main and passes with the fix?

rif7t:
> Hi, I don't see an evidence in this PR why this change is needed. I see it centralizes the logic and it uses in the check command.
> 
> Could you provide a failing reproduction on current main and ideally a test that fails on main and passes with the fix?

Hello, 
here are repro steps on current main branch

Prerequisites:

- stack-id must be set and match your real stack (validation fails otherwise and no HTTP call is made)
- Discovery cache must be cleared, cached responses skip HTTP calls entirely, so the refresh transport never fires
- Each test needs a fresh auth login, the server invalidates refresh tokens on use

```
git checkout main && go mod vendor && go build -buildvcs=false -o bin/gcx ./cmd/gcx/
./bin/gcx auth login
./bin/gcx config set contexts.default.grafana.stack-id YOUR_STACK_ID
rm -rf ~/.cache/gcx/discovery/
sed -i 's/oauth-token-expires-at:.*/oauth-token-expires-at: "2020-01-01T00:00:00Z"/' ~/.config/gcx/config.yaml
before=$(grep 'oauth-token:' ~/.config/gcx/config.yaml | head -1)
./bin/gcx config check
after=$(grep 'oauth-token:' ~/.config/gcx/config.yaml | head -1)
echo "BEFORE: $before"
echo "AFTER:  $after"
```
# bug: refresh token consumed but new one never saved

on fix/363-oauth-persistence branch

```
git checkout fix/363-oauth-persistence && go mod vendor && go build -buildvcs=false -o bin/gcx ./cmd/gcx/
./bin/gcx auth login
./bin/gcx config set contexts.default.grafana.stack-id YOUR_STACK_ID
rm -rf ~/.cache/gcx/discovery/
sed -i 's/oauth-token-expires-at:.*/oauth-token-expires-at: "2020-01-01T00:00:00Z"/' ~/.config/gcx/config.yaml
before=$(grep 'oauth-token:' ~/.config/gcx/config.yaml | head -1)
./bin/gcx config check
after=$(grep 'oauth-token:' ~/.config/gcx/config.yaml | head -1)
echo "BEFORE: $before"
echo "AFTER:  $after"
```
# fix: rotated tokens saved to disk



I also added a regression test `TestRegressionTokenPersistence_WithAndWithoutWiring` with two sub-tests:

- `without_persistence`  reproduces the bug: tokens refresh in-memory but are never saved to disk
- `with_RESTConfigForContext ` proves the fix: rotated tokens are saved to disk

wardbekker:
thanks for the contribution. As discussed, going to include the checkContext fix into https://github.com/grafana/gcx/pull/522

PR #467, opened 2026-04-13
Use "k6" instead of "K6" in docstrings/UI

federicotdn:
A PR to change many occurrences of "K6" to "k6", in order to properly follow the branding for k6.
I've updated `naming.md` to hopefully make this more permanent.


javiermolinar (approved):
LGTM

PR #473, opened 2026-04-13
feat(skills): improve list output, add installed status, and support single-skill install

annanay25:
Builds on top of the existing skills commands added by @javiermolinar 

- Replace raw tabwriter with style.NewTable for proper terminal-width rendering so descriptions no longer get cut off
- Remove redundant summary table from list output; show skills directly
- Add INSTALLED column to skills list showing install status per skill
- Change install command to accept individual skill names as arguments instead of always installing the entire bundle (use --all for that)
- Validate requested skill names against the bundle, with tab completion
- Add missing agent.token_cost annotations for list and uninstall commands
- Fix test flakiness from agent-mode auto-detection in Cursor via agent.ResetForTesting()

Made-with: Cursor

### DIFF
--- a/cmd/gcx/skills/command.go
+++ b/cmd/gcx/skills/command.go
@@ -81,11 +81,11 @@ func newInstallCommand(source fs.FS) *cobra.Command {
 		Use:   "install [SKILL]...",
 		Short: "Install bundled gcx skills into ~/.agents/skills",
 		Long:  "Install one or more bundled gcx Agent Skills into a user-level .agents directory for tools that follow the .agents skill convention. Use --all to install the entire bundle.",
-		Example: `  gcx skills install gcx
-  gcx skills install gcx setup-gcx debug-with-grafana
+		Example: `  gcx skills install setup-gcx
+  gcx skills install setup-gcx debug-with-grafana explore-datasources
   gcx skills install --all
   gcx skills install --all --dry-run
-  gcx skills install gcx --force`,
+  gcx skills install setup-gcx --force`,
 		Args: cobra.ArbitraryArgs,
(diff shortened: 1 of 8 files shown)


cmd/gcx/skills/command.go
```diff
@@ -285,6 +623,229 @@ func fileMode(source fs.FS, path string) fs.FileMode {
...
+	IO     cmdio.Options
+}
+
+func (o *uninstallOpts) setup(flags *pflag.FlagSet) {
+	o.IO.DefaultFormat("text")
+	o.IO.RegisterCustomCodec("text", &uninstallTextCodec{})
+	o.IO.BindFlags(flags)
+
+	flags.StringVar(&o.Dir, "dir", "~/.agents", "Root directory for the .agents installation")
+	flags.BoolVar(&o.All, "all", false, "Uninstall all skills in the target directory")
```
javiermolinar: This is a no go. We cannot uninstall all the skills in .agents since it will remove non-gcx skills too.

cmd/gcx/skills/command.go
```diff
@@ -160,7 +204,274 @@ func (c *installTextCodec) Decode(_ goio.Reader, _ any) error {
...
+
+	return result, nil
+}
+
+func isSkillInstalled(skillsDir string, name string) bool {
+	info, err := os.Stat(filepath.Join(skillsDir, name, "SKILL.md"))
+	return err == nil && !info.IsDir()
+}
+
+func listInstalledSkills(root string) (listResult, error) {
```
javiermolinar: This list all the possible skill, regardless if they belong to gcx or not. Thats way the uninstall --all is so destructive

cmd/gcx/skills/command.go
annanay25: oops, addressed. thanks for the review!

wardbekker (approved):

PR #475, opened 2026-04-13
fix: `gcx dev serve` does not exit on ctrl+c

### LINKED ISSUE #474: gcx dev serve does not exit on Ctrl+C
## Problem

Running \`gcx dev serve\` and pressing Ctrl+C does not terminate the process. The server continues listening on its port and the process must be killed manually (\`kill -9\`) or by closing the terminal.

This happens because \`Server.Start\` uses the package-level \`http.ListenAndServe\`, which has no context awareness and blocks indefinitely regardless of cancellation signals.

## Steps to Reproduce

1. Run `gcx dev serve .`
2. Press Ctrl+C
3. The process does not exit

To confirm the server is still running after Ctrl+C:

```
lsof -i :8080
```

## Root Cause

Two issues in the serve path:

1. `internal/server/server.go`: `http.ListenAndServe` is a package-level function with no mechanism to receive context cancellation. When SIGINT fires and the context is cancelled, the server keeps blocking.
2. `internal/server/watch/watcher.go`: The `Watch` goroutine has no `ctx.Done()` case, so the `fsnotify` watcher and its file descriptors are never released.

## Expected Behaviour

Pressing Ctrl+C exits the process immediately and cleanly (exit code 5, consistent with the rest of the CLI).

## Acceptance Criteria

- [x] Ctrl+C causes \`gcx dev serve\` to exit immediately
- [x] The port is released immediately after exit
- [ ] The file watcher goroutine is stopped on cancellation

stephan-rayner:
Fixes #474

## Problem

Pressing Ctrl+C while running \`gcx dev serve\` does not terminate the process. The server keeps listening on its port and must be killed manually.

## Root Cause

Two issues in the serve path:

1. \`internal/server/server.go\`: The package-level \`http.ListenAndServe\` has no context awareness and blocks indefinitely, ignoring cancellation signals.
2. \`internal/server/watch/watcher.go\`: The \`Watch\` goroutine has no \`ctx.Done()\` case, so the \`fsnotify\` watcher and its file descriptors are never released.

## Changes

- **\`internal/server/server.go\`**: Replace \`http.ListenAndServe\` with an \`http.Server\` that shuts down gracefully when the context is cancelled. A goroutine watches \`ctx.Done()\` and calls \`Shutdown\` with \`context.WithoutCancel(ctx)\` to preserve context values while dropping the cancellation signal. \`http.ErrServerClosed\` is treated as a clean exit. Sets \`ReadHeaderTimeout\` to satisfy gosec G112. Removes the now-unnecessary \`//nolint:gosec\` suppression (gosec G114 targets the package-level function, not \`http.Server\`).
- **\`internal/server/watch/watcher.go\`**: Add a \`context.Context\` parameter to \`Watch\`. The select loop gains a \`ctx.Done()\` case and \`defer w.notifier.Close()\` to release file descriptors on exit.
- **\`cmd/gcx/resources/serve.go\`**: Update the \`Watch\` call to pass \`cmd.Context()\`. Remove the redundant \`go\` prefix as \`Watch\` already spawns an internal goroutine.

## Testing

```bash
# Build
make build

# Run Server
./bin/gcx dev serve .

# In a second terminal, watch for the process while you Ctrl+C in the first
watch -n 1 "pgrep -l gcx"

Press Ctrl+C -- process exits immediately
# Hit Ctrl+C — the process should exit immediately with no zombie process

# Confirm port is released
lsof -i :8080
```

### DIFF
--- a/cmd/gcx/resources/serve.go
+++ b/cmd/gcx/resources/serve.go
@@ -175,7 +175,7 @@ support for file notifications.
 				}
 
 				// Start listening for events.
-				go watcher.Watch()
+				watcher.Watch(cmd.Context())
 			}
 
 			serverCfg := server.Config{
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -12,6 +12,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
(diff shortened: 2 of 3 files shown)


javiermolinar (approved):
LGTM

PR #476, opened 2026-04-13
Fix Watcher Error Channel Handling

### LINKED ISSUE #477: Fix Watcher Error Channel Handling in Watch Loop
## Description

The `Errors` channel case in `internal/server/watch/watcher.go` has two bugs:

- When the channel closes (`ok == false`), a nil error is logged and the goroutine falls through, spinning indefinitely on the closed channel
- Real fsnotify errors (`ok == true`) are silently dropped

## Expected Behavior

- Channel close should exit the goroutine (matching the `Events` case)
- Real fsnotify errors should be logged as warnings

## Steps to Reproduce

1. Run `gcx dev serve`
2. Close the watcher (e.g. on shutdown)
3. Goroutine spins; fsnotify errors are never surfaced

## Fix

Correct the condition and log placement in the `Errors` case to mirror the `Events` case. See PR #476.

stephan-rayner:
Closes: #477 

## Summary

- When the `Errors` channel closed (`ok == false`), the code logged a nil error instead of returning, and the goroutine continued spinning on the closed channel indefinitely
- Real fsnotify errors (`ok == true`) were silently dropped
- Mirrors the existing `Events` channel case: return on close, log on real errors

## Test Plan

- [x] Run `go test ./internal/server/...`
- [x] Manually verify the dev server (`gcx dev serve`) starts, detects file changes, and shuts down cleanly

### DIFF
--- a/internal/server/watch/watcher.go
+++ b/internal/server/watch/watcher.go
@@ -83,8 +83,9 @@ func (w *Watcher) Watch() {
 				}
 			case err, ok := <-w.notifier.Errors:
 				if !ok {
-					w.logger.Warn("watch error", logs.Err(err))
+					return
 				}
+				w.logger.Warn("watch error", logs.Err(err))
 			}
 		}
 	}()


javiermolinar (approved):
Lgtm

PR #490, opened 2026-04-14
Add ability to filter incidents by Tags and from / to

nachogiljaldo:
Add support for filtering labels and from / to as described in the updated docs.

Useful if you want to get recent incidents for a given team.


javiermolinar (approved):
LGTM

PR #499, opened 2026-04-15
fix(auth): skip OAuth bearer injection when Authorization header already set

### LINKED ISSUE #497: RefreshTransport unconditionally overwrites Authorization header, breaking Basic auth for adaptive metrics
## Summary

`RefreshTransport.RoundTrip` in `internal/auth/transport.go` unconditionally replaces the `Authorization` header with its OAuth bearer token on every request. This breaks any code path that sets its own auth (e.g. Basic auth for adaptive metrics API calls).

## Reproduction

1. Configure a context with both a `grafana` block (with OAuth tokens and proxy-endpoint) and a `cloud` block (with an access policy token):
   ```yaml
   ops:
     grafana:
       server: https://ops.grafana-ops.net
       token: glsa_...
       proxy-endpoint: https://assistant-ops-eu-south-0.grafana-ops.net/assistant
       oauth-token: gat_...
       oauth-refresh-token: gar_...
     cloud:
       token: glc_ops_...
       stack: ops
       api-url: https://grafana-ops.com
     providers:
       adaptive:
         metrics-tenant-id: "10428"
         metrics-tenant-url: https://prometheus-ops-03-ops-eu-south-0.grafana-ops.net
   ```
2. Run `gcx metrics adaptive rules list`
3. Observe 401 error: `"authentication error: legacy auth cannot be upgraded because the host is not found"`

## Root cause

The adaptive metrics client (`internal/providers/metrics/adaptive/client.go:61`) correctly sets Basic auth via `req.SetBasicAuth(tenantID, cloudToken)`. However, when a `grafana` config with OAuth is present, `LoadCloudConfig` creates a REST config with a `WrapTransport` that includes `RefreshTransport`.

In `internal/auth/transport.go:50-52`:
```go
clone := req.Clone(req.Context())
clone.Header.Set("Authorization", "Bearer "+token)
return t.base().RoundTrip(clone)
```

This unconditionally overwrites the `Authorization` header, replacing `Basic <base64(tenantID:cloudToken)>` with `Bearer gat_...`. The Mimir auth gateway receives the wrong token type and rejects it.

## Workaround

Remove the `grafana` block from the context so that `LoadCloudConfig` returns a nil `RESTConfig` and uses a plain HTTP client instead of the k8s transport chain.

## Suggested fix

Skip OAuth injection when an `Authorization` header is already present, matching the pattern in k8s `bearerAuthRoundTripper` (`vendor/k8s.io/client-go/transport/round_trippers.go`):

```go
func (t *RefreshTransport) RoundTrip(req *http.Request) (*http.Response, error) {
    if req.Header.Get("Authorization") != "" {
        return t.base().RoundTrip(req)
    }
    // ... existing refresh + bearer logic
}
```

wardbekker:
## Summary
- `RefreshTransport.RoundTrip` unconditionally overwrote the `Authorization` header with its OAuth bearer token, breaking code paths that set their own auth (e.g. Basic auth for adaptive metrics API calls)
- Added an early return that skips bearer injection when the request already carries an `Authorization` header, matching the k8s `bearerAuthRoundTripper` pattern
- Verified live: `gcx metrics adaptive rules list` against the ep context now succeeds (was returning 401)

## Changes
- `internal/auth/transport.go` — guard `RoundTrip` to pass through requests with existing `Authorization` header
- `internal/auth/transport_test.go` — add `TestRefreshTransport_PreservesExistingAuthorizationHeader`

## Test Plan
- [x] New test verifies Basic auth header is preserved through the transport
- [x] All existing transport tests still pass
- [x] Full test suite passes with race detection (`go test -race -count=1 ./...`)
- [x] Lint clean (`make lint` — 0 issues)
- [x] Live-verified against ep context: `gcx metrics adaptive rules list` returns `0 rule(s)` instead of 401
- [x] OAuth-authenticated commands (`gcx resources get folders`) still work

Closes #497

### DIFF
--- a/internal/auth/transport.go
+++ b/internal/auth/transport.go
@@ -39,6 +39,12 @@ type RefreshTransport struct {
 }
 
 func (t *RefreshTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// If the caller already set an Authorization header (e.g. Basic auth for
+	// adaptive metrics), respect it and skip OAuth bearer injection.
+	if req.Header.Get("Authorization") != "" {
+		return t.base().RoundTrip(req)
+	}
+
 	if err := t.maybeRefresh(req); err != nil {
 		return nil, fmt.Errorf("token refresh failed: %w", err)
 	}
--- a/internal/auth/transport_test.go
+++ b/internal/auth/transport_test.go
@@ -210,6 +210,46 @@ func TestRefreshTransport_ReturnsErrRefreshTokenExpired_On401(t *testing.T) {
 	}
 }
 
+func TestRefreshTransport_PreservesExistingAuthorizationHeader(t *testing.T) {
+	var gotHeader string
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotHeader = r.Header.Get("Authorization")
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer backend.Close()
+
+	transport := &auth.RefreshTransport{
+		Base:          http.DefaultTransport,
+		ProxyEndpoint: backend.URL,
+		Token:         "gat_test-token",
+		ExpiresAt:     time.Now().Add(1 * time.Hour),
+	}
+
+	client := &http.Client{Transport: transport}
+	req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, backend.URL+"/api/test", nil)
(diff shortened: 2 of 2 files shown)


flxbk:
FWIW, still doesn't work for me when oauth is configured in the grafana block (no issues when there's only a service account token).

wardbekker:
dupe

PR #502, opened 2026-04-15
adaptive telemetry clients work correctly when oauth client is used for grafana

### LINKED ISSUE #434: OAuth: adaptive telemetry APIs reject OAuth tokens
Parent: #179
Related: #433

## Problem

When using OAuth login, the adaptive telemetry APIs (Adaptive Metrics, Adaptive Logs, Adaptive Traces) reject the `gat_` OAuth token with `401: legacy auth cannot be upgraded because the host is not found`.

These providers make direct HTTP calls to product-specific backend endpoints, bypassing the Grafana/assistant proxy. The backend services don't recognise OAuth tokens.

**Affected resources (5):**

| Provider | Endpoint | Resource |
|---|---|---|
| Adaptive Metrics | `prometheus-blocks-prod-us-central1.grafana.net/aggregations/rules` | AggregationRule |
| Adaptive Logs | `logs-prod-us-central1.grafana.net/adaptive-logs/exemptions` | Exemption |
| Adaptive Logs | `logs-prod-us-central1.grafana.net/adaptive-logs/segments` | Segment |
| Adaptive Traces | `tempo-us-central1.grafana.net/adaptive-traces/api/v1/policies` | Policy |

**Note:** Adaptive Logs DropRules fail with both auth methods (404), so that's a separate issue. Adaptive Logs Exemptions and Segments also fail with token auth but with a different error (`invalid scope requested`), suggesting the token scopes need attention too.

## Fix options

1. **Route through assistant proxy** when OAuth auth is active — these providers already use the `adaptive` auth package (`internal/auth/adaptive/`), which could be extended to detect OAuth mode and proxy requests through the assistant endpoint instead of calling backends directly.
2. **Token exchange** — exchange the OAuth token for product-specific credentials via GCOM. More complex.

Option 1 is most consistent with how K8s resource requests already work under OAuth.

## Acceptance criteria

- [ ] `gcx resources get` returns Adaptive Metrics AggregationRules when using OAuth
- [ ] `gcx resources get` returns Adaptive Logs Exemptions and Segments when using OAuth
- [ ] `gcx resources get` returns Adaptive Traces Policies when using OAuth
- [ ] Signal provider query commands (`gcx metrics query`, `gcx logs query`, `gcx traces query`) work with OAuth

dafydd-t:
closes https://github.com/grafana/gcx/issues/434

Even though adaptive telemetry providers use a public API & the cloud.token credential, the requests were failing when we're using oauth to auth with grafana. This PR fixes that by using the provider client instead of the oauthed client, if it has a credential.

commands like `gcx resources get AggregationRule` work when using an oauthed client now (cloud.token is still required)

### DIFF
--- a/internal/auth/transport.go
+++ b/internal/auth/transport.go
@@ -39,6 +39,11 @@ type RefreshTransport struct {
 }
 
 func (t *RefreshTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Provider requests may carry their own BasicAuth credentials.
+	if req.Header.Get("Authorization") != "" {
+		return t.base().RoundTrip(req)
+	}
+
 	if err := t.maybeRefresh(req); err != nil {
 		return nil, fmt.Errorf("token refresh failed: %w", err)
 	}
--- a/internal/auth/transport_test.go
+++ b/internal/auth/transport_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


dafydd-t:
@claude review

javiermolinar (approved):
Nice!

PR #514, opened 2026-04-16
fix(slo): default definitions list limit to all; satisfy prealloc lint

annanay25:
- Default `gcx slo definitions list --limit` to 0; document trim behavior and set agent token_cost to medium with a scoped llm_hint.
- Pre-allocate slices flagged by golangci-lint prealloc (graph, linter, IRM OnCall clients, logs/metrics adaptive, banner) and use max(0, n) in collectN for iterator helpers. These changes unfortunately got mixed with the real changes in this PR, but the following changes are all lint related
	- internal/graph/chart.go internal/linter/linter.go
	- internal/providers/irm/oncall_commands_extra.go
	- internal/providers/irm/oncallpublic/client.go
	- internal/providers/logs/adaptive/segment_stats.go
	- internal/providers/metrics/adaptive/commands.go
	- internal/providers/slo/definitions/commands.go
	- internal/providers/slo/provider_test.go
	- internal/style/banner.go

Made-with: Cursor

### DIFF
--- a/internal/providers/slo/definitions/commands.go
+++ b/internal/providers/slo/definitions/commands.go
@@ -60,7 +60,7 @@ func (o *listOpts) setup(flags *pflag.FlagSet) {
 	o.IO.DefaultFormat("table")
 	o.IO.BindFlags(flags)
 
-	flags.Int64Var(&o.Limit, "limit", 50, "Maximum number of items to return (0 for all)")
+	flags.Int64Var(&o.Limit, "limit", 0, "Maximum number of items to return after fetch (0 for all; use a positive value to trim output only)")
 }
 
 func newListCommand(loader GrafanaConfigLoader) *cobra.Command {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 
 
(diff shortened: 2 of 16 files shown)


internal/providers/slo/definitions/commands.go
```diff
@@ -60,7 +60,7 @@ func (o *listOpts) setup(flags *pflag.FlagSet) {
 	o.IO.DefaultFormat("table")
 	o.IO.BindFlags(flags)
 
-	flags.Int64Var(&o.Limit, "limit", 50, "Maximum number of items to return (0 for all)")
+	flags.Int64Var(&o.Limit, "limit", 0, "Maximum number of items to return after fetch (0 for all; use a positive value to trim output only)")
```
radiohead, line 63: **nitpick (non-blocking):** The flag description ("trim output only; same API payload either way") is accurate for SLO, but the example in `docs/design/help-text.md` could mislead readers into assuming this holds across all providers — it doesn't for providers where `opts.Limit` maps to server-side pagination. Consider scoping the claim to SLO explicitly, or adding a clarifying note in the design doc.

radiohead (approved):

Reviewed the `--limit` default change and the prealloc sweep across 16 files. The limit change is safe — SLO was already fetching all records unconditionally; the previous 50 was display-only truncation. All prealloc capacity hints are correct and the docs are updated in lockstep.

PR #515, opened 2026-04-16
feat(kg): add --from/--to time range flags to all kg commands

xujiaxj:
## Summary

- All `gcx kg` commands previously only accepted `--since` (relative duration like `1h`, `7d`). This adds `--from`/`--to` support, consistent with the rest of gcx (metrics, logs, traces, slo timeline).
- `--from`/`--to` accept RFC3339, Unix timestamps, or relative expressions like `now-1h`.
- `--since` is preserved for backward compatibility and remains mutually exclusive with `--from`/`--to`.
- Commands that previously hardcoded a 1-hour window (`entities show`, `entities list`, `search entities`, `search sample`) now respect the user-specified time range.

Closes #337

## Test plan

- [x] `gcx kg entities list --from now-2h --to now` returns entities in the 2-hour window
- [x] `gcx kg insights query Service--my-service --from 2024-01-01T00:00:00Z --to 2024-01-01T01:00:00Z` uses absolute timestamps
- [x] `gcx kg entities list --since 1h` still works (backward compat)
- [x] `gcx kg entities list --since 1h --from now-1h` returns an error (mutual exclusion)
- [x] `gcx kg entities list --from now-1h` (without `--to`) returns an error

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -8,7 +8,6 @@
   ],
   "shell": {
     "init_hook": [
-      "unset GOROOT",
       "echo 'Entering Python venv' && . $VENV_DIR/bin/activate",
       "echo 'Installing dependencies...' && make deps",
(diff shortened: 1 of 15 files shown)


xujiaxj:
The reference docs were already regenerated in the second commit (5c9f042). `make reference-drift` exits clean on this branch — no drift. The reviewer may have looked at the PR before that commit was pushed.

annanay25 (commented):
nice! wondering if we need to retain the `--since` flag?

xujiaxj:
> nice! wondering if we need to retain the `--since` flag?

do we still need to support "since"? Do other commands support "since"?                                                                                                         

  Searched for 6 patterns (ctrl+o to expand)

⏺ --since is present across several providers — signal providers (via query/opts.go), SLO timeline, synth checks, and profiles. It's a consistent pattern in the codebase, not
  KG-specific.

  So yes, keep it. The convention gcx uses everywhere is: --from/--to for explicit windows, --since as a shorthand for "last N duration" (mutually exclusive with --from). The KG
  commands now match that pattern correctly.

radiohead (approved):

PR #521, opened 2026-04-17
refactor(kg): remove redundant search entities command

xujiaxj:
## Summary

- `gcx kg search entities` was a strict subset of `gcx kg entities list`: identical implementation, identical flags (`--type`, `--page`, scope flags), only differing in default output format (`json` vs `table`) — trivially overridden with `-o`
- `gcx kg entities list` is the superset, additionally supporting `--assertions-only`
- Removes the command, its `searchEntitiesListOpts` struct, and the `//nolint:dupl` suppression that was only needed because the two implementations were identical

## Test plan

- [x] `gcx kg entities list` still works
- [x] `gcx kg search` no longer shows an `entities` subcommand
- [x] `gcx kg entities list -o json` produces the same output as `gcx kg search entities` did

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


annanay25 (approved):
lgtm!

PR #522, opened 2026-04-17
fix(auth): prevent OAuth refresh lockout from concurrent gcx invocations

wardbekker:
## Summary

- OAuth proxy rotates the refresh token on every call. Any failure in the refresh/persist chain leaves the client with a now-invalid refresh token on disk — the next invocation fails with 401 and the user has to re-login. With a ~1h access-token TTL this manifests as _"tokens expire within a day"_.
- Two concurrent `gcx` processes sharing a config file would both read the same stale refresh token, race to refresh it, and one would always 401 on a just-rotated token. Reproduced locally: two parallel `gcx resources get folders` → one exits 1 with `"1 resource(s) failed to get"`. After the fix, three parallel invocations all succeed.
- A successful refresh whose response had an unparseable `expires_at` was discarded entirely — even though the server had already consumed+rotated the refresh token. The client then silently had a stale refresh token on disk.

## What changed

- `internal/auth/transport.go`: `RefreshTransport` gains `Lock` (cross-process) and `Reload` callbacks. Inside `maybeRefresh`, we now acquire the lock, re-read disk state, and — if another process already refreshed to a non-stale token — adopt those tokens and skip the network call. Parse failures on `expires_at` no longer discard the new tokens; the transport falls back to a zero expiry so the next request re-refreshes.
- `internal/config/rest.go`: `WireTokenPersistence` wires `Lock` via `gofrs/flock` (`<config>.lock`) and `Reload` via `config.Load`. Persistence uses a `context.WithoutCancel(ctx)` so disk writes complete even if the request context is cancelled.
- New dep: `github.com/gofrs/flock`.

## Test plan

- [x] `go test -race -count=1 ./...` — green, including four new tests:
  - `TestRefreshTransport_PersistsTokensWhenExpiresAtUnparseable`
  - `TestWireTokenPersistence_WritesAfterContextCancelled`
  - `TestWireTokenPersistence_ConcurrentRefreshesSerializeViaFileLock`
  - `TestWireTokenPersistence_RoundTripAcrossInvocations`
- [x] `make lint` — clean.
- [x] Live reproduction against real proxy: before fix, two parallel `gcx resources get folders` against an expired access token → one fails with `refresh token expired`. After fix, three parallel invocations all succeed and exactly one refresh hits the network.


wardbekker:

### CI Checks
| Check | Status |
|-------|--------|
| Linters | PASS |
| Tests | PASS |
| Documentation | PASS |
| claude-review | PASS ("No issues found") |
| zizmor | PASS |
| TruffleHog | PASS |
| Socket Security | PASS |

### Auto-Fixes Applied
None required — branch arrived clean.

### Pending Review
- No human review yet (no reviewers requested)
- `mergeStateStatus`: `BLOCKED` (waiting on `REVIEW_REQUIRED`)

### Escalations
None.

---
*Updated by shepherd-pr*

radiohead (approved):

PR #529, opened 2026-04-20
feat(auth): consolidate gcx auth and gcx config into gcx login

### LINKED ISSUE #518: Consolidate gcx auth + gcx config into unified gcx login
## Overview

Replace the current 6-8 step first-run dance (`config set` ×3 + `auth login` +
`config use-context` + `config check`) with a single guided command: `gcx login`.

Also deprecates `gcx auth` as a top-level command — its functionality folds into
`gcx login` (primary auth) and `gcx config` (token management primitives).

Design: ADR `login-consolidation/001-login-config-consolidation` (accepted) — lands
with the first implementation PR.

Source research: `docs/plans/2026-04-14-ux-consistency-design.md` (section D1).

## Scope

- New `internal/login` package with package-level `Run(ctx, opts)` function
  orchestrating the full lifecycle (server URL → cloud detection → Grafana auth →
  Cloud auth → validation → config write)
- New `cmd/gcx/login` Cobra command, thin wiring over `login.Run()`, interactive
  prompts via `charmbracelet/huh` (already a dep, matches `cmd/gcx/dev/scaffold.go`)
- Three-bucket cloud detection: known Cloud (`*.grafana.net`, `-dev.net`, `-ops.net`)
  → `TargetCloud`; localhost/RFC 1918 → `TargetOnPrem`; ambiguous → prompt
- Two-step auth: step 1 = Grafana (OAuth or SA token), step 2 = Cloud Access
  Policy token (optional, Cloud only)
- Config schema: one additive field (`GrafanaConfig.AuthMethod`, `omitempty`,
  backward-compatible)
- Deprecation: `gcx auth` gets `cobra.Command.Deprecated` warning, subcommand
  rewires to `login.Run()` for 2 minor versions, then hard-removed
- CI/agent path: env vars (`GRAFANA_SERVER`, `GRAFANA_TOKEN`, `GRAFANA_CLOUD_TOKEN`)
  unchanged; `--yes` disables prompts and errors on missing required values
- Flags: `--server`, `--token`, `--cloud-token`, `--cloud`, `--cloud-api-url`,
  `--context`, `--yes`

## Out of scope

- OAuth coverage for Cloud APIs (#310, #311) — this ADR wires to the existing
  OAuth flow; step 2 CAP token remains until OAuth covers GCOM
- `gcx setup` product onboarding framework (#287) — separate axis, D2 from the
  UX design

## Success Criteria

- [ ] `gcx login` completes first-run setup in one command (interactive TTY)
- [ ] `gcx login --yes --server X --token Y [--cloud-token Z]` persists a context
      non-interactively
- [ ] Env vars (`GRAFANA_SERVER`/`GRAFANA_TOKEN`/`GRAFANA_CLOUD_TOKEN`) continue
      to work without running `gcx login`
- [ ] `gcx auth login` prints a deprecation warning and still works, internally
      calling `login.Run()`
- [ ] Cloud detection correctly classifies `*.grafana.net`, `*.grafana-dev.net`,
      `*.grafana-ops.net`, localhost, RFC 1918, and prompts on ambiguous domains
- [ ] Existing config files (pre-`AuthMethod`) load and work unchanged
- [ ] Connectivity validation fails cleanly without writing partial config
- [ ] `internal/login` is covered by table-driven unit tests (no CLI integration
      tests required for core logic)

radiohead:
## Summary

Consolidates the multi-step Grafana onboarding flow into a single `gcx login` command. What used to require three `gcx config set` calls plus `gcx auth login` plus `gcx config use-context` plus `gcx config check` is now one command that handles the whole thing.

Behaviour:

- Auto-detects Grafana Cloud vs on-premises from the server URL, with an explicit `--cloud` override.
- Interactive flow: OAuth (browser) leads on Cloud, service account tokens are the only option on on-premises.
- Non-interactive flow with `--yes` for CI and automation; structured error output in agent mode when required inputs are missing.
- Connectivity validation before any context is persisted: health check, K8s API availability, Grafana version (≥12 required), and GCOM stack discovery when a Cloud Access Policy token is supplied.
- Fresh on-premises logins default `grafana.org-id` to 1, so resource commands work immediately without a manual config step.
- `gcx auth` removed. `gcx config check` now shows the resolved auth method per context.

Closes #518

## Review response

Addresses review feedback from the initial and follow-up rounds:

- OAuth is now the first option in the interactive auth-method prompt on Cloud, with softened copy (raised by @sd2k).
- On-premises targets no longer show the OAuth option (the spec already required this; raised by @wardbekker).
- Fresh on-premises logins default `grafana.org-id` to 1 (raised by @wardbekker).
- The user-facing Cloud-commands advisory uses the full "Cloud Access Policy token" term instead of the "CAP token" abbreviation (raised by @wardbekker).
- README Authenticate section rewritten to match the new command surface.

A separate follow-up for self-healing missing stack/org identifiers in pre-existing contexts is tracked in #545.

## Documentation

Adds a user-facing login reference and two architecture deep dives (login orchestration and the authentication subsystem) so the new flow is documented for both users and contributors.

## Test plan

- [x] `go test -race -count=1 ./...` passes
- [x] `make lint` passes
- [x] `gcx login --help` shows the current flag set
- [x] `gcx auth` returns an unknown-command error
- [x] `gcx config check` shows the resolved auth method
- [x] Non-interactive login with `--yes` completes without prompts
- [x] Agent mode exits non-zero with a structured error when required inputs are missing
- [x] `make docs` builds successfully


cmd/gcx/login/command.go
```diff
@@ -0,0 +1,514 @@
...
+	Config      configcmd.Options
+	Server      string
+	Token       string
+	CloudToken  string
+	CloudAPIURL string
+	Cloud       bool
+	Yes         bool
+}
+
+func (opts *loginOpts) setup(flags *pflag.FlagSet) {
```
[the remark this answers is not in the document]
radiohead: This comment has been generated by Claude Code.

Addressed in b574523.

`loginOpts` now has a `Validate(args []string) error` method at `cmd/gcx/login/command.go:71`, called from `RunE` before `runLogin`. It holds the positional-arg-vs-`--context` mutual-exclusion check that was previously inlined in `runLogin`, plus `opts.IO.Validate()` so `--output`/`--json` are vetted up front.

New `cmd/gcx/login/command_test.go` covers the mutual-exclusion cases (`TestLoginOptsValidate`) alongside tests for `structuredMissingFieldsError`, `structuredClarificationError`, and the `printResult` text codec.

cmd/gcx/login/command.go
```diff
@@ -0,0 +1,610 @@
...
+// chosen, for the token itself. When existingToken is non-empty (re-auth),
+// the token prompt allows empty input to reuse the stored token.
+func askGrafanaAuth(opts *login.Options, existingToken string) error {
+	authMethod := "token"
+	methodForm := huh.NewForm(huh.NewGroup(
+		huh.NewSelect[string]().
+			Title("Authentication method").
+			Options(
+				huh.NewOption("Service account token (requires permissions for managing service accounts)", "token"),
+				huh.NewOption("OAuth (browser) — experimental, some functionality may not work; fall back to a service account token if you hit auth issues", "oauth"),
```
sd2k: Perhaps we should rearrange these so that OAuth is first since that's the one we want to guide people towards?

sd2k (commented):
Quite a lot to review but it seems to work fine locally!

wardbekker:
I've got "Note: Cloud API commands require a CAP token" after succesful authenticating with OAuth. Is the text here correct? IIRC CAP token is the OAuth token, not the gcom / cloud access policy token that is missing.

wardbekker:
While testing this PR against a local Grafana OSS 13.0.1 container, I noticed a small usability gap:

`gcx login` successfully creates the on-prem context, stores the token, and sets it as current — but does **not** set `grafana.org-id`. The next resource command then fails hard:

```
$ gcx login local-oss --yes --server http://localhost:3000 --token glsa_…
{ "contextName": "local-oss", "server": "http://localhost:3000", "authMethod": "token", "cloud": false, "grafanaVersion": "13.0.1", "hasCloudToken": false }

$ gcx resources get folders
Invalid configuration found in '':
missing contexts.local-oss.grafana.org-id or contexts.local-oss.grafana.stack-id
```

Workaround is `gcx config set contexts.local-oss.grafana.org-id 1`, after which resource commands work.

### Suggestion

When the login target is detected as on-prem (`cloud: false`) and no `org-id` is configured on the context, default it to `1` during login. That's the only org ID on a stock Grafana OSS install, and it's the overwhelmingly common case for on-prem users who aren't running a multi-tenant setup. Users who do run multi-tenant can still override via `--org-id` (if we add the flag) or `gcx config set`.

Alternatively, query `/api/user` with the supplied token and populate `org-id` from `orgId` in the response — that also handles users whose token is scoped to a non-default org without needing a new flag.

Either way, a fresh `gcx login` against an on-prem Grafana should leave the user in a state where `gcx resources get …` works immediately, which is the obvious expectation after the "consolidate auth+config into login" framing of this PR.

wardbekker:
For an self hosted context, it should not offer OAuth login, as that's only available on Cloud stacks

> ./bin/gcx login local-oss-interactive --server http://localhost:3000
Creating new context "local-oss-interactive"
Existing contexts: default, ep-amer-cloud-demokit-grafana-com, local-oss

┃ Authentication method
┃ > Service account token (requires permissions for managing service accounts)
┃   OAuth (browser) — experimental, some functionality may not work; fall back to a service account token if you hit auth issues

wardbekker:
Follow-up to my previous comment: `gcx login --cloud-token …` doesn't populate `cloud.stack` (or `grafana.stack-id`), so Cloud-management commands still fail after a supposedly-complete login.

**Repro** (against a cloud context, fresh config):

```
$ gcx login my-ctx --server https://<stack>.grafana.net --token <sa-token> --cloud-token glc_…
# login succeeds, writes cloud.token but NOT cloud.stack

$ gcx synth checks list
SM token not configured: … cloud stack is not configured: set cloud.stack in config or GRAFANA_CLOUD_STACK env var

$ gcx k6 projects list
k6: load cloud config: cloud stack is not configured: set cloud.stack in config …

$ gcx fleet pipelines list
cloud stack is not configured: …
```

User then has to manually `gcx config set contexts.my-ctx.cloud.stack <slug>` before any of these work.

**Where:** [`internal/login/login.go:351-361`](https://github.com/grafana/gcx/blob/feat/gcx-login/internal/login/login.go#L351-L361) — `resolveCloudAuth` only writes `Token` and `APIUrl`:

```go
if opts.CloudToken != "" {
    return &config.CloudConfig{
        Token:  opts.CloudToken,
        APIUrl: opts.CloudAPIURL,
    }, nil
}
```

**Ironic part:** the same function already has everything needed. Line 274 computes `result.StackSlug` via `resolveStackSlug(opts.Server)` and ships it back to the caller for display — it's just never persisted to `CloudConfig.Stack`.

**Suggested fix (in priority order):**

1. **Fast path** — if `resolveStackSlug(opts.Server)` returns non-empty, copy it into `CloudConfig.Stack`. One-line change, covers every `*.grafana.net` server URL.
2. **Robust path** — `glc_…` tokens are base64-JSON with `{o: orgID, m: {r: region}}`. Decode the token, `GET /api/orgs/<org>/instances` against GCOM, find the instance whose `url` matches the context's server URL (handles custom domains like `*.cloud.demokit.grafana.com` where the slug can't be extracted from the hostname).
3. **Both** — 1 as the common-case fast path, 2 as the fallback.

Option 3 closes the gap for every realistic deployment topology without adding interactive prompts, and aligns with the PR's "consolidate auth + config into login" framing: a successful login with `--cloud-token` should leave the user in a fully-working state for Cloud commands.

Same argument applies to `grafana.stack-id` (the K8s namespace field) — GCOM's instance response includes the numeric ID right next to the slug.

cmd/gcx/login/command.go
radiohead: I was thinking about that and decided to keep OAuth 2nd because it's quite experimental still.

But it's a coin toss, so I'm happy to re-arrange the options!

radiohead:
> I've got "Note: Cloud API commands require a CAP token" after succesful authenticating with OAuth. Is the text here correct? IIRC CAP token is the OAuth token, not the gcom / cloud access policy token that is missing.

CAP = Cloud Access Policy, so that should be correct. But I'm happy to change the wording here if it's too ambiguous.

> For an self hosted context, it should not offer OAuth login, as that's only available on Cloud stacks

Whoops, that requirement (no OAuth in OSS / on-prem) was in the spec but somehow slipped through.

RE: orgID / stackID – all good suggestion, will implement.

wardbekker:
> > I've got "Note: Cloud API commands require a CAP token" after succesful authenticating with OAuth. Is the text here correct? IIRC CAP token is the OAuth token, not the gcom / cloud access policy token that is missing.
> 
> CAP = Cloud Access Policy, so that should be correct. But I'm happy to change the wording here if it's too ambiguous.
> 
> > For an self hosted context, it should not offer OAuth login, as that's only available on Cloud stacks
> 
> Whoops, that requirement (no OAuth in OSS / on-prem) was in the spec but somehow slipped through.
> 
> RE: orgID / stackID – all good suggestion, will implement.

CAP -> can you use "Cloud Access Policy (CAP)" wording  and perhaps also a deep link to the docs to help folks here? 
https://grafana.com/docs/grafana-cloud/security-and-account-management/authentication-and-permissions/access-policies/  Hopefully this makes it super clear for users.

cmd/gcx/login/command.go
radiohead: Done!

wardbekker:
## `--yes` bypasses the server-override guard

**File:** `internal/login/login.go:420-434`

Running `gcx login <existing-context> --yes --server different-host.example.net` silently repoints an existing context at a new server without prompting. The server-mismatch guard is intentionally skipped when `--yes` is set:

```go
if oldServer != "" && newServer != "" && oldServer != newServer &&
    !opts.Yes && !opts.AllowOverride {
    return &ErrNeedClarification{...}
}
```

This is documented in the comment ("Bypass when Yes"), but it means automated/CI scripts running `gcx login --yes` can accidentally change which Grafana instance a saved context targets — which then silently changes where subsequent `gcx push`/`pull`/`delete` commands operate.

**Suggested fix:** require a separate, explicit flag (e.g. `--allow-server-override`) to change the server on an existing context. `--yes` would continue to auto-confirm other prompts but would not override the server-identity guard. Alternatively, keep `--yes` as-is but refuse the override unless the user also passes `--server` explicitly on the command line (i.e. never silently override from a staged/cached value).

Happy to open a follow-up PR if you agree this warrants tightening.

wardbekker (approved):

sd2k (approved):
LGTM!

radiohead:

Addressed in c4e84a8. Three items closed out:

**1. `--yes` bypasses the server-override guard** (latest comment by @wardbekker): `--yes` no longer bypasses the server-mismatch guard. The guard now only checks `!opts.AllowOverride`. A new `--allow-server-override` flag is the explicit opt-in for changing a context's server target non-interactively. The structured error suggestion now says `--allow-server-override` instead of `--yes`. New test: `TestPersist_ServerMismatch_YesDoesNotBypass`.

**2. `cloud.Stack` not populated after `--cloud-token` login**: `resolveCloudAuth` now derives the stack slug from the server URL (fast path for `*.grafana.net`) and populates `CloudConfig.Stack` when non-empty. Cloud management commands should work without a manual `gcx config set` after login.

**3. CAP token wording**: Advisory now reads "Cloud Access Policy (CAP) token" with the docs deep-link.

PR #537, opened 2026-04-20
feat(homebrew): replace binary cask with source formula for Gatekeeper-free macOS installs

radiohead:
## Background

gcx release binaries are not Apple-notarised and have no Developer ID signature. On macOS this means the OS rejects the binary the first time a user runs it, with two distinct symptoms:

- **Intel macOS**: a Gatekeeper dialog — *"Apple could not verify 'gcx' is free of malware…"* — and the binary refuses to run.
- **Apple Silicon**: the kernel's code-signing check fires before Gatekeeper; the process exits immediately with `killed: 9` and no visible dialog.

Both come from the same underlying cause: the `com.apple.quarantine` extended attribute that macOS attaches to any binary downloaded via a user-space fetch. Homebrew 5.0 removed the `--no-quarantine` cask flag, so `brew install --cask` now leaves the xattr in place — every first-run of a cask-installed gcx hits one of the two symptoms above.

The prior shape of this PR shipped a cask anyway and papered over the problem with a documented `xattr -d com.apple.quarantine && codesign --sign -` recipe. That's two opaque shell commands between `brew install` and a working binary, for every user, on their very first interaction with gcx. It also diverged from every other Grafana CLI — `alloy.rb`, `grafana.rb`, `mimirtool.rb`, `cortextool.rb`, `grafana-agent.rb` in the same `grafana/homebrew-grafana` tap all ship as build-from-source formulas.

This PR switches gcx to the same model. The formula pulls GitHub's auto-generated source tarball, declares `go` as a build dependency, and compiles `gcx` on the user's machine during `brew install`. No binary crosses the wire, so no xattr is set and no signature is checked — the problem doesn't exist on this path. Manual-download and `curl | sh` installs still need the workaround, but those are for users who are already stepping outside the package-manager blessed path.

Apple notarisation is tracked separately and this change does not block on it.

### Formula template (`.github/homebrew/gcx.rb.tmpl`)

The formula is a 30-line Homebrew Ruby DSL with `${VERSION}` and `${SHA256}` shell-style placeholders that `envsubst` substitutes at publish time. Ruby's `#{...}` interpolations (`#{version}`, `#{time.iso8601}`, `#{bin}`) are preserved by invoking `envsubst` with an allow-list of only `$VERSION $SHA256`.

A few deliberate design points:

- `depends_on "go" => :build` is floating rather than pinned to `go@1.26`. `go@1.26` doesn't exist as a versioned formula in homebrew-core yet, and the default `go` is currently at `1.26.2`, compatible with gcx's `go.mod` floor. Matches Alloy. If the floor ever outruns homebrew-core, we pin then.
- `-buildvcs=false` is needed because GitHub's `archive/refs/tags/vX.Y.Z.tar.gz` tarballs contain no `.git`, and `go build` would otherwise fail trying to stamp VCS info.
- `main.commit=homebrew` is a sentinel. The real commit SHA isn't available inside the install block, and threading it through the publisher adds churn for low value. `gcx --version` showing `commit=homebrew` is a clear tell that the binary came from a Homebrew build.
- `generate_completions_from_executable(bin/"gcx", "completion")` installs bash/zsh/fish completions via Cobra's default `completion <shell>` subcommand. This replaces the `completion` caveat the cask carried.
- `head "…", branch: "main"` enables `brew install --HEAD grafana/grafana/gcx` for users who want tip-of-main.
- No `caveats`, `zap`, or `conflicts` stanzas. The cask-era stanzas don't apply: completions are auto-installed now, `zap` is a cask-only concept, and there's no naming collision.

### Publish workflow (`.github/workflows/publish-homebrew-formula.yml`)

Runs after each stable release tag, renders the formula from the template, and opens a PR against `grafana/homebrew-grafana`.

- **Checkout ref.** The workflow checks out at its *native* ref, not `inputs.tag`. When called via `workflow_call` from `release.yaml`, that's the release tag (which sits on `main` and includes the template). When dispatched manually on a feature branch, it's that branch. Older tags don't have the template; pinning checkout to `inputs.tag` would break historical re-dispatches. `inputs.tag` is used only for the sha256 computation and the version string.
- **sha256 format gate.** After computing the hash, the workflow asserts `[[ $SHA256 =~ ^[0-9a-f]{64}$ ]]`. Defends against a truncated `sha256sum` output (e.g. if upstream pipe changes break `awk`).
- **Post-render sanity checks.** Three `grep -q` assertions on the rendered `gcx.rb` — class header present, sha256 substituted, version substituted into the URL. Catches template corruption, envsubst misconfiguration, or a missing allow-list argument *before* the tap PR is opened. If `envsubst` were ever invoked without `'$VERSION $SHA256'`, the class header would still match but the sha256 check would fail on the literal `${SHA256}`.
- **Tap layout.** `gcx.rb` goes at the tap root (flat layout, next to `alloy.rb`), not under `Casks/`. The commit clone step explicitly `cp`s to the root, and a comment inline documents the divergence from the cask workflow it was cloned from.
- **Idempotency.** `git diff --cached --quiet` runs before `git commit`/`git push`, so a byte-identical re-render exits 0 without opening a duplicate PR. The existing-PR check (`gh pr list --state all`) runs before `gh pr create`, which prevents the known "branch with closed PR already exists" failure mode when re-dispatching.
- **Pre-release guard.** Tags containing a hyphen (`v*-rc.*`, `v*-dev.*`) exit 0 with a log message before any render runs. Homebrew users should only see stable releases in the tap.
- **Auth.** Unchanged from the cask workflow: GitHub App credentials via Vault OIDC, token scoped to `grafana/homebrew-grafana` only, `persist-credentials: false` on the gcx-side checkout so the installation token doesn't leak into `.git/config`.

### GoReleaser config (`.goreleaser.yaml`)

Removed the `homebrew_casks:` stanza and the `release.extra_files` glob. GoReleaser no longer participates in the Homebrew path. Anyone updating formula content should edit the template, not `.goreleaser.yaml` — this is called out explicitly in `CONTRIBUTING.md` to prevent future drift.

### Release pipeline wiring (`.github/workflows/release.yaml`)

Replaced the `publish_homebrew_cask` downstream job with `publish_homebrew_formula` pointing at the new workflow. Dropped `continue-on-error: true`. The cask path had a "missing `.rb` asset" failure mode that was reasonable to swallow (it kept the release green when GoReleaser's upload flaked); the formula path has no such mode — the source tarball is always present post-release. Any failure (Vault auth, App token mint, curl, render assertion, tap PR creation) is a real error and should fail the release run loudly. Placed after the `pypi:` job that landed on `main` during this PR's lifetime.

### Cask workflow deletion

`.github/workflows/publish-homebrew-cask.yml` removed. Its 199 lines of auth and tap-publishing logic are fully subsumed by the new formula workflow (auth plumbing is identical; the render step is the only new surface). The cask never merged to the tap, so there's no migration burden for existing users.

### Docs

- `README.md` and `docs/installation.md`: `brew install --cask grafana/grafana/gcx` → `brew install grafana/grafana/gcx`. Section title changed to "Homebrew (macOS and Linux)" — the formula works on Linuxbrew too. Added a one-line note about the ~30–60 s first-install compile cost (downloading Go modules from `proxy.golang.org`) so users aren't surprised.
- `docs/installation.md`'s macOS Gatekeeper section: scope narrowed to manual downloads and `curl | sh`. New sentence **"Homebrew users are not affected — the formula compiles gcx from source on your machine, so no pre-built binary is downloaded and no xattr is set."** The `xattr -d` / `codesign -` code block stays for the remaining affected install paths.
- `CONTRIBUTING.md`: rewrote the "GoReleaser" and "Homebrew tap integration" paragraphs. Describes the 4-step publish pipeline (checkout → compute sha256 → render → open tap PR) and explicitly states **"The formula is not generated by GoReleaser."** Added a note on how to re-run the publisher for a past tag via `workflow_dispatch`.

## Verification

Local on this branch (Apple Silicon macOS, after rebase onto current `main`):

```
$ GCX_AGENT_MODE=false make all
# lint: golangci-lint clean
# test: 87 Go packages, 45/45 linter bundle tests — all green
# build: bin/gcx produced with correct ldflags
# docs: mkdocs build in 4.5 s (pre-existing warnings only)

$ TAG=v0.2.8 \
  VERSION=${TAG#v} \
  SHA256=$(curl -fsSL "https://github.com/grafana/gcx/archive/refs/tags/${TAG}.tar.gz" | shasum -a 256 | awk '{print $1}') \
  envsubst '$VERSION $SHA256' < .github/homebrew/gcx.rb.tmpl > /tmp/gcx.rb
$ ruby -c /tmp/gcx.rb
Syntax OK
$ grep -c 'time.iso8601\|version=v#{version}' /tmp/gcx.rb
2    # Ruby interpolations preserved — envsubst allow-list works
```

GitHub Actions validation of the publish workflow against a real tag is deferred until the first stable release cut after merge (expected `v0.2.9`). The failure modes are enumerated and surfaced loudly by the workflow itself — Vault auth, App token mint, curl retry exhaustion, render assertion, tap PR creation — so a first-release flake is observable and recoverable via `workflow_dispatch` re-run rather than a silent miss.

## Follow-ups (not blocking)

- Confirm `VAULT_APP_ID_PATH` and `VAULT_APP_PRIVATE_KEY_PATH` repo Actions variables are set (they were inherited from the prior cask workflow setup; a no-op if so).
- Confirm the GitHub App installation used by Vault has `contents: write` + `pull_requests: write` on `grafana/homebrew-grafana`.
- Revisit the formula's `curl --retry 3 --retry-delay 5` budget in the sha256 step if we see real-world tarball-generation lag after tag pushes.

## References

- Design spec: [`docs/specs/2026-04-22-homebrew-formula.md`](docs/specs/2026-04-22-homebrew-formula.md)
- Implementation plan: [`docs/plans/2026-04-22-homebrew-formula-implementation.md`](docs/plans/2026-04-22-homebrew-formula-implementation.md)


sd2k (approved):
LGTM!

PR #539, opened 2026-04-21
Added InfluxDB

mattvella07:
Used the add-datasource skill to add InfluxDB

Screenshot 2026-04-21 at 4 16 58 PM


internal/query/influxdb/client.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	respBody, err := c.executeQuery(ctx, body)
+	if err != nil {
+		return nil, err
+	}
+
+	var grafanaResp GrafanaQueryResponse
+	if err := json.Unmarshal(respBody, &grafanaResp); err != nil {
+		return nil, fmt.Errorf("failed to parse response: %w", err)
```
javiermolinar, line 58: This client returns plain errors for both http failures and datasource query errors, while the other datasource query clients use queryerror.FromBody(...) / queryerror.New(...).

internal/query/influxdb/client.go
```diff
@@ -0,0 +1,340 @@
...
+	var queryExpr string
+
+	switch mode {
+	case ModeFlux:
+		if bucket == "" {
+			return nil, errors.New("--bucket is required for Flux mode measurements")
+		}
+		queryExpr = fmt.Sprintf("import \"influxdata/influxdb/schema\"\nschema.measurements(bucket: %q)", bucket)
+	default:
+		queryExpr = "SHOW MEASUREMENTS"
```
javiermolinar, line 85: This treats every non-Flux mode as InfluxQL and sends SHOW MEASUREMENTS. `gcx datasources influxdb measurement`s will fail on SQL-mode datasource.

This should either:
- implement a SQL-specific metadata query, or
- reject SQL mode explicitly with a clear unsupported-mode error.

javiermolinar:
@codex review

javiermolinar (commented):
There is an issue in how the frame response from Grafana influxdb is being handled. Right now the frames are flattened into a single table losing the information about where the data came from:

ie:


 ```json
   {
     "results": {
       "A": {
         "frames": [
           {
             "schema": {
               "fields": [
                 { "name": "Time", "type": "time" },
                 {
                   "name": "_value",
                   "type": "number",
                   "labels": { "host": "server-a" }
                 }
               ]
             },
             "data": {
               "values": [
                 [1719849600000, 1719853200000],
                 [42.1, 43.0]
               ]
             }
           },
           {
             "schema": {
               "fields": [
                 { "name": "Time", "type": "time" },
                 {
                   "name": "_value",
                   "type": "number",
                   "labels": { "host": "server-b" }
                 }
               ]
             },
             "data": {
               "values": [
                 [1719849600000, 1719853200000],
                 [91.5, 90.2]
               ]
             }
           }
         ]
       }
     }
   }
 ```

The PR flattens both frames into one table:

 ```text
   Time                  _value
   2024-07-01T16:00:00Z  42.1
   2024-07-01T17:00:00Z  43.0
   2024-07-01T16:00:00Z  91.5
   2024-07-01T17:00:00Z  90.2
 ```

when it should be something like this:

```text
   Time                  _value  host
   2024-07-01T16:00:00Z  42.1    server-a
   2024-07-01T17:00:00Z  43.0    server-a
   2024-07-01T16:00:00Z  91.5    server-b
   2024-07-01T17:00:00Z  90.2    server-b
 ```

**Comment by mattvella07** [comment]
@javiermolinar Thanks for pointing this out!  I fixed it and here is a screenshot: 

Screenshot 2026-04-30 at 11 27 27 AM

File: internal/datasources/query/codecs.go [hunk 3]
```diff
@@ -56,6 +59,8 @@ func (c *queryWideCodec) Encode(w io.Writer, data any) error {
 		return loki.FormatQueryTableWide(w, resp)
 	case *tempo.SearchResponse:
 		return tempo.FormatSearchTable(w, resp)
+	case *influxdb.QueryResponse:
+		return influxdb.FormatQueryTable(w, resp)
```
dafydd-t: Since this shows the same as -o table, can we omit it here?

internal/datasources/query/codecs.go
dafydd-t, line 81: I think we could add to the graph codec here for influx, right?

internal/datasources/query/codecs.go
```diff
@@ -112,11 +117,51 @@ func (c *queryGraphCodec) Decode(io.Reader, any) error {
 	return errors.New("graph codec does not support decoding")
 }
 
+type queryJSONCodec struct {
+	inner *format.JSONCodec
+}
```
dafydd-t, line 129: Can you add a comment explaining why we need this? Its explained in FormatQueryJSON but its a confusing addition here without context, IMO

dafydd-t: Same  for queryYAMLCodec

internal/datasources/influxdb/fieldkeys.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+
+			modeStr, err := dsquery.GetInfluxDBMode(ctx, cfg, datasourceUID)
+			if err != nil {
+				return fmt.Errorf("failed to detect influxdb mode: %w", err)
+			}
+
+			if modeStr != "InfluxQL" {
+				return fmt.Errorf("field-keys is only supported in InfluxQL mode (datasource is configured for %s mode)", modeStr)
+			}
```
dafydd-t, line 89: Could we add a suggestion for what to do instead? e.g.

```
To get fields from Flux datasources you need to query directly:

  from(bucket: "grafanainfluxdb") |> range(start: -5m) |> filter(fn: (r) => r._measurement == "cpu_metrics") |> keep(columns: ["_field"]) |> distinct(column: "_field")
```

or do you think this is too complicated?

internal/datasources/influxdb/fieldkeys.go
```diff
@@ -0,0 +1,132 @@
...
+				return fmt.Errorf("failed to get field keys: %w", err)
+			}
+
+			return opts.IO.Encode(cmd.OutOrStdout(), resp)
+		},
+	}
+
+	cmd.Annotations = map[string]string{
+		agent.AnnotationTokenCost: "small",
+		agent.AnnotationLLMHint:   "gcx datasources influxdb field-keys -d UID -o json",
```
dafydd-t: I know this is the same everywhere, but we don't really need `-o json` on the LLM hint, because JSOn is the default output when agent mode is detected 


```suggestion
		agent.AnnotationLLMHint:   "gcx datasources influxdb field-keys -d UID",
```

internal/query/influxdb/client.go
```diff
@@ -0,0 +1,390 @@
...
+	"net/http"
+	"sort"
+	"strconv"
+
+	"github.com/grafana/gcx/internal/config"
+	"github.com/grafana/gcx/internal/queryerror"
+	"k8s.io/client-go/rest"
+)
+
+const maxResponseBytes = 50 << 20 // 50 MB
```
dafydd-t, line 23: Do we surface this anywhere, and is it obvious that this is the cause of a failure when we try and read a response that's too big?

internal/query/influxdb/types.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// FieldKey represents a single field key with its type.
+type FieldKey struct {
+	FieldKey  string `json:"fieldKey"`
+	FieldType string `json:"fieldType"`
+}
+
+// GrafanaQueryResponse represents the response from Grafana's datasource query API.
+type GrafanaQueryResponse struct {
```
dafydd-t, line 68:
```suggestion
type QueryResponse struct {
```

Maybe?

dafydd-t: ah no i see, we have both

docs/reference/cli/gcx_datasources_influxdb_field-keys.md
```diff
@@ -0,0 +1,52 @@
+## gcx datasources influxdb field-keys
+
+List field keys
```
dafydd-t, line 3: should we also add tag-keys / tag-values?

internal/datasources/influxdb/fieldkeys.go
mattvella07: That makes sense

docs/reference/cli/gcx_datasources_influxdb_field-keys.md
mattvella07: Added

wardbekker (approved):
Smoke-tested all five subcommands (measurements, field-keys, tag-keys, tag-values, query) against play.grafana.org's InfluxCloudQL datasource — all return real data with the expected schemas. All inline review feedback from previous rounds appears addressed. LGTM 🎉

javiermolinar (approved):
thank you for your contribution

PR #544, opened 2026-04-22
fix(irm): pass --max-age filter to OnCall backend

oleg-kozlyuk-grafana:
## Summary

- The `gcx irm oncall alert-groups list --max-age` flag was fetching **all** alert groups and filtering client-side. Now the duration is converted to a `started_after` ISO8601 timestamp and sent as a query parameter, letting the backend handle the filtering.
- Extends `ListConfig`/`ListOption` with `WithStartedAfter(time.Time)` — both the internal API client and the public API client forward the parameter.
- Removes the `filterAlertGroupsByMaxAge` client-side filter function.

## Test plan

- [x] New unit tests verify the query param is sent when `WithStartedAfter` is provided
- [x] New unit tests verify no query param is sent by default (backward-compat)
- [x] Existing pagination and limit tests pass unchanged
- [x] Full test suite passes (`go test ./...`)
- [x] Lint clean (`make lint` — 0 issues)
- [ ] Manual: `gcx irm oncall alert-groups list --max-age 24h` against a live stack

### DIFF
--- a/internal/providers/irm/oncall_client.go
+++ b/internal/providers/irm/oncall_client.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/grafana/gcx/internal/config"
 	"github.com/grafana/gcx/internal/providers/irm/oncalltypes"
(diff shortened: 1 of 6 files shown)


javiermolinar (approved):
LGMT

PR #553, opened 2026-04-22
feat(config): add mTLS client certificate authentication support (Auth via Teleport)

aposnov:
## Summary

- Add `cert-file`, `key-file`, and `ca-file` fields to the TLS config struct, enabling mutual TLS (mTLS) authentication with file-based client certificates
- Implement the existing `// TODO: CertData, KeyData, CAData` in `ToStdTLSConfig()` to actually load and use client certificates and custom CA bundles
- Add environment variables `GRAFANA_TLS_CERT_FILE`, `GRAFANA_TLS_KEY_FILE`, `GRAFANA_TLS_CA_FILE` for CI/CD and agent usage

## Motivation

Many enterprise Grafana deployments sit behind identity-aware proxies like [Teleport](https://goteleport.com/), which authenticate users via mTLS client certificates rather than API tokens. The Grafana MCP server already supports this via `--tls-cert-file` / `--tls-key-file` flags, but gcx had no equivalent — forcing users to either create service account tokens (which may require Admin permissions they don't have) or use workarounds like local proxies.

This change closes that gap by letting users point gcx directly at their client certificate and key files.

### Via config

```bash
gcx config set contexts.myctx.grafana.tls.cert-file /path/to/cert.pem
gcx config set contexts.myctx.grafana.tls.key-file /path/to/key.pem
gcx config set contexts.myctx.grafana.tls.ca-file /path/to/ca.pem  # optional
```

### Via environment variables

```bash
export GRAFANA_TLS_CERT_FILE=/path/to/cert.pem
export GRAFANA_TLS_KEY_FILE=/path/to/key.pem
export GRAFANA_TLS_CA_FILE=/path/to/ca.pem  # optional
```

### Example: Teleport integration

```bash
tsh apps login grafana
gcx config set contexts.prod.grafana.server https://grafana.teleport.example.com
gcx config set contexts.prod.grafana.tls.cert-file $(tsh apps config grafana -f cert)
gcx config set contexts.prod.grafana.tls.key-file $(tsh apps config grafana -f key)
gcx config use-context prod
gcx api /api/org  # works via mTLS, no token needed
```

## What changed

| File | Change |
|------|--------|
| `internal/config/types.go` | Added `CertFile`, `KeyFile`, `CAFile` fields to `TLS` struct; added `ResolveFiles()` method; implemented `ToStdTLSConfig()` with cert loading (was `// TODO`) |
| `internal/config/rest.go` | Call `ResolveFiles()` before passing TLS data to k8s REST client |
| `internal/config/stack_id.go` | Handle new `ToStdTLSConfig()` error return |
| `internal/grafana/client.go` | Handle new `ToStdTLSConfig()` error return |
| `internal/server/server.go` | Handle new `ToStdTLSConfig()` error return |
| `internal/server/grafana/requests.go` | Handle new `ToStdTLSConfig()` error return |
| `internal/config/tls_test.go` | **New** — 8 unit tests for file resolution, cert loading, CA parsing, error cases |

## Test
Tested, works! :)

### DIFF
--- a/internal/config/rest.go
+++ b/internal/config/rest.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"log/slog"
 	"net/http"
 	"strings"
 	"time"
@@ -204,6 +205,10 @@ func NewNamespacedRESTConfig(ctx context.Context, cfg Context) NamespacedRESTCon
 	}
 
 	if cfg.Grafana.TLS != nil {
+		// Resolve file paths to data before passing to the k8s REST client.
+		if err := cfg.Grafana.TLS.ResolveFiles(); err != nil {
+			slog.WarnContext(ctx, "failed to resolve TLS files", "error", err)
+		}
 		// Kubernetes really is wonderful, huh.
 		// tl;dr it has its own TLSClientConfig,
 		// and it's not compatible with the one from the "crypto/tls" package.
--- a/internal/config/rest.go
+++ b/internal/config/rest.go
@@ -2,7 +2,7 @@ package config
 
 import (
 	"context"
-	"log/slog"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
@@ -193,7 +193,7 @@ func parseRFC3339OrZero(s string) time.Time {
 }
 
 // NewNamespacedRESTConfig creates a new namespaced REST config.
-func NewNamespacedRESTConfig(ctx context.Context, cfg Context) NamespacedRESTConfig {
+func NewNamespacedRESTConfig(ctx context.Context, cfg Context) (NamespacedRESTConfig, error) {
 	rcfg := rest.Config{
 		UserAgent:       version.UserAgent(),
 		Host:            strings.TrimSuffix(cfg.Grafana.Server, "/"),
@@ -207,7 +207,7 @@ func NewNamespacedRESTConfig(ctx context.Context, cfg Context) NamespacedRESTCon
 	if cfg.Grafana.TLS != nil {
 		// Resolve file paths to data before passing to the k8s REST client.
 		if err := cfg.Grafana.TLS.ResolveFiles(); err != nil {
-			slog.WarnContext(ctx, "failed to resolve TLS files", "error", err)
+			return NamespacedRESTConfig{}, fmt.Errorf("TLS configuration: %w", err)
 		}
 		// Kubernetes really is wonderful, huh.
 		// tl;dr it has its own TLSClientConfig,
@@ -291,5 +291,5 @@ func NewNamespacedRESTConfig(ctx context.Context, cfg Context) NamespacedRESTCon
 		Namespace:      namespace,
 		GrafanaURL:     strings.TrimSuffix(cfg.Grafana.Server, "/"),
 		oauthTransport: oauthTransport,
-	}
+	}, nil
 }
--- a/internal/config/stack_id.go
+++ b/internal/config/stack_id.go
@@ -23,7 +23,10 @@ func DiscoverStackID(ctx context.Context, cfg GrafanaConfig) (int64, error) {
 		return 0, err
 	}
 
-	client := newBootdataHTTPClient(cfg)
+	client, err := newBootdataHTTPClient(cfg)
+	if err != nil {
+		return 0, err
+	}
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, bootdataURL.String(), nil)
 	if err != nil {
@@ -81,7 +84,7 @@ func buildBootdataURL(server string) (*url.URL, error) {
 	return parsed, nil
 }
 
-func newBootdataHTTPClient(cfg GrafanaConfig) *http.Client {
+func newBootdataHTTPClient(cfg GrafanaConfig) (*http.Client, error) {
 	transport := &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 	}
@@ -89,7 +92,7 @@ func newBootdataHTTPClient(cfg GrafanaConfig) *http.Client {
 	if cfg.TLS != nil {
 		tlsCfg, err := cfg.TLS.ToStdTLSConfig()
 		if err != nil {
-			return &http.Client{Timeout: 5 * time.Second, Transport: transport}
+			return nil, fmt.Errorf("TLS configuration: %w", err)
 		}
 		transport.TLSClientConfig = tlsCfg
 	}
@@ -97,5 +100,5 @@ func newBootdataHTTPClient(cfg GrafanaConfig) *http.Client {
 	return &http.Client{
 		Timeout:   5 * time.Second,
 		Transport: transport,
-	}
+	}, nil
 }
--- a/internal/config/stack_id.go
+++ b/internal/config/stack_id.go
@@ -87,7 +87,11 @@ func newBootdataHTTPClient(cfg GrafanaConfig) *http.Client {
 	}
 
 	if cfg.TLS != nil {
-		transport.TLSClientConfig = cfg.TLS.ToStdTLSConfig()
+		tlsCfg, err := cfg.TLS.ToStdTLSConfig()
+		if err != nil {
+			return &http.Client{Timeout: 5 * time.Second, Transport: transport}
+		}
+		transport.TLSClientConfig = tlsCfg
 	}
 
 	return &http.Client{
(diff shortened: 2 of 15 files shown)


internal/config/rest.go
```diff
@@ -204,6 +205,10 @@ func NewNamespacedRESTConfig(ctx context.Context, cfg Context) NamespacedRESTCon
 	}
 
 	if cfg.Grafana.TLS != nil {
+		// Resolve file paths to data before passing to the k8s REST client.
+		if err := cfg.Grafana.TLS.ResolveFiles(); err != nil {
+			slog.WarnContext(ctx, "failed to resolve TLS files", "error", err)
```
javiermolinar: Should we return the error instead?

internal/config/stack_id.go
```diff
@@ -87,7 +87,11 @@ func newBootdataHTTPClient(cfg GrafanaConfig) *http.Client {
 	}
 
 	if cfg.TLS != nil {
-		transport.TLSClientConfig = cfg.TLS.ToStdTLSConfig()
+		tlsCfg, err := cfg.TLS.ToStdTLSConfig()
```
javiermolinar, line 93: We should propage the error here as well instead of fallback to a plain http client

internal/config/types.go
```diff
@@ -334,6 +336,17 @@ type TLS struct {
 	// server is used.
 	ServerName string `json:"server-name,omitempty" yaml:"server-name,omitempty"`
 
+	// CertFile is the path to a PEM-encoded client certificate file.
+	// When set, KeyFile must also be provided.
```
javiermolinar: This is not enforced in the code as far as I can see

javiermolinar (commented):
I have a couple of comments

aposnov:
> I have a couple of comments

Thanks for the review @javiermolinar ! All three comments addressed in the latest push:
  1. rest.go - NewNamespacedRESTConfig now returns (NamespacedRESTConfig, error) and
  propagates the TLS error to the caller
  2. stack_id.go - newBootdataHTTPClient now returns (*http.Client, error) and propagates the
   error instead of falling back to a plain client
  3. types.go - Added validation in ResolveFiles(): "both cert-file and key-file must be
  provided together"

  Also fixed CI:
  - Linter: fmt.Errorf → errors.New (no format args)
  - Documentation: regenerated env-var and config reference docs
  - Rebased on latest main and updated the new login.go caller for the changed signature

internal/config/types.go
aposnov: solved

internal/config/stack_id.go
aposnov: solved

internal/config/rest.go
aposnov: solved

javiermolinar:
@codex review

javiermolinar:
@aposnov we are close! Codex found a couple of issues that I think its worth to patch

internal/config/types.go
```diff
@@ -410,6 +412,15 @@ type TLS struct {
 	// server is used.
 	ServerName string `json:"server-name,omitempty" yaml:"server-name,omitempty"`
 
+	// CertFile is the path to a PEM-encoded client certificate file.
+	// This enables mutual TLS (mTLS) authentication with the server.
+	CertFile string `env:"GRAFANA_TLS_CERT_FILE" json:"cert-file,omitempty" yaml:"cert-file,omitempty"`
+	// KeyFile is the path to a PEM-encoded client certificate key file.
+	KeyFile string `datapolicy:"secret" env:"GRAFANA_TLS_KEY_FILE" json:"key-file,omitempty" yaml:"key-file,omitempty"`
```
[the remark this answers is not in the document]
javiermolinar: Good catch

internal/config/types.go
```diff
@@ -430,14 +441,72 @@ type TLS struct {
...
+
+	tlsCfg := &tls.Config{
 		//nolint:gosec
 		InsecureSkipVerify: cfg.Insecure,
 		ServerName:         cfg.ServerName,
 		NextProtos:         cfg.NextProtos,
 	}
+
+	if len(cfg.CertData) > 0 && len(cfg.KeyData) > 0 {
+		cert, err := tls.X509KeyPair(cfg.CertData, cfg.KeyData)
```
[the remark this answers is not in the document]
andreyposnov: sure, solved

internal/config/types.go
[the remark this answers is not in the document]
andreyposnov: sure, solved

wardbekker:
Hi @aposnov — _edit: trimmed this after re-checking git blame; some of what I flagged was pre-existing code, not yours. Keeping only the findings that are on lines this PR adds, plus a scope note._

Beyond the two Codex findings @javiermolinar flagged:

### P2 — `ToStdTLSConfig` doesn't pin `MinVersion: TLS 1.2`

`internal/config/types.go:478` returns a `tls.Config` with no `MinVersion`, so Go's default (TLS 1.0) applies in every consumer that uses it. Compare `internal/httputils/client.go:75`, which pins `MinVersion: tls.VersionTLS12`. Fix: set `MinVersion: tls.VersionTLS12` on the returned config.

### P2 — `ToStdTLSConfig` silently discards system roots on `SystemCertPool` error

`internal/config/types.go:498–506`:

```go
pool, err := x509.SystemCertPool()
if err != nil {
    pool = x509.NewCertPool()
}
```

On platforms where `SystemCertPool` can error (Windows, some sandboxes), this falls back to an empty pool, so the returned config verifies _only_ user‑provided `CAData`. A user who adds a corporate CA would lose the ability to verify public roots for the same host. Fix: return the error, or explicitly document that `CAData` replaces system roots (which is probably not the intent — the doc comment says "adds").

### Scope note (not asking you to fix in this PR) — mTLS isn't engaged on the `gcx login` path

`internal/login/login.go` `resolveGrafanaAuth` builds a fresh `&config.GrafanaConfig{Server: opts.Server}` with no TLS; `detectTarget` and `FetchAnonymousSettings` use `httputils.NewDefaultClient` which has no hook for a `*tls.Config`. So after this PR users can use mTLS for push/pull (via `NewNamespacedRESTConfig`), but `gcx login` against an mTLS server still won't work, and a re‑login would overwrite a manually‑set `grafana.tls.*` block. Worth tracking as a follow‑up rather than piling onto this PR.

### Suggested tests

- `ToStdTLSConfig` errors on half‑configured `CertData`/`KeyData` (covers Codex P2).
- `ToStdTLSConfig` pins `MinVersion == tls.VersionTLS12`.
- `CAData` is added to system roots, not replacing them.
- Round‑trip: `NewNamespacedRESTConfig` propagates `CertData`/`KeyData`/`CAData` into `rest.TLSClientConfig`.

Sorry for the earlier over‑reach — nice feature.

andreyposnov:
> ### Scope note (not asking you to fix in this PR) — mTLS isn't engaged on the `gcx login` path
> `internal/login/login.go` `resolveGrafanaAuth` builds a fresh `&config.GrafanaConfig{Server: opts.Server}` with no TLS; `detectTarget` and `FetchAnonymousSettings` use `httputils.NewDefaultClient` which has no hook for a `*tls.Config`. So after this PR users can use mTLS for push/pull (via `NewNamespacedRESTConfig`), but `gcx login` against an mTLS server still won't work, and a re‑login would overwrite a manually‑set `grafana.tls.*` block. Worth tracking as a follow‑up rather than piling onto this PR.

 @wardbekker Thanks for the thorough review! All addressed in the latest commit:

P2 - MinVersion TLS 1.2: Fixed. ToStdTLSConfig now pins MinVersion: tls.VersionTLS12, consistent with
  httputils/client.go:75.

P2 - SystemCertPool error: Fixed. Now returns the error instead of silently falling back to an empty pool.
  CAData is added to system roots as documented.

Codex P1 - TLS pointer for env vars: Fixed. Grafana.TLS is initialized before env.Parse() so
  GRAFANA_TLS_CERT_FILE/GRAFANA_TLS_KEY_FILE work in env-only/CI setups. Nil'd back out after parsing if no
  fields were set.

 Codex P2 - Half-configured CertData/KeyData: Fixed. ToStdTLSConfig now fails fast with "both cert-data and
  key-data must be provided together".

  Tests added:
  - TestTLS_ToStdTLSConfig_HalfConfiguredCertData
  - TestTLS_ToStdTLSConfig_PinsMinVersionTLS12
  - TestTLS_ToStdTLSConfig_CADataAddsToSystemRoots

  Scope note (login path): Agreed this is a follow-up - i can open a separate PR to track mTLS support on
  the gcx login path, after this one will be merged

javiermolinar (approved):
LGTM thank you for your contribution!

PR #554, opened 2026-04-23
feat(alert): add --datasource and --name flags to instances list

oleg-kozlyuk-grafana:
## Summary

- **`--datasource <uid>`**: queries alert rules from a Prometheus datasource's ruler API instead of only Grafana-managed rules. This makes Mimir ruler alerts (Tempo, Mimir, Loki, Pyroscope) visible to `gcx alert instances list`.
- **`--name <regex>`**: case-insensitive regex filter on rule name (e.g. `--name Tempo` to show only Tempo alerts).

Example:
```
gcx alert instances list --datasource grafanacloud-prom --state firing --name Tempo
```

## Motivation

Product team alerts are evaluated by the Mimir ruler and served through Prometheus datasources at `/api/prometheus/{uid}/api/v1/rules`. The command previously only queried `/api/prometheus/grafana/api/v1/rules` (Grafana-managed rules), making the majority of infrastructure alerts invisible.

## Test plan

- [x] `go build ./cmd/gcx/` compiles cleanly
- [x] `go test ./internal/providers/alert/...` passes
- [x] Verified default behavior (no `--datasource`) still works against Grafana-managed rules
- [x] Verified `--datasource grafanacloud-prom --state firing --name Tempo` returns Tempo Mimir ruler alerts
- [x] Verified `--name` regex is case-insensitive and filters correctly

### DIFF
--- a/docs/reference/cli/gcx_alert_instances_list.md
+++ b/docs/reference/cli/gcx_alert_instances_list.md
@@ -9,13 +9,15 @@ gcx alert instances list [flags]
 ### Options
 
 ```
-      --folder string   Filter by folder UID
-      --group string    Filter by group name
-  -h, --help            help for list
-      --json string     Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
-  -o, --output string   Output format. One of: json, table, wide, yaml (default "table")
-      --rule string     Filter by rule UID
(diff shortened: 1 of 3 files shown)


javiermolinar (approved):
LGTM

PR #561, opened 2026-04-24
fix: re-enable grafana-com instance selector

theSuess:
#529 removed functionality introduced in #481

This PR adds back this functionality by launching the interactive instance selector when no server is provided

### DIFF
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -263,12 +263,16 @@ func askForInput(e *login.ErrNeedInput, opts *login.Options, sourceCtx *config.C
 	for _, field := range e.Fields {
 		switch field {
 		case "server":
+			description := "e.g. https://my-stack.grafana.net"
+			if opts.GrafanaToken == "" {
+				description += "\nLeave empty to select your Grafana Cloud instance interactively"
+			}
 			form := huh.NewForm(huh.NewGroup(
 				huh.NewInput().
 					Title("Grafana server URL").
-					Description("e.g. https://my-stack.grafana.net").
+					Description(description).
 					Validate(func(s string) error {
-						if s == "" {
+						if opts.GrafanaToken != "" && s == "" {
 							return errors.New("server URL is required")
 						}
 						return nil
@@ -277,6 +281,10 @@ func askForInput(e *login.ErrNeedInput, opts *login.Options, sourceCtx *config.C
 			))
 			if err := form.Run(); err != nil {
 				return err
+			}
+			if opts.Server == "" {
+				opts.UseCloudInstanceSelector = true
+				return nil
 			}
 
 		case "grafana-auth":
--- a/docs/architecture/login-system.md
+++ b/docs/architecture/login-system.md
@@ -46,6 +46,7 @@ classDiagram
         Target, GrafanaToken
         CloudToken, CloudAPIURL
         UseOAuth, Yes, Writer
+        UseCloudInstanceSelector
     }
     class Hooks {
         ConfigSource
@@ -109,8 +110,7 @@ flowchart TD
     Start([Run entry]) --> Server{Server set?}
     Server -->|no| SrvSentinel[Return ErrNeedInput&#123;server&#125;]
     Server -->|yes| Scheme[Normalize scheme:<br/>default https]
-    Scheme --> Ctx[Derive context name]
-    Ctx --> Detect[Detect target]
+    Scheme --> Detect[Detect target]
     Detect --> TgtKnown{Target resolved?}
     TgtKnown -->|no, interactive| TgtSentinel[Return ErrNeedClarification&#123;target&#125;]
     TgtKnown -->|no, --yes or agent| TgtOnPrem[Force TargetOnPrem]
@@ -118,7 +118,8 @@ flowchart TD
     TgtOnPrem --> Auth
     Auth --> AuthHave{Token or OAuth?}
     AuthHave -->|no| AuthSentinel[Return ErrNeedInput&#123;grafana-auth&#125;]
-    AuthHave -->|yes| Cloud[Resolve cloud auth]
+    AuthHave -->|yes| Ctx[Derive context name]
+    Ctx --> Cloud[Resolve cloud auth]
     Cloud --> CloudNeeded{Cloud target<br/>+ no token?}
     CloudNeeded -->|yes, interactive| CloudSentinel[Return ErrNeedInput&#123;cloud-token&#125;<br/>Optional=true]
     CloudNeeded -->|no or skip| Validate[Validation pipeline]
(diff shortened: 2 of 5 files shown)


wardbekker:
Can you add test coverage for this functionality to prevent future regressions?

theSuess:
@wardbekker addressed the automated review findings & added a test case that checks whether the oauth flow gets called when `UseCloudInstanceSelector` is set

internal/login/login.go
```diff
@@ -338,6 +344,9 @@ func resolveGrafanaAuth(ctx context.Context, opts Options, target Target) (strin
 		if err != nil {
 			return "", nil, fmt.Errorf("OAuth flow failed: %w", err)
 		}
+		if grafanaCfg.Server == "" {
```
radiohead, line 357: **issue (blocking):** This is the first place `result.InstanceEndpoint` is promoted to a persisted server URL. The upstream validation in `flow.go` uses only `url.Parse(instanceEndpoint)` — no scheme check — so an attacker-controlled value (e.g. via a browser extension intercepting the `127.0.0.1` callback) can inject an `http://` URL that gets written to the user's config. Domain-allowlist validation is not viable here because custom-domain instances are legitimate, but enforcing HTTPS is the right minimal guard. The fix belongs in the callback handler in `flow.go:227-233`:

```go
u, err := url.Parse(instanceEndpoint)
if err != nil || u.Scheme != "https" {
    errCh <- fmt.Errorf("invalid instance endpoint: must use https")
    renderErrorPage(w, "Invalid instance endpoint passed")
    return
}
```

cmd/gcx/login/command.go
```diff
@@ -266,18 +266,18 @@ func askForInput(e *login.ErrNeedInput, opts *login.Options, sourceCtx *config.C
...
-						}
-						return nil
-					}).
+					Description("e.g. https://my-stack.grafana.net\nLeave empty to select your Grafana Cloud instance interactively").
 					Value(&opts.Server),
 			))
 			if err := form.Run(); err != nil {
 				return err
 			}
+			if opts.Server == "" {
```
radiohead, line 285: **issue (blocking):** When `--token` is supplied alongside an empty server, this block sets `UseOAuth = true` and returns early. `resolveGrafanaAuth` then takes the token branch (not OAuth), and `grafanaCfg.Server` stays `""` — the context is persisted with an empty server URL. Before this PR, an empty server with `--token` returned `ErrNeedInput{server}` and re-prompted. Gate the cloud-selector path on no token being present:

```go
if opts.Server == "" && opts.GrafanaToken == "" {
    opts.UseOAuth = true
    opts.Target = login.TargetCloud
    opts.UseCloudInstanceSelector = true
    return nil
}
```

internal/login/login.go
```diff
@@ -224,12 +219,23 @@ func Run(ctx context.Context, opts *Options) (Result, error) {
...
 
-	// Step 4: Grafana auth
+	// Step 3: Grafana auth
 	authMethod, grafanaCfg, err := resolveGrafanaAuth(ctx, *opts, target)
 	if err != nil {
 		return Result{}, err
 	}
 
+	// set the server if the user used the interactive instance selector
+	if opts.Server == "" {
```
radiohead, line 236: **suggestion (blocking):** If OAuth returns an empty `InstanceEndpoint`, `grafanaCfg.Server` stays `""` and `opts.Server` is never repaired. Execution continues into `ContextNameFromServerURL("")`, REST config build, and persistence — where `mergeAuthIntoExisting` can silently overwrite an existing context's `Server` with `""`. An explicit guard gives a precise error before any of that:

```go
if opts.Server == "" {
    if grafanaCfg.Server == "" {
        return Result{}, fmt.Errorf("OAuth did not return an instance endpoint")
    }
    opts.Server = grafanaCfg.Server
}
```

internal/login/login.go
```diff
@@ -179,24 +180,18 @@ type AuthFlow interface {
 // the CLI sentinel-retry flow. Callers that retry after ErrNeedInput /
 // ErrNeedClarification should reuse the same Options value.
 func Run(ctx context.Context, opts *Options) (Result, error) {
-	// Step 1: server must be set
-	if opts.Server == "" {
+	// Step 1: check if the server is set
+	if opts.Server == "" && !opts.UseCloudInstanceSelector {
```
radiohead, line 187: **suggestion (non-blocking):** `UseCloudInstanceSelector = true` implicitly requires `Target = TargetCloud` and `UseOAuth = true`. The CLI sets all three together, but a future caller setting only the flag would send `detectTarget` an empty `opts.Server`. Normalizing at the top of `Run()` makes the field safe to set in isolation:

```go
if opts.UseCloudInstanceSelector {
    opts.Target = TargetCloud
    opts.UseOAuth = true
}
```

internal/login/login.go
```diff
@@ -28,14 +28,15 @@ const (
...
-	Yes          bool
+	Server                   string
+	ContextName              string
+	Target                   Target
+	GrafanaToken             string
+	CloudToken               string
+	CloudAPIURL              string
+	UseOAuth                 bool
+	Yes                      bool
+	UseCloudInstanceSelector bool
```
radiohead, line 42: **nitpick (non-blocking):** `UseCloudInstanceSelector` has no CLI flag and is only ever set by the interactive prompt handler, but it is exported on the user-facing `Inputs` struct with no comment. A future maintainer might wire a `--use-cloud-instance-selector` flag not realizing it bypasses the server-present check.

```go
// UseCloudInstanceSelector is an internal sentinel set only by the interactive prompt handler.
UseCloudInstanceSelector bool
```

docs/architecture/login-system.md
```diff
@@ -135,25 +136,25 @@ The pipeline reads top-to-bottom in `Run()` (login.go:180). Each step returns
 early on failure; sentinel branches unwind to the CLI for interactive
```
radiohead, line 136: **suggestion (non-blocking):** The flowchart above (line 113) still shows `Scheme --> Ctx[Derive context name] --> Detect[Detect target]`, placing context-name derivation before target detection and auth. This PR moves it after Grafana auth resolution — the numbered list below was updated correctly, but the diagram needs the same treatment: `Ctx[Derive context name]` should sit between `Auth` and `Cloud`.

docs/architecture/login-system.md
```diff
@@ -135,25 +136,25 @@ The pipeline reads top-to-bottom in `Run()` (login.go:180). Each step returns
...
    yields `ErrNeedInput{grafana-auth}`.
-6. **Cloud auth resolution** (`resolveCloudAuth`, login.go:373). Only runs
+5. **Context-name derivation** (login.go:234). Falls back to
+   `config.ContextNameFromServerURL`, which returns the stack slug for known
+   Grafana Cloud URLs and a hyphenated hostname otherwise.
+6. **Cloud auth resolution** (`resolveCloudAuth`, login.go:383). Only runs
    for `TargetCloud`; a missing token yields an optional
    `ErrNeedInput{cloud-token}` that the CLI may skip.
-7. **Validation** (login.go:246). Delegated to `Validate` (see below); the
+7. **Validation** (login.go:245). Delegated to `Validate` (see below); the
```
radiohead, line 155: **nitpick (non-blocking):** `login.go:245` points at `tempCtx := config.Context{…}` (the temp context build), not at the `validateFn` call (~line 258). Worth a quick sweep of the remaining step references while the doc is open.

radiohead (changes requested):

The pipeline reorder is correct and the happy path is well-tested. A few edge cases in the new empty-server path need addressing — one is a behavioral regression when `--token` is combined with an empty server entry.

cmd/gcx/login/command.go
theSuess: Resolved by conditionally applying the validation function instead, giving the user feedback on the fact that a URL is required

internal/login/login.go
theSuess: Since the `instanceEndpoint` change is live everywhere now, I enforced validation instead, so this will never be empty, leaving us with one less case to handle

docs/architecture/login-system.md
theSuess: Honest question: Do line numbers in here help agents even though the same flow is documented in the file as well? Updating this seems like more of a burden than actually useful

internal/login/login.go
theSuess: There is no advantage in checking for https here. If an attacker were to override this url, they could point it to an arbitrary https server as well and extract the token that way which is simpler and has more impact than relying on http MITM to extract secrets that way

radiohead: Is there a _disadvantage_ to having an HTTPS check?

internal/login/login.go
radiohead: Thanks!

docs/architecture/login-system.md
radiohead: That's a very good point. I agree – these are not very useful, let's ignore this for now – I'll clean up the file and get rid of the unnecessary details.

radiohead (approved):

radiohead:
@theSuess LGTM, thank you! I've approved the PR but I'd love to hear your thoughts on https://github.com/grafana/gcx/pull/561#discussion_r3147921671, so I won't merge just yet. If you think we don't need HTTPS validation, feel free to merge this yourself.

internal/login/login.go
theSuess: The only case I could think of is BYOC with a non-https URL, but I don't think this is something we'd support anyway. I'll add the check as we can always remove it if it causes issues

radiohead: Thanks! Yeah I was thinking we should check for HTTPS just for the reasons of good hygiene – there shouldn't be anything exposed on HTTP.

theSuess:
Added the https check to be on the safe side, thanks for taking the time to review this!

PR #563, opened 2026-04-24
fix: suggest required scopes on adaptive command auth errors

dafydd-t:
## Summary

When adaptive commands fail with scope-related auth errors, the error messages now tell the user exactly which access policy scope they need to add. Previously users had to guess or look up documentation.

### Scope suggestions on "invalid scope" errors

Each adaptive provider now suggests the specific scope(s) needed:

- **Adaptive Logs** → `adaptive-logs:admin`
- **Adaptive Traces** → `adaptive-traces:admin`
- **Adaptive Metrics** → the specific scope for the subcommand (e.g. `adaptive-metrics-rules:read`, `adaptive-metrics-segments:write`, `adaptive-metrics-exemptions:delete`)

Examples:
```
# no stacks:read scope
gcx metrics adaptive exemptions list
Error: Cloud stack lookup: permission denied
│
├─ Details:
│
│ adaptive-metrics: failed to load cloud config for token: failed to get stack info for "dafyddt": gcom client: unexpected status 403 Forbidden: {
│   "code": "Forbidden",
│   "message": "You do not have permission to perform the requested action.",
│   "requestId": "fd9b090d-93a1-49fd-a513-c643a19fdc32"
│ }
│
├─ Suggestions:
│
│ • Ensure your access policy includes the stacks:read scope
│ • Adaptive Metrics commands also require an adaptive-metrics-* scope (the specific scope depends on the subcommand)
│
└─


# with stacks:read, missing adaptive-logs:admin
./bin/gcx logs adaptive patterns show
Error: Adaptive Logs: permission denied
│
├─ Details:
│
│ API request failed (HTTP 401): authentication error: invalid scope requested
│
├─ Suggestions:
│
│ • Ensure your access policy includes the adaptive-logs:admin scope
│
└─

# adaptive metrics writes require the write scope
./bin/gcx metrics adaptive rules create  --metric hello
Error: Adaptive Metrics: permission denied
│
├─ Details:
│
│ metrics: create rule: status 401: {"status":"error","error":"authentication error: invalid scope requested"}
│
├─ Suggestions:
│
│ • Ensure your access policy includes the adaptive-metrics-rules:write scope
│
└─
```

### Supporting changes

- **`internal/auth/adaptive/auth.go`**: Error prefix changed from `adaptive:` to `adaptive-<signal>:` (e.g. `adaptive-logs:`, `adaptive-metrics:`) so the fail converter can identify which product's scope to suggest.
- **`internal/providers/traces/adaptive/client.go`**: Wrapped `handleErrorResponse` calls with a `traces:` prefix so the fail converter can identify traces errors (previously returned bare "unexpected status" errors with no provider context).

### DIFF
--- a/cmd/gcx/fail/convert.go
+++ b/cmd/gcx/fail/convert.go
@@ -889,35 +889,32 @@ func adaptiveScopeSuggestionFromSignalPrefix(msg string) string {
 	}
 }
 
-type adaptiveMetricsResource struct {
-	keyword   string
-	base      string
-	reads     []string
-	writes    []string
-	deleteKey string
-}
-
-// adaptiveMetricsResources maps error message keywords to scope prefixes.
-// Operation matches are checked in priority order: delete > write > read.
-var adaptiveMetricsResources = []adaptiveMetricsResource{
-	{"rule", "adaptive-metrics-rules",
-		[]string{"list rules", "get rule", "list recommended rules"},
-		[]string{"create rule", "update rule", "sync rules", "validate rules"},
-		"delete rule"},
-	{"recommendation", "adaptive-metrics-recommendations",
-		[]string{"list recommendations"}, nil, ""},
-	{"segment", "adaptive-metrics-segments",
-		[]string{"list segments"},
-		[]string{"create segment", "update segment"},
-		"delete segment"},
-	{"exemption", "adaptive-metrics-exemptions",
-		[]string{"list exemptions", "list segmented exemptions", "get exemption"},
-		[]string{"create exemption", "update exemption"},
-		"delete exemption"},
-}
-
 func adaptiveMetricsScopeFromError(msg string) string {
-	for _, r := range adaptiveMetricsResources {
+	type resource struct {
+		keyword   string
+		base      string
+		reads     []string
+		writes    []string
+		deleteKey string
+	}
+	// Operation matches are checked in priority order: delete > write > read.
+	resources := []resource{
+		{"rule", "adaptive-metrics-rules",
+			[]string{"list rules", "get rule", "list recommended rules"},
+			[]string{"create rule", "update rule", "sync rules", "validate rules"},
+			"delete rule"},
+		{"recommendation", "adaptive-metrics-recommendations",
+			[]string{"list recommendations"}, nil, ""},
+		{"segment", "adaptive-metrics-segments",
+			[]string{"list segments"},
+			[]string{"create segment", "update segment"},
+			"delete segment"},
+		{"exemption", "adaptive-metrics-exemptions",
+			[]string{"list exemptions", "list segmented exemptions", "get exemption"},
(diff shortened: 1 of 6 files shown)


cmd/gcx/fail/convert.go
```diff
@@ -790,6 +806,32 @@ func convertCloudConfigErrors(err error) (*DetailedError, bool) {
...
+				Summary: "Adaptive Metrics: permission denied",
+				Suggestions: []string{
+					fmt.Sprintf("Ensure your access policy includes the %s scope", scope),
+				},
+				ExitCode: new(ExitAuthFailure),
+			}, true
+		}
+	}
+
 	// Fleet management not available.
```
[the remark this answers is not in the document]
dafydd-t: addressed in [513ac2e](https://github.com/grafana/gcx/pull/563/commits/513ac2e4fc210503cc05e0cc7f6ddfde8ea8b0b8)

internal/providers/traces/adaptive/client.go
```diff
@@ -39,7 +39,7 @@ func (c *Client) ListPolicies(ctx context.Context) ([]Policy, error) {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, handleErrorResponse(resp)
+		return nil, fmt.Errorf("traces: list policies: %w", handleErrorResponse(resp))
```
dafydd-t: We need to wrap these responses with `traces: *` so that we can pick that up in the suggestions logic and know which scopes to suggest

cmd/gcx/fail/convert.go
```diff
@@ -792,6 +806,33 @@ func convertCloudConfigErrors(err error) (*DetailedError, bool) {
 		}, true
 	}
 
+	// Adaptive Traces scope errors.
+	if strings.Contains(msg, "traces:") && strings.Contains(msg, "invalid scope") {
```
annanay25: Could we prefix Adaptive Traces API errors with `adaptive-traces:` instead of `traces:` and make the converter check that exact prefix?

The current `strings.Contains(msg, "traces:") && strings.Contains(msg, "invalid scope")` probably doesn’t catch normal `gcx traces query` Tempo errors today because those use typed `queryerror.APIError`, but `traces:` is broad enough that future non-adaptive traces errors could be misclassified as Adaptive Traces permission failures. Using `adaptive-traces:` would align with the new `adaptive-%s:` cloud-config prefix and make the scope-suggestion check unambiguous.

Similar comment for the "metrics:" check below.

dafydd-t: Yes good idea, I'll take a look.

dafydd-t: Applied this for metrics and traces (logs are unaffected). I also changed the suggestions to be explicit that we're talking about gcom APs, not the grafana SA roles or something.

radiohead (approved):

PR #568, opened 2026-04-24
fix(metrics): make default view usable

javiermolinar:
Current output of a metric query is a table with one columnn per dimension. Making impossible to understand what's happening. ie:

Screenshot 2026-04-24 at 17 27 35

This pr make Prometheus query table output compact by default with VALUE, TIMESTAMP, and SERIES columns

Screenshot 2026-04-24 at 17 27 13


radiohead:
@javiermolinar nice one! Do you think we should switch default outputs for anything that produces a graph to `-ograph` instead?

javiermolinar:
> @javiermolinar nice one! Do you think we should switch default outputs for anything that produces a graph to `-ograph` instead?

I think its tricky, the graph has its own problem. The representation its not as detailed as grafana and when the number of series is high there the result is difficult to read it and you cannot select single series.

It also conflicts with the DESIGN.md and docs/design/output.md say human defaults are table/text, with graph as an optional codec.

radiohead (approved):

radiohead:
> > @javiermolinar nice one! Do you think we should switch default outputs for anything that produces a graph to `-ograph` instead?
> 
> I think its tricky, the graph has its own problem. The representation its not as detailed as grafana and when the number of series is high there the result is difficult to read it and you cannot select single series.
> 
> It also conflicts with the DESIGN.md and docs/design/output.md say human defaults are table/text, with graph as an optional codec.

Okay, I'll keep it mind in the future, FWIW design is something that we ourselves shape!

PR #569, opened 2026-04-24
feat(datasources): add support for Infinity datasource

mattvella07:
### What

This PR adds support for Grafana Infinity datasource, using `/add-datasource` skill.

### Why

Infinity datasource is one of the more popular ones and it makes sense to support it explicitly.

Screenshot 2026-04-24 at 1 36 16 PM


radiohead (commented):
@mattvella07 apologies for taking a while with the review, but after looking at the PR I have a question about the overall design – can we reduce the scope of this command to querying only pre-configured datasources, i.e. the `gcx datasources infinity query -d $UID` use-case?

It feels a bit strange / wrong to me to support the other use-cases because at that point it's no longer "I want to query a _data_ source" and more of "I am using infinity DS as an RPC for processing my data" and that can be addressed without gcx at all – users can use jq / uql / etc. tools locally no problem.

By reducing the scope to just "I want to query a pre-configured infinity DS" we can also benefit from dropping config like url / method / query type / header / etc., so essentially the command can be fully rewritten to:
```sh
# EXPR depends on what format $UID is configured with:
# json / xml / html – optional selector
# graphql - required query
# csv / tsv - EXPR is not allowed, must be empty
gcx datasources infinity query -d $UID $EXPR # or --expr $EXPR
```

internal/query/infinity/client.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	resp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to execute query: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes))
```
radiohead: This would produce a cryptic error with no indication the cap was hit. Could we do something like this instead:
```suggestion
respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes+1))
if err != nil {
    return 0, nil, fmt.Errorf("failed to read response: %w", err)
}

if int64(len(respBody)) > maxResponseBytes {
    return 0, nil, fmt.Errorf("response body exceeds %d MB limit; use a narrower time range or add filters", maxResponseBytes>>20)
}
```

internal/query/infinity/client.go
```diff
@@ -0,0 +1,140 @@
+package infinity
```
radiohead, line 1: nit: could you add some unit tests for the client? That'd help catch regressions in the future.

internal/datasources/infinity/query.go
```diff
@@ -0,0 +1,121 @@
...
+	dsquery "github.com/grafana/gcx/internal/datasources/query"
+	"github.com/grafana/gcx/internal/providers"
+	"github.com/grafana/gcx/internal/query/infinity"
+	"github.com/grafana/grafana-app-sdk/logging"
+	"github.com/spf13/cobra"
+)
+
+// QueryCmd returns the `query` subcommand for an Infinity datasource parent.
+func QueryCmd(loader *providers.ConfigLoader) *cobra.Command {
+	shared := &dsquery.SharedOpts{}
```
radiohead: The command does not conform to the `opts struct + setup(flags) + Validate() + constructor` pattern required by CONSTITUTION.md – see `internal/datasources/influxdb/query.go:18–32` for the expected shape.

Could you address this? Thanks!

internal/datasources/infinity/query.go
```diff
@@ -0,0 +1,121 @@
...
+  gcx datasources infinity query -d UID
+
+  # Narrow results with a JSONPath expression
+  gcx datasources infinity query -d UID '$.items'
+
+  # Output as JSON
+  gcx datasources infinity query -d UID '$.results' -o json
+
+  # Query with a time range
+  gcx datasources infinity query -d UID --from 2024-01-01 --to 2024-01-02`,
```
radiohead: nit: this example fails at runtime — short date format is rejected, only RFC3339 (`2024-01-01T00:00:00Z`) and relative times (`now-1h`) are accepted.
```suggestion
gcx datasources infinity query -d UID --from now-24h --to now
```

internal/agent/command_annotations.go
```diff
@@ -56,9 +56,10 @@ var commandAnnotations = map[string]annotation{
 	"gcx config view":            {Cost: "medium", Hint: "--minify -o json"},
 
 	// datasources
-	"gcx datasources get":   {Cost: "medium", Hint: "<uid> -o json"},
-	"gcx datasources list":  {Cost: "small"},
-	"gcx datasources query": {Cost: "large", Hint: "Run gcx help-tree metrics (or logs, traces, profiles) to discover signal commands. Prefer gcx metrics query for PromQL, gcx logs query for LogQL, gcx traces query for TraceQL, gcx profiles query for profiling. Example: <datasource-uid> 'up' --since 1h -o json"},
+	"gcx datasources get":            {Cost: "medium", Hint: "<uid> -o json"},
+	"gcx datasources list":           {Cost: "small"},
+	"gcx datasources infinity query": {Cost: "large", Hint: "-d UID https://api.example.com/data --type json --root '$.items' -o json"},
```
radiohead: nit: this registry entry is dead code – it only fills missing annotations and `query.go` already sets the correct hint directly, so this entry is never applied. It's also misleading: flags `--type` and `--root` don't exist, and the URL positional isn't part of the command surface. Let's remove it if you don't mind.

internal/datasources/infinity/query.go
```diff
@@ -0,0 +1,121 @@
...
+			dsType, err := dsquery.GetDatasourceType(ctx, cfg, datasourceUID)
+			if err != nil {
+				return err
+			}
+			if err := dsquery.ValidateDatasourceType(dsType, "infinity"); err != nil {
+				return err
+			}
+
+			now := time.Now()
+			start, end, _, err := shared.ParseTimes(now)
```
radiohead: QQ – does infinity not support step?

mattvella07: From my understanding, it doesn't.  The [query](https://github.com/grafana/grafana-infinity-datasource/blob/main/pkg/models/query.go#L90) struct doesn't have a field for it.  I changed this to use `ParseTimeRange` instead

radiohead (approved):

PR #571, opened 2026-04-25
feat(login): engage mTLS on the gcx login path

aposnov:
## Summary

   Thread TLS client certificate configuration through the entire `gcx login` flow so that login
   works against mTLS-protected Grafana servers (e.g. behind a Teleport proxy).

   ### What changed

   **Login flow (`cmd/gcx/login/`, `internal/login/`)**
   - `login.Inputs` carries a `TLS` field populated from the existing context on re-auth (so users
    don't have to re-specify certs)
   - `detectTarget` builds a TLS-aware HTTP client for the server probe
   - `Validate` threads TLS into the health-check context
   - Interactive auth menu offers **"Client certificate (mTLS)"** as a choice when certs are
   configured; defaults to it in non-interactive (`--yes`) mode
   - Re-login preserves existing `grafana.tls.*` settings

   **Grafana client (`internal/grafana/`)**
   - `GetVersion` reuses the TLS config already resolved by `clientFromContextWithTLS`, instead of
    discarding it by replacing the transport with a bare default client
   - `FetchAnonymousSettings` accepts an optional `*http.Client` so callers can supply a TLS-aware
    client for mTLS servers

   **HTTP utils (`internal/httputils/`)**
   - New `NewDefaultClientWithTLS(ctx, *tls.Config)` helper that threads a TLS config into the
   default client pipeline

   **Fix (`internal/providers/configloader.go`)**
   - Fix env-var bug where `GRAFANA_TLS_CERT_FILE` / `GRAFANA_TLS_KEY_FILE` were silently ignored
   because the TLS pointer was not initialized before `env.Parse`

   **Tests**
   - TLS propagation through login flow
   - Re-auth TLS preservation
   - PR #553 suggested test cases: half-configured CertData/KeyData, MinVersion pinning, CAData
   added to system roots, `NewNamespacedRESTConfig` round-trip

   **Docs**
   - Updated `docs/architecture/auth-system.md` with mTLS login flow details

   ### Scope 

   Closes the scope note from #553.

   ## Test plan

   - [x] `go test ./internal/login/...` — TLS propagation and re-auth preservation
   - [x] `go test ./internal/config/...` — REST config round-trip and TLS edge cases
   - [x] `go test ./internal/grafana/...` — settings client signature change
   - [x] Manual: `gcx login --server https://mtls-grafana.example.com` with Teleport certs
   configured
   - [x] Manual: re-auth flow preserves TLS settings without re-prompting

### DIFF
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -165,6 +165,14 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 		!flags.Yes &&
 		!agent.IsAgentMode()
 
+	// Carry existing TLS settings into the login flow so that mTLS keeps
+	// working on re-auth without requiring the user to re-specify certs.
+	var existingTLS *config.TLS
+	if sourceCtx != nil && sourceCtx.Grafana != nil && sourceCtx.Grafana.TLS != nil &&
+		!sourceCtx.Grafana.TLS.IsEmpty() {
+		existingTLS = sourceCtx.Grafana.TLS
+	}
+
 	opts := login.Options{
 		Inputs: login.Inputs{
 			Server:       flags.Server,
@@ -174,6 +182,7 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 			CloudAPIURL:  flags.CloudAPIURL,
 			Yes:          flags.Yes,
 			Writer:       cmd.ErrOrStderr(),
+			TLS:          existingTLS,
 		},
 		Hooks: login.Hooks{
 			ConfigSource: flags.Config.ConfigSource(),
@@ -337,22 +346,42 @@ func askForInput(e *login.ErrNeedInput, opts *login.Options, sourceCtx *config.C
 //   - Unknown (target still ambiguous): both options are offered, token
 //     first to match the historical default.
 func askGrafanaAuth(opts *login.Options, existingToken string) error {
+	// When TLS client certs are configured, mTLS is a valid standalone auth
+	// method (e.g. Teleport proxy). Offer it as the default choice.
+	hasMTLS := opts.TLS != nil && !opts.TLS.IsEmpty() &&
+		(len(opts.TLS.CertData) > 0 || opts.TLS.CertFile != "")
+	if hasMTLS && opts.Yes {
+		// Non-interactive with certs configured: default to mTLS.
+		return nil // resolveGrafanaAuth will pick up the TLS case.
+	}
+
 	tokenOption := huh.NewOption("Service account token (requires permissions for managing service accounts)", "token")
 	oauthOption := huh.NewOption("OAuth (browser) — recommended for cloud stacks; experimental on some configurations, fall back to a service account token if you hit issues", "oauth")
+	mtlsOption := huh.NewOption("Client certificate (mTLS) — authenticate via TLS client cert (e.g. Teleport)", "mtls")
 
 	var options []huh.Option[string]
 	switch opts.Target {
 	case login.TargetOnPrem:
-		options = []huh.Option[string]{tokenOption}
+		if hasMTLS {
+			options = []huh.Option[string]{mtlsOption, tokenOption}
+		} else {
+			options = []huh.Option[string]{tokenOption}
+		}
 	case login.TargetCloud:
 		options = []huh.Option[string]{oauthOption, tokenOption}
 	default: // TargetUnknown
-		options = []huh.Option[string]{tokenOption, oauthOption}
+		if hasMTLS {
+			options = []huh.Option[string]{mtlsOption, tokenOption, oauthOption}
+		} else {
+			options = []huh.Option[string]{tokenOption, oauthOption}
+		}
 	}
 
 	authMethod := "token"
-	// On-prem has a single option; skip the one-item menu and fall through
-	// to the token-input prompt directly.
+	if hasMTLS {
+		authMethod = "mtls"
+	}
+	// Single option: skip the menu and fall through directly.
 	if len(options) > 1 {
 		methodForm := huh.NewForm(huh.NewGroup(
 			huh.NewSelect[string]().
@@ -366,6 +395,10 @@ func askGrafanaAuth(opts *login.Options, existingToken string) error {
 	}
 	if authMethod == "oauth" {
 		opts.UseOAuth = true
+		return nil
+	}
+	if authMethod == "mtls" {
+		// mTLS needs no additional input — the certs are already in opts.TLS.
 		return nil
 	}
 
@@ -483,7 +516,7 @@ func structuredMissingFieldsError(e *login.ErrNeedInput) error {
 		case "server":
 			suggestions = append(suggestions, "Pass --server <url> or set GRAFANA_SERVER")
 		case "grafana-auth":
-			suggestions = append(suggestions, "Pass --token <token> for a service account token")
+			suggestions = append(suggestions, "Pass --token <token> for a service account token, or configure TLS client certs for mTLS auth")
 		case "cloud-token":
 			suggestions = append(suggestions, "Pass --cloud-token <token> to enable Cloud features, or --yes to skip")
 		default:
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -17,6 +17,7 @@ import (
 	"github.com/grafana/gcx/internal/format"
 	"github.com/grafana/gcx/internal/login"
 	cmdio "github.com/grafana/gcx/internal/output"
+	"github.com/grafana/grafana-app-sdk/logging"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"golang.org/x/term"
@@ -171,6 +172,17 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 	if sourceCtx != nil && sourceCtx.Grafana != nil && sourceCtx.Grafana.TLS != nil &&
 		!sourceCtx.Grafana.TLS.IsEmpty() {
 		existingTLS = sourceCtx.Grafana.TLS
+		// Advisory: when --allow-server-override re-points the context at a
+		// different server, the existing TLS client cert will be presented to
+		// the new server. This is gated by explicit user opt-in.
+		if flags.AllowServerOverride && flags.Server != "" &&
+			sourceCtx.Grafana != nil && sourceCtx.Grafana.Server != "" &&
+			flags.Server != sourceCtx.Grafana.Server {
+			logging.FromContext(cmd.Context()).Warn("reusing existing TLS client certificate for a different server",
+				"previous_server", sourceCtx.Grafana.Server,
+				"new_server", flags.Server,
+			)
+		}
 	}
 
 	opts := login.Options{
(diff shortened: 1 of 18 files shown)


radiohead:
@claude Please review this PR.

internal/providers/configloader.go
```diff
@@ -93,8 +93,23 @@ func cloudEnvOverride(cfg *config.Config) error {
 	if curCtx.Cloud == nil {
 		curCtx.Cloud = &config.CloudConfig{}
 	}
+	// Initialize nested pointers so env.Parse can populate GRAFANA_TLS_* env vars.
```
radiohead: **suggestion (non-blocking):** This same init/cleanup dance — zero-init `Grafana.TLS` before `env.Parse`, then nil it out via `IsEmpty()` — appears in three places: here (`cloudEnvOverride`), in `envOverride` below at line 155+, and in `cmd/gcx/config/command.go`. The second site even has a comment pointing at the third ("same fix as cmd/gcx/config/command.go"). `configloader.go` is a churn hotspot, so future drift is likely. Consider extracting a helper into `internal/config/`, e.g. `func PrepareForEnvParse(*Context)` + `func CleanupAfterEnvParse(*Context)`, so all three sites become one call. Adding a new nested `*struct` field to `GrafanaConfig` (similar to `TLS`) would otherwise require remembering to repeat this dance everywhere or env vars silently get read into a discarded zero-struct.

internal/grafana/client.go
```diff
@@ -25,17 +26,24 @@ func (e *VersionIncompatibleError) Error() string {
 	return fmt.Sprintf("grafana version %s is not supported; gcx requires Grafana 12.0.0 or later", e.Version)
 }
 
-func ClientFromContext(ctx *config.Context) (*goapi.GrafanaHTTPAPI, error) {
+// clientResult bundles the goapi client with the resolved *tls.Config so
```
radiohead, line 29: **suggestion (non-blocking):** The fix bundles the goapi client + resolved `*tls.Config` in a private `clientResult`; exported `ClientFromContext` discards the TLS half. This works today because there are only two callers, but it institutionalizes a latent trap: any future caller wrapping the goapi client with middleware via `WithHTTPClient` and using the public `ClientFromContext` will silently lose TLS — re-creating exactly the bug this PR fixes for `GetVersion`. Nothing in the type system signals this. Consider either (a) making `ClientFromContext` always wire `httputils.NewDefaultClientWithTLS(ctx, tlsCfg)` internally so callers get a fully-configured client in one call (preferred — collapses the two-step ritual), or (b) exporting `ClientFromContextWithTLS` so the existence of TLS is visible at the API surface.

internal/grafana/client.go
```diff
@@ -2,6 +2,7 @@ package grafana
 
 import (
 	"context"
+	crypto_tls "crypto/tls"
```
radiohead: **nitpick (non-blocking):** The `crypto_tls "crypto/tls"` alias is inconsistent with the rest of the codebase — `internal/config/types.go`, `internal/config/tls_test.go`, `internal/server/server.go`, `internal/server/grafana/requests.go`, and `internal/httputils/client.go` all use plain `"crypto/tls"`. If there's a name-shadowing concern with `TransportConfig.TLS` in this package, a comment explaining it would help; otherwise drop the alias and use `*tls.Config`.

cmd/gcx/login/command.go
```diff
@@ -165,6 +165,14 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 		!flags.Yes &&
 		!agent.IsAgentMode()
 
+	// Carry existing TLS settings into the login flow so that mTLS keeps
```
radiohead, line 169: **question (non-blocking):** When `--server B --allow-server-override` is used against an existing mTLS context configured for server A, the existing TLS client cert is carried into the new login flow and presented to server B during target detection and connectivity validation. This is gated by an explicit user opt-in and server B controls whether it accepts the cert, so it's not exploitable without intent — but the behavior is surprising. Is it intentional that `--allow-server-override` silently reuses TLS material from the prior server? A log-level advisory when this combination is detected might be worthwhile.

cmd/gcx/login/command.go
```diff
@@ -329,22 +338,42 @@ func askForInput(e *login.ErrNeedInput, opts *login.Options, sourceCtx *config.C
...
+		} else {
+			options = []huh.Option[string]{tokenOption, oauthOption}
+		}
 	}
 
 	authMethod := "token"
-	// On-prem has a single option; skip the one-item menu and fall through
-	// to the token-input prompt directly.
+	if hasMTLS {
+		authMethod = "mtls"
```
radiohead, line 396: **suggestion (non-blocking):** When `Target == TargetCloud && hasMTLS == true`, the option list is `[oauthOption, tokenOption]` (mTLS deliberately excluded), but `authMethod` is pre-set to `"mtls"` by the `if hasMTLS { authMethod = "mtls" }` block. If `huh.Select` leaves the bound value untouched until the user navigates, an immediate-Enter could exit the prompt with `authMethod == "mtls"` despite the menu not offering it — and the post-menu `if authMethod == "mtls"` branch returns nil, letting `resolveGrafanaAuth` pick the mTLS path. Either reset `authMethod` to `"oauth"` for the Cloud arm before opening the form, or drop the `hasMTLS` precondition for Cloud entirely (auth-system.md only describes mTLS for on-prem behind Teleport, so Cloud+mTLS may be vestigial).

cmd/gcx/login/command.go
```diff
@@ -475,7 +508,7 @@ func structuredMissingFieldsError(e *login.ErrNeedInput) error {
 		case "server":
 			suggestions = append(suggestions, "Pass --server <url> or set GRAFANA_SERVER")
 		case "grafana-auth":
-			suggestions = append(suggestions, "Pass --token <token> for a service account token")
+			suggestions = append(suggestions, "Pass --token <token> for a service account token, or configure TLS client certs for mTLS auth")
```
radiohead: **nitpick (non-blocking):** Nice addition pointing users at mTLS. Consider also mentioning `GRAFANA_TLS_CERT_FILE`/`GRAFANA_TLS_KEY_FILE` env vars or `gcx config set contexts.<ctx>.grafana.tls.cert-file ...` for completeness with the docs.

internal/login/login.go
```diff
@@ -290,11 +299,30 @@ func Run(ctx context.Context, opts *Options) (Result, error) {
 }
 
 // detectTarget calls DetectFn or falls back to the real DetectTarget.
+// When TLS settings are present, builds a TLS-aware HTTP client for the probe.
 func detectTarget(ctx context.Context, opts Options) (Target, error) {
 	if opts.DetectFn != nil {
 		return opts.DetectFn(ctx, opts.Server)
 	}
-	return DetectTarget(ctx, opts.Server, httputils.NewDefaultClient(ctx))
+	client, err := tlsAwareClient(ctx, opts.TLS)
```
radiohead, line 325: **question (non-blocking):** `detectTarget` now returns a hard error from `tlsAwareClient` (cert-load failure) wrapped as `"TLS configuration: ..."`. The documented `probeTarget` contract is "any probe error → TargetUnknown," so this is a different failure category. A malformed `cert-file` path that previously surfaced at `Validate` time will now surface at detect time, before target classification. Is this the intended fail-fast point for cert errors? Either way, a comment documenting the intent (or a degrade to TargetUnknown letting the later validation step produce the actionable error) would clarify the contract.

internal/login/login.go
```diff
@@ -323,6 +352,12 @@ func resolveGrafanaAuth(ctx context.Context, opts Options, target Target) (strin
 		grafanaCfg.AuthMethod = "token"
 		method = "token"
 
+	case opts.TLS != nil && (len(opts.TLS.CertData) > 0 || opts.TLS.CertFile != ""):
```
radiohead, line 373: **suggestion (non-blocking):** The mTLS branch decision here checks only for client-cert presence (`CertFile`/`CertData`) but not the matching key. A misconfigured `TLS{CertFile: "x"}` without a key file would land on this branch, then fail downstream at `ResolveFiles`/`X509KeyPair`. Pre-validating cert+key pairing here would yield a friendlier error. Not blocking — `TLS.ResolveFiles` already enforces "both cert-file and key-file must be provided together" — but the failure happens further down the stack than necessary.

radiohead (approved):
Reviewed the mTLS login plumbing across the cmd/login, internal/login, internal/grafana, internal/httputils, and config layers. The implementation cleanly threads TLS through target detection, validation, the version probe, and re-auth persistence; backward compatibility holds because the new `Inputs.TLS` field is additive and absent-TLS preserves the prior code path. Tests cover the propagation paths well, and the security posture improves (TLS 1.2 pinning preserved, half-configured cert state rejected, no credential material in logs). Inline comments are mostly architectural-quality observations and minor UX/doc polish — none blocking. Two notes that fall outside the diff hunks worth flagging here:

- `internal/config/types.go` `AuthMethod` field doc-comment and `internal/login/login.go` `Result.AuthMethod` field doc-comment still enumerate `"oauth", "token", or "basic"` — `CHANGELOG`, `auth-system.md`, and `InferredAuthMethod()` were updated to mention `mtls`, but these two struct field comments were missed.
- `internal/config/rest.go:215-222` (`NewNamespacedRESTConfig`) builds `rest.TLSClientConfig` directly from raw fields rather than going through `ToStdTLSConfig`, so the K8s REST path doesn't pin `MinVersion: tls.VersionTLS12` explicitly the way the `http.Client` paths do. Modern Go runtime defaults to 1.2 so this is not a current vulnerability, but explicit pinning would close the gap if/when client-go exposes a knob.

---
**Generated by Claude Code**

internal/providers/configloader.go
andreyposnov: Done! Extracted `ParseEnvIntoContext` into `internal/config/envparse.go` — all three sites now use one call. Thanks for catching the drift risk 👍

internal/grafana/client.go
andreyposnov: Good catch — exported `ClientFromContextWithTLS` and added a doc warning on `ClientFromContext` so future callers know about the TLS gap.

internal/grafana/client.go
andreyposnov: Fixed — dropped the alias, using plain `"crypto/tls"` now.

cmd/gcx/login/command.go
andreyposnov: Added a `logging.Warn` when `--allow-server-override` reuses TLS certs from a different server. Agreed the behavior is surprising — the log makes it visible.

cmd/gcx/login/command.go
andreyposnov: Nice catch on the immediate-Enter edge case! Fixed — `authMethod` now defaults to `"token"` (not `"mtls"`) when target is Cloud.

cmd/gcx/login/command.go
andreyposnov: Added `GRAFANA_TLS_CERT_FILE` / `GRAFANA_TLS_KEY_FILE` and `gcx config set` to the suggestion text 👍

internal/login/login.go
andreyposnov: Yes, intentional fail-fast — added a doc comment explaining the contract. A broken cert should surface early rather than producing a confusing auth error downstream.

internal/login/login.go
andreyposnov: Added a doc comment noting that `TLS.ResolveFiles()` enforces cert+key pairing downstream. Kept the check here simple since the error path is already covered.

radiohead:
@aposnov appreciate you taking time to address the feedback! Thank you!

andreyposnov:
> @aposnov appreciate you taking time to address the feedback! Thank you!

Hey @radiohead thank you, could you please re-trigger checks

PR #583, opened 2026-04-27
fix(adapter): skip DeleteFn when delete options request dry-run

### LINKED ISSUE #573: gcx resources delete --dry-run can perform real deletes for provider-backed resources using TypedCRUD.
gcx resources delete --dry-run can perform real deletes for provider-backed resources using TypedCRUD.

  The CLI advertises --dry-run as a simulation/preview, and remote.Deleter correctly passes metav1.DeleteOptions{DryRun:
  []string{"All"}}. However, typedAdapter.Delete(ctx, name, _ metav1.DeleteOptions) ignores those options and calls DeleteFn
  unconditionally.

  Expected: typed provider adapters should not call the mutating delete API during dry-run.

  Actual: provider-backed resources may be deleted even when --dry-run is set.

jimohabdol:
fixes [573](https://github.com/grafana/gcx/issues/573)

Delete now uses the same isDryRun guard as Create and Update, returning without calling DeleteFn when dry-run is set.


radiohead (commented):
Thank you for your contribution! This is a reasonable hotfix, but I think that instead of completely skipping deletes, we should be correctly passing dry-run options to the underlying implementations (i.e. `DeleteFn`'s signature should be updated to take options as an extra argument) and look into the specifics of each implementation to see if we can properly handle dry runs there (at the very least report what would be deleted).

jimohabdol:
> Thank you for your contribution! This is a reasonable hotfix, but I think that instead of completely skipping deletes, we should be correctly passing dry-run options to the underlying implementations (i.e. `DeleteFn`'s signature should be updated to take options as an extra argument) and look into the specifics of each implementation to see if we can properly handle dry runs there (at the very least report what would be deleted).

@radiohead I have updated `DeleteFn`'s signature to take options as an extra argument, as suggested

jimohabdol:
@radiohead @wardbekker Kindly help review my PR

PR #587, opened 2026-04-27
feat(kg): add suppressions dry-run diff

### LINKED ISSUE #577: gcx kg supressions create should have a dry-run mode.
### What exists today?

`gcx kg supressions create` right now only pushes the config up to the remote.

### How could this be better?

I first want to see the diff before I apply something. Can we have a `--dry-run` mode that verifies that the config is correct and shows a diff?

It'll help me iterate with confidence.

Reduces errors or confusion

_No response_

None

annanay25:
Add --dry-run to gcx kg suppressions create so users can validate YAML and preview the remote-to-local diff without uploading. The remote read path uses `GET /api/plugins/grafana-asserts-app/resources/asserts/api-server/v1/config/disabled-alerts/`, which was tested against the configured stack and returned disabledAlertConfigs.

Tested with `go test ./internal/providers/kg/... ./internal/agent/...`

Fixes https://github.com/grafana/gcx/issues/577

Made-with: Cursor

### DIFF
--- a/internal/agent/command_annotations.go
+++ b/internal/agent/command_annotations.go
@@ -215,7 +215,7 @@ var commandAnnotations = map[string]annotation{
 	"gcx kg search insights":         {Cost: "medium", Hint: "--type <type> --since 1h"},
 	"gcx kg search sample":           {Cost: "small"},
 	"gcx kg status":                  {Cost: "small"},
-	"gcx kg suppressions create":     {Cost: "small"},
+	"gcx kg suppressions create":     {Cost: "small", Hint: "-f suppressions.yaml --dry-run"},
 
 	// -----------------------------------------------------------------------
 	// Logs provider
--- a/internal/providers/kg/commands.go
+++ b/internal/providers/kg/commands.go
@@ -667,14 +667,14 @@ func runSuppressionsDryRun(cmd *cobra.Command, client *Client, localData []byte)
 		return err
 	}
 	if remote == local {
-		cmdio.Info(cmd.OutOrStdout(), "[dry-run] Suppressions YAML is valid; no changes")
+		cmdio.Info(cmd.ErrOrStderr(), "[dry-run] Suppressions YAML is valid; no changes")
 		return nil
 	}
 	diff, err := unifiedYAMLDiff(remote, local)
 	if err != nil {
 		return fmt.Errorf("render suppressions diff: %w", err)
 	}
-	cmdio.Info(cmd.OutOrStdout(), "[dry-run] Suppressions YAML is valid; showing diff (remote -> local)")
+	cmdio.Info(cmd.ErrOrStderr(), "[dry-run] Suppressions YAML is valid; showing diff (remote -> local)")
 	_, err = fmt.Fprint(cmd.OutOrStdout(), diff)
 	return err
 }
--- a/internal/providers/kg/commands.go
+++ b/internal/providers/kg/commands.go
@@ -116,7 +116,7 @@ func (f *scopeFlags) validateScopes(ctx context.Context, client *Client) error {
 	if err != nil {
 		return nil //nolint:nilerr // best-effort: scope validation is advisory
 	}
-	var errs []string
+	errs := make([]string, 0, len(active))
 	for _, c := range active {
 		known := scopes[c.dim]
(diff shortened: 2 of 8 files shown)


internal/providers/kg/commands.go
```diff
@@ -602,20 +627,58 @@ func newSuppressionsCommand(loader RESTConfigLoader) *cobra.Command {
...
+		return nil
+	}
+	diff, err := unifiedYAMLDiff(remote, local)
+	if err != nil {
+		return fmt.Errorf("render suppressions diff: %w", err)
+	}
+	cmdio.Info(cmd.OutOrStdout(), "[dry-run] Suppressions YAML is valid; showing diff (remote -> local)")
+	_, err = fmt.Fprint(cmd.OutOrStdout(), diff)
+	return err
+}
```
gouthamve-but-enterprise, line 680: This does no validation imo. So if I have a faulty YAML, it'll say it's good, and I'll only find out when I do a `create`. @xujiaxj might be worth adding a backend endpoint for this.

internal/providers/kg/commands.go
```diff
@@ -602,20 +627,58 @@ func newSuppressionsCommand(loader RESTConfigLoader) *cobra.Command {
...
+	}
+	if remote == local {
+		cmdio.Info(cmd.OutOrStdout(), "[dry-run] Suppressions YAML is valid; no changes")
+		return nil
+	}
+	diff, err := unifiedYAMLDiff(remote, local)
+	if err != nil {
+		return fmt.Errorf("render suppressions diff: %w", err)
+	}
+	cmdio.Info(cmd.OutOrStdout(), "[dry-run] Suppressions YAML is valid; showing diff (remote -> local)")
```
radiohead: **suggestion (blocking):** The `[dry-run] ...` banners are written to `cmd.OutOrStdout()`, mixing diagnostic preamble with the diff result on stdout. CONSTITUTION.md: "STDOUT is the result, STDERR is the diagnostic." Pipe consumers like `| patch` or `| diffstat` will receive contaminated input from the banner lines.

Route both banners through `cmd.ErrOrStderr()` (matching the pattern in `internal/providers/traces/adaptive/commands.go`) and keep only the unified diff body on stdout. For the "no changes" case, stdout stays empty — a correct "result: no diff" signal. Also update the two dry-run tests to assert the banner against `cmd.SetErr(...)` and the diff against `cmd.SetOut(...)`.

radiohead (commented):
The PR adds a well-structured `--dry-run` flag to `gcx kg suppressions create`. The fail-fast ordering (local YAML validated before any network call), three-scenario test coverage, and the structural guard preventing accidental uploads in dry-run mode are all solid. One stdout/stderr placement concern is posted inline.

---
**Generated by Claude Code**

internal/providers/kg/commands.go
```diff
@@ -602,20 +627,58 @@ func newSuppressionsCommand(loader RESTConfigLoader) *cobra.Command {
...
+	_, err = fmt.Fprint(cmd.OutOrStdout(), diff)
```
radiohead, line 678: **suggestion (non-blocking):** The unified diff is emitted via `fmt.Fprint`, bypassing the `internal/output` codec registry. This works for the current text output, but leaves `--dry-run` behavior with `-o json` undefined. Worth a follow-up to define a structured form so agent consumers do not have to parse unified diff text.

internal/agent/command_annotations.go
```diff
@@ -215,7 +215,7 @@ var commandAnnotations = map[string]annotation{
 	"gcx kg search insights":         {Cost: "medium", Hint: "--type <type> --since 1h"},
 	"gcx kg search sample":           {Cost: "small"},
 	"gcx kg status":                  {Cost: "small"},
-	"gcx kg suppressions create":     {Cost: "small"},
+	"gcx kg suppressions create":     {Cost: "small", Hint: "-f suppressions.yaml --dry-run"},
```
radiohead, line 218: **nitpick (non-blocking):** The example hint includes `--dry-run`, making the canonical agent invocation a no-op upload. Consider whether the active form (`-f suppressions.yaml`) is more useful as the hint, with `--dry-run` documented in `--help` only. The current safe-default approach is defensible.

PR #600, opened 2026-04-29
feat: notify users about new gcx versions

javiermolinar:
In a previous pr https://github.com/grafana/gcx/pull/562 we introduced a notification system. This pr uses it to notify the user when a new version of GCX is available. As before this notification (and the check) is done only once per day and users can opt-out

### DIFF
--- a/internal/notifier/run.go
+++ b/internal/notifier/run.go
@@ -1,6 +1,7 @@
 package notifier
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"io/fs"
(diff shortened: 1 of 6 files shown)


internal/notifier/run.go
```diff
@@ -52,3 +61,33 @@ func maybeNotifySkillsAt(source fs.FS, dst io.Writer, statePath, root string, no
...
+		return nil
+	}
+
+	_, err = fmt.Fprintln(dst, msg)
+	return err
+}
+
+func httpClientWithTimeout() *http.Client {
+	return &http.Client{Timeout: 2 * time.Second}
+}
```
radiohead: **suggestion**: Could we use a shared client instead?

internal/notifier/version.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return fmt.Sprintf("A new gcx version is available: %s (current: %s)\nUpgrade: %s", release.TagName, currentVersion, upgradeURL), nil
+}
+
+func fetchLatestRelease(client *http.Client, url string) (githubRelease, error) {
+	if client == nil {
+		client = &http.Client{Timeout: 2 * time.Second}
+	}
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, url, nil)
```
radiohead: **issue(non-blocking)**: I wonder if we could wire context from the command here, so that SIGINT is properly handled here.

internal/notifier/version.go
```diff
@@ -0,0 +1,95 @@
...
+		upgradeURL = releaseTagURLPrefix + release.TagName
+	}
+
+	return fmt.Sprintf("A new gcx version is available: %s (current: %s)\nUpgrade: %s", release.TagName, currentVersion, upgradeURL), nil
+}
+
+func fetchLatestRelease(client *http.Client, url string) (githubRelease, error) {
+	if client == nil {
+		client = &http.Client{Timeout: 2 * time.Second}
+	}
```
radiohead: **suggestion**: Would be nice to use a shared client instead.

radiohead (approved):
Looks good, approved!

A few non-blocking comments – feel free to skip them if you want.

Also, I think it'd be great to add install method eventually and use it in the notifier so that we could provide commands to upgrade using the correct method (e.g. via Homebrew or `go install`, etc.) but I realize it's a non-trivial task.

PR #601, opened 2026-04-29
refactor(kg): align commands to [noun][verb] format

xujiaxj:
## Summary

- Rename `gcx kg describe` → `gcx kg meta`
- Rename `gcx kg search insights` → `gcx kg insights search` (moved as a subcommand of `insights`)
- Remove `gcx kg entities show` — the list case is covered by `entities list`, the single-entity case was minimal enough to drop

## Test plan

- [x] `gcx kg meta --help` shows correct subcommands
- [x] `gcx kg insights search --help` shows correct flags
- [x] `gcx kg entities show` is gone; `gcx kg entities list` still works
- [x] Reference docs regenerated and consistent


docs/reference/cli/gcx_kg.md
```diff
@@ -23,17 +23,16 @@ Manage Grafana Knowledge Graph rules, entities, and insights
 ### SEE ALSO
 
 * [gcx](gcx.md)	 - Control plane for Grafana Cloud operations
-* [gcx kg describe](gcx_kg_describe.md)	 - Describe the Knowledge Graph: entity types, valid env/namespace/site values, and telemetry query configs.
 * [gcx kg entities](gcx_kg_entities.md)	 - Manage Knowledge Graph entities.
 * [gcx kg health](gcx_kg_health.md)	 - Show a health summary with active insight counts.
 * [gcx kg insights](gcx_kg_insights.md)	 - Query Knowledge Graph insights.
 * [gcx kg inspect](gcx_kg_inspect.md)	 - Inspect an entity: info, insights, and summary.
+* [gcx kg meta](gcx_kg_meta.md)	 - Show Knowledge Graph metadata: entity types, valid env/namespace/site values, and telemetry query configs.
```
annanay25, line 30: can we name this `metadata`?

docs/reference/cli/gcx_kg_entities.md
```diff
@@ -24,5 +24,4 @@ Manage Knowledge Graph entities.
 
 * [gcx kg](gcx_kg.md)	 - Manage Grafana Knowledge Graph rules, entities, and insights
 * [gcx kg entities list](gcx_kg_entities_list.md)	 - List Knowledge Graph entities for a given type.
-* [gcx kg entities show](gcx_kg_entities_show.md)	 - Show entities. Without a name, lists all; with a name, shows one.
```
annanay25, line 27: Is the single entity show use-case not valuable for gcx? I thought it would be useful for agents to view a single entity.

docs/reference/cli/gcx_kg.md
xujiaxj: I debated to myself on this. Claude convinced me to go with the shorter form. What do you think?

> meta — it's shorter and fits the terse CLI style better. metadata is more explicit but every other gcx command uses abbreviated nouns (slo, synth, irm) and short subcommand names. meta reads cleanly in gcx kg meta schema vs the slightly clunky gcx kg metadata schema.

docs/reference/cli/gcx_kg_entities.md
xujiaxj: To locate a single entity, the 'list' command can do it with '--property name=foo`.

To check all the insights, graph data, `kg inspect` can do, and I plan to rename `kg inspect` to `kg entities describe`. So between these two, `show` seems a bit redundant and at its current form is a dual-mode command (list + single). I found a dual-mode command a bit confusing.

> In short: show covers the get-one case and a basic "list all" case; list is the purpose-built list command with property filtering and details. They overlap in the "list all of a type" scenario but list has more filtering power there.
                                                                                                                                                                                  
> If I want to remove "show", what do you think?                                                                                                                              

> Good candidate for removal. The list-all behavior is fully covered by list, and the single-entity lookup can be done with list --property name=<value>. The one question is whether you want a dedicated entities get [name] to replace the single-entity case — list --property name=foo works but is more verbose for that common pattern. If inspect already covers the "I want everything about one entity" case, then just deleting show with no replacement is probably fine.

docs/reference/cli/gcx_kg.md
annanay25: ok lets go with the short form 👍

docs/reference/cli/gcx_kg_entities.md
annanay25: ok this is fine to remove. can we add an additional hint for the entities list command indicating that it can query a single entity as well using this property name filter?

xujiaxj: Pushed. The hint now reads: use --property name=<value> to fetch a single entity by name | run gcx kg meta scopes first to discover valid env/namespace/site values"

annanay25 (approved):

PR #603, opened 2026-04-29
fix(linter): detect missing title/description on panels nested in collapsed rows

### LINKED ISSUE #604: Linter rule `panel-title-description` skips panels nested in collapsed rows
### gcx Version

`gcx v0.2.10` (also reproducible on `main` at commit 8baf86a)

Other (`gcx dev lint`)

### Command Executed

```shell
gcx dev lint run ./dashboard.yaml
```

### Command Output

```shell
{
  "violations": [],
  "summary": {
    "files_scanned": 1,
    "files_failed": 0,
    "num_violations": 0
  }
}
```

### What happened?

The built-in `panel-title-description` Rego rule only iterates `spec.panels[]` at the top level. In v1beta1 dashboards (`dashboard.grafana.app/v1beta1`), Grafana stores the children of a *collapsed* row under `row.panels` rather than at the dashboard root. As a result, panels inside collapsed rows are never linted.

Real-world impact: a dashboard with 22 timeseries panels missing descriptions, all nested under collapsed rows, produced **zero** violations.

One violation per nested panel missing a title or description, e.g. `panel 2 has no description`.

1. Create a dashboard with one collapsed row containing one panel without a description:

   ```yaml
   kind: Dashboard
   apiVersion: dashboard.grafana.app/v1beta1
   metadata:
     name: collapsed-row-test
   spec:
     panels:
       - id: 1
         type: row
         collapsed: true
         title: Some row
         panels:
           - id: 2
             type: timeseries
             title: Foo
             # description missing
   ```

2. Run `gcx dev lint run ./dashboard.yaml`
3. Observe: `No violations found` — expected one violation pointing at panel 2.

Every time

### Environment Details

Linux 6.17 (Ubuntu 24.04), bash 5.2, Go 1.26

Root cause is in `internal/linter/bundle/gcx/utils/dashboard.rego` — `dashboard_v1_panels` filters only by `panel.type != "row"` and never recurses into `row.panels`. v2 dashboards (`spec.elements`) are unaffected because panels are stored in a flat map.

Fix proposed in #603.

aristide-bayard:
## Summary

The built-in `panel-title-description` Rego rule only iterated `spec.panels[]` at the top level. Grafana stores the children of a *collapsed* row under `row.panels` rather than at the dashboard root, so any panel hidden in a collapsed row was invisible to the rule.

Concrete impact: a v1beta1 dashboard with 22 timeseries panels missing descriptions, all nested under collapsed rows, produced **zero** violations on `gcx v0.2.10`.

## Fix

`utils.dashboard_v1_panels` now also yields panels nested inside rows. v2 (`spec.elements`) already stores panels in a flat map, so no change there.

## Test plan

- [x] New regression test: panel nested in a collapsed row without description → violation
- [x] New test: panel nested in a collapsed row with title + description → no violation
- [x] New test: a row by itself never produces a violation
- [x] Existing tests still pass (`gcx dev lint test ./internal/linter/bundle/gcx/` → 48/48)
- [x] Reproduced manually: a dashboard with one collapsed row containing one panel without a description now reports `panel <id> has no description`

Closes #604


javiermolinar (approved):
LGTM. Thank you for your contribution

PR #607, opened 2026-04-30
feat(kg): add cypher search command

kozhuhds:
## Summary

- Adds `gcx kg cypher <query>` to run read-only Cypher queries against the Knowledge Graph
- Routes through the existing Grafana plugin proxy (`/api/plugins/grafana-asserts-app/resources/asserts/api-server/v1/search/cypher`) with the same service token auth as all other `gcx kg` commands
- Supports `--from`/`--to`/`--since`, `--page`, `--insights-only` flags; table output by default, `--output json` for raw response
- New types: `CypherSearchRequest`, `CypherSearchResponse`, `CypherEntity`, `CypherEdge`, `CypherInsight`

## Test plan

- [x] `TestClient_CypherSearch` — 7 table-driven cases covering path, request body, scope, withInsights, empty response, edges, and error codes
- [x] `go test ./internal/providers/kg/...` passes
- [x] `gcx kg cypher "MATCH (s:Service) RETURN s LIMIT 10"` verified against live stack


docs/reference/cli/gcx_kg_cypher.md
```diff
@@ -0,0 +1,48 @@
...
+### Options
+
+```
+      --env string         Environment scope (run 'gcx kg meta scopes' to see valid values)
+      --from string        Start time (RFC3339, Unix timestamp, or relative like 'now-1h')
+  -h, --help               help for cypher
+      --insights-only      Return only entities with active insights
+      --json string        Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+      --namespace string   Namespace scope (run 'gcx kg meta scopes' to see valid values)
+  -o, --output string      Output format. One of: json, table, yaml (default "json")
```
radiohead: Could you make table the default output? And agent mode should be wired to switch to JSON instead.

xujiaxj:
LGTM.

xujiaxj (dismissed):

internal/providers/kg/commands.go
```diff
@@ -93,70 +92,6 @@ func (f *scopeFlags) scopeCriteria() *ScopeCriteria {
 	return &ScopeCriteria{NameAndValues: vals}
 }
 
-// validateScopes checks that any set scope values exist in the KG scope registry.
-// If a value is not an exact match it fetches known values, finds candidates by
-// substring match, and returns an error with actionable hints so the caller
-// (human or LLM) can retry with the correct value. Validation is best-effort:
-// if the scopes API is unavailable the error is silently ignored.
-func (f *scopeFlags) validateScopes(ctx context.Context, client *Client) error {
```
annanay25, line 103: Why is this being removed from other commands?

internal/agent/command_annotations.go
```diff
@@ -204,6 +204,7 @@ var commandAnnotations = map[string]annotation{
 	// -----------------------------------------------------------------------
 	// Knowledge Graph provider
 	// -----------------------------------------------------------------------
+	"gcx kg cypher":                  {Cost: "medium", Hint: "\"MATCH (s:Service) RETURN s LIMIT 10\" [--env <env>] [--namespace <ns>] [--since 1h] | read-only Cypher query; always include LIMIT for targeted lookups; omit for broad discovery"},
```
annanay25: not seeing `[--env <env>] [--namespace <ns>] [--since 1h]` as flags to the cypher command. am i missing something?

rdubrock: Yeah scope isn't a top level option for cypher. You can specify it as part of the query itself if you want.

rdubrock:
Stale agent hint at `internal/agent/command_annotations.go:207`:

```
"gcx kg cypher": {Cost: "medium", Hint: "\"MATCH (s:Service) RETURN s LIMIT 10\" [--env <env>] [--namespace <ns>] [--since 1h] | ..."}
```

`--env` and `--namespace` were removed from this command in 16390b3, so agents following the hint will hit "unknown flag" errors. Drop them from the hint string.

rdubrock:
Claude found that stale hint, but otherwise lgtm!

rdubrock (dismissed):

internal/agent/command_annotations.go
annanay25: Gotcha. So it needs to be removed from this hint as that would lead the LLM to pass those flags to gcx.

rdubrock: Yah that's what my claude review found as well

internal/providers/kg/commands.go
kozhuhds: good catch, restoring this.

annanay25 (approved):
lgtm.

PR #610, opened 2026-04-30
feat(traces): render trace tree as table for traces get

radiohead:
## What

\`gcx traces get\` now renders a tree-indented table when \`-o table\` or \`-o wide\` is passed. The default output format for non-agent sessions is \`table\`; agent mode continues to receive JSON. Piped output renders the same table format without ANSI styling (handled by \`IsStylingEnabled()\`).

The table format includes five columns (SPAN, SERVICE, SPAN_ID, DURATION, %) where SPAN is tree-indented with connectors, DURATION uses ns-precision formatting, and % shows each span's share of total trace duration. Wide format adds KIND and START columns.

Spans whose parent is absent from the response are rendered in a separate detached-subtrees section after the attached tree, preceded by a labelled dim divider. Color thresholds on the % column highlight the critical path at a glance; spans with error status render in red with a warning prefix.

The thresholded percentage coloring and error-span cell styling are exported from \`internal/style\` so future formatters can reuse them.

\`traces get\` is wired into the shared codec pipeline (\`dsquery.RegisterCodecs\`), closing the inconsistency where it was the only signal-tier command registering codecs ad hoc. The \`-o graph\` variant is not registered for this command and returns an "unknown output format" error at flag validation time.

\`internal/style.TableBuilder\` gains a \`ColumnWidths()\` method that locks specific columns to fixed widths via lipgloss \`fixedWidth\`, so narrow terminals only compress the variable-length SPAN and SERVICE columns rather than trimming the fixed-width SPAN_ID, DURATION, and % columns.

## Why

\`gcx traces get\` was the only signal-tier command that bypassed the shared codec pipeline, emitting raw OTLP JSON regardless of the \`-o\` flag. JSON is machine-readable but gives no visual structure at the terminal — no tree, no duration percentages, no indication of detached subtrees. Users had to leave the CLI for Grafana Explore to answer "where did the time go".

The tree-table format fills that gap without replacing Explore. It renders one trace top-to-bottom in tree order with cost annotations, satisfying the most common human-facing query (critical-path identification) while preserving the agent and script contract (JSON remains the default when in agent mode).

## How it was tested

- [x] \`make lint\` exited 0 with 0 issues across all packages
- [x] \`go test -race ./...\` passed with no failures; all packages including \`internal/query/tempo\` and \`internal/style\` reported ok
- [x] Table-driven unit tests in \`internal/query/tempo/formatter_test.go\` cover: basic tree (root/child/grandchild), detached subtrees divider, async-tail header suffix, error-span red cell + warning prefix, nil trace (no panic), wide column order, \`formatDurationNanos\` inputs (500ns through 1m32s and degenerate \`<= 0\` cases), zero total-duration rows, and three-sibling connector ordering
- [x] Codec dispatch tests in \`internal/datasources/query/codecs_test.go\` cover \`*tempo.GetTraceResponse\` for table and wide cases

---
**Generated by Claude Code**


docs/adrs/traces-get-table/001-tree-table-render-for-traces-get.md
```diff
@@ -0,0 +1,260 @@
+# Tree-Table Rendering for `traces get`
```
annanay25, line 1: (drive-by comment, haven't had a chance to review thoroughly): do we need an ADR for this change?

docs/adrs/traces-get-table/001-tree-table-render-for-traces-get.md
```diff
@@ -0,0 +1,260 @@
...
+
+**Created**: 2026-04-30
+**Status**: proposed
```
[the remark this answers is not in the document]
radiohead: Fixed — ADR status updated to `accepted` and the three stale sections (Decision paragraph, Default format subsection, Consequences follow-ups) updated to reflect what actually shipped. Committed in a534142.

---
**Generated by Claude Code**

docs/adrs/traces-get-table/001-tree-table-render-for-traces-get.md
```diff
@@ -0,0 +1,260 @@
...
+borders) when stdout is piped, in agent mode, or `--no-color` is set; no
+custom branching needed.
+
+#### Default format
+
+The spike keeps JSON as the format default and surfaces the new view via
+explicit `-o table`/`-o wide`. Flipping the non-agent default to
+`table` is intentionally deferred (see Consequences) so the new
+formatter can soak in production-ish use before changing the contract for
+existing pipelines.
```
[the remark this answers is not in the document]
radiohead: Fixed in a534142 — the Decision paragraph, Default format subsection, and the three Consequences follow-up bullets (Tests, Lint, Default flip) have all been updated to say "shipped in this PR" rather than describing them as future work.

---
**Generated by Claude Code**

docs/adrs/traces-get-table/001-tree-table-render-for-traces-get.md
radiohead: IMHO it doesn't hurt to have one – it leaves a paper trail of "why did we do X and why did we do it in Y way". But if you are strongly opposed to keeping it, I am happy to drop it.

internal/datasources/tempo/get.go
```diff
@@ -26,7 +26,10 @@ type getOpts struct {
 }
 
 func (opts *getOpts) setup(flags *pflag.FlagSet) {
-	opts.IO.DefaultFormat("json")
+	dsquery.RegisterCodecs(&opts.IO, false)
+	// Non-agent TTY default is the human-readable tree table (FR-001).
+	// In agent mode, cmdio.Options.BindFlags overrides to JSON regardless.
+	opts.IO.DefaultFormat("table")
```
javiermolinar, line 33: DefaultFormat("table") is unconditional. cmdio.Options.BindFlags only overrides for agent mode, not piped stdout

internal/datasources/query/codecs.go
```diff
@@ -95,6 +99,8 @@ func (c *queryGraphCodec) Encode(w io.Writer, data any) error {
 		}
 	case *tempo.SearchResponse:
 		return errors.New("graph output is not supported for trace search results; use -o table/json/yaml")
+	case *tempo.GetTraceResponse:
```
javiermolinar: This code seems to be unreachable/dead

```
❯ go run ./cmd/gcx traces get 95a24ceec16628c302020c9c737972b9 -o graph
Error: Unknown output format 'graph'. Valid formats are
│
│ json, table, wide, yaml
│
└─

exit status 1
```

internal/query/tempo/formatter.go
```diff
@@ -163,3 +174,354 @@ func formatStartTime(startTimeUnixNano string) string {
...
+}
+
+func aggregateTotals(spans []traceSpan) traceTotals {
+	t := traceTotals{
+		start: spans[0].start,
+		end:   spans[0].end,
+	}
+	services := make(map[string]struct{})
+	for _, s := range spans {
+		if s.start > 0 && (t.start == 0 || s.start < t.start) {
```
javiermolinar: What happens with a root span at 0

javiermolinar (commented):
Looking good, very cool feature. Although I think this can breaks in small screens. I have done a simple test and the last column is trimmed

Image

internal/datasources/tempo/get.go
radiohead: Yup, updated the PR description and the comment. Good catch.

internal/datasources/query/codecs.go
radiohead: Removed, thanks!

internal/query/tempo/formatter.go
radiohead: The agent tried to be too smart. Cleaned up, this should now be more straightforward.

radiohead:
> Looking good, very cool feature. Although I think this can breaks in small screens. I have done a simple test and the last column is trimmed
> 
> Image

I've updated the table renderer to use dynamic column sizes, so it should work now, though truthfully, on super-small terminal sizes it'll look bad no matter what – that's just an inherent limitation. Still, at least no clipping now.

javiermolinar:
@codex review

internal/query/tempo/formatter.go
```diff
@@ -163,3 +174,364 @@ func formatStartTime(startTimeUnixNano string) string {
...
+	return "-"
+}
+
+func decodeIDField(v any) string {
+	s, ok := v.(string)
+	if !ok || s == "" {
+		return ""
+	}
+	if b, err := base64.StdEncoding.DecodeString(s); err == nil {
+		return hex.EncodeToString(b)
```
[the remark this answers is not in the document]
radiohead: Fixed in ac87ee66 — `decodeIDField` now disambiguates by expected wire-form length instead of preferring base64 unconditionally. The function takes an `expectedBytes` argument (16 for trace IDs, 8 for span IDs); a string of length `2*expectedBytes` that parses as hex is treated as hex (and lowercased), otherwise we attempt base64 and require the decoded length to match, otherwise we pass the raw value through. Hex is a strict subset of the base64 alphabet but a 32-char hex trace ID base64-decodes to 24 bytes — not 16 — so the length check disambiguates cleanly.

Added round-trip tests in `formatter_test.go` for both encodings, uppercase-hex normalisation, mixed encodings within a single payload, and a regression guard that asserts the rendered output never contains a 48-char hex string (the signature of the old double-decode path).

---
**Generated by Claude Code**

radiohead:
@javiermolinar @annanay25 I think I've addressed all comments, would appreciate another look, thank you!

javiermolinar (approved):

PR #612, opened 2026-04-30
feat(login): support custom oauth callback port

### LINKED ISSUE #611: 🚀 Enhancement: expose an option for oauth callback port
### What exists today?

`gcx` doesn't allow specifying the port.

https://github.com/grafana/gcx/blob/0e39a5e7eab31180f4b9e6a1dd9462b638fed75a/internal/auth/flow.go#L387

This counts up from `54321` but we would like to specify the port number. 

### How could this be better?

Give more access to users to specify port information.

Give more access to users to specify port information.

Makes it easier to use

An example improvement would look like

`gcx login --port 8250`

No, it's purely additive (new flag/option)

srstrickland:
closes https://github.com/grafana/gcx/issues/611

Adds a `--oauth-callback-port` flag to `gcx login` so users can pin the OAuth callback server to a fixed local port instead of the auto-pick range (54321–54399).

**Before:** `gcx login` always probes 54321–54399 for a free port, requiring whichever it lands on to be reachable from the browser. **After:** `gcx login --oauth-callback-port 8250` binds the callback to the user-specified port.

When developers run `gcx login` on a remote dev host, the OAuth flow only works if the callback port is forwarded back to the machine running the browser. Many organizations port-forward a fixed set of ports between local and remote dev machines; today that means adding all of 54321–54399 to the forwarding list (or hoping the right one is free). `auth.Options.Port` already supports a fixed port internally — this just exposes it as a CLI flag.

- `internal/login/login.go` — new `Inputs.OAuthCallbackPort int` field, threaded into `auth.Options.Port` when constructing the OAuth flow
- `cmd/gcx/login/command.go` — new `--oauth-callback-port` flag with range validation (1–65535, or 0 to auto-pick) wired into `login.Inputs`
- `docs/reference/cli/gcx_login.md` — regenerated

The flag is purely additive; default of 0 preserves today's auto-pick behavior.

- [x] `go test ./internal/login/... ./internal/auth/... ./cmd/gcx/login/...` passes
- [x] `go vet` on touched packages passes
- [x] `gcx login --help` shows the new flag with a clear description
- [x] `gcx login --oauth-callback-port 99999` returns the structured "Port must be between 1 and 65535" validation error
- [x] Manual e2e on a remote dev host: confirm callback binds to the user-specified port and OAuth completes through the forwarded port

### DIFF
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -47,6 +47,7 @@ type loginOpts struct {
 	Cloud               bool
 	Yes                 bool
 	AllowServerOverride bool
+	OAuthCallbackPort   int
 }
 
 func (opts *loginOpts) setup(flags *pflag.FlagSet) {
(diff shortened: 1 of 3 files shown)


javiermolinar (approved):
LGTM thank you for your contribution

PR #613, opened 2026-04-30
feat(kg): add suppressions list/delete and fix overwrite bug

xujiaxj:
## Summary

- **Bug fix**: `gcx kg suppressions create` was using the batch endpoint (`POST /v1/config/disabled-alerts`) which fully replaces all stored suppressions — silently wiping any not included in the uploaded file. Switched to the single-item endpoint (`POST /v1/config/disabled-alert`) which the backend implements as a read-modify-write upsert, leaving unrelated suppressions untouched.
- **New**: `gcx kg suppressions list` — fetch all current suppressions (table/yaml/json output)
- **New**: `gcx kg suppressions delete <name>` — delete a single suppression by name
- **UX**: `create` now accepts input via stdin (pipe YAML directly, no temp file needed); errors immediately with a helpful example if run interactively without `-f`

## Test plan

- [x] `gcx kg suppressions list` returns existing suppressions without modifying them
- [x] `gcx kg suppressions create -f file.yaml` upserts only the named suppressions, leaves others intact
- [x] `echo '...' | gcx kg suppressions create` works via stdin
- [x] `gcx kg suppressions create` with no input errors immediately (no hang)
- [x] `gcx kg suppressions delete <name>` removes the named suppression, leaves others intact
- [x] All tests pass: `go test -race -count=1 ./...`
- [x] Linter clean: `golangci-lint run`

### DIFF
--- a/docs/reference/cli/gcx_kg_suppressions_create.md
+++ b/docs/reference/cli/gcx_kg_suppressions_create.md
@@ -1,15 +1,27 @@
 ## gcx kg suppressions create
 
-Upload suppressions from a YAML file.
+Upload suppressions from a YAML file or stdin.
 
 ```
 gcx kg suppressions create [flags]
 ```
 
+### Examples
+
+```
+  gcx kg suppressions create -f suppressions.yaml
+
+  echo 'disabledAlertConfigs:
+    - name: my-suppression
+      matchLabels:
+        alertname: ErrorRatioBreach
+        job: my-service' | gcx kg suppressions create
+```
+
 ### Options
 
 ```
-  -f, --file string   Input file (YAML)
+  -f, --file string   Input file (YAML), or '-' for stdin. Reads from stdin if omitted.
   -h, --help          help for create
 ```
 
@@ -27,5 +39,5 @@ gcx kg suppressions create [flags]
 
 ### SEE ALSO
 
-* [gcx kg suppressions](gcx_kg_suppressions.md)	 - Push suppressions to the Knowledge Graph.
+* [gcx kg suppressions](gcx_kg_suppressions.md)	 - Manage alert suppressions in the Knowledge Graph.
 
--- a/docs/reference/cli/gcx_kg_suppressions_create.md
+++ b/docs/reference/cli/gcx_kg_suppressions_create.md
@@ -1,6 +1,6 @@
 ## gcx kg suppressions create
 
-Upload suppressions from a YAML file or stdin.
+Create or update one or more suppressions from a YAML file or stdin.
 
 ```
 gcx kg suppressions create [flags]
--- a/internal/providers/kg/client.go
+++ b/internal/providers/kg/client.go
@@ -26,6 +26,8 @@ const (
 	assertionsPath   = pluginResourcePath + "/asserts/api-server/v1/assertions"
 	searchPath       = pluginResourcePath + "/asserts/api-server/v1/search"
 	rulesPath        = pluginResourcePath + "/asserts/api-server/v1/config/prom-rules/"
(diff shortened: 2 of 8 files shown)


internal/providers/kg/client.go
```diff
@@ -209,9 +209,49 @@ func (c *Client) UploadModelRules(ctx context.Context, yamlContent string) error
...
+
+// Suppressions is the batch payload shape returned by GET /v1/config/disabled-alerts.
+type Suppressions struct {
+	DisabledAlertConfigs []Suppression `json:"disabledAlertConfigs" yaml:"disabledAlertConfigs"`
+}
+
+// UpsertSuppression creates or updates a single suppression without affecting others.
+// It uses the single-item endpoint so the backend performs a read-modify-write upsert.
+func (c *Client) UpsertSuppression(ctx context.Context, s Suppression) error {
+	return c.postJSON(ctx, pluginResourcePath+"/asserts/api-server/v1/config/disabled-alert", s, nil)
```
annanay25: lets move these paths to the constants defined at the top of the file.

internal/providers/kg/client.go
```diff
@@ -209,9 +209,49 @@ func (c *Client) UploadModelRules(ctx context.Context, yamlContent string) error
...
+}
+
+// DeleteSuppression deletes a single suppression by name.
+func (c *Client) DeleteSuppression(ctx context.Context, name string) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete,
+		c.host+pluginResourcePath+"/asserts/api-server/v1/config/disabled-alert/"+url.PathEscape(name), nil)
```
annanay25: same comment as above.

internal/providers/kg/client.go
```diff
@@ -209,9 +209,49 @@ func (c *Client) UploadModelRules(ctx context.Context, yamlContent string) error
...
+	if resp.StatusCode >= 400 {
+		return readError(resp)
+	}
+	return nil
+}
+
+// GetSuppressions retrieves all disabled-alert configurations for the tenant.
+func (c *Client) GetSuppressions(ctx context.Context) (*Suppressions, error) {
+	var result Suppressions
+	if err := c.getJSON(ctx, pluginResourcePath+"/asserts/api-server/v1/config/disabled-alerts", &result); err != nil {
```
annanay25: here as well.

internal/providers/kg/commands.go
```diff
@@ -599,20 +608,98 @@ func newModelRulesCommand(loader RESTConfigLoader) *cobra.Command {
...
+			}
+			client, err := NewClient(cfg)
+			if err != nil {
+				return err
+			}
+			for _, s := range suppressions.DisabledAlertConfigs {
+				if err := client.UpsertSuppression(cmd.Context(), s); err != nil {
+					return fmt.Errorf("failed to upsert suppression %q: %w", s.Name, err)
+				}
+			}
```
annanay25: if this upsert operation fails at the kth entity should we print a more clear message that 1...k-1 entities were successfully created?

docs/reference/cli/gcx_kg_suppressions_create.md
```diff
@@ -1,15 +1,27 @@
 ## gcx kg suppressions create
 
-Upload suppressions from a YAML file.
+Upload suppressions from a YAML file or stdin.
```
annanay25: Can we explicitly mention that this endpoint can be used to create more than one suppressions at a time? I was a little confused reading this.

annanay25 (commented):
left a few nits!

annanay25 (approved):
lgtm!

PR #625, opened 2026-05-04
fix(providers): honor --context across all CRUD adapters

alexander-akhmetov:
`--context` was silently ignored on multiple providers (aio11y, apo11y, k6, faro, fleet, irm, slo, synth, etc). These adapters always ran against the active context regardless of what the user passed.

The provider `ConfigLoader.BindFlags` re-bound `--context` as a persistent flag on each provider's parent command, shadowing root's binding. When parent and child register the same persistent flag, pflag picks the child's value, so root's `PersistentPreRun` (which threads the value into `context.Context`) saw an empty string.

This PR drops the duplicate binding. Root now owns `--context` exclusively and the adapters already read it via `ContextNameFromCtx(ctx)` and pick up the threaded value with no code change.

#### How to test manually

With at least two contexts:

```bash
gcx config list-contexts
gcx config use-context stackA

# Each of these must return stackB's resources, not stackA's:
gcx aio11y rules list --context=stackB
gcx --context=stackB aio11y rules list
```

Before the fix, they returned `stackA`'s data.

### DIFF
--- a/cmd/gcx/root/command_test.go
+++ b/cmd/gcx/root/command_test.go
@@ -12,6 +12,7 @@ import (
 	claudeplugin "github.com/grafana/gcx/claude-plugin"
 	"github.com/grafana/gcx/cmd/gcx/root"
 	"github.com/grafana/gcx/internal/agent"
+	internalconfig "github.com/grafana/gcx/internal/config"
 	"github.com/grafana/gcx/internal/providers"
 	"github.com/grafana/gcx/internal/resources/adapter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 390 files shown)


alexander-akhmetov:
@claude review

annanay25:
great catch @alexander-akhmetov!

how do we feel like addressing point 2 in claude review as part of this PR?

> 2. SetContextName calls are now redundant but harmless
> 
> Several adapters call loader.SetContextName(internalconfig.ContextNameFromCtx(ctx)) before passing ctx to load methods. Since resolvedContextName(ctx) already falls through to ContextNameFromCtx(ctx) when l.ctxName == "", these calls just set l.ctxName to the same value that resolvedContextName would retrieve anyway. Not a bug — just cosmetic redundancy that won't cause issues.


Happy to do it in a follow up too.

alexander-akhmetov:
@annanay25 I updated the PR :) https://github.com/grafana/gcx/pull/625/changes/f442221b325c5a975fe657ac91f3d54277f69660

annanay25 (approved):
lgtm!

PR #627, opened 2026-05-04
feat(kg): add RCA Workbench deep link to entities inspect

xujiaxj:
## Summary

Adds `--share-link` and `--open` flags to `gcx kg entities inspect`, following the same pattern as the datasource query commands (PR #567):

- **`--share-link`** — prints the RCA Workbench URL for the inspected entity to stderr (opt-in, quiet by default)
- **`--open`** — opens the entity directly in the RCA Workbench in your browser
- The URL encodes entity type, name, scope (env/namespace/site), and time range — relative (`now-24h`/`now`) when `--since` is used, epoch ms timestamps when `--from`/`--to` are used
- No URL injected into JSON output — surfaces only when explicitly requested, consistent with DESIGN.md "quiet by default"

## Test plan

- [x] `--since 24h --share-link` — verify URL on stderr with `start=now-24h&end=now`
- [x] `--from now-24h --to now --share-link` — verify URL uses epoch ms timestamps
- [x] `--open` — verify browser opens to correct entity in RCA Workbench
- [x] `--share-link --open` — verify both work together
- [x] No flags — verify clean JSON output, nothing on stderr

image

### DIFF
--- a/docs/reference/cli/gcx_kg_entities.md
+++ b/docs/reference/cli/gcx_kg_entities.md
@@ -23,6 +23,6 @@ Manage Knowledge Graph entities.
 ### SEE ALSO
 
 * [gcx kg](gcx_kg.md)	 - Manage Grafana Knowledge Graph rules, entities, and insights
-* [gcx kg entities inspect](gcx_kg_entities_inspect.md)	 - Show detailed info, insights, and summary for a single entity, with a link to open it in the RCA Workbench.
+* [gcx kg entities inspect](gcx_kg_entities_inspect.md)	 - Show detailed info, insights, and summary for a single entity, including a link to the RCA Workbench.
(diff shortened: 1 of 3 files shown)


annanay25 (commented):
tried running this and claude got confused with `gcx kg open` 

Image

Should we sunset that subcommand?

annanay25 (commented):
even cursor had the same issue

Image

xujiaxj:
> tried running this and claude got confused with `gcx kg open`
> 
> Image
> Should we sunset that subcommand?

Interesting, Claude failed to pass the `--open` flag when running the command. If it had used the flag, it should've not confused with the `open` subcommand. Can you give it another try after rebuilding the gcx command?

annanay25 (approved):
worked! i checked out the PR using gh cli but for some reason it didn't fetch the right changes locally. lgtm!

PR #631, opened 2026-05-05
feat: rename synth to synthetic-monitoring

Pokom:
I do not believe we externally communicate synth as a nickname for our synthetic monitoring product, and so let's update gcx to recognize the official name as "synthetic-monitoring" across documentation and skills. The command was updated to have two aliases:
- `sm`
- `synth`

`sm` is the preferred shorthand for the product, and I'm leaving in `synth` to not break any existing configurations workflows.


Pokom:
@claude can you review again?

Pokom:
@claude can you review again?

radiohead (approved):

PR #632, opened 2026-05-05
fix(skills): handle read-only files during skill update

dafydd-t:
Address permissions issues I saw when running `gcx skills update`

## Summary

- `os.WriteFile` doesn't override permissions on existing files, so updating a skill file that's read-only (e.g. `0o444`) fails with a permission error
- added `os.Chmod` before the write to ensure the file is writable
- added test covering the read-only file update path


annanay25 (approved):
lgtm!

PR #639, opened 2026-05-06
feat(diag): log failed agent invocations for capability-gap analysis

wardbekker:
## Summary

- Adds opt-in JSONL logging of failed agent-mode gcx invocations to disk, so unknown commands, unknown flags, and other capability gaps can be discovered and analysed
- Off by default — enable via `diagnostics.agent-invocation-log: true` in the gcx config file
- All flag values are stripped before logging; only command names and flag names are recorded — no secrets or user data reach the log

## Configuration

```yaml
diagnostics:
  agent-invocation-log: true
  log-dir: /optional/override   # defaults to $XDG_STATE_HOME/gcx/
```

Log file: `$XDG_STATE_HOME/gcx/agent-invocation-errors.jsonl`

Example log line:
```json
{"timestamp":"2026-05-06T10:30:00Z","version":"0.2.12","args":["kg","search","--format=<value>"],"error_kind":"usage_error","error":"unknown command \"search\" for \"gcx kg\"","exit_code":2}
```

## Implementation

- `internal/agentlog/` — new package: `Entry` type, `Append()`, `StripArgValues()` (strips all flag values to `<value>`), `KindFromExitCode()`
- `internal/config/types.go` — new `DiagnosticsConfig` struct + `Diagnostics` field on `Config`
- `cmd/gcx/main.go` — best-effort config load before `ValidateArgs` + log call in `handleError()` (covers both pre-validation and runtime errors)
- Config reference auto-regenerated

## Test plan

- [ ] `go test ./internal/agentlog/...` — unit tests for `StripArgValues`, `Append`, `KindFromExitCode`
- [ ] Enable in config and trigger `gcx kg nonexistent` with `CLAUDECODE=1`; confirm JSONL entry appears
- [ ] Confirm `--token secret` becomes `--token <value>` in the log (no secret leakage)
- [ ] Confirm no log is written when `diagnostics.agent-invocation-log` is absent
- [ ] Confirm no log is written outside agent mode (`CLAUDECODE=` unset)

### DIFF
--- a/cmd/gcx/main.go
+++ b/cmd/gcx/main.go
@@ -16,6 +16,8 @@ import (
 	"github.com/grafana/gcx/internal/agentlog"
 	internalconfig "github.com/grafana/gcx/internal/config"
 	appversion "github.com/grafana/gcx/internal/version"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"golang.org/x/mod/module"
 )
 
@@ -36,16 +38,22 @@ func main() {
 	// root.Command() because io.Options.BindFlags() reads agent.IsAgentMode()
 	// during command construction to set the default output format.
 	preParseAgentFlag()
-	agentlog.Configure(loadDiagnosticsConfig())
+	if agent.IsAgentMode() {
+		agentlog.Configure(loadDiagnosticsConfig())
+	}
 
 	formattedVersion := formatVersion()
 	appversion.Set(version)
 	appversion.SetBuildInfo(commit, date)
-	if err := root.ValidateArgs(root.Command(formattedVersion), os.Args[1:]); err != nil {
-		handleError(err)
+
+	cmd := root.Command(formattedVersion)
+	boolFlags := collectBoolFlags(cmd)
+	subCmds := collectSubCmds(cmd)
+	if err := root.ValidateArgs(cmd, os.Args[1:]); err != nil {
+		handleError(err, boolFlags, subCmds)
 	}
 
-	handleError(root.Command(formattedVersion).ExecuteContext(ctx))
+	handleError(cmd.ExecuteContext(ctx), boolFlags, subCmds)
 }
 
 // preParseAgentFlag scans os.Args for --agent / --agent=true / --agent=false
@@ -69,7 +77,7 @@ func preParseAgentFlag() {
 	}
 }
 
-func handleError(err error) {
+func handleError(err error, boolFlags map[string]struct{}, subCmds map[string]bool) {
 	if err == nil {
 		return
 	}
@@ -95,9 +103,9 @@ func handleError(err error) {
 		_ = agentlog.Append(agentlog.Entry{
 			Timestamp: time.Now(),
 			Version:   appversion.Get(),
-			Args:      agentlog.StripArgValues(os.Args[1:]),
+			Args:      agentlog.StripArgValues(os.Args[1:], boolFlags, subCmds),
 			ErrorKind: agentlog.KindFromExitCode(exitCode),
-			Error:     detailedErr.Summary,
+			Error:     truncate(detailedErr.Summary, 200),
 			ExitCode:  exitCode,
 		})
 	}
@@ -116,28 +124,66 @@ func handleError(err error) {
 	os.Exit(exitCode)
 }
 
-// loadDiagnosticsConfig reads diagnostics settings from the gcx config file on
-// a best-effort basis. Any error (missing file, parse failure) returns a
-// disabled config so the caller is never affected by a config read failure.
-func loadDiagnosticsConfig() agentlog.Config {
-	// GCX_CONFIG bypasses discovery, mirroring LoadLayered behaviour.
-	if explicit := os.Getenv(internalconfig.ConfigFileEnvVar); explicit != "" {
-		return loadDiagnosticsFromFile(explicit)
+// collectBoolFlags walks the full command tree and returns a set of all boolean
+// flag names (and their shorthands) so StripArgValues can skip consuming the
+// next token for flags that take no value argument.
+func collectBoolFlags(cmd *cobra.Command) map[string]struct{} {
+	bools := make(map[string]struct{})
+	var visit func(c *cobra.Command)
+	visit = func(c *cobra.Command) {
+		addBools := func(f *pflag.Flag) {
+			if f.Value.Type() == "bool" || f.NoOptDefVal != "" {
+				bools[f.Name] = struct{}{}
+				if f.Shorthand != "" {
+					bools[f.Shorthand] = struct{}{}
+				}
+			}
+		}
+		c.Flags().VisitAll(addBools)
+		c.PersistentFlags().VisitAll(addBools)
+		for _, sub := range c.Commands() {
+			visit(sub)
+		}
 	}
-	sources, err := internalconfig.DiscoverSources()
-	if err != nil || len(sources) == 0 {
-		return agentlog.Config{}
-	}
-	// Use the highest-priority source (last in slice — local overrides user overrides system).
-	return loadDiagnosticsFromFile(sources[len(sources)-1].Path)
+	visit(cmd)
+	return bools
 }
 
-func loadDiagnosticsFromFile(path string) agentlog.Config {
-	cfg, err := internalconfig.Load(context.Background(), func() (string, error) { return path, nil })
+// collectSubCmds walks the full command tree and returns a set of all registered
+// subcommand names (and aliases). Positional args matching this set are safe to
+// log; all other positionals are treated as values and redacted.
+func collectSubCmds(cmd *cobra.Command) map[string]bool {
+	names := make(map[string]bool)
+	var walk func(*cobra.Command)
+	walk = func(c *cobra.Command) {
+		for _, sub := range c.Commands() {
+			names[sub.Name()] = true
+			for _, alias := range sub.Aliases {
+				names[alias] = true
+			}
+			walk(sub)
+		}
+	}
+	walk(cmd)
+	return names
+}
+
+// loadDiagnosticsConfig reads diagnostics settings from the layered gcx config.
+// Any error (missing file, parse failure) returns a disabled config so the
+// caller is never affected by a config read failure.
+func loadDiagnosticsConfig() agentlog.Config {
+	cfg, err := internalconfig.LoadLayered(context.Background(), "")
 	if err != nil || cfg.Diagnostics == nil || !cfg.Diagnostics.AgentInvocationLog {
 		return agentlog.Config{}
 	}
 	return agentlog.Config{Enabled: true, LogDir: cfg.Diagnostics.LogDir}
+}
+
+func truncate(s string, n int) string {
+	if len(s) <= n {
+		return s
+	}
+	return s[:n]
 }
 
 func formatVersion() string {
--- a/cmd/gcx/main.go
+++ b/cmd/gcx/main.go
@@ -13,6 +13,8 @@ import (
 	"github.com/grafana/gcx/cmd/gcx/fail"
 	"github.com/grafana/gcx/cmd/gcx/root"
 	"github.com/grafana/gcx/internal/agent"
+	"github.com/grafana/gcx/internal/agentlog"
+	internalconfig "github.com/grafana/gcx/internal/config"
 	appversion "github.com/grafana/gcx/internal/version"
 	"golang.org/x/mod/module"
 )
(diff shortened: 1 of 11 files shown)


wardbekker:
## Future direction: foundation for opt-in telemetry

This local log is intentionally designed as a stepping stone. Once we have a sense of what agents are hitting, the next natural evolution would be an **anonymous, opt-in telemetry channel** that sends the same structured events to Grafana Cloud — letting the broader gcx user base collectively influence the roadmap rather than relying on a single operator's log.

Key design properties that would carry forward:
- **Opt-in only** — same config-file gate, extended with a separate `send-to-grafana` flag
- **No secrets, no PII** — `StripArgValues()` already ensures only command structure (not values) is recorded; the same guarantee holds for any future upload
- **Grafana Cloud as the backend** — events could land in a Loki stream or a Faro ingest endpoint, making the data queryable with the same tools gcx manages
- **Aggregate, not individual** — telemetry would be aggregated server-side so no single invocation is attributable

The local-log-first approach also lets us validate the data shape and privacy guarantees before any network transmission is introduced.

wardbekker:
## Shepherd Status (iteration 1 — complete, awaiting re-review)

### All CI checks: ✅ PASS

| Check | Status |
|-------|--------|
| Tests | ✅ pass (3m 43s) |
| Linters | ✅ pass (2m 48s) |
| Documentation | ✅ pass (50s) |
| TruffleHog / zizmor / Socket | ✅ pass |

### Fixes applied (addressing @radiohead's CHANGES_REQUESTED)

| Issue | Classification | Fix |
|-------|---------------|-----|
| Positional values leak to log (`config set cloud.token VALUE`) | **blocking** | `StripArgValues` now accepts `subCmds map[string]bool`; positionals not matching a known subcommand become `<value>` |
| `loadDiagnosticsConfig` reads only highest-priority source | **blocking** | Replaced with `LoadLayered` call |
| `MergeConfigs` didn't propagate `Diagnostics` across layers | **blocking** | Added `mergeDiagnosticsConfig` in `merge.go` |
| `agentlog.Configure` runs unconditionally on human hot path | suggestion | Guard with `if agent.IsAgentMode()` |
| Error summary may expose raw library strings | nitpick | Capped at 200 chars via `truncate()` |
| XDG state dir created with `0755` | suggestion | Changed to `0700` |
| Conflated godoc comment spanning two functions | nitpick | Split into per-function comments |
| Rebase conflict (`appversion.SetBuildInfo` lost) | rebase | Resolved; also fixed double `root.Command()` call |

### New tests added
- `TestStripArgValues/positional_value_after_subcommand_chain_is_redacted`
- `TestStripArgValues/nil_subCmds_keeps_all_positionals`
- `TestMergeConfigs_DiagnosticsLayering`
- `TestMergeConfigs_DiagnosticsOverride`

### Status
🟡 **Waiting for @radiohead to re-review** — no code changes needed on our end

---
*Updated by shepherd-pr*

internal/agentlog/agentlog.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		if consumeNext {
+			out = append(out, "<value>")
+			consumeNext = false
+			continue
+		}
+		isLongFlag := strings.HasPrefix(arg, "--")
+		isShortFlag := len(arg) >= 2 && arg[0] == '-' && arg[1] != '-'
+		if !isLongFlag && !isShortFlag {
+			out = append(out, arg)
```
radiohead: **issue (blocking):** `StripArgValues` keeps all positional arguments verbatim. Several `gcx` commands accept secret values as positionals rather than flags — the most direct case is `gcx config set`:

```
gcx config set cloud.token glsa_MYTOKEN    # token reaches the log verbatim
gcx config set grafana.password my-pass
```

The PR description's guarantee ("no secrets or user data reach the log") does not hold for this code path. Three options in order of robustness:

- Walk the cobra command tree: any positional that matches a registered child command at that position is a safe subcommand name; everything else becomes `<value>`.
- Heuristic: replace any positional containing `.`, `/`, `@`, or exceeding 30 characters, or following a key matching `*.token`/`*.password`.
- Conservative: drop all positionals beyond the subcommand chain entirely (guarantees safety, sacrifices some debuggability).

cmd/gcx/main.go
```diff
@@ -101,6 +120,55 @@ func handleError(err error) {
...
+	// GCX_CONFIG bypasses discovery, mirroring LoadLayered behaviour.
+	if explicit := os.Getenv(internalconfig.ConfigFileEnvVar); explicit != "" {
+		return loadDiagnosticsFromFile(explicit)
+	}
+	sources, err := internalconfig.DiscoverSources()
+	if err != nil || len(sources) == 0 {
+		return agentlog.Config{}
+	}
+	// Use the highest-priority source (last in slice — local overrides user overrides system).
+	return loadDiagnosticsFromFile(sources[len(sources)-1].Path)
```
radiohead: **suggestion (blocking):** `loadDiagnosticsConfig` reads only the highest-priority source (the last element of `DiscoverSources()`), not the layered-merged result. If the user enables `diagnostics.agent-invocation-log: true` in their user config but has a local config that omits the `diagnostics:` block, this function finds `Diagnostics == nil` in the local file and silently returns a disabled config.

The fix requires two steps together: (1) use `LoadLayered` here instead of reading the last source directly, and (2) extend `MergeConfigs` in `internal/config/merge.go` to propagate the `Diagnostics` field — without that second step, switching to `LoadLayered` alone still drops the user-layer setting. A regression test in `internal/config/merge_test.go` that verifies the user-layer value survives a local config without a `diagnostics:` block would lock this in.

cmd/gcx/main.go
```diff
@@ -87,6 +95,17 @@ func handleError(err error) {
 		exitCode = *detailedErr.ExitCode
 	}
 
+	if agent.IsAgentMode() && agentlog.IsEnabled() {
+		_ = agentlog.Append(agentlog.Entry{
+			Timestamp: time.Now(),
+			Version:   appversion.Get(),
+			Args:      agentlog.StripArgValues(os.Args[1:], boolFlags),
+			ErrorKind: agentlog.KindFromExitCode(exitCode),
+			Error:     detailedErr.Summary,
```
radiohead: **nitpick (non-blocking):** `detailedErr.Summary` is stored verbatim. Common error paths have curated summaries, but the fallback path (`splitErrorMessage`) can surface raw library error strings. If a library formats an error without the expected separator and the message contains a token value, the full string becomes `Entry.Error`. Capping at a fixed length would reduce the blast radius:

```go
Error: truncate(detailedErr.Summary, 200),
```

Also worth noting in `docs/reference/configuration/index.md` that error messages are stored as-is and may contain resource names or identifiers, so users know what to expect when sharing the log file.

internal/agentlog/agentlog.go
```diff
@@ -0,0 +1,173 @@
...
+	f.Close()
+	if writeErr != nil {
+		return writeErr
+	}
+	return trimLog(path, maxEntries)
+}
+
+// trimLog keeps only the last max entries in the JSONL file, dropping the oldest.
+// It is a no-op when the file has max or fewer entries.
+func trimLog(path string, limit int) error {
```
radiohead, line 80: **nitpick (non-blocking):** `trimLog` reads the full file then overwrites it non-atomically. Concurrent `Append` calls from two `gcx` processes running simultaneously (plausible in multi-threaded agent harnesses) can race: both append, both read, both shrink, and the second `os.WriteFile` discards the first process's append. The window is small, but silent log-entry loss is worth documenting. Either acquire a file lock around append+trim, or add a note in the package docstring that concurrent writers may lose entries from the trim head.

internal/agentlog/agentlog.go
```diff
@@ -0,0 +1,173 @@
...
+	Args      []string  `json:"args"`
+	ErrorKind string    `json:"error_kind"`
+	Error     string    `json:"error"`
+	ExitCode  int       `json:"exit_code"`
+}
+
+// Append writes entry to the log file as a JSONL record and trims to maxEntries.
+func Append(entry Entry) error {
+	path := LogPath()
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
```
radiohead: **suggestion (non-blocking):** XDG state-home convention is `0700` for application directories to avoid information disclosure. With `0755`, any local user can enumerate the directory and confirm that `agent-invocation-errors.jsonl` exists.

```go
if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
```

internal/agentlog/agentlog.go
```diff
@@ -0,0 +1,173 @@
...
+var cfg Config
+
+// Configure sets agentlog options. Called once from main() before command execution.
+func Configure(c Config) { cfg = c }
+
+// IsEnabled reports whether agent invocation logging is active.
+func IsEnabled() bool { return cfg.Enabled }
+
+// LogPath returns the full path to the log file.
+func LogPath() string {
```
radiohead, line 37: **nitpick (non-blocking):** `LogDir` accepts arbitrary paths without validation. A tampered config could redirect log output to a world-readable directory. Clamping to a path under `xdg.StateHome` contains the blast radius:

```go
if !strings.HasPrefix(filepath.Clean(dir), xdg.StateHome) {
    dir = filepath.Join(xdg.StateHome, "gcx")
}
```

cmd/gcx/main.go
```diff
@@ -34,14 +38,18 @@ func main() {
 	// root.Command() because io.Options.BindFlags() reads agent.IsAgentMode()
 	// during command construction to set the default output format.
 	preParseAgentFlag()
+	agentlog.Configure(loadDiagnosticsConfig())
```
radiohead: **suggestion (non-blocking):** `loadDiagnosticsConfig()` runs unconditionally on every invocation, but the result is only used when `agent.IsAgentMode()` is true. This adds a config-tree stat + YAML parse on the human-user hot path:

```go
if agent.IsAgentMode() {
    agentlog.Configure(loadDiagnosticsConfig())
}
```

cmd/gcx/main.go
```diff
@@ -101,6 +120,55 @@ func handleError(err error) {
 	os.Exit(exitCode)
 }
 
+// loadDiagnosticsConfig reads diagnostics settings from the gcx config file on
```
radiohead: **nitpick (non-blocking):** This comment block describes two functions (`loadDiagnosticsConfig` and `collectBoolFlags`) but sits above `collectBoolFlags`, leaving `loadDiagnosticsConfig` (defined below at line 151) with no godoc. Split so each function has its own leading comment.

internal/providers/slo/definitions/status.go
```diff
@@ -184,7 +184,7 @@ func ComputeStatus(s Slo, sli *float64, objective float64) string {
 	if s.ReadOnly != nil && s.ReadOnly.Status != nil {
 		switch strings.ToLower(s.ReadOnly.Status.Type) {
 		case "creating", "updating", "deleting", "error":
-			return strings.Title(s.ReadOnly.Status.Type) //nolint:staticcheck
+			return strings.Title(s.ReadOnly.Status.Type) //nolint:staticcheck // deprecated but no x/text dep here
```
radiohead, line 187: **nitpick (non-blocking):** This `//nolint:staticcheck` update is unrelated to the agent-invocation-log feature. Fine change on its own, but worth splitting into a separate commit to keep the PR's diff focused.

cmd/gcx/main.go
```diff
@@ -101,6 +120,55 @@ func handleError(err error) {
...
+	}
+	sources, err := internalconfig.DiscoverSources()
+	if err != nil || len(sources) == 0 {
+		return agentlog.Config{}
+	}
+	// Use the highest-priority source (last in slice — local overrides user overrides system).
+	return loadDiagnosticsFromFile(sources[len(sources)-1].Path)
+}
+
+func loadDiagnosticsFromFile(path string) agentlog.Config {
```
radiohead: **nitpick (non-blocking):** `loadDiagnosticsConfig` and `loadDiagnosticsFromFile` contain source-priority decision logic that sits closer to domain code than the CLI wiring `cmd/` is meant to hold. Moving these to `internal/agentlog` as `agentlog.LoadFromConfig()` would keep `cmd/gcx/main.go` to pure wiring and align with the strict layer-separation rule in CONSTITUTION.md. Not a hard violation — flagging for consistency.

radiohead (changes requested):
The agentlog feature is well-structured and all four issues from the prior review have been addressed. The remaining findings center on positional-argument redaction (secrets can still reach the log via `gcx config set`) and config-layer merging (the feature can silently stay disabled when a local config omits the diagnostics block).

---
**Generated by Claude Code**

radiohead (approved):

PR #642, opened 2026-05-06
fix(k6): make k6 provider work with OAuth login

fornfrey:
https://github.com/grafana/gcx-internal/issues/12

## Summary

Reworks the k6 provider to use OAuth authentication and route the requests through the `grafana-k6-app` plugin proxy instead of performing CAP to k6 token exchange. 

After this PR, `gcx k6 ...` works with any stack the user is logged into via OAuth, but it becomes the only supported auth method: SA tokens are still unsupported, and authentication with CAP tokens is removed.  

## Changes

- **Plugin-proxy routing** (`client.go`): all calls now route through `<grafana>/api/plugins/k6-app/resources/cloud/...`; the plugin injects the necessary headers.
- **Removed support for CAP tokens**: the token exchange had been misconfigured - it had been using a CAP token where k6 API expects a SA token to allow k6 backend call the Grafana API. 
- **Lazy `Token(ctx)` and `orgID(ctx)`**: both fetched on demand through the proxy (`/v3/account/me` and `/organization`) and memoised. `Token(ctx)` returns a k6 Personal API token bound to the user identity. `orgID` is now private and only used by the methods that depend on the legacy API.


internal/providers/k6/client.go
```diff
@@ -30,82 +30,99 @@ const (
...
-	stackID   int
-	token     string
-	http      *http.Client
+	grafanaURL string
+	proxyBase  string
+	http       *http.Client
+
+	cachedToken string // memoized result of /v3/account/me
+	cachedOrgID int    // memoized result of /organization, used only by env var methods
 }
```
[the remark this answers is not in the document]
fornfrey: fixed

internal/providers/k6/client.go
```diff
@@ -10,17 +10,18 @@ import (
...
+	"sync"
 
 	"github.com/grafana/gcx/internal/httputils"
 )
 
 const (
-	// DefaultAPIDomain is the default k6 Cloud API domain.
-	DefaultAPIDomain = "https://api.k6.io"
+	// pluginProxy forwards requests to api.k6.io with
+	// stack-scoped credentials injected by the GCk6 plugin.
```
dafydd-t: Can you add the full name of the plugin here? it will be useful for reference in the future

internal/providers/k6/client.go
```diff
@@ -30,82 +31,104 @@ const (
...
 // Client is an HTTP client for the k6 Cloud API.
-// It must be authenticated before use by calling Authenticate.
+// It routes every k6 API call through the grafana-k6-app plugin proxy.
 type Client struct {
-	apiDomain string
-	orgID     int
-	stackID   int
-	token     string
-	http      *http.Client
+	grafanaURL string
```
dafydd-t:
```suggestion
	host string
```

nit: this same property is called `host` in most other clients. can we keep it consistent?

dafydd-t (changes requested):
This is exciting to see! much better than before. I've verified that the OAuth flow works.

I think we still want to support SA token auth for cloud stacks - would we be able to support that too?
```
./bin/gcx k6 projects list
Error: K6 provider requires OAuth authentication -- run `gcx login` and select OAuth mode
│
└─
```

internal/providers/k6/client.go
fornfrey: Done ✅

internal/providers/k6/client.go
fornfrey: Done ✅

dafydd-t (approved):
Thanks @fornfrey !

PR #643, opened 2026-05-06
fix(adapter): honor DryRun option in typedAdapter.Delete

### LINKED ISSUE #573: gcx resources delete --dry-run can perform real deletes for provider-backed resources using TypedCRUD.

wardbekker:
## Summary
- `typedAdapter.Delete` was discarding `metav1.DeleteOptions` entirely (the parameter was `_`), causing real deletes to execute even when `--dry-run` was set
- Added the same `isDryRun` guard that `Create` and `Update` already use — returns `nil` on dry-run, matching server-side k8s dry-run semantics
- Added `TestTypedCRUD_DeleteDryRun` regression test mirroring the existing `TestTypedCRUD_CreateDryRun` and `TestTypedCRUD_UpdateDryRun`

## Changes
- `internal/resources/adapter/typed.go:428` — renamed `_` param to `opts` and added `isDryRun(opts.DryRun)` short-circuit
- `internal/resources/adapter/typed_test.go` — added `TestTypedCRUD_DeleteDryRun` with sub-tests for both the dry-run guard and normal delete path

## Test Plan
- [x] `go test ./internal/resources/adapter/ -run TestTypedCRUD_DeleteDryRun -v` — new test passes
- [x] `go test ./internal/resources/adapter/ -run "TestTypedCRUD_(Create|Update|Delete|DryRun)" -v` — all dry-run tests pass
- [x] `go test -mod=mod ./internal/resources/...` — full resources package passes
- [ ] Note: `mise run lint` is failing in main due to pre-existing vendor drift (`vendor/modules.txt` out of sync with go.mod); not caused by this change

Closes #573

### DIFF
--- a/internal/resources/adapter/typed.go
+++ b/internal/resources/adapter/typed.go
@@ -425,9 +425,13 @@ func (a *typedAdapter[T]) Update(ctx context.Context, obj *unstructured.Unstruct
 	return &u, nil
 }
 
-func (a *typedAdapter[T]) Delete(ctx context.Context, name string, _ metav1.DeleteOptions) error {
+func (a *typedAdapter[T]) Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error {
 	if a.crud.DeleteFn == nil {
 		return errors.ErrUnsupported
+	}
+
+	if isDryRun(opts.DryRun) {
+		return nil
 	}
 
 	return a.crud.DeleteFn(ctx, name)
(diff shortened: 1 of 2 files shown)


annanay25 (approved):
scary bug! thanks for catching it!

PR #644, opened 2026-05-06
fix(config): prevent env var secrets from being persisted to config file

### LINKED ISSUE #572: 🐛 Bug Report/Security Issue: env var GRAFANA_TOKEN gets written to config.yaml
### gcx Version

v0.2.9

config (set/unset/use-context/list-contexts/current-context/view/check)

### Command Executed

```shell
Not sure, I guess any command does this.
```

### Command Output

```shell
N/A
```

### What happened?

We have a common `.gcx.yaml` committed to repo so we can share to configuration with other users. This obviously has no token value present in the file as we do not want the secret committed nor do we want it somewhere as plain text on the disk. Therefore, users must set the `GRAFANA_TOKEN` env variable with their token.

And so have I set my token as the `GRAFANA_TOKEN` env variable.

I'm not sure when it happened, but `gcx` decided to write my `GRAFANA_TOKEN` value to the configuration file. Since I was trying to fix a pipeline error I did not see immediately that `gcx` had written the token to the config file. Only after I did a self-review of my MR did I see gcx had leaked my token.

I think this is really bad design and definitely a security issue, so I strongly feel this must be accidental.

When I follow what would be considered a good practice by not writing my token to a plain text file, I would not expect that same program to write it to disk just the same without warning.

1. Set a dummy value for `GRAFANA_TOKEN` as environment variable in your shell.
2. Switch context: `gcx -vvv config use-context [name]`
3. Check your configuration file.

Every time

### Environment Details

Ubuntu 24.04, Grafana 12.4

- [x] I checked my configuration with `gcx config view`
- [x] I verified I'm using the correct context with `config current-context`
- [x] I verified my authentication token/credentials are valid

_No response_

wardbekker:
## Summary

- **Bug confirmed real:** When `GRAFANA_TOKEN` (or `GRAFANA_PASSWORD`, `GRAFANA_USER`, `GRAFANA_PROVIDER_*`) was set in the environment, `gcx config use-context` leaked those values into the on-disk config YAML.
- **Root cause:** `useContextCmd` called `LoadConfigTolerant`, which merges env vars into the in-memory config via `ParseEnvIntoContext`, then `config.Write` serialised the entire mutated struct to disk.
- **Fix (use-context):** Switch to `config.LoadLayered` with no env overrides — the safe pattern already used by `set`/`unset`.
- **Fix (assistant):** `persistRefreshedTokens` re-reads the raw on-disk config before applying OAuth token updates, so env-sourced secrets in the live session config are never persisted.

## Changes

- `cmd/gcx/config/command.go`: `useContextCmd` now calls `config.LoadLayered` (no env overrides) instead of `LoadConfigTolerant`.
- `cmd/gcx/assistant/command.go`: `persistRefreshedTokens` re-reads from disk; drops the unused `*config.Config` parameter; `buildTokenRefresher` signature updated accordingly.
- `cmd/gcx/config/command_test.go`: Regression tests for `GRAFANA_TOKEN`, `GRAFANA_PASSWORD`, and `GRAFANA_PROVIDER_SLO_TOKEN` — each verifies the env secret is absent from the on-disk file after `use-context`.

## Test Plan

- [x] New regression tests pass (`go test ./cmd/gcx/config/... -run UseContext -v`)
- [x] Full test suite passes (`go test ./...` — exit 0)
- [x] Manual end-to-end verified: `GRAFANA_TOKEN=secret ./bin/gcx config use-context --config <file> new` → secret absent from file
- [x] `go vet` clean on changed packages
- [x] `gofmt` clean on changed files

Closes #572

### DIFF
--- a/cmd/gcx/assistant/command.go
+++ b/cmd/gcx/assistant/command.go
@@ -381,7 +381,7 @@ func resolveAssistantClientOptions(ctx context.Context, configOpts *cmdconfig.Op
 	switch {
 	case grafana.ProxyEndpoint != "" && grafana.OAuthToken != "":
 		// OAuth path: direct API via ProxyEndpoint
-		refresher := buildTokenRefresher(ctx, configOpts, &cfg, cfg.CurrentContext, grafana)
+		refresher := buildTokenRefresher(ctx, configOpts, cfg.CurrentContext, grafana)
 		return assistant.ClientOptions{
 			GrafanaURL:     grafana.Server,
 			Token:          grafana.OAuthToken,
@@ -426,7 +426,7 @@ func newAssistantStreamingHTTPClient(ctx context.Context, streamTimeoutSeconds i
 const refreshThreshold = 5 * time.Minute
 
 // buildTokenRefresher creates a TokenRefresher that uses gcx's auth refresh mechanism.
-func buildTokenRefresher(ctx context.Context, configOpts *cmdconfig.Options, cfg *config.Config, ctxName string, grafana *config.GrafanaConfig) assistant.TokenRefresher {
+func buildTokenRefresher(ctx context.Context, configOpts *cmdconfig.Options, ctxName string, grafana *config.GrafanaConfig) assistant.TokenRefresher {
 	var mu sync.Mutex
 	token := grafana.OAuthToken
 	refreshToken := grafana.OAuthRefreshToken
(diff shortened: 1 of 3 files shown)


wardbekker:
## Shepherd Status

### CI Checks
All 12 checks: **PASS**

### Auto-Fixes Applied
None needed — CI was already green.

### Pending Review
- Waiting for at least one reviewer approval (`REVIEW_REQUIRED`)

### Escalations
None

---
*Updated by shepherd-pr*

radiohead (approved):

PR #597, opened 2026-04-29
feat(instrumentation): migrate instrumentation to a provider

radiohead:
## What

Replaces the `gcx setup instrumentation` CRD-style facade — which mapped `Set/Get` + observed-state RPCs onto `kind: Cluster`/`kind: App` declarative resources — with a dedicated `gcx instrumentation` provider command tree built around action verbs that match the actual fleet-management API shape.

The new command tree has four subtrees:

- `gcx instrumentation setup <cluster>` — loud, mutating, idempotent onboarding; calls `SetupK8sDiscovery` then `SetK8SInstrumentation` if anything changed, and prints a parameterized helm command
- `gcx instrumentation clusters [list|get|configure|remove|wait]` and `clusters apps [list|get|configure|remove|wait]` — declared-state read/write with tri-state flag semantics on `configure` (RMW with per-namespace optimistic-lock guard)
- `gcx instrumentation services [list|get|include|exclude|clear]` — observed-state fleet sweep via `RunK8sDiscovery`, with DWIM single-workload mutation
- `gcx instrumentation status` — cross-cutting observed view across cluster→namespace→service hierarchy

Read paths are split by intent: `get`/`list` are sourced from declared-state endpoints (`GetK8SInstrumentation`, `GetAppInstrumentation`, `ListPipelines` as enumeration fallback) and may consult observed-state for STATUS only; `status`/`wait` go through observed-state endpoints (`RunK8sMonitoring`, `RunK8sDiscovery`) with the same cluster-set merge so list/get/status agree.

Boolean flags follow the canonical `--feat=true|false` idiom across the tree (no `--no-*` paired variants). Multi-word flags are kebab-case (`--cost-metrics`, `--cluster-events`, `--energy-metrics`, `--node-logs`, `--process-metrics`, `--extended-metrics`). List endpoints emit the canonical `{"items":[...]}` envelope; mutation commands emit `MutationResult` JSON in agent mode and a one-line summary otherwise. Not-found errors across the tree exit 1 with the canonical `DetailedError` schema (Summary / Details / Suggestions[]).

Several shared-infrastructure improvements accompany the provider work: reflection-based field discovery for empty typed slices in the output codec, `--json list` field discovery, fleet-management HTTP 401/403 surfaced as `DetailedError`, and the fleet provider's managed-pipeline guard upgraded to the same canonical error schema.

[...]

PR #654, opened 2026-05-07
feat(fail,output,fleet): foundation for instrumentation provider migration

radiohead:
Shared infrastructure improvements that unlock the instrumentation provider migration in subsequent stack PRs.

- cmd/gcx/fail: DetailedError JSON output strips Unicode box-drawing characters defensively, preventing render artefacts from leaking into agent-mode JSON envelopes.
- internal/output: reflection-based field discovery for empty typed slices (so --json list works even when the response array is empty); --json list field discovery generalised across codecs.
- internal/fleet: typed HTTPError carrying status, path, and trimmed body so consumers can convert non-2xx responses into structured DetailedErrors.
- internal/providers/fleet: managed-pipeline guard upgraded to canonical fail.DetailedError schema (Summary / Details / Suggestions[]) instead of an opaque fmt.Errorf string; redirects users to the new gcx instrumentation command tree.
- internal/style: TableBuilder agent-mode plain-text invariant test (no box chars / ANSI even with -o wide).
- docs/design: agent-mode, errors, and output design rules updated to describe the new schemas and conventions.
- CONSTITUTION.md: agent-mode invariant clarified (format vs non-format defaults).

Part 1 of 5 in the stack splitting #597 into reviewable units.

<hr/>

#### Stack

* **PR #654** 👈
  * **PR #655**
    * **PR #656**
      * **PR #657**
        * **PR #658**


internal/providers/fleet/provider.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/grafana/gcx/cmd/gcx/fail"
```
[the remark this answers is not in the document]
radiohead: Fixed. Created `internal/fail/` package with `DetailedError`, exit code constants, and `WriteJSON` methods — these are the shared types that both `internal/` and `cmd/` need. `cmd/gcx/fail` now re-exports via a type alias (`type DetailedError = ifail.DetailedError`) and const re-exports so all existing `cmd/` callers are unaffected. `internal/providers/fleet/` imports `internal/fail` directly. No violation remains.

---
**Generated by Claude Code**

cmd/gcx/fail/helpers.go
```diff
@@ -0,0 +1,19 @@
+package fail
+
+import "strings"
+
+// sameRenderedMessage reports whether details and parent render the same
+// message, used to suppress redundant output in error formatting.
+func sameRenderedMessage(details string, parent string) bool {
```
annanay25: Cursor pointed out

🟡 WARNING File: `cmd/gcx/fail/helpers.go` (entire file) Issue: `sameRenderedMessage` now exists in two places (`internal/fail/detailed.go:109` and `cmd/gcx/fail/helpers.go`). They are identical implementations. If one is updated without the other, behavior will diverge. Suggestion: Consider having `cmd/gcx/fail/helpers.go` call `internal/fail`'s version, or export it from `internal/fail`. Low urgency since both are stable.

annanay25 (dismissed):
cursor pointed out a repeated function but other than that lgtm!

cmd/gcx/fail/helpers.go
radiohead: Ah yeah, that's a leftover. Removed, thanks!

javiermolinar (approved):
LGTM I like the simplicity here

PR #662, opened 2026-05-07
feat(kg): support pagination for entities list

xujiaxj:
## Summary
- Surface the backend's `lastPage` and `searchResultsMaxLimitHit` pagination signals from `Client.Search()` via a new `SearchPage` struct, instead of discarding them.
- Emit a stderr hint from `kg entities list` when more pages exist, recommending `--page N+1` or narrower filters — so agents/scripts no longer silently get capped results.
- Emit a stderr hint when the client-side `--limit` is reached.
- Rename misnamed `entitiesShowOpts` → `entitiesListOpts`.

## Test plan
- [x] `mise run lint` — 0 issues
- [x] `go test ./...` — full suite passes
- [x] `GCX_AGENT_MODE=false mise run reference` — CLI reference regenerated
- [x] Live verified against ops: pagination hint fires when backend caps the page; no false positive when results fit on one page.

### DIFF
--- a/internal/providers/kg/client_test.go
+++ b/internal/providers/kg/client_test.go
@@ -216,12 +216,12 @@ func TestClient_Search(t *testing.T) {
 	defer server.Close()
 
 	client := newTestClient(t, server)
-	results, err := client.Search(t.Context(), kg.SearchRequest{
+	page, err := client.Search(t.Context(), kg.SearchRequest{
 		FilterCriteria: []kg.EntityMatcher{{EntityType: "Service"}},
 	})
 	require.NoError(t, err)
-	assert.Len(t, results, 2)
-	assert.Equal(t, "svc-1", results[0].Name)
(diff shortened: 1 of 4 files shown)


xujiaxj:
Monosnap Image 2026-05-07 15-28-04

internal/providers/kg/client_test.go
```diff
@@ -216,12 +216,12 @@ func TestClient_Search(t *testing.T) {
 	defer server.Close()
 
 	client := newTestClient(t, server)
-	results, err := client.Search(t.Context(), kg.SearchRequest{
```
annanay25, line 219: can we also add a test for multiple pages?

xujiaxj: 👍 . Added one.

annanay25 (approved):
lgtm!

PR #672, opened 2026-05-08
simplify GH issue templates

dafydd-t:
There were loads of required fields in our issue templates, they were overwhelming. Generally I prefer less structure here than more - we can always re-add structure where needed, later.

This PR should make some improvements in the right direction in my opinion:

- Remove the "config issue" template - we can use bug report for this. 
- Make "feature request" and "enhancement" only have one required field
- "enhancement":  condense it into one field, add the placeholder text from the old 3 for examples
- "feature request": leave in some potentially useful fields (workarounds, importance), but none of them are required
- remove some slop text from tips and descriptions
- remove Q&A and discussions links, since we dont use them, and fix the docs link

Easiest way to preview is to look at the files in the branch, e.g. https://github.com/grafana/gcx/blob/dafydd-t/simplify-issue-templates/.github/ISSUE_TEMPLATE/1-bug-report.yml


javiermolinar (approved):
Best contribution ever

PR #677, opened 2026-05-09
Feat/kg diagnose

danstadler-pdx:
This PR adds a new command: gcx kg diagnose. It is meant for use by people working to get Entity Graph to light up correctly, and covers 5 common scenarios people may run into during that process. 

(For Grafana Labs employees: please see my notes from 5/8 in the internal GCX channel.)


annanay25 (commented):
Hey Dan! Thanks for this PR - just checked it out and this is awesome, I can see this being useful for people trying to debug their environments. A few thoughts:

- i think the output text can be a bit more structured and descriptive. at the moment it just dumps the datasources and a Pass/Fail flag which is not very intuitive (for humans?)
- doesn’t the KG currently only work on the default prometheus datasource? so should we default to checking labels there instead of skipping prometheus checks if multiple datasources are found
- do we have the appropriate commands in gcx to follow up on the diagnostics and configure the right settings?

docs/reference/cli/gcx_kg_diagnose.md
```diff
@@ -0,0 +1,59 @@
+## gcx kg diagnose
```
annanay25, line 1: The formatting of the output seems off and very different from the rest of the gcx command family. Can you check the output of `gcx kg entities list --type Service` as an example and maybe format the result in a similar table fashion?

docs/reference/cli/gcx_kg_diagnose.md
```diff
@@ -0,0 +1,59 @@
...
+
+Run diagnostic checks to verify the Knowledge Graph is healthy.
+
+Checks stack status, sanity results, entity counts, scope values,
+telemetry drilldown configuration, and recording rule metrics in
+Mimir. Use --env to scope checks to a specific environment.
+
+Metric checks require a Prometheus datasource. The datasource UID is
+resolved from --datasource, the datasources.prometheus config key, or
+auto-discovery. If unavailable, metric checks are skipped.
```
annanay25, line 15: I find the auto-discovery confusing. For example, when I run this command against a demo cluster I see:

```
datasource auto-discovery failed, using grafanacloud-prom (multiple prometheus datasources found....
```

What does this mean here when KG just uses the default prometheus datasource everytime?

docs/reference/cli/gcx_kg_diagnose_labels.md
```diff
@@ -0,0 +1,48 @@
+## gcx kg diagnose labels
```
annanay25, line 1: There seem to be three distinct sections in the output of this command but there is no clear heading for these which makes it harder to read the output. Going back to the ask, can we also add some helpful text so humans can consume the output of diagnose?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,47 @@
+# Copilot Working Instructions — Dan's gcx Fork
```
annanay25: is this file deliberately pushed?

internal/providers/kg/diagnose_labels.go
```diff
@@ -0,0 +1,414 @@
...
+	// Mapping table.
+	if len(r.Mappings) > 0 {
+		fmt.Fprintln(w, "  LABEL MAPPING")
+		fmt.Fprintln(w, "  deployment_environment → asserts_env")
+		for _, m := range r.Mappings {
+			switch m.Status {
+			case "mapped":
+				fmt.Fprintf(w, "    %-30s → %-30s  ✓ mapped\n", m.DeploymentEnv, m.AssertsEnv)
+			case "unmapped":
+				fmt.Fprintf(w, "    %-30s → %-30s  ✗ NOT MAPPED — check relabeling rules\n", m.DeploymentEnv, "(missing)")
```
annanay25:
```suggestion
				fmt.Fprintf(w, "    %-30s → %-30s  ✗ not mapped — check relabeling rules\n", m.DeploymentEnv, "(missing)")
```

annanay25 (commented):
Reg. The skills discussion that we had on slack, can you check the `gcx agent skills` command? We have a skills bundle that we ship with gcx and I think we can include the agent playbook in there.

GCX_KG_DIAGNOSE_FORMAT.md
```diff
@@ -0,0 +1,36 @@
+# kg diagnose — Output Format Cleanup
```
annanay25: Let's remove this file.

internal/providers/kg/diagnose.go
```diff
@@ -0,0 +1,906 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"github.com/grafana/gcx/internal/style"
```
annanay25: Nit: `github.com/grafana/gcx/internal/style` is in the stdlib import group — it should be in the third-party/internal block (after the blank line separator). Same issue in `diagnose_labels.go` and `diagnose_service.go`.

internal/providers/kg/diagnose.go
```diff
@@ -0,0 +1,906 @@
...
+		types = append(types, t)
+	}
+	sort.Strings(types)
+	for _, t := range types {
+		cnt := counts[t]
+		total += cnt
+		parts = append(parts, fmt.Sprintf("%s: %d", t, cnt))
+	}
+
+	if total == 0 {
```
annanay25, line 480: `nsFilter` is computed but then blanked with `_ = nsFilter` and never used in the returned definitions. Looks like leftover code — should it be wired into the edge source gap metric queries, or removed?

annanay25 (approved):
lgtm!

PR #685, opened 2026-05-11
feat(assistant): support v2 (Lodestone) investigations API

wardbekker:
## Summary

- Adds v2 (Lodestone) Assistant investigations API alongside the existing v1 legacy surface.
- Stack capability is auto-detected via a cheap probe (`GET /investigations/lodestone?limit=1`) cached at `$XDG_STATE_HOME/gcx/lodestone-capability.json` for 24h — users never pick a version.
- Eight new v2-only subcommands plus auto-dispatch for the three commands that exist in both APIs (`list`, `get`, `create`).
- v1 commands emit `hint:` lines to stderr pointing at v2 successors when the stack is known to support Lodestone — agents read these and adjust subsequent invocations.

## Why

v2 Lodestone is the future direction for Grafana Assistant investigations — single-agent, session-based, with richer operations (pause/resume/mode/sharing/Mermaid repair). v1 isn't deprecated; both must coexist. Surfacing v2 behind transparent capability detection means existing usage keeps working on v1-only stacks while v2-enabled stacks transparently use the richer endpoints.

### Command map (under `gcx assistant investigations`)

| Command | v1 stack | v2 stack |
| --- | --- | --- |
| `list` | v1 `/summary` | v2 `/lodestone` with `includeLegacy=true` |
| `get <id>` | v1 `/investigations/{id}` | v2 `by-id` → `state`; 404 falls back to v1 |
| `create` | v1 POST, `--description` | v2 POST, `--instruction` + `--team`/`--profile-id` |
| `cancel` | v1 (unchanged; works on both) | v1 (terminal on Lodestone too) — emits hint to use `pause` |
| `todos`/`timeline`/`report` | v1 (unchanged) | v1 (legacy concepts) — each emits hint to use `state` |
| `document`/`approvals` | v1 (unchanged) | v1 (no useful v2 successor) |
| **new** `pause`, `resume`, `mode`, `share`, `regenerate-report`, `repair-mermaid`, `update-mermaid`, `state` | clear v2-required error | v2 |

`list` gains v2-only filters (`--scope`, `--team`, `--q`, `--from`, `--to`, `--sort`, `--order`, `--view`, `--label`, `--include-legacy`); setting any of them on a v1 stack returns an enum-style error consistent with #673.

### Files

- New: `capability.go`, `client_v2.go`, `types_v2.go`, `commands_v2.go` + tests
- Modified: `commands.go` (auto-dispatch + v2-successor hints in v1 commands), `internal/agent/command_annotations.go` (token-cost entries for new leaf commands), `AGENTS.md` (package-map note)
- Regenerated: CLI reference under `docs/reference/cli/`

### Capability cache

JSON file keyed by REST host. 24h TTL with refresh on expiry; probe errors fall through to v1 without poisoning the cache. Probe uses `GET /investigations/lodestone?limit=1` (the canonical list endpoint — reliable across stack versions; the `profiles` endpoint isn't deployed everywhere). Undocumented `GCX_ASSISTANT_API_VERSION=v1|v2` env var bypasses the probe (used by tests; not in help).

### v2-successor hints

v1 commands with a meaningful v2 successor (`report`, `todos`, `timeline`, `cancel`) read the cached capability (no extra network call) and print a one-line `hint:` to stderr when v2 is available, e.g.:

```
hint: Lodestone (v2) stores the report inline in session state — try `gcx assistant investigations state <id>`
```

Channel and `hint:` prefix match the existing convention in `internal/output/format.go`; emitted in all modes including agent (agents are the primary audience and learn from stderr hints).

## Live verification (against a Lodestone-enabled stack)

- `list` → v2 endpoint, includes legacy investigations
- `create --instruction="..." --title="..."` → returns `investigationId` + `chatId`
- `get <id>` / `state <id>` → full Lodestone session state (plan, hypotheses, mode, epoch)
- `pause`/`resume`/`mode <id> medium` → all wired correctly; server returns structured success or expected 400 on invalid lifecycle transitions
- `cancel` (v1 endpoint) → works on Lodestone investigations; prints v2-successor hint
- `todos`/`timeline`/`approvals` on a Lodestone investigation → return `[]` (server-side: Lodestone has no equivalents); hints point at `state`
- `report` → returns legacy summary; hint points at `state`
- `repair-mermaid` with fake element id → server 404 (wiring correct)
- `update-mermaid` without `--content` → client-side validation error

On a v1-only stack: `pause`/`resume`/`mode`/`share`/etc. return:

```
Error: lodestone (v2 investigations) is not available on <host>; use `gcx assistant investigations list` to see legacy investigations
```

## Test Plan

- [x] `go test -race -count=1 ./internal/assistant/investigations/...`
- [x] `go test -count=1 ./cmd/gcx/root` (token-cost consistency for new leaf commands)
- [x] `mise run lint` clean
- [x] `mise run reference-drift` clean against this branch
- [x] Help-tree smoke: new subcommands listed; v1 commands list new v2 filters; no `--api-version` flag (auto-detection only)
- [x] Live smoke on a Lodestone-enabled stack (`wbkprez`): create, list, get, state, pause, resume, mode, cancel, todos, timeline, report, repair-mermaid, update-mermaid, regenerate-report
- [ ] Live smoke on a strictly v1-only stack (manual): verify v2-only commands return the clear v2-required error

Pre-existing test failures in `internal/providers/aio11y/eval/{collections,savedconversations}`, `internal/providers/dashboards`, `internal/providers/dashboards/versions`, and `internal/providers/stacks` reproduce on `main` and are unrelated to this PR.

## Commits

- `feat(assistant): support v2 (Lodestone) investigations API` — initial v2 surface + auto-detection
- `fix(assistant): use lodestone list endpoint as capability probe` — switched probe off `/profiles` (not deployed on every stack) to `/lodestone?limit=1`
- `refactor(assistant): remove profiles endpoint — internal API only`
- `feat(assistant): emit v2-successor hints from v1 commands`


wardbekker:
## Shepherd Status

Rebased onto latest `main` (was 52 commits behind; branch squashed to a single clean commit, conflict in `internal/agent/command_annotations.go` resolved) and addressed all four findings from @radiohead's review.

### Review findings
| # | Finding | Status |
|---|---------|--------|
| F1 | Capability cache permanently recorded `v1`, stranding upgraded stacks | **Fixed** — only `v2` is sticky; `v1` re-probes each run. New `TestDetectAPIMode_V1NotCached`. |
| F2 | `share` skipped `resolveID`, 404ing on chat IDs | **Fixed** — resolves to canonical investigation ID first |
| F4 | `state` duplicated `get` on v2 stacks (CONSTITUTION §CLI Grammar) | **Fixed** — removed `state`; `get` is the single v2 read path. Confirmed upstream has no distinct detail payload. |
| F3 | Split v2 methods onto own type (non-blocking) | **Deferred** to follow-up; runtime `requireV2` gate retained |

### CI
All checks green after the rebase: **Tests** ✅ · **Linters** ✅ · **Documentation** ✅ (+ zizmor, trufflehog, semgrep). Local `GCX_AGENT_MODE=false mise run all` also passes.

### Merge readiness
`mergeable: MERGEABLE` — blocked only on @radiohead re-review (currently `CHANGES_REQUESTED`).

---
*Updated by shepherd-pr*

docs/reference/cli/gcx_assistant_investigations_create.md
```diff
@@ -4,7 +4,7 @@ Create a new investigation.
 
 ### Synopsis
 
-Create a new investigation with a title and optional description.
+Create a new investigation. On Lodestone-enabled stacks, uses the v2 API with --instruction; falls back to legacy create otherwise.
```
annanay25: Is lodestone the official API route? I think the lodestone endpoint will fold into /investigations at some point? I am not sure if we need to add support for legacy routes here

internal/assistant/investigations/capability.go
```diff
@@ -0,0 +1,175 @@
+package investigations
```
annanay25, line 1: This is a novel caching pattern in gcx — no other file uses a bespoke JSON cache at `$XDG_STATE_HOME`. The existing precedent for "discovered capability, cached across invocations" is `SaveProviderConfig` (used by adaptive-telemetry, faro, and synth for the same purpose). Consider using that instead:

```go
loader.SaveProviderConfig(ctx, "assistant", "lodestone-v2", "true")
```

Benefits: automatic context-switching support, config discoverability, no new file-management code, and future `gcx cache clear`-style tooling wouldn't need to discover this path separately.

If the bespoke file is kept, it also lacks an env-var override for its path (`GCX_OPENAPI_CACHE_DIR` / `GCX_DISCOVERY_CACHE_DIR` exist for the other disk caches).

internal/assistant/investigations/capability.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/adrg/xdg"
+	"github.com/grafana/gcx/internal/assistant/assistanthttp"
+)
+
+const (
+	capabilityCacheFileName = "lodestone-capability.json"
+	capabilityCacheTTL      = 24 * time.Hour
+	envAPIVersionOverride   = "GCX_ASSISTANT_API_VERSION"
+)
+
```
annanay25, line 23: The HTTP probe at line 72 runs under `detectorMu`, so any concurrent caller (goroutine or parallel `gcx` process sharing the lock) blocks for the full network round-trip. Hold the lock only around file I/O and release it during the probe:

```go
detectorMu.Lock()
cache := loadCapabilityCache(path)
if entry, ok := cache.Entries[key]; ok && now.Sub(entry.CheckedAt) < capabilityCacheTTL {
    detectorMu.Unlock()
    return entry, nil
}
detectorMu.Unlock()

v2, err := probe(ctx, base)
if err != nil {
    return Capability{}, err
}

detectorMu.Lock()
defer detectorMu.Unlock()
cache = loadCapabilityCache(path) // re-read to avoid clobbering concurrent writes
entry := Capability{V2: v2, CheckedAt: now}
cache.Entries[key] = entry
_ = saveCapabilityCache(path, cache)
return entry, nil
```

internal/assistant/investigations/commands.go
```diff
@@ -151,44 +291,105 @@ func newGetCommand(loader *providers.ConfigLoader) *cobra.Command {
...
 }
 
+func (o *createOpts) validateForV2() error {
+	if o.setFlags["instruction"] && o.setFlags["description"] {
+		return errors.New("--instruction and --description are mutually exclusive")
+	}
+	if o.Instruction == "" && o.Description == "" {
+		return errors.New("--instruction is required")
+	}
+	return nil
```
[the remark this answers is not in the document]
wardbekker: Could you clarify which review you're replying to here? I want to make sure I don't misread the threading — pinging on this since the in-reply-to was the (since-superseded) Validate-pattern comment, but I assume you meant the legacy-routes question.

If you can confirm, my read on the v1 path:

- The v1 `/api/plugins/grafana-assistant-app/resources/investigations` endpoint is still the only one exposed on stacks that don't yet ship Lodestone. The capability probe falls back to v1 transparently on 404.
- If/when Lodestone fully replaces v1 on every stack we ship, removing the v1 client and dispatch path is a one-line change in `commands.go` (drop the `!v2` branches) plus deleting `v1_client.go`.
- Until that happens, dropping v1 here would break any non-Lodestone customer.

Happy to remove the v1 path if you'd prefer — just want to confirm the rollout assumption before yanking the fallback.

docs/reference/cli/gcx_assistant_investigations_create.md
wardbekker: let me check with the dev

internal/assistant/investigations/capability.go
wardbekker: Done in 8fe26b72 — migrated to `SaveProviderConfig` keyed at
  `providers.assistant.lodestone-v2`. Side benefit: drops the
  `detectorMu`-held-across-probe concern from your third comment, since
  there's no shared in-process mutex anymore. Also dropped the 24h TTL
  (Lodestone is stable per stack, matching synth/faro/adaptive's
  no-TTL precedent).

internal/assistant/investigations/capability.go
wardbekker: Done in 8fe26b72 — migrated to `SaveProviderConfig` keyed at
  `providers.assistant.lodestone-v2`. Side benefit: drops the
  `detectorMu`-held-across-probe concern from this comment, since
  there's no shared in-process mutex anymore. Also dropped the 24h TTL
  (Lodestone is stable per stack, matching synth/faro/adaptive's
  no-TTL precedent).

wardbekker:
Layered #693 onto this PR (commit 77daf9f3) — adds four read-only subcommands that surface the chat thread that backs every Lodestone investigation:

- `chat <id>` — full message stream with tool_use/tool_result blocks (default condensed view; `-o json` for raw); `--role` + `--include-hidden` filters
- `narrative <id>` — assistant-authored prose only
- `tools <id>` — tool calls paired with results (`--name` filter)
- `skills <id>` — `search_skills` results: matched Skills + chunk snippets

All four require v2 (Lodestone) and reuse the existing `requireV2` + `resolveChatID` helpers; ID → chatId is resolved via `ResolveByID`, then `GET /chats/{chatId}/all-messages` returns the full message list.

### Why fold into this PR
The chat-thread surface only makes sense alongside the v2 commands shipping here — on v1-only stacks there is no chatId to resolve. Folding lets the v2 surface land complete: state + control (pause/resume/mode) **and** content visibility, which is what the webinar demo flow actually needs.

### Files added
- `internal/assistant/investigations/chat.go` — types + `GetChatThread` + `Narrative` / `ExtractToolCalls` / `ExtractSkillMatches` (tolerant of both flat and Anthropic-style nested-text payload shapes)
- `internal/assistant/investigations/commands_chat.go` — four cobra commands + table codecs
- `internal/assistant/investigations/{chat,commands_chat}_test.go` — covers payload variants, filtering, codec output, error markers

### Updated
- `commands.go` — registers the four commands alongside other v2-only commands
- `internal/agent/command_annotations.go` — token-cost entries for the new leaves
- Reference docs auto-regenerated under `docs/reference/cli/`

### Tests
- `go test -race -count=1 ./internal/assistant/investigations/...` ✓
- `go test -race -count=1 ./internal/agent/... ./cmd/gcx/root/...` ✓ (token-cost consistency)
- `mise run lint` ✓
- `mise run tests` (full suite) ✓

Closes #693.

wardbekker:
Fixed the three broken subcommands in commit 047fc52a, based on the captured live payload shape.

### Root cause
The previous commit (77daf9f3) used Anthropic-style content-block JSON tags (`name`, `input`, `tool_use_id`, `content`). The Grafana Assistant plugin actually emits its own camelCase shape:

```json
// tool_use
{"type":"tool_use", "toolId":"...", "toolName":"...", "toolInput":{...}}
// tool_result — paired by toolUseId == toolId
{"type":"tool_result", "toolUseId":"...", "toolName":"...", "durationMs":N, "isError":false,
 "toolResult":[{"type":"text","text":"..."}]}
// thinking
{"type":"thinking", "thinking":"...", "signature":"..."}
// artifact (inside messages where message-level type=="artifact")
{"type":"artifact", "artifactType":"panel", "panel":[{"panelId":"p5","panel":{...}}]}
```

Also: `search_skills` input is `{"queries":[{"keywords","query"}]}` (array, not a single `query`); result text is loose XML — `<results count="N"><chunk skill_id="..." title="...">body</chunk>...</results>`.

### Live verification against the same investigation (`08599227-...` on `wbkprez`)

| Command | Before | After |
|---|---|---|
| `skills <id>` | `null` | 3 chunks, `skill_id=78ffce9d-...`, `title="Investigate payment error spikes"` |
| `tools <id> -o json \| jq length` | 27 with `{"id":"","name":""}` | 27 with populated `name`/`input`/`durationMs`/`result` — six distinct tool names |
| `tools <id> --name search_skills` | `[]` | 1 row with the full query input |
| `chat <id> -o json` | tool_use/result fields stripped | All 53 messages round-trip with toolName, toolInput, toolUseId, durationMs, isError; 27 tool_results, 17 artifact blocks preserved |
| `narrative <id>` (agent mode) | JSON-quoted markdown | Raw markdown |

### Other changes
- `chat` table view now shows `~` for thinking blocks and `◆ artifact panel panels=p5,p7` for artifact blocks
- `SkillMatch` shape simplified to `{toolUseId, query, skillId, title, chunk}` (no fictional `score`/`source` fields — server doesn't emit them)
- Side fix in `internal/output/format.go`: `Options.allowedCodecs()` was duplicating names when a command registered a custom codec under a builtin format name. With the new `NarrativeCodec` registered under `"agents"`, this would have shown `agents, agents` in `--help`. Now deduplicated.

Tests: full `go test -race -count=1 ./...` green, lint clean.

@-original-reporter — if you re-test against `08599227-81f6-4882-a972-5a40fcc01a7e` on `wbkprez`, all four commands should now match the priorities you listed.

docs/reference/cli/gcx_assistant_investigations_create.md
wardbekker: Confirmed with the dev — the official surface is `/api/v2/investigations/*` (introduced in grafana-assistant-app#6645), and the temporary `/investigations/lodestone/*` paths are going away.

Switched the v2 client to `/api/v2/investigations/*` in c7f939e6. The user-facing capability is now described as "v2 investigations API" rather than the internal "Lodestone" codename.

On legacy support: the v1 surface (`/api/v1/...`) stays for stacks that don't yet expose `/api/v2`. Capability auto-detection picks the right one (cached at `providers.assistant.api-mode`), so users on older stacks transparently fall through to v1. We can drop v1 once the rollout is complete on every stack — happy to file a follow-up issue tracking that cleanup.

internal/assistant/investigations/client_v2.go
```diff
@@ -0,0 +1,247 @@
...
+		return fmt.Sprintf(v2ResumeFmt, url.PathEscape(id))
+	case "mode":
+		return fmt.Sprintf(v2ModeFmt, url.PathEscape(id))
+	case "share":
+		return fmt.Sprintf(v2ShareFmt, url.PathEscape(id))
+	case "regen-report":
+		return fmt.Sprintf(v2ReportRegenFmt, url.PathEscape(id))
+	}
+	panic(fmt.Sprintf("investigations: unknown v2 op %q", op))
+}
```
annanay25: why do we need this? can we just use the paths directly in the functions as needed?

internal/agent/command_annotations.go
```diff
@@ -28,15 +28,27 @@ var commandAnnotations = map[string]annotation{
...
+	"gcx assistant investigations repair-mermaid":    {Cost: "medium", Hint: "<id> <element-id> [--message=...] -o json"},
+	"gcx assistant investigations report":            {Cost: "medium", Hint: "<id> -o json"},
+	"gcx assistant investigations resume":            {Cost: "small"},
+	"gcx assistant investigations share":             {Cost: "small", Hint: "<id> --team=<name> (repeatable)"},
+	"gcx assistant investigations skills":            {Cost: "medium", Hint: "<id> -o json; Skill matches (search_skills tool results)"},
+	"gcx assistant investigations state":             {Cost: "large", Hint: "<id> -o json; returns full Lodestone session state (plan, hypotheses, mode, report)"},
+	"gcx assistant investigations timeline":          {Cost: "medium", Hint: "<id> -o json"},
+	"gcx assistant investigations todos":             {Cost: "medium", Hint: "<id> -o json"},
+	"gcx assistant investigations tools":             {Cost: "medium", Hint: "<id> [--name=<tool>] -o json; tool calls with results"},
+	"gcx assistant investigations update-mermaid":    {Cost: "small", Hint: "<id> <element-id> --content=path|-"},
```
annanay25: does it make sense to expose the `repair-mermaid` / `update-mermaid` commands to an agent? does this really help?

docs/reference/cli/gcx_assistant_investigations_skills.md
```diff
@@ -0,0 +1,36 @@
+## gcx assistant investigations skills
+
+Show Skill matches from a v2 investigation.
+
+### Synopsis
+
+Extract the search_skills tool results from a v2 investigation's chat thread: which Skills semantically matched and the chunks returned. The substantive piece for the Skills GA story.
```
annanay25: do we need to expose this?

internal/assistant/investigations/client_v2.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, assistanthttp.HandleErrorResponse(resp)
+	}
+	var envelope struct {
+		Data T `json:"data"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&envelope); err != nil {
+		return nil, fmt.Errorf("decode %s response: %w", op, err)
+	}
+	return &envelope.Data, nil
+}
```
annanay25: this feels like it should be moved up a level instead of being tucked away in investigations client, wdyt?

internal/agent/command_annotations.go
```diff
@@ -28,15 +28,27 @@ var commandAnnotations = map[string]annotation{
...
-	"gcx assistant investigations create":    {Cost: "small", Hint: "Use for deep cross-signal root cause analysis. Dispatches specialist agents for metrics, logs, traces, and profiles in parallel — more efficient than chaining individual gcx query commands. Example: --title=\"Checkout latency spike after deploy\""},
-	"gcx assistant investigations document":  {Cost: "medium", Hint: "<investigation-id> <document-id> -o json"},
-	"gcx assistant investigations get":       {Cost: "medium", Hint: "<id> -o json"},
-	"gcx assistant investigations list":      {Cost: "small"},
-	"gcx assistant investigations report":    {Cost: "medium", Hint: "<id> -o json"},
-	"gcx assistant investigations timeline":  {Cost: "medium", Hint: "<id> -o json"},
-	"gcx assistant investigations todos":     {Cost: "medium", Hint: "<id> -o json"},
+	"gcx assistant investigations approvals":         {Cost: "medium", Hint: "<id> -o json"},
+	"gcx assistant investigations cancel":            {Cost: "small"},
+	"gcx assistant investigations chat":              {Cost: "large", Hint: "<id> [--role=user|assistant|tool] [--include-hidden] -o json; full Lodestone chat thread with tool calls and results"},
```
annanay25: Do we want to use the name "Lodestone" in our code? No strong opinion, just checking.

internal/assistant/investigations/commands_chat.go
```diff
@@ -0,0 +1,530 @@
...
+func newChatCommand(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &chatOpts{}
+	cmd := &cobra.Command{
+		Use:   "chat <id>",
+		Short: "Show the chat thread for a v2 investigation.",
+		Long: "Stream the chat thread that backs a v2 investigation: " +
+			"assistant prose, tool calls (search_skills, prometheus_query_handler, " +
+			"loki_query_handler_investigator, tempo_query_handler, ...), and tool " +
+			"results. The legacy report/timeline/todos endpoints return empty stubs " +
+			"on v2 — this command is the substantive view.",
```
annanay25, line 45: seems like a few commands in this file should be in `commands_v2.go`?

internal/assistant/investigations/client_v2.go
wardbekker: Agreed — inlined in b697b139. `pathFor` is gone; each method now uses `fmt.Sprintf(v2…Fmt, url.PathEscape(id))` directly. Removed the runtime `panic` on typo, and the URL each method hits is visible at the call site.

internal/assistant/investigations/client_v2.go
wardbekker: Done in b697b139 — moved to `internal/assistant/assistanthttp` as `DoEnvelopeRequest[T]`. The `{"data": T}` envelope is the v2 plugin-API convention, so it belongs at the base layer where other future v2 clients can reuse it. Callers now pass `c.base` (the `*assistanthttp.Client`).

internal/agent/command_annotations.go
wardbekker: Good catch — fixed in b697b139. Replaced the remaining "Lodestone" references in the user-facing hints (`chat` and `state`) with "v2". Kept the internal Go type/method names (`CreateLodestoneRequest`, `ListLodestone`, `LodestoneState`) since they're the upstream payload-type names from grafana-assistant-app — happy to rename those in a follow-up if you'd prefer.

internal/agent/command_annotations.go
wardbekker: Agreed — removed the `repair-mermaid` and `update-mermaid` commands entirely in b697b139 (not just the annotations). The Mermaid persist and LLM-repair endpoints serve a human-curating-report workflow that the CLI doesn't need to expose; agents and scripts don't call them. Dropped the Client methods, request/response types, path constants, command wiring, tests, and reference docs.

docs/reference/cli/gcx_assistant_investigations_skills.md
wardbekker: Removed the `skills` command entirely in b697b139. Skill-match extraction was a niche debug surface, and the underlying `search_skills` tool calls are already visible through `gcx assistant investigations tools --name=search_skills`. Dropped `SkillMatch` / `ExtractSkillMatches` / `parseSkillResultXML`, `SkillsTableCodec`, the command, its tests, and the reference doc.

internal/assistant/investigations/commands_chat.go
wardbekker: After removing `skills` (per your other comment), `commands_chat.go` now holds `chat`, `narrative`, and `tools` — all three operate on the chat-thread payload (`GetChatThread` + chat-codec rendering), so they're cohesive as a unit. `commands_v2.go` holds the investigation-lifecycle commands (`pause`, `resume`, `mode`, `share`, `regenerate-report`, `state`) that operate on the v2 `/api/v2/investigations/{id}` surface. Keeping them split by data source rather than by API version, but happy to fold them together if you'd prefer a single `commands_v2.go`.

internal/assistant/investigations/capability.go
```diff
@@ -0,0 +1,151 @@
...
+	cfg, _, err := loader.LoadProviderConfig(ctx, capabilityProviderName)
+	if err != nil {
+		return "", false
+	}
+	raw, ok := cfg[capabilityCacheKey]
+	if !ok {
+		return "", false
+	}
+	switch APIMode(raw) {
+	case APIModeLegacy, APIModeV2Standard:
```
radiohead: **issue (blocking):** `loadCachedMode` treats a cached `"v1"` as a permanent hit — once it is written, the probe never re-runs. The `"v2"` direction being permanent is correct (stacks do not get downgraded), but the `"v1"` direction is not: this PR exists to support a rolling Lodestone deployment where v1 stacks are actively being upgraded. A user who probes before their stack upgrades gets `api-mode: v1` written to `~/.gcx.yaml` and is silently stuck on legacy endpoints indefinitely, with no error and no hint path out.

The difference from the `synth`/`faro`/`adaptive` precedent: those only cache positive discovery values (a URL that was issued, a token that was registered). This also caches a negative capability result.

A two-key fix fits the existing `map[string]string` schema — expire `"v1"` entries after 24h, keep `"v2"` entries permanent:

```go
// save — alongside the existing SaveProviderConfig call
if mode == APIModeLegacy {
    _ = loader.SaveProviderConfig(ctx, capabilityProviderName, "api-mode-probed-at",
        time.Now().UTC().Format(time.RFC3339))
}

// load — in loadCachedMode, after the raw lookup
if mode == APIModeLegacy {
    probedAt, ok := cfg["api-mode-probed-at"]
    if !ok {
        return "", false
    }
    t, err := time.Parse(time.RFC3339, probedAt)
    if err != nil || time.Since(t) > 24*time.Hour {
        return "", false // expired — trigger a fresh probe
    }
}
```

internal/assistant/investigations/commands_v2.go
```diff
@@ -0,0 +1,334 @@
...
+				return err
+			}
+			if err := opts.IO.Validate(); err != nil {
+				return err
+			}
+			client, err := requireV2(cmd, loader)
+			if err != nil {
+				return err
+			}
+			resp, err := client.Scope(cmd.Context(), args[0], opts.Teams)
```
radiohead: **issue (blocking):** `args[0]` is passed directly to `client.Scope` without going through `resolveID`. The `/api/v2/investigations/{investigationId}/share` endpoint requires an investigation ID, but users commonly pass a chat ID — the exact case `resolveID` handles for every other v2 command (pause:92, resume:131, mode:190, regenerate-report:278, state:321). `gcx assistant investigations share <chatId>` will 404 silently while `gcx assistant investigations pause <chatId>` works fine.

```go
investigationID, err := resolveID(cmd.Context(), client, args[0])
if err != nil {
    return err
}
resp, err := client.Scope(cmd.Context(), investigationID, opts.Teams)
```

internal/assistant/investigations/client_v2.go
```diff
@@ -0,0 +1,173 @@
...
+	v2SnapshotFmt        = v2InvestigationsBase + "/%s/snapshot"
+	v2PauseFmt           = v2InvestigationsBase + "/%s/pause"
+	v2ResumeFmt          = v2InvestigationsBase + "/%s/resume"
+	v2ModeFmt            = v2InvestigationsBase + "/%s/mode"
+	v2ShareFmt           = v2InvestigationsBase + "/%s/share"
+	v2ReportRegenFmt     = v2InvestigationsBase + "/%s/report/regenerate"
+)
+
+// CreateLodestone starts a new investigation against /api/v2/investigations.
+func (c *Client) CreateLodestone(ctx context.Context, req CreateLodestoneRequest) (*CreateLodestoneResponse, error) {
```
radiohead, line 29: **suggestion (non-blocking):** All v2 methods (`CreateLodestone`, `ListLodestone`, `ResolveByID`, `GetState`, `Pause`, `Resume`, `RegenerateReport`, `SetMode`, `Scope`) are added to the same `*Client` type as the v1 methods. The capability gate is an honour-system convention enforced at the command layer via `requireV2` — nothing in the type system prevents a future caller from invoking a v2 method on a v1-stack client, which surfaces as a runtime 404 rather than a compile error. As the v2 surface grows this becomes harder to maintain.

Preferred fix: introduce `ClientV2 struct { *Client }` and move all v2 methods onto it. `requireV2` returns `*ClientV2`; v1 commands keep `*Client`. The capability invariant becomes a type invariant — misuse is a compile error, not a silent 404.

If a full type split is too disruptive right now, a `V2` suffix on v2 method names (`PauseV2`, `ResumeV2`, …) at least makes the requirement self-documenting at call sites. The current asymmetry (`CreateLodestone` vs `Pause`) is the worst of both worlds.

internal/assistant/investigations/commands_v2.go
```diff
@@ -0,0 +1,334 @@
...
+			}
+			client, err := requireV2(cmd, loader)
+			if err != nil {
+				return err
+			}
+			chatID, err := resolveID(cmd.Context(), client, args[0])
+			if err != nil {
+				return err
+			}
+			state, err := client.GetState(cmd.Context(), chatID)
```
radiohead: **issue (blocking):** On a v2-enabled stack, `gcx assistant investigations get <id>` and `gcx assistant investigations state <id>` run identical code paths — both call `client.ResolveByID` then `client.GetState` and encode the same `LodestoneState` response. The only behavioural delta is that `state` hard-errors on v1 stacks/investigations via `requireV2`, while `get` falls back to the legacy detail endpoint.

CONSTITUTION.md §CLI Grammar: *"Extensions must not duplicate CRUD semantics — if it can be done with `list`/`get`/`push`/`pull`/`delete`, it is not an extension."* Per the compliance hierarchy, a violation requires either a fix or explicit human approval and an ADR/design-doc carve-out.

The in-file comment's rationale ("`state` always uses the v2 endpoint; `get` may fall back to v1") is defensible for strict v2-enforcement semantics, but it does not change that both commands return the same data on a v2 investigation. A `--require-v2` flag on `get` achieves that without a duplicate verb.

radiohead (changes requested):
The v2 capability-detection and command surface is well structured — the SaveProviderConfig migration, the url.PathEscape discipline, and the client/codec test coverage are all solid. Two correctness issues need attention before merge: the capability cache permanently records "v1" with no re-probe path, which will silently strand users whose stacks upgrade to Lodestone after their first probe; and the share command skips the ID-resolution step that all its sibling v2 commands use. Two further findings are posted inline.

---
**Generated by Claude Code**

internal/assistant/investigations/capability.go
wardbekker: Fixed. The capability cache is now sticky **only** for `v2` (stacks are never downgraded). A `v1` probe result is no longer persisted, so a stack upgraded to Lodestone after the first probe is re-detected on the next invocation instead of being stranded on legacy endpoints. `loadCachedMode` returns a hit only for `v2`; any other value (missing, `v1`, or a stale codename) falls through to a fresh probe. Added `TestDetectAPIMode_V1NotCached` covering the upgrade path.

internal/assistant/investigations/commands_v2.go
wardbekker: Fixed. `share` now calls `resolveID(cmd.Context(), client, args[0])` before `client.Scope`, matching pause/resume/mode/regenerate-report. Passing a chat ID to `share` now resolves to the canonical investigation ID instead of 404ing.

internal/assistant/investigations/commands_v2.go
wardbekker: Fixed by removing the `state` command. I confirmed against grafana-assistant-app: the v2 surface exposes only two single-investigation reads — `GET /api/v2/investigations/{id}` (`resolveLodestoneInvestigation`, returns just `{investigationId, chatId}` — a thin resolver) and `GET /api/v2/investigations/{id}/snapshot` (`getLodestoneSnapshotV2`, the full session state). There is no distinct rich detail payload to back a separate command, so `state` and `get` rendered the identical snapshot. Removed `state` (command, opts, annotation, reference doc); `get` is the single v2 read path and still falls back to v1 detail for legacy IDs. Resolves the CONSTITUTION §CLI Grammar concern.

internal/assistant/investigations/client_v2.go
wardbekker: Acknowledged — agree the type-system guarantee would be stronger than the `requireV2` honour-system gate. Leaving it for a follow-up since it is non-blocking and the refactor touches every v2 call site (larger diff on an already-large PR). Tracking separately; the runtime gate plus the friendly v1-stack error keeps the current surface safe in the meantime.

radiohead (approved):

PR #687, opened 2026-05-11
fix(kg): repoint insights search at /v1/assertions/search (#670)

### LINKED ISSUE #670: `gcx kg insights search` throws a 500 error
### gcx Version

v0.2.13

resources (get/list/push/pull/delete/edit/serve/validate)

### Command Executed

```shell
gcx kg insights search
```

### Command Output

```shell
Error: Knowledge Graph API request failed (HTTP 500)
│
│ kg: search assertions
│
│ Cannot invoke "java.util.List.stream()" because "filterCriteria" is null
│
└─
```

### What happened?

See above

It works, or gives me help text on what to do right.

Run `gcx kg insights search`

Every time

### Environment Details

macOS

- [ ] I ran the command with -v, -vv, or -vvv flag for verbose output
- [ ] I checked my configuration with `gcx config view`
- [ ] I verified I'm using the correct context with `config current-context`
- [ ] I verified my authentication token/credentials are valid
- [ ] I checked the documentation for this command

_No response_

xujiaxj:
## Summary

- `gcx kg insights search` was hitting `/v1/search/assertions` (generic graph search) with a `SearchRequest` payload. With no flags this returned HTTP 500 (`filterCriteria null`); with `--type` alone it returned HTTP 422. That's the wrong endpoint — the Asserts UI's "Entities with Insights" panel uses `/v1/assertions/search` with a label-matcher rule DSL.
- Repointed the command at `/v1/assertions/search` and reshaped flags to match the UI's model (`--type`, `--insight op=value` repeatable, `--severity` comma-list). Added required-flag guard so empty calls fail fast with a usage error instead of a server 500.

## Flag surface

| Flag | Purpose |
|---|---|
| `--type` | Root entity type, default `Service` |
| `--insight` | Insight-name rule `op=value` (`contains`/`starts-with`/`equals`); repeatable, ORed |
| `--severity` | Comma-list of `critical,warning,info`; ANDed into every rule |
| `--env`/`--namespace`/`--site`/`--from`/`--to`/`--since` | Shared scope/time flags |

`MarkFlagsOneRequired("insight", "severity")` shields users from the server's 500 on empty `searchCriteria` (the backend validator allows empty but PromQL fails on the resulting empty query). Severity-only calls seed an `asserts_assertion_name IS NOT NULL` matcher so severity has a group to attach to.

## Test plan

- [x] `gcx kg insights search` → "At least one of the flags in the group [insight severity] is required"
- [x] `gcx kg insights search --insight foo=bar` → "Invalid --insight op \"foo\": must be one of contains, starts-with, equals"
- [x] `gcx kg insights search --insight contains=Saturation` → returns entity list
- [x] `gcx kg insights search --severity critical` alone → returns entity list (severity-only seeding works)
- [x] `gcx kg insights search --type Namespace --insight starts-with=Latency --since 1h` → returns namespace results
- [x] Unit tests added for `buildInsightSearchRequest` (8 cases covering op mapping, severity AND-ing, seed behavior, validation errors)
- [x] `GCX_AGENT_MODE=false mise run all` — tests, build, docs all pass; lint findings limited to sibling worktrees

Fixes #670

### DIFF
--- a/docs/reference/cli/gcx_kg_insights.md
+++ b/docs/reference/cli/gcx_kg_insights.md
@@ -27,7 +27,7 @@ Query Knowledge Graph insights.
 * [gcx kg insights example](gcx_kg_insights_example.md)	 - Print an example insights request YAML.
 * [gcx kg insights graph](gcx_kg_insights_graph.md)	 - Query insights with graph topology.
 * [gcx kg insights query](gcx_kg_insights_query.md)	 - Query insights for a time range.
-* [gcx kg insights search](gcx_kg_insights_search.md)	 - Search for insights matching a query.
+* [gcx kg insights search](gcx_kg_insights_search.md)	 - Find entities with active insights matching the given rules.
 * [gcx kg insights source-metrics](gcx_kg_insights_source-metrics.md)	 - Get source metrics for a specific insight.
 * [gcx kg insights summary](gcx_kg_insights_summary.md)	 - Get insights summary for a time range.
 
--- a/docs/reference/cli/gcx_kg_insights_search.md
+++ b/docs/reference/cli/gcx_kg_insights_search.md
@@ -1,24 +1,41 @@
 ## gcx kg insights search
 
-Search for insights matching a query.
+Find entities with active insights matching the given rules.
+
+### Synopsis
+
+Find entities with active insights matching the given rules.
+
+Backed by the same endpoint the Asserts UI's "Entities with Insights" panel uses.
+Each --insight flag is a separate rule (ORed together); severities are ANDed
+into every rule.
 
 ```
 gcx kg insights search [flags]
 ```
 
+### Examples
+
+```
+  gcx kg insights search --insight contains=Saturation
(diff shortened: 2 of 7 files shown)


xujiaxj:
The current endpoint `/v1/search/assertions` is actually NOT insight/assertion search. It's an entity search endpoint. To search for insights, i.e., from the other around, we should use `v1/assertions/search` instead. This PR repurpose this command to do insight search. Essentially it implements this UI feature:

Monosnap Image 2026-05-11 15-36-50

xujiaxj:
I found the insight filter flag rather awkward, i.e, `--insight contains=Saturation`, but I don't have a better idea. Please let me know if you have some suggestions.

xujiaxj:
Monosnap Image 2026-05-11 15-50-26

radiohead (approved):
Nice work! RE: UX clunkiness – I don't really have a suggestion here, so let's keep it as-is for now.

PR #694, opened 2026-05-12
refactor(kg)!: remove UI-centric insights query/summary/graph commands

xujiaxj:
## Summary

- Remove `gcx kg insights query`, `summary`, and `graph` — these three subcommands mirror the Asserts Workbench's Summary, Timeline, and Graph panels. Their API responses are shaped for the UI, not agent consumption, and we don't plan to maintain them as a stable agent surface.
- `gcx kg entities inspect` already covers ~80% of the data they returned. The remaining gap — aggregate alert counts, a few extra Prometheus labels also present elsewhere, and multi-entity request batching — isn't material to agentic workflows.
- Also removes the now-orphaned `gcx kg insights example` helper (its only purpose was to print the request YAML for the three removed commands).

Remaining under `gcx kg insights`: `entity-metric`, `search`, `source-metrics`.

## Test plan

- [x] `mise run lint` clean for the live tree (remaining 29 hits are all in other agents' `.claude/worktrees/`, pre-existing)
- [x] `go test ./...` passes (`TestConsistency_NoOrphanedRegistryEntries` updated by removing the four registry entries in `internal/agent/command_annotations.go`)
- [x] `mise run reference` regenerates — deletes four orphaned CLI reference pages and updates the parent `gcx_kg_insights.md`
- [x] `mise run docs` builds successfully; no warnings referencing the removed commands
- [x] Verify `gcx kg insights --help` shows only `entity-metric`, `search`, `source-metrics`


xujiaxj:
Addressed in 16e0f27c — updated both tables to reference `insights search` and `entities list` (the surviving real commands; note: the original `search entities` in the table was itself stale and didn't match any real command, so I used `entities list` instead of the bot's suggested `search entities`).

annanay25 (approved):
lgtm. should we also considering merging the remaining commands under insights into `gcx kg entities`? (can do in a future PR).

xujiaxj:
> lgtm. should we also considering merging the remaining commands under insights into `gcx kg entities`? (can do in a future PR).

This is the part I need some help next. What's the best way to organize these commands? We can discuss offline.

PR #696, opened 2026-05-13
docs(profiles): hint at profile-types command in --profile-type flag help

simonswine:
## Summary

- Adds a hint to the `--profile-type` flag help text, pointing users to `gcx datasources pyroscope profile-types` to discover available profile type IDs
- Regenerates CLI reference docs to reflect the updated flag description

## Notes

Flag descriptions use single quotes (not backticks) to avoid pflag's metavar substitution, which would otherwise replace the `string` type token with the backtick-quoted text and drop it from the description.

### DIFF
--- a/docs/reference/cli/gcx_datasources_pyroscope_query.md
+++ b/docs/reference/cli/gcx_datasources_pyroscope_query.md
@@ -40,7 +40,7 @@ gcx datasources pyroscope query [EXPR] [flags]
       --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
       --max-nodes int         Maximum nodes in flame graph (default 1024)
   -o, --output string         Output format. One of: agents, graph, json, table, wide, yaml (default "table")
-      --profile-type string   Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds') (required)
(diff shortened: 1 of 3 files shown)


docs/reference/cli/gcx_datasources_pyroscope_query.md
```diff
@@ -40,7 +40,7 @@ gcx datasources pyroscope query [EXPR] [flags]
       --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
       --max-nodes int         Maximum nodes in flame graph (default 1024)
   -o, --output string         Output format. One of: agents, graph, json, table, wide, yaml (default "table")
-      --profile-type string   Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds') (required)
+      --profile-type string   Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds'); use 'gcx datasources pyroscope profile-types' to list available (required)
```
marcsanmi: Should we mentioned at least both? I'm not sure but I think this is the recommended one (even you can still access it through datasources).
```gcx profiles profile-types```

marcsanmi (approved):
LGTM

PR #699, opened 2026-05-13
feat(irm): add incidents contexts list command

kozhuhds:
## Summary

- Adds `gcx irm incidents contexts list <incident-id>` (alias `context`, `ctx`), exposing `IncidentContextService.QueryIncidentContext`. Supports `--type`, `--status`, `--alert-group-id`, and `--limit` filters and renders a table by default with a `wide` variant.
- Adds the matching client method, types, table codec, and tests (client + codec).
- Registers the new leaf under `internal/agent/command_annotations.go` (`small` cost) so the agent-mode consistency test passes.

## Why

The incident-context endpoint is the only way to discover the alert groups (and other context surfaces) linked to an incident. Until now, anyone needing this from the CLI had to drop down to a raw `curl` against the IRM plugin proxy with a hand-managed bearer token. This makes it a normal `gcx` call.

One non-obvious thing the smoke-test surfaced: alert-group links come back as `type=genericURL` contexts with the `alertGroupID` field set — there is no dedicated `alertGroup` type value in the IRM API. The `--type` flag help calls this out and points users to `--alert-group-id` for that case.

## Test plan

- [x] `go test ./internal/providers/irm/...` — new client + codec tests pass
- [x] `go test ./cmd/gcx/root/...` — consistency test passes (new leaf has token-cost annotation)
- [x] `mise run lint` — clean
- [x] `GCX_AGENT_MODE=false mise run reference` — regenerated CLI reference docs are committed

### DIFF
--- a/docs/reference/cli/gcx_irm_incidents_contexts.md
+++ b/docs/reference/cli/gcx_irm_incidents_contexts.md
@@ -1,0 +1,27 @@
+## gcx irm incidents contexts
+
+Manage incident contexts (linked alert groups, dashboards, etc.).
+
+### Options
+
+```
+  -h, --help   help for contexts
+```
+
+### Options inherited from parent commands
+
+```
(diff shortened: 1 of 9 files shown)


xujiaxj (approved):
LGTM. With the context, we can then retrieve the alert group, and with alert group, we can retrieve alerts.

PR #702, opened 2026-05-13
refactor(kg)!: move insights search to entities find-by-insight

xujiaxj:
## Summary
- Move `gcx kg insights search` to `gcx kg entities find-by-insight` — the command returns entity keys, not insights, so it belongs under the entities subtree.
- Reserves `gcx kg entities search` for future use.
- Updates agent annotations and regenerates CLI reference docs.

## Test plan
- [x] `mise run lint` clean
- [x] `go test ./internal/providers/kg/... ./internal/agent/...` pass
- [x] `GCX_AGENT_MODE=false mise run reference` regenerated
- [x] `GCX_AGENT_MODE=false mise run docs` builds
- [x] Confirm no downstream callers (scripts, runbooks) still invoke `gcx kg insights search`


xujiaxj:
This is the follow-up PR from an internal naming discussion.

internal/providers/kg/commands.go
```diff
@@ -878,7 +878,61 @@ func newEntitiesCommand(loader RESTConfigLoader) *cobra.Command {
...
+		searchNames      []string
+		searchSeverities []string
+		searchScope      scopeFlags
+	)
+	cmd := &cobra.Command{
+		Use:   "find-by-insight",
+		Short: "Find entities with active insights matching the given rules.",
+		Long: `Find entities with active insights matching the given rules.
+
+Backed by the same endpoint the Asserts UI's "Entities with Insights" panel uses.
```
annanay25: do we need this in the description?

internal/providers/kg/commands.go
```diff
@@ -878,7 +878,61 @@ func newEntitiesCommand(loader RESTConfigLoader) *cobra.Command {
...
+Each --insight flag is a separate rule (ORed together); severities are ANDed
+into every rule.`,
+		Example: `  gcx kg entities find-by-insight --insight contains=Saturation
```
annanay25: i think the syntax is confusing`find-by-insight --insight contains=Saturation`

i would like to write something along the lines of

`find-by-insight <insight name>`
or
`find-by-insight --regex <insight regex>`

xujiaxj: Good point! Also the filter format is also confusing as I mentioned in the previous PR, now the command name makes it worse. Let me come back with a better approach.

xujiaxj: The issue is that this command also take `--type` which is the entity type, and `--severity` which is the severity, so the `=Saturation` or `contains Saturation` should be after a flag, not after the command. It's not search for an insight, it's find an entity that has some insights.

xujiaxj: how about this:

```
    gcx kg entities find-by-insight --insight name=~Saturation                
      gcx kg entities find-by-insight --insight name=ErrorRatioBreach --      
  severity critical                                                           
      gcx kg entities find-by-insight --severity critical,warning --namespace 
  mimir-prod-01                                                               
      gcx kg entities find-by-insight --type Namespace --insight name=~Latency --
  since 1h   
  
        --insight stringArray   Filter by insight property: name=value (exact) or name=~value (contains); repeatable, rules are ORed. Only "name" is supported today.
```
        
This aligns with `--property` syntax in `entities list`, so later on we can merge `list` and `find-by-insights`, so `--insight` is just another flag with similar syntax as `--property`.

```
      --property stringArray   Filter by property: name=value (exact) or name=~value (contains); repeatable (run 'gcx kg meta schema' to list property names)
```

xujiaxj: (I don't want to combine them just yet, as their return shape is different, so that needs more work to bring the return shape to the same.)

xujiaxj:
Close as I think we should do https://github.com/grafana/gcx/pull/730 instead.

PR #703, opened 2026-05-14
feat(metrics): add --time flag for instant queries at a specific timestamp

dafydd-t:
## Summary

`gcx metrics query` has no way to run an instant query at a historical timestamp. The only options were no time flags (meaning query now) or a range query (`--from`/`--to`). This forced users and agents into range queries with arbitrary step sizes when they needed a single point-in-time value, which caused errors in sampling values

The new `--time` flag evaluates an instant query at a specific timestamp:

```
gcx metrics query 'max_over_time(cache_refresh_lag[6h])' --time 2026-05-13T16:30:00Z
```

It accepts RFC3339, Unix timestamps, and relative times (`now-5m`). Mutually exclusive with `--from`/`--to`/`--since`.

## Context

This came out of benchmarking gcx-based agents against MCP-based agents on o11y-bench. The lack of instant query support was the single largest contributor to GCX agent score regression. You can see the analysis [here](https://github.com/grafana/gcx-skills-bench/blob/2026-06-02-opus/comparisons/2026-06-02-gcx-0-3-0-vs-mcp/2026-06-02-gcx-0-3-0-vs-mcp-comparison.md#1-promql-numeric-accuracy-failures-4pp-impact)

## Followups

the other MTLP datasources also support instant queries, so I'll make another PR to add instant query support for those datasources too.

## Result

Here's a comparison report, from running the promQL tasks from o11y-bench, using this branch, and main:

**Date**: 2026-06-03  
**Runs compared**:
- `jobs/2026-06-02-opus-4-6-gcx-0-3-0-n3` (**gcx-0-3-0**) — stable 0.3.0, built 2026-05-26
- `jobs/2026-06-03-opus-4-6-gcx-instant-flag-updated-range-guidance-2-promql-n3` (**updated-guidance-2**) — devel 3d022b95, 2026-06-03, with `--time` flag + instant/range guidance

**Model**: claude-opus-4-6, reasoning_effort: off, GcxOpenCodeAgent  
**Tasks**: 16 (prometheus_query only, common to both runs), 3 trials each

### Headline numbers

| Metric | gcx-0-3-0 (B) | updated-guidance-2 (D) |
|---|---|---|
| Overall mean score | **0.792** | **0.913** |
| 95% CI | [0.686, 0.891] | [0.843, 0.971] |
| Perfect tasks (all trials 1.0) | 5 | 10 |
| Mean cost per trial | $0.369 | $0.272 |
| Total cost | $17.72 | $13.03 |
| Mean wall-clock time | 99.1s | 75.2s |
| Mean trial-to-trial stdev | 0.133 | 0.102 |

*95% CI on delta (D−B): [+0.007, +0.237] — statistically significant at n=16 (CI excludes zero, barely).*

D scores 12.1pp higher, runs 24% faster, and costs 26% less per trial. D has 10/16 perfect tasks vs B's 5/16. The gap is primarily explained by the `--time` flag: without it, B's agent either fails with "Unknown flag" or falls back to range queries that return wrong time-point values.

### DIFF
--- a/docs/reference/cli/gcx_datasources_prometheus_query.md
+++ b/docs/reference/cli/gcx_datasources_prometheus_query.md
@@ -23,6 +23,9 @@ gcx datasources prometheus query [EXPR] [flags]
   # Instant query using configured default datasource
   gcx datasources prometheus query 'up{job="grafana"}'
 
+  # Instant query at a specific time
+  gcx datasources prometheus query 'rate(http_requests_total[5m])' --time 2026-01-15T10:30:00Z
+
   # Range query with explicit datasource UID
   gcx datasources prometheus query -d UID 'rate(http_requests_total[5m])' --from now-1h --to now --step 1m
 
@@ -49,6 +52,7 @@ gcx datasources prometheus query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
       --to string           End time (RFC3339, Unix timestamp, or relative like 'now')
 ```
 
--- a/docs/reference/cli/gcx_datasources_prometheus_query.md
+++ b/docs/reference/cli/gcx_datasources_prometheus_query.md
@@ -52,7 +52,7 @@ gcx datasources prometheus query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
-      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). Mutually exclusive with --from/--to/--since
       --to string           End time (RFC3339, Unix timestamp, or relative like 'now')
 ```
 
--- a/docs/reference/cli/gcx_metrics_billing_query.md
+++ b/docs/reference/cli/gcx_metrics_billing_query.md
@@ -46,6 +46,7 @@ gcx metrics billing query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
       --to string           End time (RFC3339, Unix timestamp, or relative like 'now')
 ```
 
--- a/docs/reference/cli/gcx_metrics_billing_query.md
+++ b/docs/reference/cli/gcx_metrics_billing_query.md
@@ -46,7 +46,7 @@ gcx metrics billing query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
-      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). Mutually exclusive with --from/--to/--since
       --to string           End time (RFC3339, Unix timestamp, or relative like 'now')
 ```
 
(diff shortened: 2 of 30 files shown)


annanay25 (commented):
Nice work — the motivation is solid and the Prometheus implementation looks clean.

One concern: `--time` is registered on the **shared** `TimeRangeOpts.SetupTimeFlags()`, so it shows up in the help text for every command that uses shared time flags (Tempo, Loki, Pyroscope, etc.). But only the Prometheus query command actually reads `shared.Time` and acts on it — in every other command, `--time` would be silently ignored.

For example, Tempo's `buildMetricsRequest` in `internal/datasources/tempo/metrics.go` calls `shared.ParseTimes(now)` and `shared.IsRange()` but never checks `shared.Time`. So `gcx traces metrics '{ } | rate()' --time 2026-05-13T16:30:00Z` would pass validation but silently ignore the timestamp.

Tempo *does* have instant query support (`api/metrics/query` vs `api/metrics/query_range`) and an existing `--instant` flag, so wiring `--time` in there would be straightforward. Same goes for Loki which has similar instant/range semantics.

A couple of options:
1. Wire `--time` into Tempo/Loki/Pyroscope commands too (since they all support instant queries)
2. Move `--time` out of the shared `SetupTimeFlags` and register it only on commands that implement it (Prometheus for now)

Either way, we should alter the current state where the flag appears in help text but is silently ignored on most commands — especially for agents that will see it in `--help` and assume it works everywhere.

dafydd-t:
Thanks @annanay25 , and apologies - I meant to create this in draft 😮‍💨 . It needs more thought for sure!

dafydd-t:
went with option 2, and we'll add --time support to the other MTLP datasources in another PR

docs/reference/cli/gcx_datasources_prometheus_query.md
```diff
@@ -49,6 +52,7 @@ gcx datasources prometheus query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
```
dafydd-t:
```suggestion
      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). Mutually exclusive with --from/--to/--since
```

docs/reference/cli/gcx_metrics_billing_query.md
```diff
@@ -46,6 +46,7 @@ gcx metrics billing query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
```
dafydd-t:
```suggestion
      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). Mutually exclusive with --from/--to/--since
```

docs/reference/cli/gcx_metrics_query.md
```diff
@@ -49,6 +52,7 @@ gcx metrics query [EXPR] [flags]
       --share-link          Print the Grafana Explore URL for the executed query to stderr
       --since string        Duration before --to (or now if omitted); mutually exclusive with --from
       --step string         Query step (e.g., '15s', '1m')
+      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since
```
dafydd-t:
```suggestion
      --time string         Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). Mutually exclusive with --from/--to/--since
```

internal/datasources/query/opts.go
```diff
@@ -25,8 +26,25 @@ func (opts *TimeRangeOpts) SetupTimeFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&opts.Since, "since", "", "Duration before --to (or now if omitted); mutually exclusive with --from")
 }
 
+// SetupInstantFlag registers the --time flag for instant queries at a specific timestamp.
+// Separated from SetupTimeFlags so commands can opt in only when they support instant queries.
+func (opts *TimeRangeOpts) SetupInstantFlag(flags *pflag.FlagSet) {
+	flags.StringVar(&opts.Time, "time", "", "Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'); mutually exclusive with --from/--to/--since")
```
dafydd-t:
```suggestion
	flags.StringVar(&opts.Time, "time", "", "Evaluation time for an instant query (RFC3339, Unix timestamp, or relative like 'now-5m'). mutually exclusive with --from/--to/--since")
```

docs/reference/cli/gcx_datasources_prometheus_query.md
dafydd-t: Addressed — fixed in source (), docs regenerated.

docs/reference/cli/gcx_metrics_billing_query.md
dafydd-t: Addressed — fixed in source (), docs regenerated.

docs/reference/cli/gcx_metrics_query.md
dafydd-t: Addressed — fixed in source (), docs regenerated.

internal/datasources/query/opts.go
dafydd-t: Addressed — fixed in source (), docs regenerated.

radiohead (approved):

PR #706, opened 2026-05-14
docs(explore-datasources): document --time flag for instant queries

dafydd-t:
## Summary

- adds `--time` example for instant queries at a specific point in time
- adds guidance on choosing instant (`--time`) vs range (`--from`/`--to`) queries
- specifically calls out the pattern for point-in-time comparisons: two instant queries with different `--time` values

## Context

ran o11y-bench with the gcx branch that supports `--time` for metrics queries. the model only used `--time` in 4/12 trials — and those 4 all scored 1.0, while the 8 that fell back to range queries scored 0.23–0.55. the skill didn't document `--time` at all, so the model had no reason to reach for it.


wardbekker (approved):
LGTM - docs only

PR #709, opened 2026-05-15
refactor(kg)!: move cypher command under kg entities

xujiaxj:
## Summary

- Moves `gcx kg cypher` to `gcx kg entities query`, grouping it with `list` and `inspect` under the entities subcommand
- Renames the subcommand from `cypher` to `query`, aligning with the established signal provider pattern (`gcx metrics query`, `gcx logs query`, `gcx traces query`, `gcx profiles query`); the short description and examples still mention Cypher by name for discoverability

## Test plan

- [x] `gcx kg entities query "MATCH (s:Service) RETURN s LIMIT 10"` works
- [x] `gcx kg cypher` and `gcx kg entities cypher` no longer exist
- [ ] `gcx kg entities --help` lists `query` as a subcommand


docs/reference/cli/gcx_kg_entities.md
```diff
@@ -23,6 +23,7 @@ Manage Knowledge Graph entities.
 ### SEE ALSO
 
 * [gcx kg](gcx_kg.md)	 - Manage Grafana Knowledge Graph rules, entities, and insights
+* [gcx kg entities cypher](gcx_kg_entities_cypher.md)	 - Run a read-only Cypher query against the Knowledge Graph.
```
annanay25: should we name this `kg entities query` instead? cypher is the implementation detail so that should be in help text.

kozhuhds:
thinking if we should use word "query" somehow. 2 options in mind: 'kg entities query' or 'kg entities cypher-query' wdyt?

xujiaxj:
`query` is fine. cypher is the query language, so in the command description we could mention that.

docs/reference/cli/gcx_kg_entities.md
xujiaxj: `query` sounds good. It also aligns with `gcx metrics/logs/traces query`.

annanay25 (approved):

PR #683, opened 2026-05-11
feat(irm): add OnCall alert-groups rich shape and SRE triage surface

radiohead:
## What

Rebuilds the OnCall `AlertGroup` and `Alert` types around a K8s-style subject + dimensions envelope and adds a filterable SRE triage surface to the CLI.

New and reworked commands:

- `alert-groups list` — filterable by state, team, integration, resolved/silenced status; defaults to active groups only (`firing,acknowledged,silenced` + `is_root=true`); supports `--limit` with cursor-aware truncation hint
- `alert-groups list-alerts` — shows individual alerts under a group with a link-focused wide table; replaces the removed `alerts` / `alerts get` sub-commands
- `alert-groups get` — retrieves a single group; emits three drilldown hints (list-alerts, rule dashboard, rule alerts) with real identifiers populated from the result
- Action verbs `acknowledge` / `unacknowledge` / `resolve` / `unresolve` / `silence` / `unsilence` / `delete` — all emit the `MutationResult` envelope, support bulk-by-filter, and require `--yes` in agent mode when matched count exceeds one

Output codec alignment: all list and action-verb commands register `table` as their default codec; `wide` is the expanded layout. All list paths wrap results in an `{"items": [...]}` envelope.

`emitWarn` and `emitNote` helpers added alongside the existing `emitHint` pattern; agent mode routes them as `{"class":"warning",...}` / `{"class":"note",...}` JSONL on stderr (FR-104). `EmitHint`, `EmitWarn`, and `EmitNote` are now exported from `internal/output` so providers delegate to a single implementation. `NewOrderedYAMLCodec` is extracted to `internal/format` for reuse.

`DiscoverFields` is now recursive, making `--json` field projection work for nested paths such as `status.state` and `status.subject.labels.*`. `ErrorToDetailedError` in `cmd/gcx/fail` is updated to match both value-typed and pointer-typed `DetailedError` returns, restoring exit-code passthrough for login and resources commands.

The `gcx irm oncall alerts` sub-command (and `gcx irm oncall alerts get`) is removed; `alert-groups list-alerts` is the replacement.

## Why

The previous `AlertGroup` and `Alert` types were shallow stubs without the field coverage needed for SRE triage: no decoded state, no severity, no integration name, no links to Grafana alerting rules or dashboards. The new rich shape surfaces enough context for an on-call engineer to acknowledge, silence, or resolve a group directly from the CLI without a browser round-trip.

[...]

PR #720, opened 2026-05-15
feat(output): add --jq flag for inline JSON transformation

### LINKED ISSUE #705: add --jq flag for inline JSON transformation
When agents use gcx for investigations, they often pipe output to Python scripts for data transformation — grouping, reducing, filtering beyond what `--json` field selection offers. This has two problems:

1. **Safety**: users are nervous auto-approving arbitrary Python scripts that agents generate
2. **Reliability**: these scripts sometimes error out and need retries, slowing down the investigation loop + causing additional query requests

A built-in `--jq` flag would let agents (and humans) do inline transformations without leaving the gcx process, removing the need for external scripting in most cases.

wardbekker:
## Summary

- Add a `--jq` flag (per-command, bound via `Options.BindFlags`, sibling to `--json`) that applies a [jq](https://jqlang.github.io/jq/) expression to JSON output before stdout, removing the need for external Python pipelines during agent investigations.
- New `JQCodec` (uses [`github.com/itchyny/gojq`](https://github.com/itchyny/gojq) — pure Go, no CGO, zero runtime deps) emits real-jq-compatible NDJSON (one pretty-printed value per line).
- Mutually exclusive with `--json`; requires JSON-shaped output; auto-flips `-o` to `json` when unset; jq expression is parsed at `Validate()` so syntax errors fail fast before any data is fetched.

## Why

Agents using gcx for investigations pipe JSON output to ad-hoc Python scripts for grouping, reducing, and filtering — unsafe to auto-approve, and unreliable enough to need retries. A built-in jq path keeps transformation in-process: no scripts to approve, no retry loops, no extra query traffic.

## Changes

- `internal/output/jq.go` — `JQCodec` implementing `format.Codec`. Bypasses the agents-codec spill behavior (callers asking for `--jq` want the data, not a `/tmp` pointer).
- `internal/output/format.go` — `JQExpression` / `jqQuery` fields on `Options`, `--jq` flag in `BindFlags`, new `applyJQFlag()` validator, branch in `Encode()`, agent-hint suppression when jq is active.
- `internal/output/jq_test.go` — table-driven codec tests plus options-level integration tests covering auto-flip, mutual exclusion, parse-time errors.
- `docs/design/output.md` — new subsection 1.6 mirroring the `--json` section structure.
- `docs/reference/cli/*` — regenerated CLI reference (every command now lists `--jq`).
- `go.mod` / `go.sum` — `github.com/itchyny/gojq` v0.12.19.

## Example

```bash
gcx resources get dashboards --jq '.items | length'
gcx resources get dashboards --jq '.items[] | .metadata.name'
gcx resources get dashboards --jq '[.items[] | {name: .metadata.name, title: .spec.title}]'
```

Errors fail fast:
```bash
gcx resources get dashboards --jq '.' -o yaml       # --jq requires JSON output
gcx resources get dashboards --jq '.' --json name   # --jq and --json cannot be used together
gcx resources get dashboards --jq 'broken['         # invalid --jq expression
```

## Test Plan

- [x] Unit + integration tests pass (`go test ./internal/output/...`)
- [x] `GCX_AGENT_MODE=false mise run all` passes (lint + tests + build + docs)
- [x] Reference docs regenerated cleanly
- [x] Live smoke against a real stack — happy path, all three error paths, NDJSON streaming all confirmed
- [ ] Reviewer: spot-check that no existing JSON-encoding bypass (e.g. `cmd/gcx/resources/get.go:281`) needs to thread through `--jq` — mutual exclusion with `--json` keeps those branches dormant.

Closes #705


wardbekker:
## Shepherd Status (iteration 2, polling)

### Ready to merge once approved
All required checks green, branch rebased on top of latest `main`, CLA signed.

### CI Checks
| Check | Status |
|-------|--------|
| Linters | PASS |
| Tests | PASS |
| Documentation | PASS |
| TruffleHog | PASS |
| zizmor | PASS |

### Changes in this iteration
- Rebased onto `main` — resolved conflicts in `internal/output/format.go` (merged the new `emitHint`/FR-104 framework from #683 with this PR's `--jq` feature), `go.mod`/`go.sum` (added `gojq` deps alongside the `jaeger-idl` v0.6.0 bump), and regenerated CLI reference docs for the `alert-groups` changes.
- Per author request: the agent-mode hint now **also fires when `--json field1,field2` is in use**, nudging callers toward `--jq` for transformations (group_by / filter / count) that field selection can't do. Still suppressed for `--jq` itself and `--json list` discovery.

### Pending
- Awaiting human review/approval

---
*Updated by shepherd-pr*

internal/output/format.go
```diff
@@ -148,6 +158,49 @@ func (opts *Options) applyJSONFlag() error {
...
+		return fmt.Errorf("--jq requires JSON output, but -o %s was specified", outputFlag.Value.String())
+	}
+
+	expr := jqFlag.Value.String()
+	query, err := gojq.Parse(expr)
+	if err != nil {
+		return fmt.Errorf("invalid --jq expression: %w", err)
+	}
+
+	opts.JQExpression = expr
```
radiohead, line 198: This is actually not used anywhere, do we still need it?

internal/output/jq.go
```diff
@@ -0,0 +1,101 @@
...
+			return fmt.Errorf("jq runtime: %w", e)
+		}
+		if err := encoder.Encode(v); err != nil {
+			return err
+		}
+	}
+}
+
+func (c *JQCodec) Decode(src io.Reader, value any) error {
+	return format.NewJSONCodec().Decode(src, value)
```
radiohead, line 59: Small nit, we don't need to create a new instance of `format.NewJSONCodec()` upon each `Decode`, we could just create one instance per `JQCodec` instead.

radiohead (approved):
I think overall this looks alright, but I'd like us to get rid of `--json` in the future. Having both `--json` and `--jq` feels like feature bloat to me.

PR #724, opened 2026-05-15
feat(config): interactive picker and previous-context for use-context

Pokom:
## Summary

Two small ergonomic additions to `gcx config use-context`, both kubectx-flavored:

- **No-arg picker** — `gcx config use` (no arguments) drops into a `huh` select list of available contexts, highlighting the current one. Rejects cleanly with a structured error in agent mode or when stdout is not a TTY.
- **Previous-context** — `gcx config use -` switches back to the previously active context. The name is persisted atomically under `$XDG_STATE_HOME/gcx/previous-context` so the affordance survives shell invocations. Same-context switches are a no-op and don't pollute the history.

Stays in-grammar (`$AREA $VERB` under `config`) — no top-level shortcut, to honor `CONSTITUTION.md` §CLI Grammar. The `use` alias is unchanged.

**Status: draft — gathering feedback** on whether the UX feels right before promoting. If you've used `kubectx` for cluster switching, this aims to be the same muscle memory in `gcx`.

## Test plan

- [x] Unit tests for `internal/config/previous.go` (round-trip, overwrite, empty-rejection, state-dir creation, whitespace trimming)
- [x] Integration tests via `CommandTestCase`: previous-switch toggle (a→b→-→a→-→b), error when no previous recorded, no-op self-switch doesn't clobber history
- [x] Existing `use-context` tests isolated under per-test `XDG_STATE_HOME` so they don't pollute the developer's state dir
- [x] `mise run lint` clean
- [x] `mise run reference` regenerated `docs/reference/cli/gcx_config_use-context.md`
- [x] Smoke-tested end-to-end against a synthetic 3-context config
- [x] Manual: try the picker on your own config and report whether the labeling (`<name> (current)`) and sort order feel right
- [x] Manual: confirm `-` round-trips between two contexts as expected

## Notes for reviewers

- The picker isn't unit-tested directly (huh is interactive-only); the rejection path is covered, the happy path is manual.
- Previous-context write failures emit a stderr warning but don't fail the switch — it's bookkeeping, not user data.
- A companion shell function (`gctx`) layered on top is in my dotfiles for keystroke economy; not part of this PR.

### DIFF
--- a/cmd/gcx/config/command.go
+++ b/cmd/gcx/config/command.go
@@ -2,10 +2,13 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
+	"github.com/charmbracelet/huh"
 	"github.com/grafana/gcx/cmd/gcx/fail"
 	"github.com/grafana/gcx/internal/agent"
 	"github.com/grafana/gcx/internal/config"
@@ -16,6 +19,7 @@ import (
 	"github.com/grafana/gcx/internal/resources/discovery"
 	"github.com/grafana/gcx/internal/secrets"
 	"github.com/grafana/gcx/internal/style"
+	"github.com/grafana/gcx/internal/terminal"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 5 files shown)


cmd/gcx/config/command_test.go
```diff
@@ -5,10 +5,20 @@ import (
...
 )
 
+// isolateStateHome points xdg.StateHome at a per-test tempdir so use-context
+// invocations don't pollute the developer's real previous-context state file.
+func isolateStateHome(t *testing.T) {
+	t.Helper()
+	t.Setenv("XDG_STATE_HOME", t.TempDir())
+	t.Cleanup(func() { xdg.Reload() })
+	xdg.Reload()
+}
```
Pokom: I'm not sure how user friendly this is across platforms.

radiohead:
@claude Please review this pull request

internal/config/previous.go
```diff
@@ -0,0 +1,58 @@
...
+	if name == "" {
+		return errors.New("previous context name cannot be empty")
+	}
+
+	path := PreviousContextPath()
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return fmt.Errorf("create previous-context dir: %w", err)
+	}
+
+	tmp := path + ".tmp"
```
radiohead: question: this could theoretically break if there are two invocations at the same time, do you think it's worth it trying to protect against that scenario? Realistically I don't expect this to happen but who knows.

cmd/gcx/config/command.go
```diff
@@ -464,42 +468,65 @@ func useContextCmd(configOpts *Options) *cobra.Command {
...
 			layered, err := config.LoadLayered(cmd.Context(), configOpts.ConfigFile)
 			if err != nil {
 				return err
 			}
-			if !layered.HasContext(args[0]) {
-				return config.ContextNotFound(args[0])
+
+			target, err := resolveUseContextTarget(layered, args)
+			if err != nil {
+				return err
```
radiohead, line 507: suggestion: could we handle SIGINT gracefully here?
```suggestion
				if errors.Is(err, huh.ErrUserAborted) {
				        cmdio.Info(cmd.OutOrStdout(), "Aborted.")
				        return nil
				}

				return err
```

cmd/gcx/config/command.go
```diff
@@ -464,42 +468,65 @@ func useContextCmd(configOpts *Options) *cobra.Command {
...
-		Args:    cobra.ExactArgs(1),
+		Use:     "use-context [CONTEXT_NAME]",
+		Args:    cobra.MaximumNArgs(1),
 		Aliases: []string{"use"},
 		Short:   "Set the current context",
-		Long: `Set the current context and updates the configuration file.
+		Long: `Set the current context and update the configuration file.
+
+Run without arguments to pick a context interactively, or pass "-" to switch
+back to the previously active context.
```
radiohead, line 479: nit: I think we should help the agent here:
```suggestion
Run without arguments to pick a context interactively, or pass "-" to switch back to the previously active context.

In agent mode or when stdout is not a TTY, a context name is required.
```

cmd/gcx/config/command.go
```diff
@@ -510,6 +537,65 @@ user config), use --file to choose which layer to update.`,
...
+		return pickContextInteractively(cfg)
+	}
+	name := args[0]
+	if name == "-" {
+		prev, err := config.ReadPreviousContext()
+		if err != nil {
+			return "", err
+		}
+		if prev == "" {
+			return "", errors.New("no previous context recorded")
```
radiohead: suggestion: let's add a bit of help here.
```suggestion
			return "", errors.New("no previous context recorded — switch contexts at least once with 'gcx config use-context <name>' to enable '-'")
```

cmd/gcx/config/command.go
```diff
@@ -510,6 +537,65 @@ user config), use --file to choose which layer to update.`,
...
+		}
+		return prev, nil
+	}
+	return name, nil
+}
+
+func pickContextInteractively(cfg config.Config) (string, error) {
+	if agent.IsAgentMode() || !terminal.StdoutIsTerminal() {
+		return "", errors.New("interactive picker requires a TTY; pass a context name")
```
radiohead: suggestion: should we be a bit more verbose here?
```suggestion
if agent.IsAgentMode() {
    return "", errors.New("interactive picker disabled in agent mode; pass a context name")
}
if !terminal.StdoutIsTerminal() {
    return "", errors.New("interactive picker requires a TTY; pass a context name")
}
```

radiohead (commented):
This is a nice feature, thanks! A few suggestions / nits, could you take a look?

Pokom:
> This is a nice feature, thanks! A few suggestions / nits, could you take a look?


@radiohead I just got back from vacation, let me rebase this next week and take a peak at your nits. Appreciate the review!

internal/config/previous.go
Pokom: Yeah I think we should add protection regardless, it's not that much extra effort to mitigate the problem.

Pokom:
@radiohead mind taking a peak again? It took me a bit longer to get back, but I believe I've addressed all of your feedback. Appreciate it!

dafydd-t (approved):
LGTM, thanks for addressing those comments!

PR #726, opened 2026-05-16
diagnose-entity-graph skill: add propagation, collision, and split-env coverage

danstadler-pdx:
## Summary

Five content additions to the `diagnose-entity-graph` skill, each
codifying a diagnostic pattern that helps the agent reach the right
conclusion faster on common Entity Graph failure modes. Pure
content additions — no metadata, structure, or frontmatter changes.

The skill is positioned as a lightweight coordinator: where a check
is deterministic and already available (or about to be available)
in the `gcx kg diagnose` command itself, the skill describes how to
read the verdict and what to recommend — not how to replicate the
check by hand. Where a recommendation requires language-specific
or context-specific knowledge the command can't capture, the skill
provides that context.

## What's added

- **Phantom-edge filter callout** in Step 3. Interpretive rule:
  Tempo's metrics generator synthesizes `client="user"` /
  `connection_type="virtual_node"` edges for SERVER spans without
  incoming `traceparent`. The skill tells the agent to filter them
  out when counting real inter-service edges.

- **Self-loop interpretation rule** in Step 3. Service-graph
  self-loops (`client == server`) for a service that shouldn't be
  calling itself are a strong indicator of `service.name` collision.

- **New Step 3.5: Trace Context Propagation.** Pairs with the
  deterministic `Trace context propagation` check landing in the
  companion PR. Skill content is language-specific remediation —
  what to look for in the OTel SDK config (Python, Node.js, Go) when
  the diagnose check FAILs.

- **Service-name collision diagnostic** in Step 7. One PromQL query
  catches collisions; the Step 3 self-loop heuristic reinforces.
  Names the common cause (untemplated Helm chart value) and the
  fix (per-workload distinct `OTEL_SERVICE_NAME`).

- **Env-scope split diagnostic** in Step 7. One Cypher query returns
  each service's `scope.env`, surfacing the case where workloads in
  the same k8s namespace land in different env scopes and the graph
  appears disconnected.

- **Eventual-consistency caveat** in "Producing a Report." When
  recommending a fix, the skill tells the agent to set expectations:
  the `asserts:*` recording rules and `traces_*` metrics have a
  5–15 minute query window; old broken-state data will keep
  appearing until it ages out.

## Validation

- `scripts/validate-skills` passes (`validated claude-plugin/skills`).
- No changes to skill metadata, structure, or frontmatter — purely
  content additions inserted at the right places in the existing
  flow.

## What's not in this PR

- No changes to `gcx kg diagnose` itself. Step 3.5 pairs with the
  propagation check in the companion code PR.
- No new skills.
- No changes to other skills.

### DIFF
--- a/claude-plugin/skills/diagnose-entity-graph/SKILL.md
+++ b/claude-plugin/skills/diagnose-entity-graph/SKILL.md
@@ -66,6 +66,21 @@ and exposes the env on both sides as `client_deployment_environment` and
 `server_deployment_environment` — there is no unified `deployment_environment`
 label.
 
+**Filter out phantom edges.** When Tempo's metrics generator sees a SERVER-kind
+span with no incoming `traceparent`, it synthesizes a placeholder edge
+with `client="user"` and `connection_type="virtual_node"`. These represent
+"some external caller" — they are not real service-to-service edges. When
+counting inter-service edges for diagnosis, filter them out:
+
+```promql
+traces_service_graph_request_total{client!="user"}
+# or equivalently
+traces_service_graph_request_total{connection_type!="virtual_node"}
+```
+
+If the *only* edges in the env have `client="user"`, treat this as a strong
+signal that outgoing trace propagation is broken — see Step 4.
+
 ```bash
 # Service identity (OTel traces)
 gcx metrics query 'count(traces_target_info)' --since 1h
@@ -78,15 +93,53 @@ gcx metrics query 'count(traces_service_graph_request_total{server_deployment_en
 ```
 
 **Interpret:**
-- Both have data → traces are flowing. Continue to Step 4.
-- `traces_target_info` exists but `traces_service_graph_request_total` doesn't →
-  Tempo server-side metrics generation may not be enabled.
+- Both have data → traces are flowing. Continue to Step 5.
+- `traces_target_info` exists but `traces_service_graph_request_total` doesn't
+  (with `client!="user"` filter applied) → either Tempo server-side metrics
+  generation isn't enabled, or outgoing trace context isn't being propagated
+  by your services. Go to Step 4.
+- Service-graph contains only `client="user"` edges → broken trace context
+  propagation. Go to Step 4.
+- Service-graph shows self-loop edges (`client == server`) for a service
+  that shouldn't be calling itself → strong indicator of `service.name`
+  collision across multiple workloads. Inspect the colliding service in
+  Step 8.
 - Both empty → no OTel traces for this environment. Entities may still exist
-  via Prometheus scraping. Continue to Step 4.
+  via Prometheus scraping. Continue to Step 5.
 
 **Shortcut:** `gcx kg diagnose --env ENV` checks all five metrics automatically.
 
-## Step 4: Recording Rules
+## Step 4: Trace Context Propagation
+
+When `gcx kg diagnose --env <env>` emits a
+**`Trace context propagation: FAIL`** check, services are emitting
+spans but the calls between them aren't carrying `traceparent`
+headers — so Tempo's metrics generator can't link the spans into
+edges. Only phantom `client="user"` edges (Tempo's synthetic
+"external caller") form.
+
+The diagnose check's `Recommendation` already names the common
+causes and per-language remediation hints — read it as the
+primary playbook. Supplementary notes for less-common cases:
+
+### When propagation looks broken but auto-instrumentation IS installed
+
+- The application bypasses the instrumented HTTP client (raw socket
+  calls, a third-party SDK using its own transport, a non-standard
+  client library). Audit outbound network calls in the upstream
+  service's code path.
+- No `TracerProvider` was registered, so context isn't tracked at
+  all. Confirm via `gcx traces query` that spans on the *receiving*
+  service have a parent SpanID.
+
+### Convergence after a fix
+
+After the propagation fix lands, expect
+`traces_service_graph_request_total{client!="user"}` to populate
+within 1–2 minutes; recording-rule output in Step 5 follows ~5–10
+minutes after that.
+
+## Step 5: Recording Rules
 
 Recording rules convert raw metrics into the `asserts:*` metrics that Entity
 Graph consumes. These use `asserts_env`, not `deployment_environment`.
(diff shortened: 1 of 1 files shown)


claude-plugin/skills/diagnose-entity-graph/SKILL.md
```diff
@@ -79,13 +94,63 @@ gcx metrics query 'count(traces_service_graph_request_total{server_deployment_en
...
+- Service-graph shows self-loop edges (`client == server`) for a service
+  that shouldn't be calling itself → strong indicator of `service.name`
+  collision across multiple workloads. Inspect the colliding service in
+  Step 7.
 - Both empty → no OTel traces for this environment. Entities may still exist
   via Prometheus scraping. Continue to Step 4.
 
 **Shortcut:** `gcx kg diagnose --env ENV` checks all five metrics automatically.
 
+## Step 3.5: Trace Context Propagation
```
annanay25: Why step 3.5 when there are no 3.1/3.2 etc? :) 

Agree that we should split this (rather large) section into smaller subunits though.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
```diff
@@ -79,13 +94,63 @@ gcx metrics query 'count(traces_service_graph_request_total{server_deployment_en
...
+
+If `traces_target_info` exists but `traces_service_graph_request_total{client!="user"}`
+is empty (or you see only phantom `client="user"` edges), the services
+themselves are emitting spans but the calls between them aren't carrying
+`traceparent` headers. Tempo's metrics generator can't link the spans
+into edges. This is canonical Entity Graph failure mode #3 — "entities
+exist but no edges."
```
annanay25: What is the failure mode "3"? There's no reference to that or any other failure modes in the rest of the doc.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
```diff
@@ -79,13 +94,63 @@ gcx metrics query 'count(traces_service_graph_request_total{server_deployment_en
...
+
+If non-phantom edges are 0 and phantom edges are > 0, propagation is broken.
+
+**Common causes:**
+
+- The OTel SDK isn't injecting `traceparent` on outgoing requests. In
+  Python, the auto-instrumentation libraries for the HTTP client in use
+  (e.g. `urllib`, `urllib3`, `requests`) handle this. Check whether
+  `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` excludes them, or whether the
+  relevant `opentelemetry-instrumentation-*` package is installed.
```
annanay25: should we make this more generic? we don't know if the user is running python.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
danstadler-pdx: Fair — there's no 3.1–3.4 to anchor against. I'll renumber to slot it as its own step (likely Step 4, renumbering the rest down) and split the content into smaller sub-headings inside it. The slim already happened in the latest force-push; this is a structural change on top of that.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
danstadler-pdx: Good catch — that's leakage from an internal taxonomy. I'll remove the phrase entirely. The diagnose check already names the condition; the skill doesn't need a numbered reference.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
danstadler-pdx: Agreed. I'll rewrite to lead with the language-agnostic mechanism (OTel SDK isn't injecting `traceparent` on outgoing calls) and list Python / Node.js / Go / Java as equal examples, not Python-first.

danstadler-pdx:
Pushed the changes from the three replies above (commit `7905395`):

- **Renumbered** `## Step 3.5: Trace Context Propagation` to `## Step 4: Trace Context Propagation`. The original Steps 4–7 shift to 5–8. All cross-references in the prose are updated accordingly.
- **Restored Step 5's heading** (`## Step 5: Recording Rules`) — that header was accidentally swallowed in an earlier slim. The Recording Rules content was still there, just unlabeled.
- **De-Python-ified** the propagation section. Python is no longer the lead example. The deterministic per-language remediation (Python / Node.js / Go) now lives in the diagnose check's `Recommendation` field (#727). The skill keeps only the supplementary cases that don't fit there cleanly: bypassed HTTP clients, missing `TracerProvider`, and convergence-time expectations.
- **Removed the "canonical Entity Graph failure mode #3" phrase** — that was leakage from an internal taxonomy as you flagged.

Companion change on the code side: #727's `Recommendation` field now covers Python / Node.js / Go HTTP-client cases explicitly (it already did, the new push doesn't change that). The skill's Step 4 explicitly defers to it as the primary playbook.

claude-plugin/skills/diagnose-entity-graph/SKILL.md
```diff
@@ -78,15 +93,53 @@ gcx metrics query 'count(traces_service_graph_request_total{server_deployment_en
...
+- `traces_target_info` exists but `traces_service_graph_request_total` doesn't
+  (with `client!="user"` filter applied) → either Tempo server-side metrics
+  generation isn't enabled, or outgoing trace context isn't being propagated
+  by your services. Go to Step 4.
+- Service-graph contains only `client="user"` edges → broken trace context
+  propagation. Go to Step 4.
+- Service-graph shows self-loop edges (`client == server`) for a service
+  that shouldn't be calling itself → strong indicator of `service.name`
+  collision across multiple workloads. Inspect the colliding service in
+  Step 8.
```
annanay25: shouldn't this be part of the diagnose command?

danstadler-pdx:
Pushed commit `3bfe1ad` to slim the skill per your guidance.

**Net effect:** PR was +108 / -11 (= +97 lines). Now +29 / -12 (= +17 lines).

What stayed in the skill:
- Two-bullet 'Interpret' rewrite in Step 3 + a 5-line pointer at `gcx kg diagnose` for the specific verdicts (Tempo metrics generation, propagation broken, collision via self-loops).
- A 5-line pointer in Step 7 naming the two patterns (`Service-name collision`, `Env-scope split`) that present as "missing entities", with the fix referenced via the diagnose check's `Recommendation`.
- The convergence-time note in 'Producing a Report' (kept; not something the command can enforce).

What moved into / is already covered by the code:
- The phantom-edge filter explanation → `checkTracePropagation` (PR #727) encodes the filter and emits the recommendation.
- The entire "Step 4: Trace Context Propagation" prose section → `checkTracePropagation`'s `Recommendation` field.
- The collision and env-split per-service workflows → `checkSplitIdentity` in `diagnose_service.go`.

**Deferred to a follow-up PR:** dedicated per-service `Split identity` and `Disconnected clusters` checks. Those would add net LOC to `diagnose.go`, which from #748 you're also pushing back on. Once that constraint is resolved we can replace the remaining skill pointers with command output.

annanay25 (approved):

PR #727, opened 2026-05-16
kg diagnose: detect broken trace context propagation

danstadler-pdx:
## Summary

Adds a new check to `gcx kg diagnose` that detects the "entities
exist but no edges" failure mode in Entity Graph.

Natural follow-up to [PR #677](https://github.com/grafana/gcx/pull/677)
which shipped the diagnose framework — and pairs with PR #726,
which adds the matching agent-facing skill content.

## What's new

A new check function `checkTracePropagation`, registered alongside
the existing per-env metric checks. Triggers when **all three**
PromQL conditions hold:

1. `count(traces_target_info{deployment_environment="ENV"}) > 0` —
   services are emitting telemetry
2. `count(traces_service_graph_request_total{server_deployment_environment="ENV", client!="user"}) == 0` —
   no real inter-service edges
3. `count(traces_service_graph_request_total{server_deployment_environment="ENV", client="user"}) > 0` —
   only phantom edges exist

The `client="user"` series is Tempo's synthetic "external caller"
edge, emitted for any SERVER span that arrives with no incoming
`traceparent`. Its presence alongside the absence of real edges is
the unambiguous fingerprint of broken outgoing trace propagation on
the upstream service — usually `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS`
disabling the HTTP client auto-instrumentation, an HTTP client that
bypasses the auto-instrumented one, or a missing `TracerProvider`
on the SDK side.

The check skips itself (returns nil) in three cases so it doesn't
add noise:
- No `--env` scope (the PromQL is per-env)
- No target info (telemetry isn't flowing at all — earlier checks
  flag that)
- Real edges exist (no problem to report)

## How it presents

When the bug is active:

```
Trace context propagation                   FAIL    services emit telemetry for env "ENV" but only phantom
                                                    client="user" edges exist — no real inter-service edges

Recommendations:
  Trace context propagation: Outgoing HTTP calls aren't carrying traceparent
  headers, so the metrics generator can't link the spans into edges. Check the
  OTel SDK config on your services: is the auto-instrumentation for your HTTP
  client (e.g. urllib, urllib3, requests, net/http) enabled?
  `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` is a common culprit. Once outgoing
  propagation is fixed, expect `traces_service_graph_request_total{client!="user"}`
  to populate within 1–2 minutes and `asserts:relation:calls` to follow ~5–10
  minutes after that.
```

On healthy or unrelated envs, the check is silent.

### Unit tests

Four new tests covering each path of the function
(`go test ./internal/providers/kg/...`):

- `TracePropagationBroken` — full signature triggers FAIL with the
  expected detail and recommendation strings
- `TracePropagationHealthy` — real edges present, check absent
- `TracePropagationNoTelemetry` — target info empty, check absent
- `TracePropagationNoEnvSkipped` — no `--env` scope, check absent

All pass. Full `go test ./...` is green. `gofmt` + `go vet` clean.

## Companion PR

The user-facing skill that documents this failure mode and tells the
agent how to recognize it lives in
[PR #726](https://github.com/grafana/gcx/pull/726). The two PRs are
independent and can land in either order — the skill points at
`gcx kg diagnose` as a shortcut, so the new check folds in naturally
once both ship.

## What's not in this PR

- No changes to the existing checks
- No changes to the diagnose command's flags or invocation
- No new commands

### DIFF
--- a/internal/providers/kg/diagnose.go
+++ b/internal/providers/kg/diagnose.go
@@ -331,6 +331,15 @@ func runDiagnose(ctx context.Context, client *Client, scope *scopeFlags, promCli
 				return nil
 			})
 		}
+		// Check N: Trace context propagation — detect the "entities exist
+		// but no edges" failure mode where outgoing OTel SDKs aren't
+		// injecting traceparent on cross-service calls.
+		g.Go(func() error {
+			if c := checkTracePropagation(ctx, promClient, datasourceUID, scope.env); c != nil {
+				addCheck(*c)
+			}
+			return nil
+		})
 	}
 
 	_ = g.Wait() // errors are captured in CheckResults, not returned
@@ -370,6 +379,9 @@ func checkOrder(name string) int {
 	}
 	if strings.HasPrefix(name, "Edge source:") {
 		return 8
+	}
+	if name == "Trace context propagation" {
+		return 9
 	}
 	return 50
 }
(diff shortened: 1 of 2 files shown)


internal/providers/kg/diagnose.go
```diff
@@ -864,6 +876,83 @@ func extractInstantValue(s prometheus.Sample) string {
...
+//     client="user"} > 0                                    (only phantom edges)
+//
+// Tempo's metrics generator synthesizes client="user" edges for SERVER spans
+// that arrive with no incoming traceparent header — the unambiguous fingerprint
+// of broken outgoing trace context propagation on the upstream service.
+//
+// Returns nil when env is unset (the check is scoped per-env), when target
+// info is empty (telemetry isn't flowing yet — let earlier checks flag that),
+// or when real edges exist (no problem to report).
+func checkTracePropagation(ctx context.Context, client *prometheus.Client, datasourceUID, env string) *CheckResult {
```
annanay25, line 899: if we are adding the code here should it be a much smaller section in the skill? https://github.com/grafana/gcx/pull/726/changes

in the skill we should just be asking to look for this output from the diagnose command

danstadler-pdx: Agreed — this is exactly the boundary feedback I've been working on with the companion PR #726. The skill side is being slimmed in the same direction. I'm also expanding this check's `Recommendation` field so the diagnose output is self-contained (covers Node.js, Go, missing TracerProvider, bypassed-client cases), which lets the skill stay short. Will push the updated recommendation alongside the #726 slim.

danstadler-pdx:
Closing the loop on the threaded reply above — the expanded `Recommendation` is already on this branch at commit `9021cfb`. Phrasing was forward-looking in the reply; the push had already happened.

The recommendation now covers, with equal billing:

- **Python** — `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS`, or the relevant `opentelemetry-instrumentation-*` package not installed
- **Node.js** — missing `@opentelemetry/instrumentation-*` registration
- **Go** — outgoing calls must be wired through `otelhttp.Transport` (no auto-instrumentation)
- **Bypassed-client case** — raw socket calls, third-party SDK with its own transport
- **Missing `TracerProvider`** — verify with `gcx traces query` that received spans have a parent

The companion slim on #726 has now also shipped (commit `7905395`). The skill side defers to this check's `Recommendation` as the primary playbook and keeps only the supplementary cases that don't fit there cleanly.

internal/providers/kg/diagnose.go
annanay25: > The skill side is being slimmed in the same direction

Can you show me where that's happening?

danstadler-pdx: Three concrete pointers:

**1. The renumber + slim of the trace-propagation section on #726**
([commit `7905395`](https://github.com/grafana/gcx/pull/726/commits/7905395)).
That commit:

- Renames `Step 3.5` → `Step 4: Trace Context Propagation`
  (responding to your other thread about the orphaned numbering).
- Cuts the manual `gcx metrics query` recipes that taught the agent
  to be the detection algorithm by hand. The body now says: "when
  `gcx kg diagnose --env <env>` emits a `Trace context propagation:
  FAIL` check, here's what it means; the check's `Recommendation`
  is the playbook."
- Drops the Python-first ordering and gives Python / Node.js / Go
  equal billing in the supplementary notes.
- The summary of the changes is in [this comment on #726](https://github.com/grafana/gcx/pull/726#issuecomment-4492946162).

**2. The recommendation expansion in this PR**
([commit `9021cfb` on this branch](https://github.com/grafana/gcx/pull/727/commits/9021cfb)).
The check's `Recommendation` field is now self-contained — it
covers Python, Node.js, Go, missing `TracerProvider`, and
bypassed-client cases. That lets the skill side stay short
because the agent reads the recommendation directly from the
command output rather than reaching for the skill to fill in the
gaps. I confirmed the push [in this follow-up comment](https://github.com/grafana/gcx/pull/727#issuecomment-4493046603)
after my earlier reply was forward-looking.

**3. PR #748** ([just opened](https://github.com/grafana/gcx/pull/748))
applies the same pattern to four more diagnose checks (DB
instrumentation, Container label drift, Resource coverage, Split
identity). Each one ships a tightly-scoped skill subsection that
points at the check's WARN and defers to the `Recommendation` for
the fix path — same shape as what we did here for trace
propagation.

The pre-existing Steps 3–6 in `diagnose-entity-graph` still have
some PromQL recipes that are now partly redundant with the new
checks. I called that out in #748's description as a follow-up
rather than expanding #748's scope to fix it — that slimming
belongs on #726's branch since you're already reviewing the same
content there. Happy to do it on either branch you'd prefer.

danstadler-pdx:
Pushed commit `54afa61` to fix the failing `unparam` lint on `findCheckByName`. CI is now green (Linters / Tests / Documentation all pass).

The helper had one caller pattern (all 4 sites passed "Trace context propagation"), so I dropped the parameter and inlined the constant inside the helper. Test behavior is unchanged.

Per your stacking guidance, this is ready to merge before #726.

annanay25 (approved):
lgtm, thanks!

Issue #605, opened 2026-04-29
Add first-class datasource support for ClickHouse, CloudWatch, and Elasticsearch


tehranian:
### What problem would this solve?

gcx today ships first-class support for **Prometheus, Loki, Tempo, and Pyroscope**, but has zero footprint for **ClickHouse, CloudWatch, and Elasticsearch** — three datasources that are widely deployed alongside the core four in real-world Grafana stacks.

For our team specifically: our observability platform is ClickHouse-backed. Without ClickHouse support in gcx, we cannot query our primary log/metric store through the CLI — it's an adoption blocker.

More broadly, users already using the sibling [grafana/mcp-grafana](https://github.com/grafana/mcp-grafana) project get all three datasources today. If gcx is intended to be a viable alternative for CLI-oriented workflows, parity on datasource coverage matters.

### Describe your ideal solution

Extend the existing `DatasourceProvider` pattern (`internal/datasources/provider.go`, `internal/datasources/registry.go`, `internal/datasources/providers/*.go`) to include three new providers so commands like these become possible:

- `gcx datasources clickhouse query <uid> "SELECT ..."`
- `gcx datasources cloudwatch query <uid> ...`
- `gcx datasources elasticsearch query <uid> ...`

Plus equivalent auto-detection via `NormalizeKind()` in `internal/datasources/query/resolve.go` so `gcx datasources query <uid> <expr>` routes to the right provider.

Capability shape (informed by what already ships in mcp-grafana — see prior art below):

- **ClickHouse**: list tables, describe table schema, run SELECT, optional log-search helper
- **CloudWatch**: list namespaces, list metrics, list dimensions, run metric queries
- **Elasticsearch**: Lucene/KQL query

Worth noting: all three are reachable through standard Grafana APIs (`/api/datasources/proxy/uid/{uid}` and `/api/ds/query`), so no third-party SDK dependencies are strictly required on the client side — mcp-grafana proves this pattern works.

### Prior art in mcp-grafana

All three datasources were added to mcp-grafana by @ataran-dev. These PRs are useful references for API surface and tool shape:

| Datasource | mcp-grafana PR | Tools / capabilities |
|---|---|---|
| **ClickHouse** | [grafana/mcp-grafana#535](https://github.com/grafana/mcp-grafana/pull/535) | `list_clickhouse_tables`, `describe_clickhouse_table`, `query_clickhouse`, `search_logs` |
[...]

PR #728, opened 2026-05-16
feat(clickhouse): add ClickHouse datasource provider support

imankurpatel000:
## Summary

Adds a ClickHouse DatasourceProvider with three subcommands and auto-detect integration, enabling agents and developers to query ClickHouse datasources through Grafana's query API. Feature set and implementation approach inspired by [mcp-grafana's ClickHouse tools](https://github.com/grafana/mcp-grafana/blob/main/tools/clickhouse.go) — same LIMIT defaults (100/1000), identifier validation, macro support, and query API integration.

- `gcx datasources clickhouse query` — execute arbitrary SQL with LIMIT enforcement, `$__timeFilter` macro support, and time-range binding
- `gcx datasources clickhouse list-tables` — discover tables across databases (filterable with `--database`)
- `gcx datasources clickhouse describe-table` — inspect column schema for a specific table
- Auto-detect via `gcx datasources query <uid> <sql>` routes ClickHouse datasources automatically

## Architecture

- **HTTP client** (`internal/query/clickhouse/client.go`): Uses K8s query API (`/apis/query.grafana.app/v0alpha1/namespaces/{ns}/query`) with inline fallback to legacy `/api/ds/query` — same iterative pattern as Prometheus and Loki clients.
- **Provider registration** (`internal/datasources/providers/clickhouse.go`): Standard DatasourceProvider interface with `init()` self-registration.
- **Type safety**: SQL identifier validation via regex allowlist prevents injection in `list-tables` and `describe-table`. Raw SQL in `query` is passed through (user responsibility).
- **LIMIT enforcement**: Default 100, max 1000, with bail-out for complex SQL patterns (LIMIT BY, LIMIT OFFSET, FORMAT, SETTINGS). `--limit 0` disables.
- **Output consistency**: Timestamps formatted as RFC3339 (matching Prometheus/Loki/Pyroscope precedent). JSON output uses lowercase struct tags (`columns`, `rows`, `name`, `type`).

## Test plan

- [x] Unit tests for types, client response parsing, formatters, and LIMIT enforcement
- [x] All tests pass with race detection (`go test -race ./internal/query/clickhouse/...`)
- [x] Verified against real ClickHouse datasource (`email-clickhouse-otel-logs-dev-us-east-1`)
- [x] Tested: query, list-tables, describe-table, auto-detect path, error handling, injection prevention, macro expansion
- [x] `mise run all` passes (lint + tests + build + docs)

Part of #605


javiermolinar:
@imankurpatel000 do you mind to fix the conflicts? thank you!

imankurpatel000:
@javiermolinar conflict resolved.

internal/query/clickhouse/client.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+
+	resp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute query: %w", err)
+	}
+	defer resp.Body.Close()
```
dafydd-t, line 85: I think we need this `defer` before any of the branches that return early, otherwise we won't close it

```suggestion
	resp, err := c.httpClient.Do(httpReq)
	defer resp.Body.Close()
	if err != nil {
		return nil, fmt.Errorf("failed to execute query: %w", err)
	}
```

internal/query/clickhouse/client.go
```diff
@@ -0,0 +1,179 @@
...
+		httpReq, err = http.NewRequestWithContext(ctx, http.MethodPost, c.restConfig.Host+apiPath, bytes.NewBuffer(body))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create request: %w", err)
+		}
+		httpReq.Header.Set("Content-Type", "application/json")
+		resp, err = c.httpClient.Do(httpReq)
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute query: %w", err)
+		}
+		defer resp.Body.Close()
```
dafydd-t, line 105: same issue with the defer here

internal/query/clickhouse/client.go
```diff
@@ -0,0 +1,179 @@
...
+	resp, err := c.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute query: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response: %w", err)
+	}
```
dafydd-t, line 90: should we move all this to after the 404 check? we could probably deduplicate this respBody call then

internal/query/clickhouse/client.go
```diff
@@ -0,0 +1,179 @@
...
+	resp := &QueryResponse{}
+	for i, frame := range result.Frames {
+		if i == 0 {
+			for _, f := range frame.Schema.Fields {
+				resp.Columns = append(resp.Columns, Column(f))
+			}
+		} else {
+			if len(frame.Schema.Fields) != len(resp.Columns) {
+				continue
+			}
```
dafydd-t: we're silently dropping rows with columns that are different to the first one - is that what we want to do?

dafydd-t: we might want to consider collecting all seen fields first, [like in influx](https://github.com/grafana/gcx/blob/main/internal/query/influxdb/client.go#L435)

internal/query/clickhouse/client.go
imankurpatel000: The current ordering is intentional, when `Do()` returns an error, `resp` is `nil`, so `defer resp.Body.Close()` before the check will panic on connection errors.

From the [Go docs](https://pkg.go.dev/net/http#Client.Do): 

> On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when CheckRedirect fails, and even then the returned [Response.Body] is already closed.

internal/query/clickhouse/client.go
imankurpatel000: same as above.

internal/query/clickhouse/client.go
imankurpatel000: Good call! You are right the ReadAll can move below the 404 check to deduplicate.

Though, this pattern is shared across all the query clients (Prometheus, Loki, InfluxDB, ClickHouse) so I was thinking of extracting a common helper that handles the k8s-api-then-legacy-fallback flow in one place. That would fix this and also give a single spot to wire in [response-too-large detection](https://github.com/grafana/gcx/pull/733#discussion_r3270870432) which I was planning in a follow-up PR.

That said, if you would rather avoid the coupling and keep this logic in each datasource, happy to do the dedup here instead.

internal/query/clickhouse/client.go
imankurpatel000: Good question. I dug into the upstream to understand what ClickHouse actually returns:

The [clickhouse-datasource plugin](https://github.com/grafana/clickhouse-datasource/blob/c56e3af64308ffc4cf304c8f4dd9ce545d6b2063/pkg/plugin/driver.go#L448-L453) uses `sqlds` with `FormatOptionTable` (format=1), which produces **one frame per SQL result set**. A single SELECT always returns exactly one frame. Multiple frames with different schemas shouldn't happen in practice for this datasource type, unlike InfluxDB/Flux where each series is its own frame with different label columns.

For reference, [mcp-grafana's clickhouse.go](https://github.com/grafana/mcp-grafana/blob/9517bc0fc86b4d4f8a5978f28498bf376dc51d65/tools/clickhouse.go#L326-L349) handles this by iterating all frames and blindly overwriting `result.Columns` on each iteration, it doesn't handle schema mismatches at all. Our code is slightly more defensive in that we at least skip mismatched frames rather than silently corrupting.

Options:
1. **Keep as-is**: the skip is a defensive guard for a case that shouldn't fire with this datasource type
2. **Only use the first frame**: simplest, matches reality, makes the "single frame" assumption explicit
3. **Log a warning on mismatch**: keeps the defensive behavior but makes it observable
4. **Adopt the InfluxDB union approach**: overkill here since ClickHouse doesn't have heterogeneous frame semantics

I'm leaning towards option 2 since it makes the assumption explicit and removes the dead code path. Thoughts?

internal/query/clickhouse/client.go
dafydd-t: doh, thanks for the pointer!

internal/query/clickhouse/client.go
dafydd-t: common handler sounds good, but we can do it separately. With the knowledge that that change is coming, I don't mind whether we dedupe here now, or leave it until the later PR. your call 👍

internal/query/clickhouse/client.go
dafydd-t: yeah I agree with option 2. please leave a comment with those references, it will help explain your decision. Thank you!

imankurpatel000: Done

internal/query/clickhouse/client.go
imankurpatel000: I will do this in the follow up PR.

cmd/gcx/datasources/query.go
```diff
@@ -117,14 +114,7 @@ that do not have a dedicated subcommand.`,
...
 
-				switch shared.IO.OutputFormat {
-				case "table":
-					return loki.FormatQueryTable(cmd.OutOrStdout(), resp)
-				case "wide":
-					return loki.FormatQueryTableWide(cmd.OutOrStdout(), resp)
-				default:
-					return shared.IO.Encode(cmd.OutOrStdout(), resp)
-				}
+				return shared.IO.Encode(cmd.OutOrStdout(), resp)
```
imankurpatel000, line 129: Didn't really want to make this change in this PR, but the lint failures pushed me. The original
dispatcher was flagged by the `maintidx` linter:

```
cmd/gcx/datasources/query.go:19:1: Function name: QueryCmd,
  Cyclomatic Complexity: 29, Halstead Volume: 4769.86, Maintainability Index: 19 (maintidx)
```

I chose the least disruptive change here to keep the linter happy and unblock this PR. Though, I would prefer to refactor this function in a separate PR because the lint failure will happen again as more cases are added for new datasources. 

## Why I removed these OutputFormat cases
These branches are redundant. They produced the same result as just calling `shared.IO.Encode` directly. 

`shared.Setup(cmd.Flags(), true)` calls `RegisterCodecs(&opts.IO, true)`
(in `internal/datasources/query/opts.go`), which registers `queryTableCodec` and
`queryWideCodec` on `shared.IO`. When `shared.IO.Encode` is called, it:

1. Looks up the codec registered for the active `OutputFormat`
2. Calls that codec's `Encode` method
3. The codec type-switches on the response and calls the per-type formatter

imankurpatel000: I tested this for Prometheus, Loki, Pyroscope and ClickHouse and all are working fine.

imankurpatel000:
@javiermolinar @dafydd-t 
This is ready for one more round of review whenever you have time. Thanks!

cmd/gcx/datasources/query.go
dafydd-t: Thanks for being thorough and testing those!

AGENTS.md
```diff
@@ -123,14 +123,16 @@ internal/
...
 │   ├── cloudwatch/  CloudWatch CLI commands (query, list-namespaces, list-metrics, list-dimensions, list-regions, list-accounts)
 │   ├── influxdb/  InfluxDB datasource command layer (query, field-keys, measurements)
 │   └── query/   Shared query CLI utils (time parsing, codecs, opts, resolve helpers — used by signal providers and GenericCmd)
 ├── query/       Datasource query clients
 │   ├── cloudwatch/  CloudWatch HTTP query client (metric queries, resource listing)
 │   ├── prometheus/  Prometheus HTTP query client
 │   ├── influxdb/    InfluxDB HTTP query client
-│   └── loki/        Loki HTTP query client
+│   ├── loki/        Loki HTTP query client
+│   └── clickhouse/  ClickHouse HTTP query client (SQL queries via Grafana datasource API)
```
dafydd-t:
```suggestion
│   └── clickhouse/  ClickHouse HTTP query client
```

docs/architecture/project-structure.md
```diff
@@ -82,12 +82,14 @@ gcx/
...
 │   ├── datasources/          # Datasource HTTP client (legacy REST API)
+│   │   ├── clickhouse/       # ClickHouse datasource commands (query, list-tables, describe-table, explore)
 │   │   ├── cloudwatch/       # CloudWatch CLI commands (query, list-namespaces/metrics/dimensions/regions/accounts)
 │   │   └── query/            # Shared query CLI utils (time parsing, codecs, opts, resolve helpers)
 │   ├── query/                # Datasource query clients
 │   │   ├── cloudwatch/       # CloudWatch HTTP client (metric queries, resource listing)
 │   │   ├── prometheus/       # Prometheus HTTP client (instant + range queries)
-│   │   └── loki/             # Loki HTTP client (log + metric queries)
+│   │   ├── loki/             # Loki HTTP client (log + metric queries)
+│   │   └── clickhouse/       # ClickHouse HTTP client (SQL queries via Grafana datasource API)
```
dafydd-t:
```suggestion
│   │   └── clickhouse/       # ClickHouse HTTP client
```

dafydd-t:
@imankurpatel000 thanks a lot - two tiny nits and then we're good to go 🚀

imankurpatel000:
@dafydd-t Accepted the suggestions. Thanks a lot! 🚀

dafydd-t (approved):
Thanks a lot @imankurpatel000  🙌

PR #733, opened 2026-05-18
feat(cloudwatch): add CloudWatch datasource provider support

imankurpatel000:
## Summary

Adds a CloudWatch DatasourceProvider with six subcommands, an auto-detect short-circuit for the generic query command, and graph rendering for time-series results. Closes part of #605.

- `gcx datasources cloudwatch query` — structured CloudWatch metric query (`--namespace`, `--metric`, `--region`, `--statistic`, `--period`, `--dimensions`, `--account-id`)
- `gcx datasources cloudwatch list-namespaces` — namespaces in a region
- `gcx datasources cloudwatch list-metrics` — metrics in a namespace
- `gcx datasources cloudwatch list-dimensions` — dimension keys for a metric
- `gcx datasources cloudwatch list-regions` — AWS regions visible to the datasource
- `gcx datasources cloudwatch list-accounts` — linked accounts on cross-account monitoring datasources

The generic `gcx datasources query <cw-uid>` returns a clear "use the structured subcommand" error instead of "expression is required" — CloudWatch queries are tuples, not expressions. **No behavior change for prom/loki/pyroscope/influxdb users** (regression-pinned by `TestGenericQueryPrometheusStillRequiresExpr`).

The wire payload to Grafana's CloudWatch plugin uses three plugin-specific shapes: `period` as a JSON string (the plugin returns 500 on numeric), `intervalMs` as a JSON number, and `dimensions` as `map[string][]string` (the multi-valued shape Grafana's Explore frontend serializes). All three are pinned by `TestClient_Query/request_body_shape`.

### Sample output

```
$ gcx datasources cloudwatch query -d <cw-uid> --region us-east-1 \
    --namespace AWS/EC2 --metric CPUUtilization --account-id all --since 30m
TIMESTAMP             VALUE   SERIES
2026-05-18T11:54:00Z  6.534   eks-default20240515-...
2026-05-18T11:59:00Z  6.896   eks-default20240515-...
2026-05-18T12:04:00Z  6.485   eks-default20240515-...
...
```

`-o graph` renders the same data as a terminal line chart; `-o json/yaml/agents` return structured frames with lowerCamelCase keys (`frames`, `timestamps`, `values`).

### 1. Flags-only query shape (no positional `NAMESPACE METRIC`)

`CONSTITUTION.md` says: *"Positional arguments are the subject, flags are modifiers."* CloudWatch's query subject is four fields (`namespace`, `metric`, `region`, `dimensions`) — no single field identifies the series. Three options:

| Option | Shape | Verdict |
|---|---|---|
| **(A) Chosen** | `--namespace AWS/EC2 --metric CPUUtilization --region us-east-1` | Self-documenting; matches mcp-grafana's tool schema; matches CloudWatch's data model. Sibling precedent: `gcx datasources pyroscope series` is also flag-only with a similar field count. |
| (B) | `query NAMESPACE METRIC --region us-east-1 --dimensions ...` | Splits the four fields across positional and flag. Doesn't actually resolve the constitutional concern, just shifts two fields. |
| (C) | `query 'AWS/EC2:CPUUtilization{InstanceId=i-abc}@us-east-1'` | Invents a CloudWatch DSL that doesn't exist anywhere in the ecosystem. |

Open to alternatives — happy to switch if maintainers see a fourth shape or prefer (B)/(C).

### 2. `list-*` naming

Sibling precedent for "list this resource type" is noun-only (`prometheus labels`, `loki series`, `tempo labels`, `pyroscope profile-types`). We chose verb-prefixed (`list-namespaces`, `list-metrics`, …) because every CloudWatch resource lookup needs scope flags (`list-metrics` requires `--region --namespace`, `list-dimensions` adds `--metric`, `list-accounts` requires `--region`). Closer precedent for scoped lookups: `gcx irm oncall alert-groups list-alerts <id>` (verb-prefixed).

`list-regions` is the one unscoped command and would more naturally be `regions`; kept verb-prefixed for symmetry with its four scoped peers in the same subtree.

Happy to flip to noun-only if maintainers prefer the prom/loki precedent over the irm/oncall one.

### 3. `matchExact` derived from `--dimensions` (not exposed as a flag)

`matchExact` controls whether the CloudWatch plugin matches the user's exact metric+dimension tuple (`true`) or runs `ListMetrics` discovery and aggregates (`false`). Three consumers, three different calls:

| Consumer | Behavior |
|---|---|
| Grafana Explore UI | Exposes a "Match exact" toggle (defaults `true`) |
| mcp-grafana | Hard-codes `true` |
| **gcx (chosen)** | Derives `matchExact = len(opts.Dimensions) > 0` |

The user's `--dimensions` flag is the same signal: dimensions set → user named a specific series → exact match. Empty → user wants the metric "across whatever exists" → discovery. The other two combinations are footguns:

| `--dimensions` | `matchExact` | Result |
|---|---|---|
| empty | `true` | `"values":[[],[]]` (empty — what mcp-grafana ships) |
| empty | `false` | aggregate across all matching series ✓ |
| set | `true` | exact match ✓ |
| set | `false` | broader match than the user asked for |

Open to exposing `--match-exact` to mirror Grafana's UI if maintainers prefer.

### 4. Auto-detect RunE reorder (`cmd/gcx/datasources/query.go`)

The generic command previously ran `shared.ResolveExpr` before `GetDatasourceType`, so `gcx datasources query <cw-uid>` errored "expression is required" instead of redirecting to the structured subcommand. The new order runs flag validation first, then the type lookup, then the cloudwatch short-circuit, then `ResolveExpr` for everything else. Existing prom/loki/pyroscope/influxdb behavior is preserved (covered by tests).

## Test plan

**Query happy path**
- [x] `query --since 1h -o json` returns data; all five base statistics (`Average`, `Sum`, `Maximum`, `Minimum`, `SampleCount`) succeed; percentiles (`p95`/`p99`/`tm95`) reach the backend (no client-side enum check)
- [x] `matchExact=true` with `--dimensions` set; `false` without
- [x] Cross-account: both `--account-id <id>` and `--account-id all` return data; the cross-account hint fires on empty-result + dimensions + no `--account-id`, silent in the three negative cases (account-id supplied, data present, no dimensions)
- [x] `--share-link` panes carry the full structured query (`namespace`, `metricName`, `dimensions`, `matchExact`, `region`, `statistic`, `period`)
- [x] `-o table/wide/json/yaml/graph/agents` all work for `query`; `-o graph` for the five list-* commands fails cleanly with "format graph not registered"

**Resource-listing commands (live against a real cloudwatch datasource)**
- [x] `list-namespaces`, `list-metrics`, `list-dimensions`, `list-regions`, `list-accounts` all return real data
- [x] Missing required flags surface "X is required" before any HTTP (matrix in `internal/datasources/cloudwatch/list_test.go`)
- [x] `list-accounts` returns linked accounts on a monitoring datasource; **404 path on a non-monitoring datasource is covered by `TestClient_ListAccounts_404_CleanError` only** (no non-monitoring CloudWatch datasource was available on the verification stack)

**Generic command (`gcx datasources query <uid>`)**
- [x] CloudWatch UID returns the structured-subcommand redirect, not "expression is required" (`TestGenericQueryCloudWatchShortCircuit`)
- [x] Flag-validation errors surface before the cloudwatch redirect (`TestGenericQueryFlagValidationPrecedence`)

**Contract pinning (unit tests)**
- [x] Wire payload: `period` JSON string, `intervalMs` JSON number, `dimensions` array-valued (`TestClient_Query/request_body_shape`)
- [x] Output JSON keys are lowerCamelCase (`frames`, `timestamps`, `values`, `id`, `arn`)
- [x] Errors surface as `queryerror.APIError` with `datasource=cloudwatch` and `operation=query`

**Quality gates**
- [x] `mise run all` and `mise run reference-drift` pass (`GCX_AGENT_MODE=false`)

Part of #605


internal/datasources/cloudwatch/explore.go
```diff
@@ -0,0 +1,49 @@
...
+		"queryType":  "timeSeriesQuery",
+		"namespace":  req.Namespace,
+		"metricName": req.MetricName,
+		"region":     req.Region,
+		"statistic":  req.Statistic,
+		"matchExact": req.MatchExact,
+		"dimensions": dimensions,
+	}
+	if req.Period > 0 {
+		query["period"] = req.Period
```
javiermolinar, line 37: This should be a string

`strconv.Itoa(req.Period)`

internal/query/cloudwatch/client.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call %s: %w", path, err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxResourceResponseBytes))
```
javiermolinar, line 202: We should handle the “response too large” error gracefully

imankurpatel000: I was thinking to handle this in a separate PR like https://github.com/grafana/mcp-grafana/pull/884 because the same issue exists in multiple existing datasources as well. But let me know if you still prefer to handle it here.

internal/datasources/cloudwatch/explore.go
imankurpatel000: This is fixed.

imankurpatel000:
@javiermolinar fixed the conflict.

javiermolinar:
@codex review

internal/query/cloudwatch/client.go
javiermolinar: I'm ok with a follow up pr

imankurpatel000:
@javiermolinar Looks like it didn't really trigger any codex review.

internal/agent/command_annotations.go
```diff
@@ -60,6 +60,14 @@ var commandAnnotations = map[string]annotation{
 	"gcx datasources list":  {Cost: "small"},
 	"gcx datasources query": {Cost: "large", Hint: "Run gcx help-tree metrics (or logs, traces, profiles) to discover signal commands. Prefer gcx metrics query for PromQL, gcx logs query for LogQL, gcx traces query for TraceQL, gcx profiles query for profiling. Example: <datasource-uid> 'up' --since 1h -o json"},
 
+	// datasources cloudwatch
+	"gcx datasources cloudwatch query":           {Cost: "medium", Hint: "gcx datasources cloudwatch query -d UID --region us-east-1 --namespace AWS/EC2 --metric CPUUtilization --since 1h -o json"},
```
javiermolinar: In my tests this is "large" due to the potentially unbound query

internal/datasources/cloudwatch/query.go
```diff
@@ -0,0 +1,228 @@
...
+	Dimensions map[string]string
+	AccountID  string
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	dsquery.RegisterCodecs(&opts.IO, true)
+	opts.IO.BindFlags(flags)
+	opts.SetupTimeFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.cloudwatch is configured)")
```
javiermolinar, line 42: There is no validation if the datasource is in fact a cloudwatch one. When using a different type ie prometheus you get a 404 instead of a proper error

internal/query/cloudwatch/types.go
```diff
@@ -0,0 +1,313 @@
...
+			status = 400
+		}
+		return nil, queryerror.New("cloudwatch", "query", status, result.Error, result.ErrorSource)
+	}
+
+	resp := &QueryResponse{
+		Frames: make([]Frame, 0, len(result.Frames)),
+	}
+
+	for _, df := range result.Frames {
```
javiermolinar, line 120: When the plugin returns a single placeholder frame with no rows, gcx surfaces it as a frame named "1 -" with labels.Series = "1 -". We should skip frames if there are no results

javiermolinar (dismissed):
I've been testing it with our cloudwatch datasources and it works very well just a few things we need to polish. We should add a --limit command because big datasources display huge tables when query

radiohead:
@claude review

imankurpatel000:
@javiermolinar I am skipping `--limit` and going with `--period auto` as the default instead.

## Why no `--limit`

1. **The CloudWatch plugin has no result-count limiter on the metric path.** [`get_metric_data_executor.go`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/pkg/cloudwatch/get_metric_data_executor.go) and [`metric_data_input_builder.go`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/pkg/cloudwatch/metric_data_input_builder.go) never read `MaxDataPoints`; only the SQL path does ([`sql.go#L164`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/pkg/cloudwatch/sql.go#L164)). A `--limit` would be a client-side trim after the bytes are already on the wire.
2. **`mcp-grafana` doesn't expose one either** — just a 10 MB hard-error cap ([`response_utils.go#L8`](https://github.com/grafana/mcp-grafana/blob/4316c7c6be623b2130d27258c01a7eda940fe454/tools/response_utils.go#L8)).
3. **Other gcx datasources have `--limit` only because their backends honor it** (Loki, Tempo). Prom/InfluxDB skip it because their query languages express it. CloudWatch is neither.

## Why `--period auto` is the better fix

`auto` targets ~2000 datapoints per series via [`calculatePeriodBasedOnTimeRange`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/pkg/cloudwatch/models/cloudwatch_query.go#L446-L459) (`datapoints := int(math.Ceil(deltaInSeconds / 2000))`), walking the CloudWatch retention ladder.

Tested on our internal CloudWatch monitoring datasource:

| Window | Fixed period | `--period auto` | Δ |
|---|---|---|---|
| 24h | 1.4 MB / 24,970 pts (`--period 300`) | 267 KB / 4,470 pts | **5.4× smaller, 5.6× fewer pts** |
| 30d | 0 series (`--period 60`, outside 15d retention) | 515 KB / 8,727 pts | fixed period returned no data; `auto` returns results |
| 90d | 0 series (`--period 300`, outside 63d retention) | 103 KB / 1,740 pts | fixed period returned no data; `auto` returns results |

Total wire size scales with series count (cardinality), not time-range × resolution.

## Parity with the plugin

Plugin's frontend default is [`period: ''`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/src/defaultQueries.ts#L22), which [`getPeriod`](https://github.com/grafana/grafana-cloudwatch-datasource/blob/f8ca1e6f2dcae6869b85ac485690c473a82a5062/pkg/cloudwatch/models/cloudwatch_query.go#L461-L481) normalises to `auto`. gcx's `--period auto` matches that server-side behavior. Validation mirrors the plugin.

internal/agent/command_annotations.go
imankurpatel000: Done.

internal/datasources/cloudwatch/query.go
imankurpatel000: Added validation.

internal/query/cloudwatch/types.go
imankurpatel000: Fixed.

imankurpatel000:
# Reply to claude additional findings

4. **`list-*` type validation gap.** Fixed. All five list-* commands now use `ResolveValidateAndSaveDatasource("cloudwatch")`.
5. **`io.LimitReader` truncates silently.** As discussed previously, will file a separate follow-up PR for this.
6. **`orEmptyDimensions` is dead code.** Not dead after the latest changes, the CLI layer now leaves `dimensions` nil when empty, so `orEmptyDimensions` is what enforces the `dimensions: {}` wire shape. Redundant pre-allocation removed.
7. **`Statistic` enum check.** Skipping. Plugin accepts arbitrary strings and surfaces its own error; client-side list would drift.
8. **`canonicalCloudDatasourceSuffix` for cloudwatch.** Skipping. CloudWatch on Grafana Cloud isn't auto-provisioned with a `grafanacloud-<stack>-cloudwatch` convention; users bring their own AWS creds. `-d` or `datasources.cloudwatch` is the right path.
9. **Cross-account hint copy.** Fixed. Hint now reads: "Hint: no data returned. If this is a cross-account monitoring datasource and you intended to query a linked account, try --account-id all (or pass a specific ID; use `list-accounts` to discover them)."

javiermolinar (approved):
Thank you for your contribution!

PR #734, opened 2026-05-18
fix: exit cleanly on Ctrl+C instead of showing noisy error

dafydd-t:
## Problem

Ctrl+C during a running command shows a full error box with suggestions instead of exiting silently:

```
^CError: Authentication failed
│
├─ Details:
│
│ context canceled
│
├─ Suggestions:
│
│ • Check that the GCOM API URL is correct
│ • Ensure you are logged in to Grafana Cloud in your browser
│ • Try again with: gcx cloud login --api-url <url>
│
└─
```

## Fix

`handleError` in `main.go` already has a fast-path for `context.Canceled` that exits cleanly with code 5 and no output. But `DetailedError` had no `Unwrap()` method, so `errors.Is(err, context.Canceled)` couldn't see through commands that wrapped the cancellation in a `DetailedError` (via the `Parent` field). The fast-path was skipped and the full error renderer kicked in.

Adding `Unwrap()` to `DetailedError` lets `errors.Is` traverse through the wrapper and hit the fast-path.

Now, with Ctrl+C while waiting for the oauth response, we don't see anything.

### DIFF
--- a/internal/gcxerrors/detailed.go
+++ b/internal/gcxerrors/detailed.go
@@ -55,8 +55,8 @@ type DetailedError struct {
 	ExitCode *int
 }
 
-// Unwrap returns the parent error, enabling errors.Is/As to traverse through
-// DetailedError wrappers (e.g. detecting context.Canceled for clean SIGINT exit).
+// Unwrap lets errors.Is/As traverse through DetailedError to detect wrapped
+// sentinels (e.g. context.Canceled for clean SIGINT exit).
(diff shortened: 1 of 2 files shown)


radiohead (dismissed):

radiohead:
@dafydd-t could you resolve conflicts? Otherwise looks good, approved!

javiermolinar (approved):
LGMT

PR #738, opened 2026-05-18
fix(kg): correct insights chart/sources request and response schemas

xujiaxj:
## Summary

The `gcx kg insights` subcommands previously had two structural bugs that made them effectively unusable:

1. **Wrong request body** — sent `{assertionId, startTime, endTime}` but the server matches on `{startTime, endTime, labels}`. Empty `labels` produced empty results.
2. **Wrong response schema** — decoded `labels` as `map[string]string`, but the server actually returns `[]{label, op, value}` (an array of PromQL matchers). When the request was made right, response decode then failed with a vague `kg: assertion source metrics` error.

Both fixed. While here, the UX is aligned and the subcommands renamed for clarity:

| Before | After | Returns |
|---|---|---|
| `entity-metric` | `chart` | Time series + thresholds (renderable chart payload) |
| `source-metrics` | `sources` | Underlying PromQL metric names + label matchers |

Both subcommands now accept the same flag set:

- Positional `Type--Name` or `--type`/`--name`
- `--insight <name>` (sets the `alertname` label)
- `--label k=v` (repeatable; for assertion-specific labels copied from `kg entities inspect` → `timeLines[].labels`)
- `--env`/`--namespace`/`--site` (validated against `/entity_scope`)
- `--from`/`--to`/`--since`
- `-f/--file` for YAML body

Example usage:

```
gcx kg insights chart Service--model-builder \
  --insight ResourceRateAnomaly \
  --label asserts_resource_type=jvm:live_threads \
  --env ops-eu-south-0 --namespace asserts --since 1h

gcx kg insights sources Service--model-builder \
  --insight LatencyP99ErrorBuildup \
  --label asserts_request_context='...' \
  --label asserts_request_type=method \
  --env ops-eu-south-0 --since 1h
```

## Test plan

- [x] `go test ./internal/providers/kg/... ./internal/agent/...` passes
- [x] `GCX_AGENT_MODE=false mise run reference` produces clean diff (old `entity-metric.md`/`source-metrics.md` deleted; new `chart.md`/`sources.md` added; index updated)
- [x] `GCX_AGENT_MODE=false mise run docs` builds
- [x] End-to-end verified against ops Asserts for two insight classes (`LatencyP99ErrorBuildup` narrows correctly; `ResourceRateAnomaly` returns coarse server response — server-side quirk, noted in comments)
- [x] Reviewer to confirm naming (`chart`/`sources`) reads well alongside the rest of `gcx kg insights`

Notes:
- `mise run lint` is blocked locally by a pre-existing `vendor/modules.txt` inconsistency unrelated to this change; CI will run it.
- A handful of unrelated tests in `aio11y/eval`, `dashboards`, and `stacks` fail on main due to a stale `"use --force"` error-message assertion. Not introduced here.


xujiaxj:
TBH I am not sure these two commands are useful on their own, but I figure there is a use case here, so decided to keep them for now, but fix all the bugs and make the names and flags more consistent.

The names (`chart` and `sources`) are suggested by Claude, and I found its arguments quite convincing.

```
Honest reassessment:

  chart — still excellent. The response is literally a renderable chart payload: {timeWindow, timeStepIntervalMs, thresholds[], metrics[].values}. Each metrics entry is one
  series of {time, value} points. There's no better one-word name for that shape.

  source — fits less well than I thought. The response is a list of PromQL selector skeletons (metricName + array of {label, op, value} matchers). Each entry is structurally a
  partial query, and the list is usually >1 element. Singular source makes it sound like a single thing is returned.
```

The use case is to answer follow-up questions after a user fetch health data for a certain entity. In that case, the context already has all the necessary labels for the insights. For example, a user may first use `entities inspect` command to check the health of a service:

Monosnap Image 2026-05-18 16-32-43

Then the user can ask Claude to explain the insights:
Monosnap Image 2026-05-18 16-34-52

radiohead (approved):

PR #746, opened 2026-05-19
fix(kg): correctness fixes for kg diagnose and related commands

danstadler-pdx:
## Summary

Three correctness fixes to `gcx kg diagnose` and the
`diagnose-entity-graph` skill. Each commit is independently
reviewable.

### `kg diagnose service` no longer says "looks healthy" when checks fail

The verdict was hard-coded to the presence of any edge, producing
internally contradictory output:

```
4/8 checks passed, 4 failed
Diagnosis:
  Service "db" looks healthy — found in graph with 3 edge(s).
```

The verdict is now derived from `Summary.Failed`. When failures
exist, the diagnosis line names the failing checks directly so an
agent or human can act without scrolling back to the table.

### Top-of-skill guardrail for reading diagnose verdicts

Adds a "Reading Diagnose Output" subsection to
`diagnose-entity-graph` that explains the two non-obvious check
classifications the diagnose command can emit (the new WARN
introduced by the next commit, plus the existing FAIL semantics).
Tells the agent to treat command verdicts as authoritative for the
queries the command ran, and to extend the same scope-skepticism
to ad-hoc PromQL queries it authors itself.

### `kg diagnose` re-probes without env filter on `FAIL`

When a scoped metric check returns no data and a scope is active,
the diagnose pipeline now re-probes the same metric without the
env/namespace filter. If the unscoped probe finds data, the result
is reclassified from `FAIL — no data` to
`WARN — metric exists (N series unscoped) but no series match the
requested env/namespace scope` with a label-pipeline hint pointing
at `gcx metrics labels --label deployment_environment` /
`--label asserts_env`.

This catches a common false-negative case where the Asserts label
pipeline derives `asserts_env` from a different source label than
the raw Tempo metric carries via `deployment_environment`. A
scoped query returns nothing even though the data is flowing —
the new WARN shape tells the user that's what happened.

Single extra PromQL query per failing scoped check.

## What's not in this PR

- A fourth commit had made `--type` optional on `gcx kg entities
  list` for cross-type entity search. After review, that change
  is being dropped: `--limit` is client-side truncation only and
  fan-out pagination doesn't compose, so the change widened cost
  without bounding it. The use case (find an entity by name when
  the type is uncertain) is covered better by
  `gcx kg entities query "MATCH (n) WHERE n.name CONTAINS '...'
  RETURN n LIMIT N"`, which enforces the limit at the graph layer.
  Agent-facing guidance for that pattern is going into PR #748.

## Validation

- `go test ./...` — clean across the whole repo
- `gofmt -d ./internal/providers/kg/...` — clean
- `go run ./scripts/validate-skills` — `validated claude-plugin/skills`

New tests:
- `TestInterpretServiceResults_AllChecksPassSaysHealthy`
- `TestInterpretServiceResults_FailedChecksSuppressesHealthyVerdict`
- `TestInterpretServiceResults_SingleCheckFailureIsReported`
- `TestCheckMetric_ReclassifiesNoDataAsLabelMismatchWhenUnscopedHasData`
- `TestCheckMetric_NoFallbackWhenScopeUnset`


internal/providers/kg/commands.go
```diff
@@ -870,12 +870,13 @@ func newEntitiesCommand(loader RESTConfigLoader) *cobra.Command {
 	listOpts := &entitiesListOpts{}
 	listCmd := &cobra.Command{
 		Use:   "list",
-		Short: "List Knowledge Graph entities for a given type, env, site, namespace.",
+		Short: "List Knowledge Graph entities, optionally filtered by type, env, site, namespace.",
```
xujiaxj: Can you explain a little more on the use case of using `list` to search across types? I was worried about the size of return so had made `--type` required.

danstadler-pdx: The use case is the "I see entity X in the graph but don't know what it is" question — an agent (or human) wants to look up an entity by name without first knowing whether it's a Service, Pod, Database, Namespace, etc. Today that's a fan-out over every type from `gcx kg meta schema`; making `--type` optional collapses it to one call.

On the size concern: `--limit` (default 50, configurable) still applies to the cross-type response the same way it does for single-type. The backend `search` endpoint per type already has its own paging, and `entities list` truncates client-side after merging. So the worst case isn't unbounded — it's the same `--limit` cap as today, just sourced across types.

If the size risk is the blocker, two alternatives: (a) keep `--type` required but add a new `gcx kg entities search <name>` subcommand that fans out internally, or (b) keep this change but lower the default `--limit` when `--type` is omitted (e.g. 20). Which would you prefer?

xujiaxj: `--limit` is a client-side truncation after all server calls, which means the search actually happens without limit for all the types in the backend one-by-one, so it can be quite expensive.

I am also worried about the pagination mechanism. The pagination mechanism is per API call, so it's per type. When we search across types, the whole pagination mechanism does not make sense any more.

For your use case, can I suggest the cypher query command?

`gcx kg entities query "MATCH (n) WHERE n.name='foo'"" RETURN n LIMIT 10"

This way the limit is enforced on the graph layer.

danstadler-pdx: Both points are right — `--limit` is client-side truncation after
all backend calls complete, and per-type pagination doesn't compose
sensibly across a fan-out. Thanks for the careful reading.

`gcx kg entities query` with Cypher is the better answer. I verified
on a live stack: `MATCH (n) WHERE n.name CONTAINS 'mysql' RETURN n
LIMIT 5` returns hits across Volume / Pod / ServiceInstance / Service
in one server-bounded call. That covers the original use case (find
an entity by name without knowing its type) without the cost profile
of `--type`-optional fan-out.

I'll drop the A.12 commit from this PR (revert `08789b1` so `--type`
stays required as you originally designed it) and update the PR
description. The remaining three commits — the "looks healthy"
verdict fix, the count-before-concluding skill guardrail, and the
metric re-probe on FAIL — are unaffected.

The agent-facing guidance for using `entities query` for cross-type
name searches fits naturally in PR #748's theme (skill expansions),
so I'll add it there as a separate commit rather than widening this
PR's scope at review time. That way #746 closes as a pure
correctness-fixes PR.

Will push the revert shortly.

danstadler-pdx:
Pushed the revert (commit `53d9be7` is now the tip). `--type`
required is restored; the three remaining commits are unchanged:

- `3d85ac7` — "looks healthy" verdict fix
- `2a8dbbf` — reading-diagnose-verdicts skill guardrail
- `53d9be7` — metric re-probe on FAIL

PR description updated to reflect the scope. The cypher-based
cross-type search guidance is going into PR #748 as a separate
commit.

xujiaxj (dismissed):

internal/providers/kg/diagnose.go
```diff
@@ -668,32 +669,48 @@ func metricChecks(env, namespace string) []metricCheckDef {
...
 
+	// unscoped returns the bare count() query (no env/namespace filter).
+	// Used as the re-probe target so that when a scoped query returns
+	// "no data", checkMetric can distinguish "metric is genuinely absent"
+	// from "metric exists but doesn't match the scoped filter".
+	//
+	// Only attach an unscoped fallback when a filter was actually applied —
+	// otherwise the scoped and unscoped queries are identical and the
+	// reclassification logic in checkMetric would be a no-op.
+	envOrNs := env != "" || namespace != ""
```
radiohead: nit: if `env` is empty but `--namespace` is set, the scoped and unscoped queries are identical and the fallback fires as a no-op round-trip. How about gating the fallback per-check on whether a filter was actually applied instead?

internal/providers/kg/diagnose.go
```diff
@@ -668,32 +669,48 @@ func metricChecks(env, namespace string) []metricCheckDef {
...
+	withFallback := func(d metricCheckDef, metric string) metricCheckDef {
+		if envOrNs {
+			d.UnscopedQuery = fmt.Sprintf("count(%s)", metric)
```
radiohead: suggestion: we prefer using promql builders instead. That said, it's a tiny query, so it's prob OK to leave it as is.
```suggestion
			d.UnscopedQuery = promql.Count(promql.Metric(metric)).String()
```

radiohead (commented):
Looks good, except for me it seems like `gcx kg diagnose service --env <env>` still returns `FAIL — no data` for the label-mismatch case.

Also, I've just noticed that we use `text` output format instead of `table`, could you fix that?

danstadler-pdx:
Thanks for the careful review — here's how each point landed:

**1. `text` → `table` output format.** Done. The 3 diagnose codecs now register as `"table"`. Types renamed `*TextCodec` → `*TableCodec` for clarity.

**2. Per-check gating of the fallback.** Done. Replaced the single `envOrNs` boolean with per-check inspection of whether that check's scoped selector is non-empty.

**3. `FAIL — no data` on label-mismatch case.** Found the root cause and fixed it. The headline issue was actually in `runServiceMetricChecks()` (in `diagnose_service.go`), not `metricChecks()` — the original PR added the unscoped-fallback reclassification to the top-level `kg diagnose` path but missed the per-service path that `kg diagnose service` uses. Reproduced on a real stack (`gcx kg diagnose service NAME --env <wrong-env>` returned 5 × `FAIL — no data` despite data clearly flowing under a different env value), then confirmed the fix produces 5 × `WARN — metric exists (N series unscoped) but no series match the requested env/namespace scope` with the asserts_env label-pipeline hint.

**4. `promql.Count` builder.** Done in both the top-level and per-service fallback queries.

New tests cover all three behaviour changes including a regression test for the per-service headline fix (`TestServiceDiagnose_ReclassifiesLabelMismatch`).

`go test ./...` clean, `golangci-lint run -c .golangci.yaml ./internal/providers/kg/...` 0 issues, `gofmt -d ./internal/providers/kg/` clean.

annanay25 (commented):
Nice PR — the three correctness fixes are logically sound, well-tested, and the commit structure is clean. Two things to address:

---

**1. Bug: parameter confusion in `TestCheckMetric_NoFallbackProbeWhenSelectorEmpty`**

`NewTestScopeFlags(env, site, namespace)` takes `site` as the second parameter, but the test passes `"demo-ns"` as `site` instead of `namespace`:

```go
// Only namespace set. traces_target_info doesn't use namespace, so its
// scoped selector is empty and no fallback should attach.
scope := kg.NewTestScopeFlags("", "demo-ns", "")
```

This sets `site="demo-ns"` and `namespace=""`. Since `site` isn't used by `metricChecks()`, this effectively tests "no scope at all" rather than the intended "namespace-only scope". The test passes accidentally because both scenarios produce empty selectors for `traces_target_info`.

Should be:
```go
scope := kg.NewTestScopeFlags("", "", "demo-ns")
```

With that fix, the test would exercise the *real* per-check gating: `traces_service_graph_request_total` gets a fallback (its selector includes `client_service_namespace`) while `traces_target_info` does not — which is what the comment describes.

---

**2. Stale `"text"` references in error messages after the text→table codec rename**

The `Decode()` methods still return `"text format does not support decoding"` and the `Encode()` methods still return `"invalid data type for text codec: expected ..."`. Since the codec is now registered as `"table"`, these should say `"table"` for consistency. Affects all three files:
- `diagnose.go` (`DiagnoseTableCodec`)
- `diagnose_service.go` (`ServiceDiagnoseTableCodec`)
- `diagnose_labels.go` (`LabelsDiagnoseTableCodec`)

danstadler-pdx:
Thanks for the careful catches — both fixed in 02006e8:

1. **`NewTestScopeFlags` parameter order:** you're absolutely right. The test was passing accidentally because both "no scope" and "namespace-only" produce empty selectors for `traces_target_info`. Swapped the args so the test now actually exercises the namespace-only case it claims to. Also added a brief inline comment noting the parameter order so this doesn't bite again.

2. **Stale `"text"` references in error messages:** fixed in all three diagnose codecs (`DiagnoseTableCodec`, `ServiceDiagnoseTableCodec`, `LabelsDiagnoseTableCodec`). Left `DescribeTextCodec` in `commands.go` alone since it's still legitimately registered as `"text"`.

`go test ./...` clean, `golangci-lint run -c .golangci.yaml ./internal/providers/kg/...` 0 issues.

annanay25 (approved):

PR #748, opened 2026-05-19
feat(kg): expand kg diagnose coverage + new manage-known-errors skill

danstadler-pdx:
## Summary

Expands `gcx kg diagnose` with four new deterministic checks for
common Knowledge Graph / Asserts UI failure modes, and adds two
companion skill improvements: a new `manage-known-errors` skill for
acknowledging known/expected insights, plus targeted updates to
`diagnose-entity-graph`.

Each commit is independently reviewable. The four new code checks
follow the existing best-effort pattern: they stay silent when the
condition doesn't apply, emit `WARN` with a concrete remediation
hint when it does.

This PR builds on top of #746 (which landed correctness fixes) and
is independent of #726 / #727 (the trace-context-propagation pair).
Each pair has its own scope; no rebase dependencies.

### `DB instrumentation` (commit `d9db730`)

Probes Tempo's tags API for any span attribute in the
OpenTelemetry `db.*` family. When none exist, no service on the
stack has installed a database-client instrumentation library —
which guarantees that trace-derived Service→Database CALLS edges
cannot form, regardless of how many database entities exist in
the graph.

`WARN` recommendation gives equal billing to Python / Node.js /
Go / Java instrumentation packages, plus Beyla as the no-code
alternative.

Adds best-effort Tempo client plumbing (`resolveTempoClient`) to
the diagnose pipeline, mirroring the existing Prometheus client
pattern. The trace-side check is silently skipped when no Tempo
datasource can be resolved.

### `Container label drift` (commit `7ee5e9f`)

Detects cAdvisor `container_*` metrics arriving with `image=""`
on real workload containers. The downstream `asserts:resource`
recording rule for `cpu:usage` and `memory:usage` filters on
`image!=""` to exclude pause containers — so dropping the
`image` label silently empties the Asserts UI's Kubernetes / CPU
/ Memory tabs for that namespace.

Common cause: an Alloy `extraMetricProcessingRules` block with
`labeldrop regex = "image.*"`, often added for cardinality
reduction without realising the rule is load-bearing for
downstream Asserts views.

Stays silent when only `kube-system` shows the signature (pause
containers there legitimately have no image label).

### `Resource coverage` (commit `4386725`)

Probes which `asserts_resource_type` values are populated in
`asserts:resource` for the scoped env/namespace and flags any
missing from the canonical expected set:
`cpu:usage`, `cpu:throttle`, `memory:usage`, `disk:usage`,
`disk:inode_usage`.

Composes with `Container label drift`: the new check says
**which** Asserts UI tab will be empty; the drift check says
**why**. Both checks point at each other in their recommendations.

### `Split identity` (commit `cbfc123`)

Detects when one physical workload is discovered as two distinct
Service entities because OTel `service.name` disagrees with the
k8s workload name. KG can't merge them — the OTel-named entity
has no k8s metadata; the k8s-named entity has no trace data —
so neither view is fully usable.

Detection: `group by (job, service) (asserts:mixin_workload_job)`
scoped to the env/namespace. A single `job` mapping to multiple
distinct `service` values is a collision.

Recommendation names the fix path: align `OTEL_SERVICE_NAME` (or
`service.name` via `OTEL_RESOURCE_ATTRIBUTES`) with the k8s
workload, rename the deployment, or configure Asserts relabel
rules to reconcile.

## Skill updates (`diagnose-entity-graph`)

- **`4e6ef42`** — Step 7 now opens with an explicit pre-flight:
  before diagnosing why a service is broken, confirm it exists.
  If a user names an entity that isn't in the graph, that's the
  finding — report it directly rather than fabricate a diagnosis.
- **`e890432`** — Expand the skill's `description` frontmatter
  with empty-UI-tab phrasings so the skill auto-loads on
  "Kubernetes tab is empty" / "no CPU/memory data in Asserts" /
  "Memory tab shows nothing" / "saturation chart is blank". No
  new content in the skill body — the existing Step 6 sub-sections
  already cover those failure modes once the agent finds the skill.

Each of the four new code checks also contributes a small skill
update (one short subsection per check) that names the WARN
shape and defers to the check's `Recommendation` for the fix
path — following the lightweight-coordinator pattern the
maintainers have been asking for on #726 / #727.

## New skill: `manage-known-errors` (commit `a233a0d`)

Covers a workflow the existing skill bundle doesn't address:
acknowledging a known/expected Asserts insight so it stops
creating noise. Two mechanisms, both shipped with gcx and both
reversible:

- **Suppression** via `gcx kg suppressions create` — silences
  the insight regardless of value.
- **Custom threshold** via `gcx kg rules create` — pushes a
  `asserts:<family>:custom_threshold` recording rule that the
  Asserts assertion rule reads before falling back to the
  automatic baseline.

Names the six standard custom-threshold rules (`asserts:error:ratio:custom_threshold`,
the three latency variants, request rate, request cardinality)
and the `* 0 + <threshold>` Prometheus idiom for preserving
labels while overriding the value.

Includes an explicit anti-pattern callout: do **not** edit or
disable the Asserts-managed base alert rule. Those rules are
plugin-managed and will be overwritten on plugin update — the
`:custom_threshold` mechanism is the supported override path.

Skill follows the lightweight-coordinator pattern: routes the
agent through existing `gcx kg suppressions` and `gcx kg rules`
commands, defers to on-stack rule definitions (read via
`gcx alert rules get`) as the source of truth for which
`:custom_threshold` recording rule to target.

## What's not in this PR

- **Sibling-edges soft check** — was considered as an `INFO`
  check in `diagnose service` to flag missing edges between
  same-namespace siblings. Dropped after design review: risk of
  noise on stacks where missing edges are intentional. Can be a
  follow-up if anyone requests it.
- **`alert rules list` integration with plugin-managed rules** —
  surfacing Asserts assertion rules in the default
  `alert rules list` output would require a new provisioning-API
  client and a merge layer between two distinct rule response
  schemas. That's meaningful new code worth its own PR rather
  than bundled here.
- **`gcx kg insights explain <NAME>` subcommand** — would
  collapse a four-command discovery chain into one CLI call. Out
  of scope here; if useful, can ship as a small follow-up PR
  along with the matching skill content.
- **Slimming the pre-existing Steps 3-6 in `diagnose-entity-graph`** —
  the new code checks in this PR make several PromQL recipes in
  those steps partially redundant. Slimming should happen on
  #726's branch since the maintainer's review comments there
  already target the same content; doing it here would cause
  cross-PR conflicts.

## Validation

- `go test ./...` — clean across the whole repo
- `gofmt -d ./internal/providers/kg/...` — clean
- `go run ./scripts/validate-skills` — `validated claude-plugin/skills`
- 11 new tests across the 4 new checks (4 for DB instrumentation,
  4 for Container label drift, 3 for Resource coverage) plus 4
  for Split identity = 15 new tests total

## Diff size

5 files changed, 1335 insertions(+), 13 deletions(-)

### DIFF
--- a/claude-plugin/skills/diagnose-entity-graph/SKILL.md
+++ b/claude-plugin/skills/diagnose-entity-graph/SKILL.md
@@ -236,7 +236,22 @@ drift, or another label-pipeline issue).
 
 ## Step 7: Per-Service Investigation
 
-For a specific missing or edge-less service:
+Before investigating *why* a service is broken, confirm it exists.
+If a user names an entity that doesn't appear in the graph at all,
+that is itself the finding — report it directly, don't fabricate a
+diagnosis. Cross-type search if the type is uncertain:
+
+```bash
+gcx kg entities list --property name=~SERVICE
+```
+
+If `entities list` returns no results across any type and
+`traces_target_info{service_name="SERVICE"}` is also empty, the
+entity is not on this stack. State that conclusion plainly; the
+correct next question is whether the user meant a different stack,
+env, or spelling.
+
+For a service that *does* exist but appears missing or edge-less:
 
 ```bash
 # Find in graph
--- a/claude-plugin/skills/diagnose-entity-graph/SKILL.md
+++ b/claude-plugin/skills/diagnose-entity-graph/SKILL.md
@@ -210,6 +210,19 @@ Extra `asserts_env` values (like AWS account IDs) that don't match any
 
 **Shortcut:** `gcx kg diagnose labels` automates this cross-reference.
 
+### The `image` label is also load-bearing
+
+When `gcx kg diagnose` emits a **`Container label drift: WARN`**
+check, cAdvisor `container_*` metrics in the named namespace are
+arriving with `image=""`. The downstream `asserts:resource`
+recording rule for `cpu:usage` and `memory:usage` filters on
+`image!=""` to exclude pause containers, so dropping the `image`
+label silently empties the Asserts UI's Kubernetes / CPU / Memory
+tabs for that namespace. The check's `Recommendation` field names
+the typical source (an Alloy `extraMetricProcessingRules` block
+with `labeldrop regex = "image.*"`, often added for cardinality
+reduction) and the fix.
+
 ## Step 7: Per-Service Investigation
 
 For a specific missing or edge-less service:
--- a/claude-plugin/skills/diagnose-entity-graph/SKILL.md
+++ b/claude-plugin/skills/diagnose-entity-graph/SKILL.md
@@ -180,8 +180,20 @@ Environment page as above.
 server-side with `asserts_env` already populated, bypassing the Mimir
 relabeling pipeline entirely.
 
-## Step 6: Label Pipeline
+### Service→Database edges: a special case
 
+Service→Database CALLS edges require the calling service to emit
+OpenTelemetry database-client spans (so Tempo's service graph
+processor can see `db.system` / `db.statement` / `db.operation` on
+outbound calls). HTTP auto-instrumentation alone is not enough —
+the language-specific DB instrumentation library must be installed
+and registered too.
+
+When `gcx kg diagnose` emits a **`DB instrumentation: WARN`** check
+(present when a Tempo datasource is available), no service on the
+stack is emitting DB-client spans. The check's `Recommendation`
+field names the relevant library per language and the Beyla
+ROUTES alternative; read it as the primary playbook.
 The most common issue: `deployment_environment` isn't mapped to `asserts_env`.
 
 ```bash
--- a/claude-plugin/skills/diagnose-entity-graph/SKILL.md
+++ b/claude-plugin/skills/diagnose-entity-graph/SKILL.md
@@ -243,13 +243,15 @@ drift, or another label-pipeline issue).
 Before investigating *why* a service is broken, confirm it exists.
 If a user names an entity that doesn't appear in the graph at all,
 that is itself the finding — report it directly, don't fabricate a
-diagnosis. Cross-type search if the type is uncertain:
+diagnosis. When the type is uncertain, use a cross-type Cypher
+query so the limit is enforced server-side rather than after a
+client-side fan-out:
 
 ```bash
-gcx kg entities list --property name=~SERVICE
+gcx kg entities query "MATCH (n) WHERE n.name CONTAINS 'SERVICE' RETURN n LIMIT 10" --since 1h
 ```
 
-If `entities list` returns no results across any type and
+If that query returns no results and
 `traces_target_info{service_name="SERVICE"}` is also empty, the
 entity is not on this stack. State that conclusion plainly; the
 correct next question is whether the user meant a different stack,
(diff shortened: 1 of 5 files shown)


danstadler-pdx:
Added commit `3643dab` (the eighth on this PR) — adopts the
`gcx kg entities query` cypher pattern for cross-type entity
lookup in `diagnose-entity-graph` Step 7.

Why it lives here rather than in #746: xujiaxj flagged that the
`--type`-optional change in #746 widened backend cost without
bounding it (`--limit` is client-side, fan-out pagination doesn't
compose). They suggested the existing `gcx kg entities query`
command with a Cypher `MATCH (n) WHERE n.name CONTAINS '...'`
clause as a better fit — the limit lives at the graph layer.

The change in #746 has been reverted; the agent-facing guidance
for the better pattern goes here, where the PR theme is skill
expansions.

The commit also corrects two pre-existing `gcx kg cypher` examples
in the same section that should have been `gcx kg entities query`
(the rename happened in #709 but the skill wasn't updated, so
those examples would error out at runtime).

Discussion: [#746 thread](https://github.com/grafana/gcx/pull/746#discussion_r3270332929).

danstadler-pdx:
Closing per the maintainer feedback above ("push back on this and check if we need this").

I want to record what was in this PR so the work isn't lost, and so the team
asking for these checks can come back with a more targeted ask later.

## What this PR was going to add

### Four new diagnostic checks in `gcx kg diagnose` (~477 lines in `diagnose.go`)

| Check | Detects | UI symptom it would explain |
|---|---|---|
| `DB instrumentation` | No `db.*` span attributes exist in any Tempo span scope | "Service → Database edge is missing even though both entities exist" |
| `Container label drift` | `container_*` cAdvisor metrics have `image=""` (typically from an Alloy `labeldrop image.*` rule added for cardinality reduction) | "K8s / CPU / Memory tabs in the Asserts UI are silently empty for a namespace" |
| `Resource coverage` | `asserts:resource` is populated for some `asserts_resource_type` values but missing others (e.g. has `cpu:throttle` but no `cpu:usage`) | "CPU panel in the K8s tab is empty even though throttle data is there" |
| `Split identity` | One physical workload appears as two distinct Service entities — OTel `service.name` disagrees with the k8s workload name | "I have one deployment but I see two Services with the same pods, and neither has both k8s metadata and trace data" |

Each check is silent on healthy stacks (returns `nil`) so the cost on a clean run is zero; they only surface when the specific failure mode is present.

### A new skill: `manage-known-errors` (238 lines)

Two workflows for users who already know an Asserts insight is firing and want to manage it rather than fix it:

1. **Suppression** via `gcx kg suppressions create` — silence a known false positive forever.
2. **Custom threshold** via `gcx kg rules create` — push a `asserts:<family>:custom_threshold` recording rule to raise the bar on a specific `(env, service, request_context)` tuple. The Asserts assertion rules already prefer custom-threshold values when present; this PR documented that mechanism end-to-end.

Both workflows are reversible via the corresponding `delete` commands.

### Skill body additions to `diagnose-entity-graph`

- Empty-UI-tab phrasings in the `description` frontmatter (so phrases like "Kubernetes tab is empty", "no CPU/memory data in Asserts", "Memory tab shows nothing", "saturation chart is blank" trigger the skill).
- Callouts inside the skill body for each of the four new checks above.
- Guidance for "entity not present on this stack" terminal-finding handling.
- Cross-type entity-lookup pattern using `gcx kg entities query` (per the boundary you and @xujiaxj converged on in #746).

## Why I'm closing rather than slimming

Two of your three concerns can be addressed by trimming, but the third can't:
- Diagnose.go size (>1200 LOC) — could be split into separate files, but the underlying objection is about the volume of new code, which trimming doesn't change.
- `manage-known-errors` AX clarity — could be reworded, but the deeper question is whether a separate skill is the right surface area for this content.
- The "do we need this" question — that's a product question, not a code question. Closing is the honest answer; if the case for the checks is strong, it should be made separately and on its own merits.

## What this means for users today

- The four detection patterns above remain undetected by `gcx kg diagnose` for now. Users hitting them will see the symptoms in the Asserts UI without a CLI hint about the cause.
- The known-errors workflows (suppress / custom threshold) still work via `gcx kg suppressions` and `gcx kg rules` — they're just undocumented in skill form.

## What's not closed

The branch `feat/kg-diagnose-coverage-expansion` will remain on my fork as a reference. If a separately-scoped follow-up is wanted later (e.g. one new check on its own merits, or the `manage-known-errors` workflows folded into `diagnose-entity-graph` as a small section), that work can be cherry-picked from these 8 commits.

PR #732, opened 2026-05-18
feat(aio11y): add experiments command group

alexander-akhmetov:
Adds a new command group under `gcx aio11y`: `experiments`.

`gcx aio11y experiments` manages eval experiment runs: create a run from a JSON/YAML spec, inspect status, patch mutable fields, cancel active runs, and view the scores/report produced by the run.

- `list`, `get`, `create`, `update`, `cancel`
- `scores <run-id>` — list scores produced by an experiment
- `report <run-id>` — aggregate report with pass rate, mean score, cost/tokens, and breakdowns

**Create/update examples:**

```bash
gcx aio11y experiments list
gcx aio11y experiments get <run-id>
gcx aio11y experiments create -f experiment.yaml
cat experiment.json | gcx aio11y experiments create -f -
gcx aio11y experiments update <run-id> --name "new name"
gcx aio11y experiments cancel <run-id>
```

**Scores and reports:**

```bash
gcx aio11y experiments scores <run-id>
gcx aio11y experiments scores <run-id> --limit 50
gcx aio11y experiments report <run-id>
gcx aio11y experiments report <run-id> -o json
```

PR #749, opened 2026-05-20
feat(aio11y): gate experiments cancel on --force + confirmation

alexander-akhmetov:
Follow-up for https://github.com/grafana/gcx/pull/732

`experiments cancel` discards a partially-run experiment. Align it with the rest of the `aio11y` changing commands which all gate destructive operations with an explicit `--force` flag.


alexander-akhmetov:
@claude review

radiohead (approved):

PR #755, opened 2026-05-20
fix(kg): align rules list/get/delete with backend contract

xujiaxj:
## Summary

The \`gcx kg rules\` subcommands were decoding response shapes that don't match the Asserts backend (\`PromRulesConfigController\`), so all three of \`list\`, \`get\`, and \`delete\` were broken in different ways.

### \`list\` — always returned empty
The backend's \`GET /v1/config/prom-rules\` returns just a list of rule file names: \`{\"ruleNames\": [...]}\`. The Go client was decoding \`{\"rules\": [...]Rule}\`, which never matched, so the slice was always nil and the table showed zero rows.

**Fix:** new \`ListRuleNames\` helper for the names endpoint, then a bounded fan-out (\`errgroup\` with limit 10, matching the project default) calling the per-name GET for each. This way \`list -o yaml/json\` produces fully-populated K8s envelopes that round-trip with \`gcx resources push\`. The Go \`Rule\` type was also reshaped to match the backend's \`PrometheusRulesDto\` — it now models a rule **file** (\`Name\` + \`Groups\` of \`PromRule\`), which is what the API actually returns.

### \`get\` — returned a misleading empty envelope for deleted rules
\`GET /v1/config/prom-rules/{name}\` returns a bare \`PrometheusRulesDto\`, but the client was unwrapping a non-existent \`{\"rules\":[...]}\` envelope. A fallback (\`if rule.Name == \"\" { rule.Name = name }\`) then synthesized a fake response from the requested name, so \`get <deleted>\` returned a populated-looking object instead of an error.

**Fix:** decode directly into the file shape, and treat an empty body (no name, no groups) as not-found — some backend versions return 200 with an empty payload instead of 404. Added a regression test.

### \`delete\` — sent PUT with no body
The old command did \`UploadPromRules(ctx, \"\")\` (PUT with empty string) intending to \"clear all rules\", but the PUT endpoint requires a \`PrometheusRulesDto\` body, so it always failed with HTTP 400. There's no bulk-delete endpoint anyway.

**Fix:** \`delete\` now takes a required \`<name>\` arg and calls \`DELETE /v1/config/prom-rules/{name}\`, matching the controller's \`deletePromRules\` handler.

### Other cleanup
- Dropped dead \`Active\` and \`ManagedBy\` fields that were never read in gcx.
- Simplified the \`list\` table codec to a single NAME column (the list endpoint can't populate record/alert/expr).

## Test plan
- [x] \`go test ./internal/providers/kg/...\`
- [x] \`golangci-lint run ./internal/providers/kg/...\`
- [x] \`GCX_AGENT_MODE=false mise run reference\` — CLI ref regenerated for the new \`delete <name>\` signature
- [x] Live stack: \`bin/gcx kg rules list\` populates specs via fan-out
- [x] Live stack: \`bin/gcx kg rules get <missing>\` exits non-zero with \"not found\"
- [x] Live stack: \`bin/gcx kg rules delete <name>\` succeeds

### DIFF
--- a/internal/providers/kg/commands.go
+++ b/internal/providers/kg/commands.go
@@ -478,23 +478,23 @@ func newRulesCommand(loader RESTConfigLoader) *cobra.Command {
 			}
 
 			// Extract rules from TypedObject
-			rules := make([]Rule, len(typedObjs))
+			files := make([]Rule, len(typedObjs))
 			for i := range typedObjs {
-				rules[i] = typedObjs[i].Spec
+				files[i] = typedObjs[i].Spec
 			}
 
 			// Table codec operates on raw []Rule for direct field access.
 			// Other formats (yaml/json) convert to K8s envelope Resources
 			// for consistency with get and round-trip support.
 			if rulesListOpts.IO.OutputFormat == "table" {
-				return rulesListOpts.IO.Encode(cmd.OutOrStdout(), rules)
+				return rulesListOpts.IO.Encode(cmd.OutOrStdout(), files)
 			}
 
 			var objs []unstructured.Unstructured
-			for _, rule := range rules {
-				res, err := RuleToResource(rule, cfg.Namespace)
+			for _, f := range files {
+				res, err := RuleToResource(f, cfg.Namespace)
 				if err != nil {
-					return fmt.Errorf("failed to convert rule %s to resource: %w", rule.Name, err)
+					return fmt.Errorf("failed to convert rule %s to resource: %w", f.Name, err)
 				}
 				objs = append(objs, res.ToUnstructured())
 			}
@@ -562,9 +562,11 @@ func newRulesCommand(loader RESTConfigLoader) *cobra.Command {
 	_ = createCmd.MarkFlagRequired("file")
 
 	deleteCmd := &cobra.Command{
-		Use:   "delete",
-		Short: "Delete all Knowledge Graph rules (upload empty).",
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		Use:   "delete <name>",
+		Short: "Delete a Knowledge Graph prom rule by name.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			name := args[0]
 			cfg, err := loader.LoadGrafanaConfig(cmd.Context())
 			if err != nil {
 				return err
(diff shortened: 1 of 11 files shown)


internal/providers/kg/commands.go
```diff
@@ -469,23 +478,23 @@ func newRulesCommand(loader RESTConfigLoader) *cobra.Command {
...
+				files[i] = typedObjs[i].Spec
 			}
 
 			// Table codec operates on raw []Rule for direct field access.
 			// Other formats (yaml/json) convert to K8s envelope Resources
 			// for consistency with get and round-trip support.
 			if rulesListOpts.IO.OutputFormat == "table" {
-				return rulesListOpts.IO.Encode(cmd.OutOrStdout(), rules)
+				return rulesListOpts.IO.Encode(cmd.OutOrStdout(), files)
 			}
```
radiohead: This actually goes against "Format-agnostic data fetching: Commands fetch all data regardless of --output format; codecs control display, not data acquisition." rule.

Since we anyway fan-out to fetch rule details for YAML / JSON, I suggest doing the same for table output.

Doing so will make table output useful, as opposed to now, where it's practically useless (since it's just a list of names).

I also suggest adding support for `wide` codec. Not 100% sure what would be a good idea to include in table vs wide, but generally speaking the idea is 80/20 rule: table for overview (80% of invocations), wide for a more detailed view (20% of invocations).

internal/providers/kg/commands.go
```diff
@@ -564,10 +575,10 @@ func newRulesCommand(loader RESTConfigLoader) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			if err := client.UploadPromRules(cmd.Context(), ""); err != nil {
+			if err := client.DeleteRule(cmd.Context(), name); err != nil {
```
radiohead: Could you wire this via `TypedCRUD` client instead and call it here? That way we can ensure that `gcx resources delete kgrules/$RULE` also works.

radiohead (commented):
Looks good, just a couple of things!

internal/providers/kg/commands.go
xujiaxj: I can put the rule group counts and rule counts as additional columns for table view. The content is more like a tree structure: 1 name -> m groups -> m*n rules, so I am not sure displaying all the rules in the table is a good idea.

image

internal/providers/kg/commands.go
xujiaxj: Done.

internal/providers/kg/commands.go
radiohead: Yeah I think that's a good start. Thank you!

radiohead (approved):

PR #758, opened 2026-05-20
feat(kg): kg diagnose orientation block

danstadler-pdx:
## Summary

Adds a top-level **Orientation block** to `gcx kg diagnose` output that
surfaces which of the five original Entity Graph scenarios a user's run
matches:

1. "I see no entities at all"
2. "Some expected entities are missing"
3. "I see entities with no edges"
4. "I see disconnected clusters of entities" *(detector deferred; v1 emits a "see also" hint)*
5. "I can't filter to the entities I want"

These scenarios were the founding product framing for `kg diagnose` but had
drifted out of the user-visible output as we layered on sub-checks. This PR
restores the top-level orientation so the user (or their AI agent) doesn't
have to infer the scenario from a 10-line check table.

## Commits

| Commit | Adds |
|---|---|
| `fef190d` | Orientation types, four scenario detectors, 14 table-driven tests |
| `8b122fe` | Wires Orientation into `runDiagnose`; both text and JSON codecs surface it |
| `eb514ff` | Skill gains a 16-line "Start here" pointer at the Orientation block |
| `fac5ff8` | **Round 2:** scope-aware `detectEntitiesNoEdges` (no more false-HIGH on empty scopes) |
| `fbfff82` | **Round 2:** label stack-wide vs scoped numbers in the codec header; multi-dimension scope hint |
| `761e90f` | **Round 2:** explicit `slices.SortStableFunc` on matched scenarios (was relying on call order) |
| `5639f6f` | **Round 2:** extract `DiagnoseSummary` named type + direct unit test on `pipelineHealthFromSummary` |
| `3cef4e9` | **Round 2:** mark none-bucket detection inert until producer is wired; drop dead codec/test paths |

## Round 2 review notes (resolved)

Round 2 review from @radiohead surfaced two correctness issues + three
minors + a test-coverage gap. All addressed:

- **M2 — `detectEntitiesNoEdges` mixed stack-wide and scoped signals.**
  Gate is now `AssertsMixinWorkloadJobSeries` (scoped). Regression test
  added with the live numbers from the review.
- **M3 + m1 — codec amplified the scope mismatch and hard-coded `(env=)`.**
  Stack-wide numbers labeled `(stack-wide)`, scoped trace count on its own
  clause. Multi-dimension scope hint built from whichever of
  env/namespace/site is set.
- **m2 — accidental scenario ordering.** Replaced with explicit
  `slices.SortStableFunc` by `(Confidence, ScenarioID)`.
- **m3 — anonymous-struct param.** Extracted `DiagnoseSummary` (JSON shape
  unchanged); added direct unit test.
- **M1 + T1 — `NoneBucketHasEntities` was never produced.** Producing it
  truthfully needs a new `CountEntityTypes(env="none")` API call. Marked
  the field reserved, removed the dead codec branch and detector operand,
  dropped the two test cases that exercised the unreachable path. NOTE
  comments mark where to restore them when the producer is wired (a
  separate follow-up).

## Design rules followed

Direct response to boundary feedback from #726/#745/#748:

- **Skill/code boundary.** All detection logic in Go. The skill addition is
  16 lines — names the five scenarios verbatim and tells the agent to read
  the Orientation block first.
- **Numbers are runtime data, not authored thresholds.** Everything printed
  comes from the current `DiagnoseResult`. No constants like `total > N`.
- **Detection is binary.** No knobs, no severity scoring, no judgment phrasing.
- **Scenarios named verbatim every time.** Skill, JSON `label`, text codec —
  all use the same strings so an agent can match by exact string.
- **Cost-aware.** Zero new API calls. `computeOrientation` reads data the
  existing per-check goroutines already collected. (M1 deferred precisely
  because wiring it would add a new API call.)
- **No new file size in `diagnose.go`.** The new detector code lives in
  `orientation.go`.

## Deliberately out of scope

- **Preflight check** for the 5-confidently-wrong "no data" verdicts
  failure mode — still under investigation.
- **Scenario 4 detector** (disconnected clusters) — needs per-env
  comparison the stack-level run doesn't currently do. Text codec emits a
  "See also" hint pointing at the per-env comparison workflow.
- **Four checks proposed in #748** — kept out per the size pushback there.
- **None-bucket producer** — see M1 above; needs a new API call.

### DIFF
--- a/internal/providers/kg/diagnose.go
+++ b/internal/providers/kg/diagnose.go
@@ -49,7 +49,14 @@ type CheckResult struct {
 
 // DiagnoseResult is the full output of the diagnose command.
 type DiagnoseResult struct {
-	Env     string        `json:"env,omitempty"`
+	Env string `json:"env,omitempty"`
+
+	// Orientation is the top-of-output summary that anchors the user in
+	// one of the five Entity Graph scenarios. Computed from the same data
+	// the per-check goroutines collect; no extra API calls. Omitted from
+	// JSON when nil (e.g. legacy callers that bypass the runner).
+	Orientation *Orientation `json:"orientation,omitempty"`
+
 	Checks  []CheckResult `json:"checks"`
 	Summary struct {
 		Total  int `json:"total"`
@@ -283,24 +290,39 @@ func runDiagnose(ctx context.Context, client *Client, scope *scopeFlags, promCli
 		mu.Unlock()
 	}
 
+	// Raw data shared with the Orientation block. Each contributing
+	// goroutine fills its own field; computeOrientation reads the
+	// composed struct after g.Wait() returns. Series counts default to
+	// zero, which is the safe value for absence in the detectors.
+	var orientationInput OrientationInput
+
 	// Check 1: Stack status + sanity checks.
 	g.Go(func() error {
-		checks := checkStackStatus(ctx, client)
+		checks, enabled := checkStackStatus(ctx, client)
 		mu.Lock()
 		result.Checks = append(result.Checks, checks...)
+		orientationInput.StackEnabled = enabled
 		mu.Unlock()
 		return nil
 	})
 
 	// Check 2: Entity counts.
 	g.Go(func() error {
-		addCheck(checkEntityCounts(ctx, client))
+		c, counts := checkEntityCounts(ctx, client)
+		mu.Lock()
+		result.Checks = append(result.Checks, c)
+		orientationInput.EntityCounts = counts
+		mu.Unlock()
 		return nil
 	})
 
 	// Check 3: Scope values.
 	g.Go(func() error {
-		addCheck(checkScopeValues(ctx, client))
+		c, scopes := checkScopeValues(ctx, client)
+		mu.Lock()
+		result.Checks = append(result.Checks, c)
+		orientationInput.Scopes = scopes
+		mu.Unlock()
 		return nil
 	})
 
@@ -317,7 +339,20 @@ func runDiagnose(ctx context.Context, client *Client, scope *scopeFlags, promCli
 	if promClient != nil && datasourceUID != "" {
 		for _, mc := range metricChecks(scope.env, scope.namespace) {
 			g.Go(func() error {
-				addCheck(checkMetric(ctx, promClient, datasourceUID, mc))
+				c, series := checkMetric(ctx, promClient, datasourceUID, mc)
+				mu.Lock()
+				result.Checks = append(result.Checks, c)
+				// Capture series counts the Orientation detectors need.
+				// metricChecks emits these in a fixed order; match by name.
+				switch mc.Name {
+				case "Metric: traces_target_info":
+					orientationInput.TracesTargetInfoSeries = series
+				case "Metric: asserts:relation:calls":
+					orientationInput.AssertsRelationCallsSeries = series
+				case "Metric: asserts:mixin_workload_job":
+					orientationInput.AssertsMixinWorkloadJobSeries = series
+				}
+				mu.Unlock()
 				return nil
 			})
 		}
@@ -351,7 +386,33 @@ func runDiagnose(ctx context.Context, client *Client, scope *scopeFlags, promCli
 	})
 
 	result.computeSummary()
+
+	// Build the Orientation block from the data collected above.
+	orient := computeOrientation(orientationInput, scope)
+	orient.PipelineHealth = pipelineHealthFromSummary(result.Summary)
+	result.Orientation = &orient
+
 	return result
+}
+
+// pipelineHealthFromSummary derives the Orientation health verdict from
+// the existing pass/warn/fail counts the per-check goroutines produce.
+// Kept here (not in orientation.go) because it depends on the
+// DiagnoseResult.Summary shape.
+func pipelineHealthFromSummary(s struct {
+	Total  int `json:"total"`
+	Passed int `json:"passed"`
+	Failed int `json:"failed"`
+	Warned int `json:"warned"`
+}) PipelineHealth {
+	switch {
+	case s.Failed > 0:
+		return PipelineFailed
+	case s.Warned > 0:
+		return PipelineDegraded
+	default:
+		return PipelineHealthy
+	}
 }
 
 // checkOrder returns a sort key for deterministic output ordering.
@@ -391,7 +452,7 @@ func checkOrder(name string) int {
 // Individual checks
 // ---------------------------------------------------------------------------
 
-func checkStackStatus(ctx context.Context, client *Client) []CheckResult {
+func checkStackStatus(ctx context.Context, client *Client) ([]CheckResult, bool) {
 	status, err := client.GetStatus(ctx)
 	if err != nil {
 		return []CheckResult{{
@@ -399,10 +460,11 @@ func checkStackStatus(ctx context.Context, client *Client) []CheckResult {
 			Status:         CheckFail,
 			Detail:         fmt.Sprintf("API error: %v", err),
 			Recommendation: "Verify the Grafana instance is reachable and the Asserts plugin is installed.",
-		}}
+		}}, false
 	}
 
 	var results []CheckResult
+	enabled := status.Enabled && status.Status == "complete"
 
 	// Main status check.
 	if status.Enabled && status.Status == "complete" {
@@ -452,10 +514,13 @@ func checkStackStatus(ctx context.Context, client *Client) []CheckResult {
 		results = append(results, c)
 	}
 
-	return results
+	return results, enabled
 }
 
-func checkEntityCounts(ctx context.Context, client *Client) CheckResult {
+// checkEntityCounts returns the user-facing check result plus the raw
+// per-type count map so the Orientation block can compute entity totals
+// without a second API call. counts is nil on API error.
+func checkEntityCounts(ctx context.Context, client *Client) (CheckResult, map[string]int64) {
 	now := time.Now()
 	counts, err := client.CountEntityTypes(ctx, now.Add(-1*time.Hour).UnixMilli(), now.UnixMilli(), nil)
 	if err != nil {
@@ -464,7 +529,7 @@ func checkEntityCounts(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckFail,
 			Detail:         fmt.Sprintf("API error: %v", err),
 			Recommendation: "Failed to retrieve entity counts. Check connectivity to the Asserts API.",
-		}
+		}, nil
 	}
 
 	if len(counts) == 0 {
@@ -473,7 +538,7 @@ func checkEntityCounts(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckFail,
 			Detail:         "no entity types found",
 			Recommendation: "No entities discovered. Verify that traces_target_info or asserts:mixin_workload_job metrics are being produced.",
-		}
+		}, counts
 	}
 
 	var total int64
@@ -496,17 +561,20 @@ func checkEntityCounts(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckFail,
 			Detail:         "all entity type counts are 0",
 			Recommendation: "Entity types exist but have no instances. Check that recording rules are producing data and the graph ingestion pipeline is running.",
-		}
+		}, counts
 	}
 
 	return CheckResult{
 		Name:   "Entity counts",
 		Status: CheckPass,
 		Detail: fmt.Sprintf("%d total (%s)", total, strings.Join(parts, ", ")),
-	}
+	}, counts
 }
 
-func checkScopeValues(ctx context.Context, client *Client) CheckResult {
+// checkScopeValues returns the user-facing check result plus the raw
+// scopes map so the Orientation block can detect known/unknown scope
+// values without a second API call. scopes is nil on API error.
+func checkScopeValues(ctx context.Context, client *Client) (CheckResult, map[string][]string) {
 	scopes, err := client.ListEntityScopes(ctx)
 	if err != nil {
 		return CheckResult{
@@ -514,7 +582,7 @@ func checkScopeValues(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckFail,
 			Detail:         fmt.Sprintf("API error: %v", err),
 			Recommendation: "Failed to retrieve scope values. Check connectivity to the Asserts API.",
-		}
+		}, nil
 	}
 
 	if len(scopes) == 0 {
@@ -523,7 +591,7 @@ func checkScopeValues(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckWarn,
 			Detail:         "no scope dimensions returned",
 			Recommendation: "No env/site/namespace values found. Entities may exist without scope labels.",
-		}
+		}, scopes
 	}
 
 	var parts []string
@@ -539,14 +607,14 @@ func checkScopeValues(ctx context.Context, client *Client) CheckResult {
 			Status:         CheckWarn,
 			Detail:         "scope dimensions present but env/site/namespace are empty",
 			Recommendation: "The asserts_env label may not be set. Verify that deployment_environment is configured in your OTel SDK and that Mimir relabeling rules map it to asserts_env.",
-		}
+		}, scopes
 	}
 
 	return CheckResult{
 		Name:   "Scope values",
 		Status: CheckPass,
 		Detail: strings.Join(parts, "; "),
-	}
+	}, scopes
 }
 
 func checkTelemetryConfigs(ctx context.Context, client *Client) []CheckResult {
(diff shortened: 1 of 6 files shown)


danstadler-pdx:
All three commits in, flipped from Draft to ready for review.

## What's in the PR

| Commit | Adds |
|---|---|
| `fef190d` | Orientation types, four scenario detectors, 14 table-driven tests |
| `8b122fe` | Wires Orientation into `runDiagnose`; both text and JSON codecs surface it |
| `eb514ff` | Skill gains a 16-line "Start here" pointer at the Orientation block |

Net: +1159 / -23 lines across 4 files. ~440 of the additions are tests; the
production code change is roughly 280 lines split across the new `orientation.go`
(detectors) and small edits to `diagnose.go` (wiring + render).

## Live output on a healthy stack (unscoped)

```
Pipeline health: healthy  (unscoped: --env / --namespace / --site not set)
Entity overview: 542 entities — 110 Service entities, of which 22 emit traces.
Scope: 4 env(s) known; 22 namespace(s) known.

→ Stack looks healthy. For per-service investigation:
    gcx kg diagnose service NAME --env ENV

Ways to scope this further:
  • By namespace (you have 22 namespace(s) on this stack):
      gcx kg diagnose --namespace NS
  • By service name pattern:
      gcx kg entities query "MATCH (s:Service) WHERE s.name CONTAINS 'X' RETURN s LIMIT 20"
  • By single service + neighbors:
      gcx kg diagnose service NAME

See also: to detect disconnected clusters across environments,
re-run scoped to each --env separately and compare.

[existing check table follows]
```

## Live output when checks fail

When `Pipeline health` is `degraded` or `failed` and no scenario matched,
the "no match" line tailors the wording instead of saying "looks healthy":

```
→ No top-level scenario matched, but some checks failed.
  See the table and recommendations below for details.
```

## Design rules followed

Direct response to the boundary feedback we converged on in #726/#745/#748:

- **Skill/code boundary.** All detection logic in Go. The skill addition is
  16 lines — names the five scenarios verbatim and tells the agent to read
  the Orientation block first. No restated PromQL, no failure-mode taxonomy.
- **Numbers are runtime data, not authored thresholds.** Everything printed
  comes from the current `DiagnoseResult`. No constants like `total > N`.
- **Detection is binary.** No knobs, no severity scoring, no judgment phrasing.
- **Scenarios named verbatim every time.** Skill, JSON `label`, text codec —
  all use the same strings so an agent can match by exact string.
- **Cost-aware.** Zero new API calls. `computeOrientation` reads data the
  existing per-check goroutines already collected.
- **No new file size in `diagnose.go`.** The new code lives in
  `orientation.go`. Reflects your size pushback on #748.

## Deliberately out of scope

- **No preflight check.** A separate failure mode (5 confidently-wrong
  "no data" verdicts on a customer stack) is still under investigation;
  if it turns out to need defensive code, that's a future PR.
- **No scenario 4 detector.** Needs per-env comparison the stack-level
  run doesn't currently do. The text codec emits a "See also" hint
  pointing the user at the per-env comparison workflow.
- **No new checks added to `diagnose.go`.** The four checks proposed in
  #748 stay out of scope here.

## What this enables for future work

The Orientation block becomes the top-level integration surface for any
future per-scenario work. If a new sub-check needs to surface, it adds
itself as a `MatchedScenario` rather than as a new row in the check
table — keeping the user's mental model anchored on the five scenarios
instead of growing the table.

xujiaxj (dismissed):
I don't have full context here, but since this is self-contained functionality, and looks okay style-wise, I think it's fine to move forward and continue iterating.

radiohead (commented):
Nice feature — the pure-detector / IO-in-`runDiagnose` split is clean, the table-driven tests are thorough, and the concurrency is race-safe (every `orientationInput` field is written under `mu` and read only after `g.Wait()`). I built the branch and ran `gcx kg diagnose` against a live multi-tenant stack to validate; the findings below are confirmed against real output, not just the diff.

Two correctness issues I'd resolve **before flipping out of draft**, both about the data feeding the detectors rather than the design itself:

1. **`NoneBucketHasEntities` is read in three places but never written in `runDiagnose`** → scenario 5's none-bucket path and its codec callout are inert in production, while the struct comment claims the field is computed.
2. **Stack-wide entity counts are combined with scoped metric series** in `detectEntitiesNoEdges` and the codec header → any `--env`/`--namespace` pointing at an empty/sparse/bogus scope produces a *false HIGH-confidence* "I see entities with no edges" verdict that outranks the correct MEDIUM "Some expected entities are missing" one.

Plus three minors and a test-coverage gap (the unit tests are green over the unreachable none-bucket path, so CI won't catch finding #1). Inline comments below.

**Housekeeping:** the PR description still says *"Draft. Commit 1 of 3 is in"* with commits 2–3 unchecked, but all three (wiring + `SKILL.md`) are on the branch. Please refresh the body/checklist so reviewers know the full scope is in.

---
**Generated by Claude Code**

internal/providers/kg/orientation.go
```diff
@@ -0,0 +1,481 @@
...
+
+	// Scopes is the ListEntityScopes response (env / site / namespace
+	// dimension values). May include "none" as a literal value for the
+	// unscoped bucket.
+	Scopes map[string][]string
+
+	// NoneBucketHasEntities indicates that the "none" env value is
+	// present in Scopes AND has entities behind it. Detected by
+	// runDiagnose at the API boundary and passed in here.
+	NoneBucketHasEntities bool
```
radiohead, line 233: 🔴 **Major — this field is never produced.**

The comment says it's *"Detected by `runDiagnose` at the API boundary and passed in here,"* but `runDiagnose` only assigns `StackEnabled`, `EntityCounts`, `Scopes`, `TracesTargetInfoSeries`, `AssertsRelationCallsSeries`, `AssertsMixinWorkloadJobSeries` (diagnose.go:304–353). This field keeps its zero value `false`, which cascades:

- `buildScopeSummary` (L314) → `NoneBucketPresent` is always `false`.
- The codec callout `if o.Scope.NoneBucketPresent` (diagnose.go:1187) never renders.
- In `detectCantFilter` (L383) the gate `if !in.NoneBucketHasEntities && !scopeValueUnknown(...)` collapses to "fire only on unknown scope value," so the none-bucket reasoning string just below it is also dead (always overwritten by the `scopeValueUnknown` branch).

**Verified on a live stack:** `"none"` is present in `scope.envsKnown`, yet `scope.noneBucketPresent` is `false` on every run and the callout never appears; scenario 5 only fired via a bogus `--env`.

Computing it honestly needs to know whether the `"none"` env has entities behind it, which the unscoped `CountEntityTypes(…, nil)` (diagnose.go:525) doesn't break down — so it needs a derived signal or one scoped query. **Either wire the producer, or — if deferring — change the comment to "reserved; not yet populated," drop the `NoneBucketPresent` codec branch, and remove the `NoneBucketHasEntities` operand from `detectCantFilter` (or add a test asserting it's inert)** so the code stops advertising a capability it doesn't have.

internal/providers/kg/orientation.go
```diff
@@ -0,0 +1,481 @@
...
+			"gcx kg diagnose --env <env>",
+		},
+	}
+}
+
+// detectEntitiesNoEdges → scenario 3, "I see entities with no edges".
+// Triggers when Service entities exist for the scope but
+// asserts:relation:calls has zero series.
+func detectEntitiesNoEdges(in OrientationInput, overview EntityOverview, scope *scopeFlags) *MatchedScenario {
+	if overview.TotalServiceCount == 0 {
```
radiohead: 🔴 **Major — this detector mixes a stack-wide signal with a scoped one.**

`overview.TotalServiceCount` comes from `CountEntityTypes(…, nil)` (diagnose.go:525) and is **always stack-wide** — `checkEntityCounts` doesn't even take `scope`. But `in.AssertsRelationCallsSeries` comes from `metricChecks(scope.env, scope.namespace)` (diagnose.go:722) and **is scoped**. So the predicate is "stack-wide services exist AND scoped edges == 0," and L363 then injects `scope.env` into a claim (L369) backed by the unscoped count.

**Verified** with `--env azure-westeurope-1` (an env with no data on a busy stack):
```
Pipeline health: degraded  (env=azure-westeurope-1)
Entity overview: 96774 entities — 6427 Service entities, of which 0 emit traces.
→ "I see entities with no edges"        (high)    ← false: this env has no services
→ "Some expected entities are missing"  (medium)  ← true
```
The HIGH verdict is the wrong one and, by append order in `computeOrientation`, it's shown first. This bites any scope filter pointing at an empty/sparse/nonexistent env or namespace — routine on a 26-env / 399-namespace stack — and points the operator (or an agent reading the JSON) at relabeling/service-graph debugging when the real issue is an empty scope.

**Suggested fix** — gate "services exist" on a *scoped* signal so both gates share a scope; this also makes the two detectors mutually exclusive and removes the contradiction:
```go
// Replace the stack-wide TotalServiceCount gate with the scoped
// workload signal, so "services exist for env=X" is actually about env=X.
if in.AssertsMixinWorkloadJobSeries == 0 {
    return nil // no workloads in scope -> defer to detectMissingEntities
}
if in.AssertsRelationCallsSeries > 0 {
    return nil
}
```

internal/providers/kg/diagnose.go
```diff
@@ -1069,3 +1147,102 @@ func (c *DiagnoseTableCodec) Encode(w io.Writer, v any) error {
...
+	scopeHint := "(unscoped: --env / --namespace / --site not set)"
+	if o.Scope.FilterSet {
+		scopeHint = fmt.Sprintf("(env=%s)", env)
+	}
+	fmt.Fprintf(w, "Pipeline health: %s  %s\n", o.PipelineHealth, scopeHint)
+
+	// Entity overview.
+	fmt.Fprintf(w, "Entity overview: %d entities", o.EntityOverview.Total)
+	if o.EntityOverview.TotalServiceCount > 0 {
+		fmt.Fprintf(w, " — %d Service entities, of which %d emit traces",
```
radiohead: 🔴 **Major (codec amplification of the scope mismatch).**

The header prints `(env=%s)` (L1171), then this line prints `Total` and `TotalServiceCount` (stack-wide) alongside `TracedServiceCount` (which *is* scoped, via `TracesTargetInfoSeries`). The live output *"6427 Service entities, of which 0 emit traces"* puts a stack-wide numerator over a scoped "of which" in one sentence — it reads like 6427 services stopped emitting traces, when the truth is "this env has ~no services." Even after fixing the detector, either don't print stack-wide totals under an `(env=X)` header, or label them, e.g. `Stack-wide: %d entities`, so the scope provenance is unambiguous.

internal/providers/kg/diagnose.go
```diff
@@ -1069,3 +1147,102 @@ func (c *DiagnoseTableCodec) Encode(w io.Writer, v any) error {
...
+//   - When scenario 4 is not detected at stack level, a "see also" hint
+//     points the user at the per-env comparison workflow.
+//
+// All numeric values come from the live OrientationInput. No authored
+// thresholds.
+func renderOrientation(w io.Writer, o Orientation, env string) {
+	// Header: one-line stack health verdict and scope.
+	scopeHint := "(unscoped: --env / --namespace / --site not set)"
+	if o.Scope.FilterSet {
+		scopeHint = fmt.Sprintf("(env=%s)", env)
```
radiohead: 🟡 **Minor — `(env=)` is printed for non-env filters.**

`o.Scope.FilterSet` is true for any of env/namespace/site, but the hint hard-codes `env`. `--namespace asserts` (no `--env`) renders `Pipeline health: degraded  (env=)` (verified). Reflect the dimension actually set, e.g. build the hint from whichever of env/namespace/site is non-empty rather than always `env=`.

internal/providers/kg/orientation.go
```diff
@@ -0,0 +1,481 @@
...
+	AssertsMixinWorkloadJobSeries int64
+}
+
+// computeOrientation produces the user-facing Orientation block from raw
+// API data and the scope flags. It runs after the per-check goroutines
+// complete; it adds no new API calls of its own.
+//
+// Scenario ordering in the output: high-confidence detectors first
+// (no-entities, entities-no-edges, cant-filter), then medium-confidence
+// (missing-entities). Within a tier, ScenarioID alphabetical order is
```
radiohead: 🟡 **Minor — ordering comment is inaccurate, and the order relies on append sequence.**

The comment says *"Within a tier, ScenarioID alphabetical order is stable for testability,"* but the append order is `no-entities, entities-no-edges, cant-filter, missing-entities` — not alphabetical (`cant-filter` would sort first). More importantly, "ordered by confidence (high first)" isn't a sort — it's an accident of call order, guarded only by `TestComputeOrientation_MultipleScenarios_OrderByConfidence`. A reordered or newly-added detector would silently break the invariant. Either add an explicit `slices.SortStableFunc` by confidence after the appends, or correct the comment to describe the real (call-order) behavior.

internal/providers/kg/diagnose.go
```diff
@@ -351,9 +386,35 @@ func runDiagnose(ctx context.Context, client *Client, scope *scopeFlags, promCli
...
+	result.Orientation = &orient
+
 	return result
 }
 
+// pipelineHealthFromSummary derives the Orientation health verdict from
+// the existing pass/warn/fail counts the per-check goroutines produce.
+// Kept here (not in orientation.go) because it depends on the
+// DiagnoseResult.Summary shape.
+func pipelineHealthFromSummary(s struct {
```
radiohead: 🟡 **Minor — anonymous struct parameter is brittle.**

Coupling this signature to `DiagnoseResult.Summary`'s unnamed inline type means any change to the summary shape silently breaks the match or forces a duplicate literal. Extract a named `Summary` type and use it in both places. (Related: `computeOrientation` returns a throwaway `PipelineHealthy` that `runDiagnose` always overwrites, so the health derivation is never exercised through `computeOrientation` in tests — worth a direct unit test on `pipelineHealthFromSummary`.)

internal/providers/kg/orientation_test.go
```diff
@@ -0,0 +1,421 @@
...
+			in:            kg.OrientationInput{Scopes: map[string][]string{"namespace": {"team-a"}}},
+			wantFilterSet: true,
+			wantNoneBkt:   false,
+		},
+		{
+			name:  "none bucket present and populated",
+			scope: kg.NewTestScopeFlags("", "", ""),
+			in: kg.OrientationInput{
+				Scopes:                map[string][]string{"env": {"prod", "none"}},
+				NoneBucketHasEntities: true,
```
radiohead: 🧪 **Test coverage — green over a production-unreachable path.**

This case (and `TestDetectCantFilter/"none bucket has entities"` at L199) sets `NoneBucketHasEntities: true` directly and passes — giving false confidence that scenario 5's none-bucket detection works end-to-end. Because `runDiagnose` never populates that field (see the Major comment on `orientation.go`), the suite is green over code production can't reach. Once the producer is wired, add a `runDiagnose`-level (or `computeOrientation`-composition) test that exercises the none-bucket path from real inputs rather than a hand-set flag.

danstadler-pdx:
Thanks for the careful round 2, @radiohead — and to @xujiaxj for the approve. Pushed 5 follow-up commits (`fac5ff8..3cef4e9`); disposition per item:

| Item | Disposition | Commit |
|---|---|---|
| **M2** — stack-wide vs scoped mismatch in `detectEntitiesNoEdges` | Accept (your suggested snippet) — gated on `AssertsMixinWorkloadJobSeries`; regression test uses your live numbers | `fac5ff8` |
| **M3** — codec amplified the scope mismatch | Accept (your "label them" option) — `(stack-wide)` + `Tracing in scope:` clauses; unscoped output unchanged | `fbfff82` |
| **m1** — `(env=)` for non-env filters | Accept — hint built from whichever of env/namespace/site is set (e.g. `(env=prod, namespace=asserts)`) | `fbfff82` |
| **m2** — accidental scenario ordering | Accept — explicit `slices.SortStableFunc` by `(Confidence, ScenarioID)`; comment corrected | `761e90f` |
| **m3** — anonymous-struct param | Accept — extracted `DiagnoseSummary`; JSON shape unchanged; direct unit test added | `5639f6f` |
| **M1** — `NoneBucketHasEntities` never produced | **Defer.** Wiring it needs a new `CountEntityTypes(env="none")` API call, which the cost-aware rule we agreed on in #726/#745 says belongs in a separate PR. Took your "defer cleanly" option: field doc-commented "reserved; not yet populated", dead codec branch removed, dead operand removed from `detectCantFilter` (now triggers on `scopeValueUnknown` alone — the path that fires in prod). JSON `noneBucketPresent` stays in the schema (always `false` today) to avoid a shape change. | `3cef4e9` |
| **T1** — tests green over unreachable none-bucket path | Accept (paired with M1) — dropped both `NoneBucketHasEntities: true` cases; NOTE comments mark where to restore them when the producer lands | `3cef4e9` |
| **H1** — PR body still said "Draft, commit 1 of 3" | Done — body refreshed above. |

`go test ./...` and `golangci-lint run ./internal/providers/kg/...` clean. Ready for re-review.

annanay25 (approved):

PR #759, opened 2026-05-21
feat(traces): add adaptive config show/set commands

csmarchbanks:
## Summary

- Adds `gcx traces adaptive config show` and `gcx traces adaptive config set` against the singleton `/adaptive-traces/api/v1/config` endpoint.
- `set` replaces the entire payload via PUT — the API doesn't support partial patches, mirroring the `notification-policies set` precedent.
- The endpoint isn't registered as a typed adapter (no list/create/delete semantics); per CONSTITUTION's provider-only rules the commands use the allowed alternative verbs `show` and `set` rather than `get`/`update`.
- Payload modeled as opaque `Config map[string]any` so server-introduced fields round-trip without CLI changes.

## Test plan

- [x] `go test ./internal/providers/traces/...` — new client and command tests pass (table-driven + a GET/PUT/GET round-trip integration test)
- [x] `go test ./cmd/gcx/root/` — `TestConsistency_AllLeafCommandsHaveTokenCost` still green after adding annotations for the two new leaves
- [x] `go test ./...` (with `GCX_AGENT_MODE=false` to avoid agent-mode auto-detection tripping pre-existing destructive-prompt tests)
- [x] Manual: `gcx traces adaptive config show` and `gcx traces adaptive config set -f config.yaml` against a real tenant

### DIFF
--- a/docs/reference/cli/gcx_traces_adaptive_config_set.md
+++ b/docs/reference/cli/gcx_traces_adaptive_config_set.md
@@ -1,11 +1,23 @@
-## gcx traces adaptive
+## gcx traces adaptive config set
 
-Manage Adaptive Traces resources
+Replace the Adaptive Traces tenant configuration.
+
+### Synopsis
+
+Replace the Adaptive Traces tenant configuration with the contents of the supplied file. The API does not support partial patches — the entire payload is required and the existing document is overwritten.
+
+```
+gcx traces adaptive config set [flags]
+```
 
 ### Options
 
 ```
-  -h, --help   help for adaptive
+  -f, --filename string   File containing the full configuration payload (use - for stdin)
+      --force             Skip confirmation prompt
+  -h, --help              help for set
+      --json string       Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+  -o, --output string     Output format. One of: agents, json, yaml (default "yaml")
 ```
 
 ### Options inherited from parent commands
@@ -22,8 +34,5 @@ Manage Adaptive Traces resources
 
 ### SEE ALSO
 
-* [gcx traces](gcx_traces.md)	 - Query Tempo datasources and manage Adaptive Traces
 * [gcx traces adaptive config](gcx_traces_adaptive_config.md)	 - Manage the Adaptive Traces tenant configuration.
-* [gcx traces adaptive policies](gcx_traces_adaptive_policies.md)	 - Manage Adaptive Traces sampling policies.
-* [gcx traces adaptive recommendations](gcx_traces_adaptive_recommendations.md)	 - Manage Adaptive Traces recommendations.
 
--- a/docs/reference/cli/gcx_traces_adaptive_config_set.md
+++ b/docs/reference/cli/gcx_traces_adaptive_config_set.md
@@ -6,6 +6,8 @@ Replace the Adaptive Traces tenant configuration.
 
 Replace the Adaptive Traces tenant configuration with the contents of the supplied file. The API does not support partial patches — the entire payload is required and the existing document is overwritten.
 
+To avoid clobbering fields you did not intend to change, run `gcx traces adaptive config show` first, edit the returned document, and pass the full result back to `set`. Any field omitted from the payload is dropped, not preserved.
+
 ```
 gcx traces adaptive config set [flags]
 ```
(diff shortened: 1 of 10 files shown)


docs/reference/cli/gcx_traces_adaptive_config_set.md
```diff
@@ -0,0 +1,38 @@
+## gcx traces adaptive config set
+
+Replace the Adaptive Traces tenant configuration.
+
+### Synopsis
+
+Replace the Adaptive Traces tenant configuration with the contents of the supplied file. The API does not support partial patches — the entire payload is required and the existing document is overwritten.
```
yvrhdn, line 7: I wonder if we need to repeat here to first fetch the existing config and update it in place. Just an extra reminder/warning for LLMs to not just overwrite existing config.

yvrhdn (dismissed):

docs/reference/cli/gcx_traces_adaptive_config_set.md
csmarchbanks: :+1: Good idea, done!

internal/agent/command_annotations.go
```diff
@@ -495,6 +495,8 @@ var commandAnnotations = map[string]annotation{
 	"gcx traces adaptive recommendations apply":   {Cost: "small"},
 	"gcx traces adaptive recommendations dismiss": {Cost: "small"},
 	"gcx traces adaptive recommendations show":    {Cost: "small"},
+	"gcx traces adaptive config show":             {Cost: "small"},
```
radiohead, line 565: Hm, judging by maxConfigFileSize is the maximum size of a config file (10 MB). this is definitely not small.

internal/agent/command_annotations.go
```diff
@@ -495,6 +495,8 @@ var commandAnnotations = map[string]annotation{
...
+	"gcx traces adaptive config set":              {Cost: "small"},
```
radiohead: I think we could add the hint about `show-before-set` mechanics here.

radiohead (commented):
Looks good, just a couple of small details.

internal/agent/command_annotations.go
csmarchbanks: :+1: I reduced the max size of the config file. In most cases it should be very small, though users can add conditions that are of variable length.

internal/agent/command_annotations.go
csmarchbanks: Done!

csmarchbanks:
I just rebased this, @radiohead could I get another review please when you have a moment? Thanks!

internal/providers/traces/adaptive/commands.go
```diff
@@ -605,3 +606,184 @@ func ReadPolicyFromReader(reader io.Reader) (*Policy, error) {
...
+		defer f.Close()
+		reader = f
+	}
+
+	return ReadConfigFromReader(reader)
+}
+
+// ReadConfigFromReader decodes a Config from an io.Reader (YAML or JSON). Exported for testing.
+func ReadConfigFromReader(reader io.Reader) (Config, error) {
+	data, err := io.ReadAll(io.LimitReader(reader, maxConfigFileSize))
```
radiohead, line 771: Using `LimitReader` is a good idea here but we should actually error out on over-the-limit files. It's possible (I've checked) to read an oversized file that gets truncated but is still valid YAML syntactically – this might result in someone unintentionally bricking their setup (e.g. if a piece of config gets cut off).

internal/providers/traces/adaptive/commands.go
```diff
@@ -605,3 +606,184 @@ func ReadPolicyFromReader(reader io.Reader) (*Policy, error) {
...
+			if err := opts.IO.Validate(); err != nil {
+				return err
+			}
+
+			cfg, err := readConfigFromFile(opts.File, cmd.InOrStdin())
+			if err != nil {
+				return err
+			}
+
+			proceed, err := providers.ConfirmDestructive(cmd.InOrStdin(), cmd.ErrOrStderr(), opts.Force,
```
radiohead, line 718: This breaks use-case of reading the file from STDIN with `cat file.yaml | gcx` if gcx is not invoked with `--foce` or `GCX_AUTO_APPROVE`.

internal/providers/traces/adaptive/commands.go
```diff
@@ -605,3 +606,184 @@ func ReadPolicyFromReader(reader io.Reader) (*Policy, error) {
...
+func (o *configShowOpts) setup(flags *pflag.FlagSet) {
+	o.IO.DefaultFormat("yaml")
+	o.IO.BindFlags(flags)
+}
+
+func (h *tracesHelper) configShowCommand() *cobra.Command {
+	opts := &configShowOpts{}
+	cmd := &cobra.Command{
+		Use:   "show",
+		Short: "Show the Adaptive Traces tenant configuration.",
```
radiohead, line 643: This is on me for taking so long with this PR but we've changed our command naming guidelines, so this should be `get` instead. Sorry!

radiohead (changes requested):
Apologies for the very late review (I was on a leave), but I finally got around to it. Could you rebase the PR and address a few small issues? Thank you! 🙇

PR #766, opened 2026-05-22
fix(k6): use grafana.token for auth and cache the v3 token

radiohead:
## What

Switches the token sent in `X-Grafana-Service-Token` during the k6 `/v3/account/grafana-app/start` exchange from the CAP token (`cloud.token`) to the Grafana service-account token (`grafana.token`). The CAP token is ~190 characters; k6 backend PR #2537 added a 100-character length check that now rejects it. The `glsa_*` service-account token is what the k6 backend expects (it stores it as `Organization.grafana_service_token` to call back to the stack for folder/RBAC management).

Adds a cross-invocation cache for the exchanged v3 token and org ID, stored in the provider config under `providers.k6.{cached-token,cached-org-id,cached-stack-id}`. Subsequent invocations skip the `/start` round-trip. The cache is keyed by stack ID; a stack change invalidates it, mirroring the synth provider's `LoadSMConfig` pattern.

Adds a `doWithReauth` retry path: on a 401 response from any k6 API call, the cached credentials are dropped, `/start` is re-exchanged, and the request is retried once with fresh credentials.

## Why

k6 backend introduced a 100-character length limit on the `X-Grafana-Service-Token` header. The CAP token stored in `cloud.token` (~190 chars) now fails that check. Previously the check did not exist and already-provisioned stacks short-circuited, so the header value never mattered.

Caching the v3 token avoids an extra `/start` round-trip on every `gcx k6` invocation, which was always wasteful but is now more noticeable since the token exchange is on the hot path for every command.

## How it was tested

- [x] `go test ./internal/providers/k6/...` passed (all tests, 0 failures, 0.492s)
- [x] `TestLoadCache` exercises six hit/miss cases: full cache hit, nil map, empty token, stack mismatch, non-numeric org ID, and non-numeric cached stack ID
- [x] `TestAuthenticatedClient_CachePath` verifies that a valid cache hit skips the HTTP exchange entirely — `SaveProviderConfig` is not called
- [x] `TestAuthenticatedClient_ColdPath` verifies that a cold start performs the exchange, populates the client, and persists all three cache fields
- [x] `TestClient_401WithReauthRetries` verifies that a 401 triggers reauth and the retry uses the refreshed `Authorization` header
- [x] `TestClient_doRaw_401WithReauthRetries` verifies the same 401-retry path for the raw (multipart/form-data) upload path

---
**Generated by Claude Code**


radiohead:
Closing — superseded by #771.

After PR #642 (`fix(k6): make k6 provider work with OAuth login`) landed, the flat-client approach in this PR was no longer applicable: the CAP/SA token exchange path that this PR was patching was deleted entirely and replaced with OAuth via the `grafana-k6-app` plugin proxy. #642's body explicitly notes "SA tokens are still unsupported".

#771 takes a different shape — a dual-client architecture: a `ProxyClient` (unchanged from #642's OAuth path) and a new `DirectClient` (this PR's SA-token + cache + 401-retry logic), with `authenticatedClient` branching on `IsOAuthProxy()`. Both implement a shared `k6.API` interface with compile-time assertions.

---
**Generated by Claude Code**

PR #768, opened 2026-05-22
feat(kg): expose read-only relabel-rules get for prologue, epilogue, generated

xujiaxj:
## Context

`gcx kg relabel-rules create` only uploads the **prologue** rule group, even though the backend supports three distinct types — `prologue`, `epilogue`, and `generated` — with full GET/PUT/DELETE on each. The existing command exposes a third of the surface, write-only, with no way to inspect what's there.

The Asserts UI for relabel rules is also behind a feature flag and not on by default. Shipping write paths ahead of the UI risks users mutating config they cannot see or roll back from, so we defer all writes (apply/create, delete, epilogue PUT/DELETE) until the flag flips on. For now we ship `get` across all three rule types so users and on-call can inspect the rules that `kg diagnose` points at when `asserts_env` is missing, and troubleshoot relabel-pipeline issues without UI access.

## Summary

- Replace `gcx kg relabel-rules create` with `gcx kg relabel-rules get --type {prologue|epilogue|generated}` (default `generated`).
- Wire up `Client.GetRelabelRules` covering all three backend GET endpoints; 204 No Content surfaces a friendly "no rules configured" stderr message.
- Drop the `UploadRelabelRules` client + the `create` subcommand and registry entry; replace with a token-cost annotation for `get`.

## Test plan

- [x] `go test -race ./internal/providers/kg/...`
- [x] `go test -race ./...` (with `GCX_AGENT_MODE=false`)
- [x] `golangci-lint run ./internal/providers/kg/...` — 0 issues
- [x] `GCX_AGENT_MODE=false mise run reference` — regenerated CLI reference docs included in this PR
- [x] `bin/gcx kg relabel-rules -h` shows only `get`; `bin/gcx kg relabel-rules get -h` shows `--type` with default `generated`
- [ ] Manual run against a stack: `gcx --context <stack> kg relabel-rules get --type generated -o yaml`


internal/providers/kg/commands.go
```diff
@@ -854,20 +852,28 @@ func (c *SuppressionTableCodec) Decode(_ io.Reader, _ any) error {
...
-	var fileFlag string
-	createCmd := &cobra.Command{
-		Use:   "create",
-		Short: "Upload relabel rules from a YAML file.",
+
+	var (
+		ruleType string
+		io       cmdio.Options
+	)
+	io.DefaultFormat("yaml")
```
radiohead: Can we use `table` by default and `agents` for agent mode?

xujiaxj: Done in 77ed5cc9 — added RelabelRuleTableCodec and switched the default to table. Columns: SOURCE LABELS | TARGET LABEL | ACTION | REGEX | REPLACEMENT. Agent mode auto-flips to agents; use -o yaml for full fidelity (separator, modulus, group-level fields).

xujiaxj: Good point!. We have it now:

image

radiohead (approved):

PR #770, opened 2026-05-23
fix(output): remove agent-mode hint banner that breaks JSON parsing

annanay25:
## Summary

- Remove the `hint: use --json list...` banner emitted to stderr in agent mode — it pollutes combined stdout+stderr in agent environments (Cursor, Claude Code, Copilot), breaking `jq` and `python3 -c json.load()` on every `-o json` response
- Drop the now-unused `agentHintShown` field from `Options` struct
- Fix `TestEncode_AgentModeHint` to capture stderr via `ErrWriter` and assert no hint on either stream (previously only checked stdout, so the leak was invisible)

Agents discover `--json list` via the `--help` tip, the gcx `SKILL.md` output flags table, and the `--json` flag description itself — the runtime hint was redundant.

## Test plan

- [x] `go test ./internal/output/...` passes
- [x] `golangci-lint run ./internal/output/...` clean (pre-existing prealloc warning only)
- [x] `GCX_AGENT_MODE=true bin/gcx providers list -o json 2>&1 | jq '.[0]'` returns valid JSON
- [x] `gofmt -l` reports no formatting issues

### DIFF
--- a/internal/output/format.go
+++ b/internal/output/format.go
@@ -40,11 +40,10 @@ type Options struct {
 	// ErrWriter is the writer for hints and diagnostics (defaults to os.Stderr).
 	ErrWriter io.Writer
 
-	customCodecs        map[string]format.Codec
-	defaultFormat       string
-	flags               *pflag.FlagSet
-	jsonFieldValidator  func(fields []string) error // optional; invoked before field extraction when --json is used
-	jsonFieldsHintShown bool
+	customCodecs       map[string]format.Codec
+	defaultFormat      string
+	flags              *pflag.FlagSet
+	jsonFieldValidator func(fields []string) error // optional; invoked before field extraction when --json is used
 }
 
 // SetJSONFieldValidator registers an optional validator invoked before field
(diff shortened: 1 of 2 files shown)


radiohead:
@annanay25 I don't necessarily think that we should straight up remove this hint. Besides, other commands have their own hints too (e.g. try `gcx irm oncall alert-groups list`). What we could do instead is to make sure that this hint respects the output format (i.e. in agent mode it prints JSON instead).

WDYT?

annanay25:
Thanks for taking a look Igor! In my experience agents almost always discover the `--json` flag by running `help-tree` before running the specific command, so they have it in context already. It seems a bit redundant but maybe we should run this via o11y-bench to prove the point. 

For the others as you mentioned (`gcx irm oncall alert-groups list`, etc) yes I agree the direction should be to make the output JSON compatible.

dafydd-t:
I have also never seen an agent use --json - [see an example o11y-bench run in slack](https://raintank-corp.slack.com/archives/C0AM98F2ZDH/p1780413757891289)

xujiaxj:
Just dropping by to say I like the idea of removing the banner from JSON output. Out of curiosity, I tested this PR against an eval suite (`asserts-eval`), and it helped reduce some parsing friction.

Edit: My workaround now is to explicitly ask Claude to do `2>/dev/null` to not show the stderr, thus not having the hint in the output.

wardbekker:
IIRC root issue fixed by https://github.com/grafana/gcx/pull/683. see also https://raintank-corp.slack.com/archives/C0AHTU3ELH5/p1780393597216409?thread_ts=1780061685.216069&cid=C0AHTU3ELH5

PR #772, opened 2026-05-25
fix: add ReadResponseBody helper to detect oversized responses

imankurpatel000:
@javiermolinar as [promised](https://github.com/grafana/gcx/pull/733#discussion_r3270870432), here is the PR to handle "response too large" for all datasources.

## Summary

Add `httputils.ReadResponseBody` helper that detects oversized responses instead of silently truncating. Wired into all query clients and the datasources client.

- Replaces `io.ReadAll(io.LimitReader(...))` with overflow detection
- Consolidates duplicated `maxResponseBytes` constants into `httputils.DefaultResponseLimit`
- Adds 50 MB limit to Tempo (previously unbounded)
- Removes InfluxDB's local `readLimited` and Infinity/Pyroscope inline overflow checks

Same pattern as [mcp-grafana#884](https://github.com/grafana/mcp-grafana/pull/884).

## Test plan

- [x] `httputils.ReadResponseBody` unit tests (within limit, at limit, overflow, reader error, empty body)
- [x] All query client tests pass
- [x] Lint clean
- [x] Manual verification: overflow error renders correctly in CLI

### DIFF
--- a/internal/datasources/client.go
+++ b/internal/datasources/client.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 
 	"github.com/grafana/gcx/internal/config"
+	"github.com/grafana/gcx/internal/httputils"
 	"k8s.io/client-go/rest"
 )
 
@@ -69,7 +69,7 @@ func (c *Client) List(ctx context.Context) ([]*Datasource, error) {
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes))
+	body, err := httputils.ReadResponseBody(resp.Body, maxResponseBytes)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response: %w", err)
 	}
@@ -108,7 +108,7 @@ func (c *Client) get(ctx context.Context, path, identifier string) (*Datasource,
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes))
+	body, err := httputils.ReadResponseBody(resp.Body, maxResponseBytes)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response: %w", err)
 	}
--- a/internal/httputils/read.go
+++ b/internal/httputils/read.go
@@ -1,0 +1,26 @@
+package httputils
+
+import (
+	"fmt"
+	"io"
+)
+
+// DefaultResponseLimit is the standard response body size cap (50 MB).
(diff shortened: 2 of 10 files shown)


bboreham:
Note that many backends, for instance Prometheus, already have a configurable limit on how much they will return.
This is more efficient as the truncation takes place server-side.

See `limit=` parameter on https://prometheus.io/docs/prometheus/latest/querying/api/

imankurpatel000:
Good point, but this PR doesn't introduce new limits. All these clients already had the same 50 MB cap via io.LimitReader. The difference is that exceeding it previously produced truncated JSON and a confusing parse error. This just refactors to a shared helper and surfaces a clear message instead.

Server-side limits (like Prometheus's `limit=`) are definitely the better first line of defense, gcx doesn't expose that parameter currently, but I believe that should be a separate PR.

wardbekker:
## Shepherd Status

Rebased onto latest `main` and resolved conflicts. Branch is now **mergeable**; CI (`semgrep`) is green.

### Conflict resolution
`main` migrated `clickhouse`, `infinity`, and `influxdb` query clients to the shared `internal/query/grafanaquery` client (per the "datasource query reuse" convention). That shared client **already performs oversized-response detection** (`readLimited`), so this PR's per-client `ReadResponseBody` wiring for those three was superseded — I took `main`'s version for them. The helper remains wired into the clients that talk to product-specific REST APIs directly (`cloudwatch`, `loki`, `prometheus`, `prometheus/series`, `pyroscope`, `tempo`) and the datasources client.

### Added commit
`refactor(grafanaquery)`: routed `grafanaquery`'s own `readLimited` through `httputils.ReadResponseBody`/`DefaultResponseLimit`, completing the PR's "consolidate duplicated `maxResponseBytes`" goal across the whole codebase (a single overflow-detection implementation).

### Verification
- `go build`, `golangci-lint` (0 issues), full `go test` suite — all pass
- Live smoke test against a Grafana Cloud stack: prometheus (instant + range), loki, tempo (search + labels), pyroscope all return valid responses — no regressions

Blocked only on review approval.

---
*Updated by shepherd-pr*

dafydd-t (approved):
LGTM, thanks!

Created an issue to use datasource API `limit` parameters, when they exist [here](https://github.com/grafana/gcx/issues/849)

Also noticed that this error is not surfaced when GCX_AGENT_MODE=true. Created an issue to track that [here](https://github.com/grafana/gcx/issues/855)

dafydd-t:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wardbekker (approved):

PR #779, opened 2026-05-26
remove some small direct dependencies

dafydd-t:
This PR removes some direct dependencies that were easy to replace with internal implementations.

This research doc ([in the PR](https://github.com/grafana/gcx/pull/779/changes#diff-d3bcfa3b678184c70d3f0b2bd124a1b0317ca1b2b94f5cfd7d6712f30ea9a20f)) shows the analysis of each dependency and highlighting those which would be easier to rip out. This PR tackles the 5/6 easiest ones of those.

- remove `huandu/xstrings` - replace with `internal/strcase` (there are 3 case-conversion functions)
- remove `caarlos0/env/v11` - replace with reflection-based env tag parser in `internal/config/envparse.go`
- remove `google/uuid` (direct) - replace with `crypto/rand` UUID generation + hex format validation
- remove `pkg/browser` - replace with `os/exec` platform switch in `internal/deeplink/browser.go`
- remove `adrg/xdg` - replace with `internal/xdg` package that reads env vars directly - I verified this has no impact on the tests
- remove `gofrs/flock` - replace with `syscall.Flock` wrapper in `internal/config/flock.go`

`google/uuid` remains as an indirect dep (pulled by k8s and others).

The binary size remains at ~120MB
```
ll -lah ./bin/gcx
-rwxr-xr-x@ 1 uuu  ggg   120M 26 May 16:24 ./bin/gcx
```


internal/notifier/path_test.go
```diff
@@ -24,8 +20,6 @@ func TestStatePath_RoundTripsStateViaLoadAndSave(t *testing.T) {
 	home := t.TempDir()
 	t.Setenv("HOME", home)
 	t.Setenv("XDG_STATE_HOME", filepath.Join(home, "state"))
-	t.Cleanup(func() { xdg.Reload() })
-	xdg.Reload()
```
dafydd-t, line 28: without `xdg`, only the env vars that are in the test case are available (i.e. set with `t.Setenv`). this seemed like a weird addition anyway

annanay25 (approved):
nice! i was wondering if we should add something into the agents.md to prevent small dependency additions in the future?

dafydd-t:
> nice! i was wondering if we should add something into the agents.md to prevent small dependency additions in the future?

Addressed in https://github.com/grafana/gcx/pull/782!

PR #789, opened 2026-05-28
feat(appo11y): add services list command for OTel target_info inventory

cedricziel:
## Summary

Adds `gcx appo11y services list` — an inventory of the services Grafana Cloud Application Observability has discovered from telemetry. Mirrors the plugin's services scene: instrumented services come from `target_info`, uninstrumented ones from service-graph metrics, namespace is parsed from the `<namespace>/<service>` job convention, and the environment column falls back across the two `deployment.environment[.name]` semconv variants.

Default columns: **NAME · NAMESPACE · ENVIRONMENT · LANGUAGE · STATUS**.

## Flags

- `-d, --datasource <uid>` — Prometheus datasource (auto-resolves from config)
- `--instrumentation all|instrumented|uninstrumented` *(default: all)*
- `--env <name>` — filter by `deployment_environment`
- `--language <name>` — filter by `telemetry_sdk_language`
- `--filter <label><op><value>` — raw PromQL label matchers (repeatable; `=`, `!=`, `=~`, `!~`)
- `--columns a,b,c` — extra `target_info` labels to surface as table columns
- `--limit N` — cap result count
- `--count` — per-language summary instead of the row list
- Output: `-o table | wide | json | yaml`

## Test plan

- [x] Unit tests cover query construction, filter parsing, response parsing, merge, summary aggregation, and the --instrumentation modes
- [x] Lint clean
- [x] Smoke-tested against a live Grafana Cloud stack across all flag combinations
- [x] CI green (linters, tests, documentation, scans)

## Example

```bash
gcx appo11y services list -d grafanacloud-prom --limit 100 
┌───────────────────────────────┬───────────────────────────────┬───────────────────────────────┬───────────────────────────────┬───────────────────────────────┐
│ NAME                          │ NAMESPACE                     │ ENVIRONMENT                   │ LANGUAGE                      │ STATUS                        │
├───────────────────────────────┼───────────────────────────────┼───────────────────────────────┼───────────────────────────────┼───────────────────────────────┤
│ flagd                         │ -                             │ -                             │ go                            │ instrumented                  │
│ frontend-web                  │ -                             │ -                             │ webjs                         │ instrumented                  │
│ imageprovider                 │ -                             │ -                             │ -                             │ instrumented                  │
│ otel-demo-flagd               │ -                             │ -                             │ -                             │ uninstrumented                │
│ otel-demo-valkey:6379         │ -                             │ -                             │ -                             │ uninstrumented                │
│ unknown                       │ -                             │ -                             │ -                             │ uninstrumented                │
│ adservice                     │ oteldemo01                    │ oteldemo01                    │ java                          │ instrumented                  │
│ cartservice                   │ oteldemo01                    │ oteldemo01                    │ dotnet                        │ instrumented                  │
│ checkoutservice               │ oteldemo01                    │ oteldemo01                    │ go                            │ instrumented                  │
│ currencyservice               │ oteldemo01                    │ oteldemo01                    │ cpp                           │ instrumented                  │
│ emailservice                  │ oteldemo01                    │ oteldemo01                    │ ruby                          │ instrumented                  │
│ flagd                         │ oteldemo01                    │ oteldemo01                    │ go                            │ instrumented                  │
│ frauddetectionservice         │ oteldemo01                    │ oteldemo01                    │ java                          │ instrumented                  │
│ frontend                      │ oteldemo01                    │ oteldemo01                    │ nodejs                        │ instrumented                  │
│ frontendproxy                 │ oteldemo01                    │ oteldemo01                    │ -                             │ instrumented                  │
│ kafka                         │ oteldemo01                    │ oteldemo01                    │ java                          │ instrumented                  │
│ loadgenerator                 │ oteldemo01                    │ oteldemo01                    │ python                        │ instrumented                  │
│ paymentservice                │ oteldemo01                    │ oteldemo01                    │ nodejs                        │ instrumented                  │
│ productcatalogservice         │ oteldemo01                    │ oteldemo01                    │ go                            │ instrumented                  │
│ quoteservice                  │ oteldemo01                    │ oteldemo01                    │ php                           │ instrumented                  │
│ recommendationservice         │ oteldemo01                    │ oteldemo01                    │ python                        │ instrumented                  │
│ shippingservice               │ oteldemo01                    │ oteldemo01                    │ rust                          │ instrumented                  │
└───────────────────────────────┴───────────────────────────────┴───────────────────────────────┴───────────────────────────────┴───────────────────────────────┘
```


internal/providers/appo11y/services/query.go
```diff
@@ -0,0 +1,328 @@
...
+	case "!~":
+		return v.LabelNotMatchRegexp(m.Label, m.Value)
+	default: // "="
+		return v.Label(m.Label, m.Value)
+	}
+}
+
+// buildServicesQuery returns a PromQL expression that groups the target_info
+// inventory. It mirrors the plugin query at
+// plugin/src/modules/services/utils/servicesQueryBuilder.ts, expanded to
```
annanay25: lets remove this reference

internal/providers/appo11y/services/commands.go
```diff
@@ -0,0 +1,404 @@
...
+			}
+			resp, err := client.Query(ctx, datasourceUID, prometheus.QueryRequest{Query: expr})
+			if err != nil {
+				return fmt.Errorf("service-graph query failed: %w", err)
+			}
+			graph, err = parseServiceGraphResponse(resp)
+			if err != nil {
+				return fmt.Errorf("failed to parse service-graph response: %w", err)
+			}
+		}
```
annanay25: can these be run in parallel? doesn't look like the two queries depend on each other

annanay25 (commented):
lgtm overall, left 2 minor comments.

internal/providers/appo11y/services/query.go
```diff
@@ -0,0 +1,328 @@
...
+	"strings"
+
+	"github.com/grafana/gcx/internal/query/prometheus"
+	"github.com/grafana/promql-builder/go/promql"
+)
+
+// defaultTargetInfoMetric is the canonical OpenTelemetry target_info metric.
+// The app-observability-app plugin abstracts this behind `${metricName:targetInfo}`
+// to support alternative metric modes; for now gcx hardcodes the default.
+const defaultTargetInfoMetric = "target_info"
```
javiruiz01: No `traces_target_info` support? It's fine adding it in a different PR, just highlighting it in case you forgot about it

internal/providers/appo11y/services/query.go
```diff
@@ -0,0 +1,328 @@
...
+// buildServiceGraphQuery returns a PromQL expression that lists every service
+// observed as a `server` in the Tempo service-graph metric, projecting
+// server_service_namespace when present (the plugin uses this label as the
+// uninstrumented-side namespace — see ServicesScene.tsx:622-648 fallback).
+// metric defaults to "traces_service_graph_request_total".
+func buildServiceGraphQuery(metric string) (string, error) {
+	if metric == "" {
+		metric = defaultServiceGraphMetric
+	}
+	expr, err := promql.Group(promql.Vector(metric)).By([]string{"server", "server_service_namespace"}).Build()
```
javiruiz01: If we're using this query for checking which services are uninstrumented, I think we should add a filter for `connection_type!=''`.

javiruiz01 (dismissed):
Love it, thanks a lot! Just a couple of comments, but I think it's looking good!

internal/providers/appo11y/services/commands.go
cedricziel: Done in d9ae4bfe — both queries now run in parallel via errgroup. As part of that the target_info query also runs in `--instrumentation uninstrumented` mode, since the service-graph results have to be diffed against it to be meaningful.

internal/providers/appo11y/services/query.go
cedricziel: Done in d9ae4bfe — scrubbed the file paths; rationale kept inline.

internal/providers/appo11y/services/query.go
cedricziel: Done in d9ae4bfe — added `connection_type!=""` to the service-graph query. On this stack it drops the uninstrumented set from 19 to 3, which matches reality.

internal/providers/appo11y/services/query.go
cedricziel: Good catch — not in this PR. Filed for follow-up: make `--target-info-metric` accept `traces_target_info` cleanly or add a first-class flag.

javiruiz01 (dismissed):
🚀

internal/providers/appo11y/services/query.go
cedricziel: Added `traces_target_info` support in 8c7c642a — now unions it with `target_info` by default. Bumped the inventory from 19 to 22 on the demo stack (3 services were only in `traces_target_info`). Both queries run in parallel via the existing errgroup; dedup is handled by the existing (namespace, name, language) merge in `parseServicesResponse`. I also dropped the `--target-info-metric` flag since it's now misleading — easy to add back if there's a use case.

cedricziel: @javiruiz01 I snuck in traces_target_info support

javiruiz01 (dismissed):
Love it, thanks for adding support for `traces_target_info` as well!

internal/providers/appo11y/services/commands.go
```diff
@@ -0,0 +1,413 @@
...
+		if err := eg.Wait(); err != nil {
+			return err
+		}
+
+		instrumented, err := parseServicesResponses(instrumentedResponses)
+		if err != nil {
+			return fmt.Errorf("failed to parse services response: %w", err)
+		}
+
+		items := resolveItems(opts.Instrumentation, instrumented, graph)
```
radiohead: The filtered `instrumented` set is used as both the display rows and the baseline for the uninstrumented diff. With e.g. `--language go` only Go services are in `instrumented` so `instrumentedIndex` ends up missing every non-Go instrumented service. Those services appear in the service graph diff and get classified as `uninstrumented`.

I suggest running a separate unfiltered `target_info` query to build `instrumentedIndex` and then apply user filters only to the displayed instrumented rows.

internal/providers/appo11y/services/commands.go
```diff
@@ -0,0 +1,413 @@
...
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, parsed)
+	}
+	if o.Language != "" {
+		out = append(out, Matcher{Label: "telemetry_sdk_language", Op: "=", Value: o.Language})
+	}
+	if o.Env != "" {
+		out = append(out, Matcher{Label: "deployment_environment", Op: "=", Value: o.Env})
```
radiohead: Looks like `buildFilters` will create a matcher only on `deployment_environment` but there's also a fallback to `deployment_environment_name` for newer-SDK stacks – those will end up showing an environment value in the ENVIRONMENT column but pass through env filtering unconditionally.

I suggest applying the env filter as a post-parse predicate over `environmentValue(s.Labels)` rather than a PromQL matcher, so that the filter and the display semantics match.

internal/providers/appo11y/services/query.go
```diff
@@ -0,0 +1,341 @@
...
+func parseFilter(raw string) (Matcher, error) {
+	m := matcherPattern.FindStringSubmatch(raw)
+	if m == nil {
+		return Matcher{}, fmt.Errorf("invalid --filter %q: expected <label><op><value> where op is = != =~ !~", raw)
+	}
+	label, op, val := m[1], m[2], m[3]
+	if len(val) >= 2 && val[0] == '"' && val[len(val)-1] == '"' {
+		val = val[1 : len(val)-1]
+	}
+	return Matcher{Label: label, Op: op, Value: val}, nil
```
radiohead, line 131: IIUC this will allow PromQL query injection – we'll need to do some char stripping.

Something like this:
```go
func escapePromql(v string) string {
    v = strings.ReplaceAll(v, `\`, `\\`)
    v = strings.ReplaceAll(v, `"`, `\"`)
    return v
}
```

internal/providers/appo11y/services/commands.go
```diff
@@ -0,0 +1,413 @@
...
+
+		instrumented, err := parseServicesResponses(instrumentedResponses)
+		if err != nil {
+			return fmt.Errorf("failed to parse services response: %w", err)
+		}
+
+		items := resolveItems(opts.Instrumentation, instrumented, graph)
+
+		if opts.Limit > 0 && len(items) > opts.Limit {
+			items = items[:opts.Limit]
```
radiohead, line 309: nit – `--count --limit=10` will end up reporting counts for the first 10 services, which feels weird. I suggest adding a check for `--count --limit` and returning an error.

Side-note, I also think we should apply a default limit value (IIRC most of the list commands use `--limit=50` by default) and emit a hint to STDERR for fetching more, similar to e.g. `gcx irm oncall alert-groups list`.

internal/providers/appo11y/services/commands.go
```diff
@@ -0,0 +1,413 @@
...
+	out := make([]Matcher, 0, len(o.Filters)+2)
+	for _, f := range o.Filters {
+		parsed, err := parseFilter(f)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, parsed)
+	}
+	if o.Language != "" {
+		out = append(out, Matcher{Label: "telemetry_sdk_language", Op: "=", Value: o.Language})
```
radiohead, line 125: question / nit – one could do something like `--language=go --filter 'telemetry_sdk_language=java'` – what should be the expected output here?

radiohead (commented):
Overall LGTM! But I did find a couple of issues and a few smaller nits / questions.

internal/providers/appo11y/services/commands.go
cedricziel: Fixed in 44f884f3 — `resolveItems` and `mergeServiceSets` now take a separate `baseline` slice for the diff index. When user filters are set and the diff is needed we issue an extra unfiltered target_info pair in parallel (skipped in the common no-filters case so the cost is opt-in). Added a regression test that asserts non-display instrumented services don't leak into the uninstrumented bucket.

internal/providers/appo11y/services/commands.go
cedricziel: Done in 44f884f3 — `--env` is no longer a PromQL matcher. It's now a post-parse predicate using the same `environmentValue` helper as the column renderer, so the filter and the displayed value always agree. The buildFilters test was updated to reflect that env doesn't emit a matcher anymore.

internal/providers/appo11y/services/query.go
cedricziel: Confirmed: I ran a quick local probe and promql-builder does indeed embed values verbatim — `bar"; drop_table--` came out as `foo="bar"; drop_table--"`. Added `escapePromqlValue` (backslash first, then quote) and apply it in `Matcher.apply` for all four operators. Regression test added with embedded quote + backslash payloads.

internal/providers/appo11y/services/commands.go
cedricziel: Done in 44f884f3 — `--count` plus an explicit `--limit > 0` now errors out (a per-language summary over a truncated row set would lie about totals). When the user only sets `--count`, the new default limit of 50 is silently dropped so the summary covers the full inventory. Also added the truncation hint via `cmdio.EmitHint` mirroring the IRM alert-groups list pattern: stderr gets `hint: showing first N services: gcx appo11y services list --limit M`.

internal/providers/appo11y/services/commands.go
cedricziel: Current behaviour: both filters AND together, so `--language=go --filter telemetry_sdk_language=java` returns the empty intersection. I considered detecting that specific case but leaning toward leaving it: `--language` is just sugar for the matcher, and the AND semantics match how PromQL itself composes matchers (and what users get if they pass two `--filter` args). Happy to add a warn-on-conflict in a follow-up if you'd rather.

wardbekker (dismissed):

internal/providers/appo11y/services/commands.go
radiohead: I think adding a warn-on-conflict would be nice. Not a blocking issue, feel free to raise a follow-up for this.

radiohead (approved):

PR #790, opened 2026-05-28
tweaks to dashboards skills

dafydd-t:
tweak the create-dashboard and manage-dashboard skills to try and:

- make choosing one skill vs the other more obvious (add more guidance to the description)
- add guidance on how to look for labels or metadata in datasources, so it avoids falling back to `gcx api` or curl
- know how to list datasources - `gcx datasources list` is more unversal than `gcx resources get datasources`
- Get rid of some cruft from the manage-dashboards skill in particular

These are bases on o11y-bench runs using the gcx agent, which I'll post next week


javiermolinar (approved):
LGTM

PR #795, opened 2026-05-29
ci: harden Claude workflows' Vault usage (export_env: false + author guard)

awsome-o:
## Summary

Two CI-only hardening changes across the workflows that pull secrets from Vault via `grafana/shared-workflows/actions/get-vault-secrets` — the two Claude workflows in this repo (`claude.yml`, `claude-code-review.yml`):

- **`export_env: false`** on every `get-vault-secrets` call. The action defaults to `export_env: true`, which writes each fetched secret into `$GITHUB_ENV` so it's readable by *every* later step in the job (including the third-party `anthropics/claude-code-action`). Both workflows now read the secret from the step output via `fromJSON(steps.get-secrets.outputs.secrets).ANTHROPIC_API_KEY` instead of `${{ env.ANTHROPIC_API_KEY }}`, so the global env export is no longer needed — this removes it and keeps the secret scoped to the step output.
- **`author_association` guard** (`OWNER`/`MEMBER`/`COLLABORATOR`) on the comment/review/issue-triggered workflow that fetches Vault secrets *and* runs Claude: `claude.yml`. It fires on four event types, so each branch is guarded with its own association field (`comment`/`comment`/`review`/`issue`). The existing `@claude` body/title checks (including this repo's issue-title check) are preserved.

## Why

`get-vault-secrets` defaulting to `export_env: true` widens every fetched secret's blast radius to the whole job; scoping it to the step output is least-privilege with no functional change.

`claude.yml` previously gated only on the comment/issue *body* (e.g. `contains(comment.body, '@claude')`), so **any commenter** could trigger a job that mints a Vault secret and runs Claude with `contents`/`pull-requests`/`issues: write`. The guard restricts that to trusted repo associations. **`grafana/gcx` is a public repo**, so this matters more here than in an internal repo: the guard is transparent for org members / collaborators and only blocks bots, non-members, and former contributors.

`claude-code-review.yml` is triggered by `pull_request` (no `pull_request_target`). GitHub does not pass secrets to `pull_request` runs from forks and Vault OIDC rejects fork refs, so an untrusted fork PR cannot fetch the secret anyway — it gets only the `export_env: false` change, and keeps its existing `user.type != 'Bot'` + `draft == false` guards.

Applies the pattern from grafana/grafana-assistant-app#7020. Part of grafana/grafana-assistant-app#6947 (Security Hardening Week — GitHub Actions guardrails; the scope list includes `grafana/gcx`).

## Test plan

- [x] `actionlint` (1.7.12) clean on both edited workflows. The only repo-wide finding is a pre-existing, info-level shellcheck note in the untouched `publish-homebrew-formula.yml` (intentional `envsubst '$VAR'` single-quoting) — unrelated.
- [ ] **`export_env: false` — post-merge:** as a member, `@claude …` on an issue/PR and confirm Claude still authenticates, validating the secret resolves via `fromJSON(steps.get-secrets.outputs.secrets)` with the env export removed. For `claude-code-review.yml`, open/ready a non-draft PR from a repo branch and confirm the review still runs and authenticates.
- [ ] **Guard — post-merge:** `issue_comment`/`issues` definitions load from the default branch, so after merge, comment `@claude …` as a member and confirm the job runs (not skipped); confirm a non-member/external `@claude` comment is skipped.
- [ ] No behavior change for members; the guard only blocks non-/former-members and bots.

## Scope / follow-ups

CI config only — no product code. Fork-PR guards are intentionally deferred: Vault OIDC already rejects fork `pull_request` events, and `issue_comment` runs in base-repo context (a fork guard wouldn't cover the comment path anyway).

### DIFF
--- a/claude-plugin/skills/debug-with-grafana/SKILL.md
+++ b/claude-plugin/skills/debug-with-grafana/SKILL.md
@@ -278,11 +278,9 @@ gcx traces query -d <tempo-uid> \
   '{ resource.service.name = "<service-name>" && duration > 1s }' \
   --from now-1h --to now
 
-# Fetch a specific trace by ID (from search results or log trace IDs)
-gcx traces get -d <tempo-uid> <trace-id>
-
-# LLM-friendly trace output for analysis
-gcx traces get -d <tempo-uid> <trace-id> --llm
+# Fetch a specific trace by ID for analysis (from search results or log trace IDs)
+# Always use --llm so Tempo returns its token-efficient LLM trace encoding.
+gcx traces get -d <tempo-uid> <trace-id> --llm -o json
 ```
 
 **TraceQL attribute scoping**: Tempo requires scoped attribute names. Use
@@ -293,6 +291,10 @@ gcx traces get -d <tempo-uid> <trace-id> --llm
 Use `name` (unscoped) for the span name, `duration` for span duration,
 and `status` for span status. Use `trace:rootService` and `trace:rootName`
 for root span attributes (not `rootServiceName` or `rootTraceName`).
+
+When inspecting trace bodies, use `gcx traces get <trace-id> --llm -o json`. Do not fetch the
+OTLP-shaped default trace and manually compact it unless the user explicitly
+needs raw trace JSON for schema/debugging work.
 
 Discover available labels:
 ```bash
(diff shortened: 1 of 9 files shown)


annanay25 (approved):

awsome-o:
missing signing, will do it here: https://github.com/grafana/gcx/pull/803

PR #801, opened 2026-06-01
feat(assistant): add conversation list/get and relax context resume

svennergr:
## Summary

Makes it easy to hand off a Grafana Assistant web conversation to a coding agent (or continue it in gcx):

- **`gcx assistant conversation list`** — discover conversation IDs. Defaults `--source` to `assistant,slack,cli` (filtered server-side, so `--limit`/`--offset` stay honest), with `all`/comma-list, pagination, and archived filters. Output: table (default), wide, or json.
- **`gcx assistant conversation get <id>`** — pull a conversation transcript as plain text (or json) to paste into a coding agent before continuing.
- **Relaxed `--context-id` resume** — instead of rejecting chats not created by the CLI, gcx now resumes any accessible chat and prints a soft notice for non-CLI sources, matching backend behavior (the backend never filtered on source).
- Ephemeral `inline-assistant` chats are excluded from listings.

## Why

Previously gcx defensively blocked resuming non-CLI chats even though the backend allowed it, and there was no way to discover conversation IDs or pull a transcript from the CLI. This rounds out the web → CLI/coding-agent handoff workflow.

## Test plan

- [x] `go test ./internal/assistant/... ./cmd/gcx/assistant/...` passes
- [x] `go build ./cmd/gcx/` succeeds
- [ ] Manual: `gcx assistant conversation list` shows the same conversations as the web sidebar
- [ ] Manual: `gcx assistant conversation get <id>` renders a readable transcript
- [ ] Manual: `gcx assistant prompt --context-id <web-uuid>` resumes a web chat with a notice

### DIFF
--- a/cmd/gcx/assistant/command.go
+++ b/cmd/gcx/assistant/command.go
@@ -78,6 +78,7 @@ Service account tokens are not supported.`,
 	configOpts.BindFlags(cmd.PersistentFlags())
 	cmd.AddCommand(promptCommand(configOpts))
 	cmd.AddCommand(dashboardCommand(configOpts))
+	cmd.AddCommand(conversationCommand(configOpts))
 
 	// Create a ConfigLoader for investigations that shares the same --config/--context
 	// flags already bound by configOpts. Wire the values via PersistentPreRunE so that
@@ -260,11 +261,15 @@ func runPrompt(cmd *cobra.Command, message string, opts *promptOpts, configOpts 
 
 	// Validate context ID if provided
 	if contextID != "" {
-		if err := c.ValidateCLIContext(ctx, contextID); err != nil {
+		notice, err := c.ValidateCLIContext(ctx, contextID)
+		if err != nil {
 			if opts.jsonOut {
 				return jsonError(err)
 			}
 			return err
+		}
+		if notice != "" && !opts.jsonOut {
+			cmdio.Info(errW, "%s", notice)
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


annanay25:
**`--limit`/`--offset` honesty for `--source all`** (re: `internal/assistant/api.go` L172-178)

The client-side `inline-assistant` filter runs *after* the server applies `limit`/`offset`, so for `--source all` a page can return fewer rows than `--limit` even when more non-inline chats exist on later offsets — i.e. the "keeps `--limit`/`--offset` honest" claim doesn't hold for `all`. It's a non-issue for the default and explicit comma-lists (inline is never requested, so the loop is just a safety net); only `all` is affected.

I'd avoid asking the backend for an `excludeSource` param — not worth coupling gcx to a backend change for this marginal case. That leaves two client-side options:

1. **Expand `all` into an explicit non-inline source list** (e.g. `assistant,slack,cli,a2a`) before sending the request. Pagination becomes honest; the drop loop stays as a safety net. Cost: brittle — a new backend source won't show under `all` until the constant is updated.

2. **Just document it** — keep the client-side filter, scope the "honest pagination" claim in the doc comment to non-`all` sources, and leave `all` as best-effort.

Leaning toward option 2: the default is already honest, `--source all` is the power-user/debug path, and silently hiding a future source under `all` (option 1) is arguably worse than a slightly short page on an explicit `all`.

svennergr:
> Leaning toward option 2: the default is already honest, --source all is the power-user/debug path, and silently hiding a future source under all (option 1) is arguably worse than a slightly short page on an explicit all.


Thanks @annanay25 - agree. Done in 4b60b71!

annanay25 (approved):

PR #802, opened 2026-06-01
feat(tempo): add LLM tag values output

javiermolinar:
When using the header` Accept: application/vnd.grafana.llm` Tempo returns a simplified JSON format optimized for LLM consumption. 

This is around ~49.8% smaller than regular response:

 ```json
   {
     "tagValues": [
       {
         "type": "string",
         "value": "gme-ingester"
       },
       {
         "type": "string",
         "value": "grafana-security-docs"
       },
       {
         "type": "string",
         "value": "cloudconfig.grafana.app"
       }
     ]
   }
 ```

 With --llm:

 ```json
   {
     "metrics": {
       "inspectedBytes": 485118972
     },
     "tagValues": {
       "string": [
         "gme-ingester",
         "grafana-security-docs",
         "cloudconfig.grafana.app"
       ]
     }
   }
```

This PR enables the --llm command in the same way we already do for the traces get command. It also improves existing skills to favor it in agentic workflows

### DIFF
--- a/claude-plugin/skills/create-dashboard/SKILL.md
+++ b/claude-plugin/skills/create-dashboard/SKILL.md
@@ -121,7 +121,7 @@ gcx datasources list -o json
 |------|-----------|---------------------------|
 | Prometheus | `gcx datasources prometheus labels -d <uid> --label __name__` | `gcx datasources prometheus labels -d <uid> --label <label>` |
 | Loki | `gcx datasources loki labels -d <uid>` | `gcx datasources loki labels -d <uid> --label <label>` |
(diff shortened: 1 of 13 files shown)


radiohead (approved):
@javiermolinar could we set LLM mode automatically based on the agent detection mechanism?

Otherwise looks good.

javiermolinar:
> @javiermolinar could we set LLM mode automatically based on the agent detection mechanism?
> 
> Otherwise looks good.

I will try it for a week or two. If agents keep using the canonical form I will change it.

PR #809, opened 2026-06-03
Fix: dashboard pagination

javiermolinar:
The command `dashboard list` has several problems:

- Default limit (500) is huge, leading to extremely slow respose
- Pagination is supported by the endpoint but not used by GCX

This PR enables pagination, allowing GCX to use the next token to iterate over the next set of dashboards.

The default limit is 50. If there are more dashboards to list the user will get a hint

`hint: showing 10 dashboards; more pages are available: gcx dashboards list --limit 10 --continue eyJuIjoiLVhmekJ2Z0d6IiwidiI6MjA2MjExMDM1MjIyMDM1NzQ2OH0=`

in the json mode:

```
 "metadata": {
    "continue": "eyJuIjoiMDAwMDAwMTkxIiwidiI6MjA2MjEwNsU4OTA3ODIyNTQ2OH0=",
    "resourceVersion": "2062106589078225468"
  },
```

It also fix:

- -o json  was returning the whole dashboard. Now it returns a summary. This summary is also used in the table view
- Instead of showing the dashboard name it also show the dashboard path


internal/providers/dashboards/codec.go
```diff
@@ -91,38 +90,85 @@ func (c *dashboardTableCodec) Encode(w io.Writer, v any) error {
...
+			return &dashboardSummaryList{Kind: "DashboardSummaryList", Items: []dashboardSummary{}}, nil
+		}
+		return val, nil
+	case dashboardSummaryList:
+		return &val, nil
+	case *dashboardSummary:
+		if val == nil {
+			return &dashboardSummaryList{Kind: "DashboardSummaryList", Items: []dashboardSummary{}}, nil
+		}
+		return single(*val), nil
```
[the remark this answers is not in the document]
javiermolinar: Fair enough

internal/providers/dashboards/folder_paths.go
```diff
@@ -0,0 +1,111 @@
...
+	ParentUID string
+}
+
+func loadFolderPaths(ctx context.Context, cfg config.NamespacedRESTConfig, client *dynamic.NamespacedClient) (map[string]string, error) {
+	desc, err := resolveFoldersDescriptor(ctx, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	list, err := client.List(ctx, desc, metav1.ListOptions{})
```
radiohead: This is potentially problematic – we are listing all folders every time we are listing dashboards, the APIs won't be happy about it.

I suggest keeping folders as-is for now, or alternatively we could build some sort of on-disk cache with folder names / paths.

internal/providers/dashboards/summary.go
```diff
@@ -0,0 +1,107 @@
...
+	FolderUID  string   `json:"folderUID,omitempty"`
+	Tags       []string `json:"tags,omitempty"`
+	PanelCount *int     `json:"panelCount,omitempty"`
+}
+
+func dashboardListSummary(list *unstructured.UnstructuredList) *dashboardSummaryList {
+	return dashboardListSummaryWithFolderPaths(list, nil)
+}
+
+func dashboardListSummaryWithFolderPaths(list *unstructured.UnstructuredList, folderPaths map[string]string) *dashboardSummaryList {
```
radiohead: I don't agree with the change to summarize `gcx dashboards list -ojson` here. Table / wide output is for summaries – JSON / YAML should use full data. We could add a `-osummary` for summarized JSON output if we wanted to, but I am not 100% sure it'd be useful.

radiohead (commented):
Nice work! A couple of comments though – I don't think we should be summarizing JSON outputs and rendering folder paths is potentially very heavy for the API.

internal/providers/dashboards/summary.go
javiermolinar: agree, we can leave it like this for now

internal/providers/dashboards/folder_paths.go
wardbekker: Reverted folder-path resolution per your suggestion to keep folders as-is for now. The FOLDER column again shows the folder UID directly (empty → "General") from the `grafana.app/folder` annotation, so `dashboards list` no longer lists all folders. `folder_paths.go` and its wiring are removed. An on-disk folder name/path cache can be a follow-up if we want readable paths later. Pagination and the JSON-no-summary fixes are retained.

wardbekker:

### Review Feedback
| Thread | Status |
|--------|--------|
| `claude[bot]` — nil guards in `codec.go` | ✅ Resolved (prior commit) |
| `claude[bot]` — nil guards in `crud.go` | ✅ Resolved (prior commit) |
| `radiohead` — JSON summarization (`summary.go`) | ✅ Resolved (prior commit) |
| `radiohead` — listing all folders (`folder_paths.go`) | ✅ Resolved this iteration |

### Change Applied
- `revert(dashboards): drop folder-path resolution from list output` (87be3954)
- Reverted folder-path rendering per @radiohead — `dashboards list` no longer lists all folders on every call. FOLDER column shows the folder UID directly (empty → `General`), matching prior behaviour.
- `folder_paths.go` + wiring removed (214 deletions). Pagination and JSON-no-summary fixes retained.
- Local quality gate green: `lint` clean, full `go test ./...` passes, reference docs no drift.

### Blocking
- Awaiting an **approving review** (`reviewDecision` empty, `mergeState` BLOCKED). All review threads now resolved.

---
*Updated by shepherd-pr*

wardbekker (approved):

dafydd-t (approved):

PR #810, opened 2026-06-03
feat(skills): add 'gcx agent skills get' and surface bundled skills in help/hints

alexander-akhmetov:
When `gcx` is used without installed skills, agents need to figure out how to use it. They can check `help` and examples, but we also already have useful skills that could be used too.

This PR adds a `gcx agent skills get <name>` command that prints a bundled `SKILL.md` without installing it, and show the available skills through the help banner, command catalog, and agent hint annotations. With this the agent while, for example, listing the dashboards, gets a hint that the dashboard-related skills can be retrieved and does so when needed.

### DIFF
--- a/internal/style/help.go
+++ b/internal/style/help.go
@@ -5,11 +5,26 @@ import (
 	"strings"
 
 	"github.com/charmbracelet/glamour"
+	"github.com/grafana/gcx/internal/agent"
 	"github.com/spf13/cobra"
 )
 
 // jsonDiscoveryTip is the help text shown for commands that support --json field selection.
 const jsonDiscoveryTip = "Use --json list to discover available fields, --json field1,field2 to select specific fields."
+
+// relatedSkillFooter returns the "Related skill" footer lines for a command, or
+// nil when no skill is mapped to the command or its ancestors.
+func relatedSkillFooter(cmd *cobra.Command) []string {
+	skills := agent.SkillsForCommand(cmd)
+	if len(skills) == 0 {
+		return nil
+	}
+	lines := []string{"Related skill(s): " + strings.Join(skills, ", ")}
+	for _, name := range skills {
+		lines = append(lines, "  Load with: gcx agent skills get "+name)
+	}
+	return lines
+}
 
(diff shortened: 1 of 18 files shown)


alexander-akhmetov:
@claude review

alexander-akhmetov:
Maybe not only the name, but also the skill description should be in the hint too, not sure.

alexander-akhmetov:
I ran a few tests with https://o11ybench.ai/ and added skill description too. Without it, models (especially haiku, opus-4-8 is fine and isn't affected) load heavy skills like `debug-with-grafana` even for simple tasks when it can be done without them and faster.

internal/style/help.go
```diff
@@ -5,26 +5,54 @@ import (
...
+	}
+	lines := []string{"Related skill(s):"}
+	for _, name := range skills {
+		if desc := skillops.ShortDescription(claudeplugin.SkillsFS(), name); desc != "" {
+			lines = append(lines, "  "+name+" — "+desc)
+		} else {
+			lines = append(lines, "  "+name)
+		}
+	}
+	lines = append(lines, "Read a skill only if it fits your task: gcx agent skills get <name>")
```
radiohead: A couple of small comments from smoke-testing this:

1. Let's add `-otext` (since otherwise the output will default to `-oagent` and will overflow).
2. Let's explicitly mention that they don't need to be installed.

Something like this:
```suggestion
	lines = append(
		lines,
		"Read a skill only if it fits your task",
		"These skills are bundled — no install needed",
		"gcx agent skills get <name> -otext",
	)
```

radiohead (dismissed):
LGTM! Just one small issue with the encoding / hint wording from the smoke test. Not a blocker.

internal/style/help.go
alexander-akhmetov: Makes sense, updated https://github.com/grafana/gcx/pull/810/commits/b4bc34c42743bf726e5401ce41a08fc09a2f7779

dafydd-t (approved):

PR #812, opened 2026-06-04
feat(aio11y): add `gcx aio11y login` to provision sigil credentials

svennergr:
Adds a one-command setup for the Grafana AI Observability coding-agent plugins (Cursor, Claude Code, Codex, Copilot, OpenCode, pi). It resolves the Sigil endpoint, OTLP gateway, and tenant ID from the current Cloud context and writes `~/.config/sigil/config.env`.

By default it mints a dedicated, correctly-scoped Cloud Access Policy token (`sigil:write`, `metrics:write`, `traces:write`) realm-scoped to the stack, reusing the access policy and rotating the token on re-runs. When the context token lacks `accesspolicies:write` it falls back to writing the context token and prints guidance. Flags: `--no-provision`, `--token`, `--token-name`, `--token-expiry`, `--content-capture-mode`, `--dry-run`.

Adds GCOM access-policy/token client methods (`Create/List` access policies, `Create/List/Delete` tokens) and extends `LoadCloudConnections` to return the authenticated GCOM client.

### Reviewer notes

- Main behavior lives under `internal/providers/aio11y/login`; the Cloud changes are small GCOM access-policy/token client helpers.
- Token secrets are written only to `config.env`; command output reports the source/metadata and intentionally does not echo token values.
- Re-run behavior intentionally reuses the stack-scoped access policy and rotates a matching token name.
- The generated CLI reference page is included because this adds a new command.

### How to test / review

- Focused tests: `go test ./internal/providers/aio11y/login ./internal/cloud ./cmd/gcx/root`
- Full local gates: `GCX_AGENT_MODE=false mise run all` and `GCX_AGENT_MODE=false mise run reference-drift`
- Manual smoke: start with `gcx aio11y login --dry-run -o json`; use `--config-path <tmp>/config.env` for live testing so you do not touch your real sigil config.

### DIFF
--- a/docs/reference/cli/gcx_aio11y.md
+++ b/docs/reference/cli/gcx_aio11y.md
@@ -31,6 +31,7 @@ Manage Grafana AI Observability resources
 * [gcx aio11y generations](gcx_aio11y_generations.md)	 - Inspect individual LLM generations.
 * [gcx aio11y guards](gcx_aio11y_guards.md)	 - Manage synchronous policy guards (hook rules) that evaluate generations on the request path.
 * [gcx aio11y judge](gcx_aio11y_judge.md)	 - List LLM providers and models available for LLM-judge evaluators.
+* [gcx aio11y login](gcx_aio11y_login.md)	 - Provision sigil credentials for coding-agent plugins from the current context.
 * [gcx aio11y rules](gcx_aio11y_rules.md)	 - Manage rules that route generations to evaluators.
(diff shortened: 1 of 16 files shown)


alexander-akhmetov:
What do you think about making this a bit simpler in `gcx` and exposing a few public methods so that we can import it in sigil CLI and reuse its login flow? It already knows where to save the config file, and we will benefit from having one command to run agents: `sigil {agent}`. It currently opens a login window on the first run, and we could make it simpler with this change.

svennergr:
[codex] Closing this stale draft because review redirected the login flow toward a different shared architecture. Any follow-up should start from that current direction.

PR #823, opened 2026-06-08
feat(faro): add session replay inspection commands

d0ugal:
## Summary

- Adds `list-replay-sessions`, `inspect-replay-session`, and `inspect-replay-segment` under `gcx frontend apps` for inspecting Faro session replay data
- `list-replay-sessions` queries Loki for `faro.session_recording.started` events to discover regular session IDs with replay recordings; it does not list all Frontend Observability sessions
- `inspect-replay-session` lists replay recordings attached to a regular session ID, enriching each with manifest data (segment count, inactivity periods) via bounded errgroup concurrency
- `inspect-replay-segment` inspects individual replay segment events with a table summary, JSON output, or file save mode; supports `--recording-id` for multi-recording sessions

Includes client methods (`ListRecordings`, `GetManifest`, `GetSegment`), response types, output codecs, agent annotations, generated CLI reference, and tests.

A follow-up PR will add `play-session` (local rrweb-player server) — split out for separate review due to security surface (local HTTP server, CDN dependencies, WebSocket control channel).

## Test plan

- [x] `go test ./internal/providers/faro/... -race`
- [x] `go test ./cmd/gcx/root -run TestConsistency_AllLeafCommandsHaveOutputClass -count=1`
- [x] `GCX_AGENT_MODE=false mise run lint`
- [x] `GCX_AGENT_MODE=false mise run reference-drift`
- [ ] Manual test: `gcx frontend apps list-replay-sessions <app>` against a stack with replay data
- [ ] Manual test: `gcx frontend apps inspect-replay-session <app> <session-id>`
- [ ] Manual test: `gcx frontend apps inspect-replay-segment <app> <session-id> <segment-id> -o json`
- [ ] Manual test: `gcx frontend apps inspect-replay-segment <app> <session-id> <segment-id> --save events.json`

Local `GCX_AGENT_MODE=false mise run docs` regenerates references, then fails in this environment because MkDocs cannot import `material.extensions.emoji`; the GitHub Documentation check runs the full docs build with CI dependencies.


internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+// ---------------------------------------------------------------------------
+
+type inspectReplaySegmentOpts struct {
+	IO          cmdio.Options
+	Save        string
+	RecordingID string
+}
+
+func (o *inspectReplaySegmentOpts) setup(flags *pflag.FlagSet) {
+	flags.StringVar(&o.Save, "save", "", "Write full replay event JSON to a file")
```
dafydd-t, line 307: Do we need this, or is there a precedent for this already? I see it does some marshalling with indentation, but nothing special. could users rely on a simple redirect e.g. `> result.json` instead?

dafydd-t: --save also overrides `-o`, so either error if both are set,  or respect it

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+
+			segment, err := client.GetSegment(ctx, appID, sessionID, recordingID, segmentID)
+			if err != nil {
+				return err
+			}
+
+			// Warn if segment has dependency.
+			manifest, manifestErr := client.GetManifest(ctx, appID, sessionID, recordingID)
+			if manifestErr == nil {
+				segID, _ := strconv.ParseInt(segmentID, 10, 64)
```
dafydd-t, line 369: what if the segmentID positional arg cannot be parsed to an int? `segID` would be 0 in that case

dafydd-t: maybe its worth not discarding the error here?

docs/reference/cli/gcx_frontend_apps_inspect-replay-session.md
```diff
@@ -0,0 +1,47 @@
+## gcx frontend apps inspect-replay-session
```
dafydd-t, line 1: this lists session recordings, so would e.g `list-replay-recordings` be a more accurate name? https://github.com/grafana/gcx/blob/main/docs/design/command-naming.md#place-each-operation-by-the-identity-it-requires

internal/providers/faro/session_list.go
```diff
@@ -0,0 +1,249 @@
...
+func parseNanoTimestamp(s string) time.Time {
+	ns, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return time.Time{}
+	}
+	return time.Unix(0, ns)
+}
+
+// parseDuration parses durations like "1h", "24h", "7d".
+func parseDuration(s string) (time.Duration, error) {
```
dafydd-t, line 240: can we use https://github.com/grafana/gcx/blob/main/internal/shared/dates.go#L94 instead?

internal/providers/faro/session_list.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+
+			appID := resolveAppID(args[0])
+			escapedAppID := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(appID)
+			query := fmt.Sprintf(`{app_id="%s"} | logfmt | event_name=%sfaro.session_recording.started%s`, escapedAppID, "`", "`")
+
+			since, err := parseDuration(opts.Since)
+			if err != nil {
+				return fmt.Errorf("invalid --since value: %w", err)
+			}
```
dafydd-t, line 140: I think we've already validated this in Validate()

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+				}
+			}
+
+			segment, err := client.GetSegment(ctx, appID, sessionID, recordingID, segmentID)
+			if err != nil {
+				return err
+			}
+
+			// Warn if segment has dependency.
+			manifest, manifestErr := client.GetManifest(ctx, appID, sessionID, recordingID)
```
dafydd-t, line 367: what about if fetching manifests fails? it IIUC it will look like there are no dependencies. we should highlight when the fetch fails i think

internal/providers/faro/session_list_test.go
```diff
@@ -0,0 +1,152 @@
+package faro_test
```
dafydd-t, line 1: We're exporting quite a few types mainly so we can have a separate test package. Should we either test in-package, or test at a different layer of abstraction? wdyt?

dafydd-t: FormatRecordingDuration in the other file looks like a prime example of this

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+						DurationHuman:     durationHuman,
+						Segments:          len(manifest.Segments),
+						InactivityPeriods: len(manifest.InactivityPeriods),
+					}
+					return nil
+				})
+			}
+			if err := g.Wait(); err != nil {
+				return err
+			}
```
dafydd-t, line 170: if *any* fetches fail, none are returned. is that what we want?

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+
+			rows := make([]ReplayRecordingRow, len(resp.Items))
+			g, gctx := errgroup.WithContext(ctx)
+			g.SetLimit(10)
+
+			for i, item := range resp.Items {
+				g.Go(func() error {
+					manifest, err := client.GetManifest(gctx, appID, sessionID, item.ID)
```
dafydd-t, line 151: its fetching a lot of data to populate two columns SEGMENTS and INACTIVITY PERIODS. can we fetch less data for the default view here, and add this more expensive-to-fetch data to `-o wide`? Your judgement, ultimately.

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+)
+
+// FormatText is the output format name for replay text codecs.
+const FormatText = format.Format("text")
+
+// ReplayRecordingRow holds pre-computed data for one replay recording row.
+type ReplayRecordingRow struct {
+	RecordingID       string        `json:"recording_id"`
+	Status            string        `json:"status"`
+	Duration          time.Duration `json:"-"`
```
dafydd-t, line 29: do we need this property at all? the table renders DurationHuman, and its ignored in json ouptut

internal/providers/faro/session_commands_test.go
```diff
@@ -0,0 +1,313 @@
...
+	require.NotNil(t, segmentCmd, "inspect-replay-segment command should be registered under apps")
+
+	f := segmentCmd.Flags().Lookup("recording-id")
+	require.NotNil(t, f, "inspect-replay-segment should have --recording-id flag")
+	assert.Empty(t, f.DefValue)
+}
+
+func TestInspectReplaySegmentRawFlagRemoved(t *testing.T) {
+	cmd := inspectReplaySegmentCommand(t)
+	assert.Nil(t, cmd.Flags().Lookup("raw"))
```
dafydd-t, line 255: is this related? there is no --raw flag in this PR

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+	return errors.New("text format does not support decoding")
+}
+
+// EventsToSummaryRows converts a slice of RRWebEvent to summary rows for display.
+func EventsToSummaryRows(events []RRWebEvent) []EventSummaryRow {
+	rows := make([]EventSummaryRow, len(events))
+	for i, e := range events {
+		typeName := EventTypeName(e.Type)
+		source := "-"
+		if e.Type == 3 {
```
dafydd-t, line 278:
```suggestion
		if e.Type == IncrementalSnapshot {
```
can we use named constants do you think?

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,400 @@
...
+	return fmt.Sprintf("Unknown(%d)", t)
+}
+
+//nolint:gochecknoglobals // Static lookup table for rrweb incremental source types.
+var incrementalSourceNames = map[int]string{
+	0: "Mutation", 1: "MouseMove", 2: "MouseInteraction", 3: "Scroll",
+	4: "ViewportResize", 5: "Input", 6: "TouchMove", 7: "MediaInteraction",
+	8: "StyleSheetRule", 9: "CanvasMutation", 10: "Font", 11: "Log",
+	12: "Drag", 13: "StyleDeclaration", 14: "Selection",
+}
```
dafydd-t, line 222: can we keep these to one line per entry?

internal/providers/faro/session_list.go
```diff
@@ -0,0 +1,249 @@
...
+
+			lokiClient, err := loki.NewClient(cfg)
+			if err != nil {
+				return fmt.Errorf("creating Loki client: %w", err)
+			}
+
+			appID := resolveAppID(args[0])
+			escapedAppID := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(appID)
+			query := fmt.Sprintf(`{app_id="%s"} | logfmt | event_name=%sfaro.session_recording.started%s`, escapedAppID, "`", "`")
```
dafydd-t, line 135: Is there a nice way to build this string?

dafydd-t (changes requested):
Thanks for the PR! I think the --save issue and the segID parsing  need fixing, and the rest is your call ultimately.

PR #826, opened 2026-06-09
feat(instrumentation): add otel-checker

maryliag:
Add [Otel Checker](https://github.com/grafana/otel-checker) as a dependency, making it possible to run its commands from here. 
I added the commands under the existing `instrumentation` because I believe it was a good match, but if there is a better place or if it should be its own command, I'm happy to make the changes.

The OTel Checker also has a webserver option, but that was not included here, since it didn't quite fit with how GCX works.

Example for `gcx instrumentation check --language=js`:

Screenshot 2026-06-09 at 9 18 41 AM

### DIFF
--- a/cmd/gcx/instrumentation/check/codec.go
+++ b/cmd/gcx/instrumentation/check/codec.go
@@ -1,0 +1,58 @@
+package check
+
+import (
+	"errors"
+	"io"
+
+	"github.com/grafana/gcx/internal/format"
+	"github.com/grafana/gcx/internal/style"
+	otelutils "github.com/grafana/otel-checker/checks/utils"
+)
+
+// CheckTableCodec renders otelutils.Results as a grouped status/component/
+// message table.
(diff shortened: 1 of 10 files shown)


radiohead:
@claude Review this PR

maryliag:
@radiohead looks like the action has failed: https://github.com/grafana/gcx/actions/runs/27273941434

radiohead (approved):

radiohead:
> @radiohead looks like the action has failed: https://github.com/grafana/gcx/actions/runs/27273941434

Yeah, sadly that failed because said action doesn't work for forks.

radiohead:
@maryliag everything looks good from my end, but could you please sign your commits? That'd require resetting & redoing the commits unfortunately.

maryliag:
@radiohead awesome! they're signed now and showing the "verified" badge!

PR #831, opened 2026-06-10
feat(appo11y): add services get with auto-detected span-metrics

cedricziel:
## Summary

- Adds `gcx appo11y services get <name>` — drills into a single service from `services list`, returning metadata (language, env, k8s labels) plus a RED snapshot (rate, error rate, error %, p50/p95/p99 latency) over a configurable window.
- Auto-detects the span-metrics family for the requested service (probes `traces_span_metrics_*`, `traces_spanmetrics_*`, and bare `calls_total` in parallel; prefers the modern family when a stack double-emits during a v2→v3 migration).
- Manual override via `--metrics-mode {auto,v3,tempo,otel}` (plus aliases `legacy`, `beyla`, `otel-109`). The resolved mode is reported on the RED block so the source of the numbers is always visible.
- Service identity uses `job="<ns>/<name>"` — matches the canonical encoding already used by `target_info` discovery, so the query is portable across every metrics-mode this command supports.

## Why this shape

- `job=` is the canonical label filter across all metrics-modes. Using `service` + `service_namespace` instead would have broken on the legacy Tempo and bare OTel families, which don't emit those labels.
- The active span-metrics family is normally a stack-level setting. Since the CLI has no direct read on that setting, auto-detection probes the stack instead — it picks the family that has data for *this* service, with a stable preference order for double-emitting stacks.

## Flags

- `--window` (default `5m`)
- `--kind`: `inbound` (default: SERVER+CONSUMER), `server`, `consumer`, `all`, or a comma list of `SPAN_KIND_*` literals
- `--metrics-mode`: `auto` (default), `v3`, `tempo`, `otel`
- `--namespace`/`-n`, `--datasource`/`-d`, `--output`/`-o`

## Test plan

- [x] Unit tests cover the kind resolver, arg parser, metrics-mode resolver, all three query builders across all three families, the mode probe builder, instant scalar extraction, error-percent math, metadata selection, opts validation, and codec output.
- [x] `mise run lint` clean.
- [x] `GCX_AGENT_MODE=false go test ./...` passes.
- [x] Reference docs regenerated (`mise run reference`).
- [x] Live verification against a Grafana Cloud stack:
  - **auto** picks the modern family on a v3 stack and returns rate + p50/p95/p99 latency for an instrumented service.
  - **`--metrics-mode tempo` (override)** demonstrates the mismatch case: rate returns from the legacy `_calls_total` series, but latency is empty because the legacy `_latency_bucket` family isn't emitted on the same stack — exactly the scenario the help text predicts.
  - **nonexistent / uninstrumented service in auto** returns a clean no-data snapshot plus a stderr hint pointing at `services list`, no crash.

## Follow-ups

- `services operations <name>` — span-name × RED breakdown.
- `services map <name>` — service-graph slice (callers + callees).

### DIFF
--- a/docs/reference/cli/gcx_appo11y_services_get.md
+++ b/docs/reference/cli/gcx_appo11y_services_get.md
@@ -15,17 +15,17 @@ can't accidentally target the wrong service. Pass --namespace or use the
 
 Metadata comes from the same target_info/traces_target_info union used by
 "gcx appo11y services list". RED numbers are computed from span metrics
-over --window (default 5m), restricted to inbound spans (SERVER + CONSUMER).
+over --since (default 5m), restricted to inbound spans (SERVER + CONSUMER).
 
 The span-metric family is selected by --metrics-mode:
   auto   probe each family for this service and pick the one with data
-         (default — prefers v3 > tempo > otel when a stack double-emits)
+         (prefers v3 > tempo > otel when a stack double-emits)
   v3     traces_span_metrics_calls_total / _duration_seconds_bucket
-         (modern Tempo / OTel Collector ≥ 1.0.9)
+         (OTel Collector >= 0.109, Grafana Alloy >= 1.5.0)
   tempo  traces_spanmetrics_calls_total  / _latency_bucket
-         (legacy Tempo metrics-generator, Beyla)
+         (Tempo metrics-generator — Grafana Cloud default — and Beyla)
   otel   calls_total / duration_seconds_bucket
-         (bare OTel Collector spanmetrics connector)
+         (OTel Collector 0.94–0.108, Alloy 1.0–1.4.3, Grafana Agent >= 0.40)
 The resolved mode is reported in the snapshot output so you can confirm
 which family produced the numbers.
 
@@ -41,7 +41,7 @@ gcx appo11y services get <service> [--namespace ns] [flags]
   gcx appo11y services get checkoutservice
 
   # Same service, explicit namespace (skips the lookup)
-  gcx appo11y services get payments/checkoutservice --window 1h
+  gcx appo11y services get payments/checkoutservice --since 1h
 
   # JSON for scripting
   gcx appo11y services get checkoutservice -o json
@@ -49,7 +49,7 @@ gcx appo11y services get <service> [--namespace ns] [flags]
   # Restrict to server-side traffic only
   gcx appo11y services get checkoutservice --kind server
 
-  # Stack still on the legacy Tempo metrics-generator
+  # Stack double-emits both families; force the Tempo metrics-generator numbers
   gcx appo11y services get checkoutservice --metrics-mode tempo
 ```
 
@@ -60,10 +60,10 @@ gcx appo11y services get <service> [--namespace ns] [flags]
   -h, --help                  help for get
       --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
       --kind string           Span kinds to include: inbound (server+consumer), server, consumer, all, or a comma list of SPAN_KIND_* literals (default "inbound")
-      --metrics-mode string   Span-metrics family: auto (default, probes the stack), v3 (traces_span_metrics_*), tempo (traces_spanmetrics_*), or otel (bare calls_total + duration_seconds_bucket) (default "auto")
+      --metrics-mode string   Span-metrics family: auto (probes the stack), v3 (traces_span_metrics_*), tempo (traces_spanmetrics_*), or otel (bare calls_total + duration_seconds_bucket) (default "auto")
   -n, --namespace string      Service namespace (only needed when the argument is the bare service name and multiple namespaces are in play)
   -o, --output string         Output format. One of: agents, json, table, wide, yaml (default "table")
-      --window string         Rate/quantile window (e.g. 1m, 5m, 1h) (default "5m")
+      --since string          Rate/quantile window applied to span metrics (e.g. 1m, 5m, 1h, 1d) — PromQL duration syntax (default "5m")
 ```
 
 ### Options inherited from parent commands
--- a/docs/reference/cli/gcx_appo11y_services_get.md
+++ b/docs/reference/cli/gcx_appo11y_services_get.md
@@ -7,9 +7,11 @@ Inspect a single Application Observability service: metadata + RED snapshot.
 Show metadata and a rate/errors/duration snapshot for one service.
 
 The argument is either the bare service name (matching the OTel service.name
-resource attribute) or the canonical "<namespace>/<name>" form. When multiple
-namespaces have a service with the same name, either supply the namespace in
-the argument or pass --namespace.
+resource attribute) or the canonical "<namespace>/<name>" form. When a bare
+name is given, the namespace is resolved automatically from target_info;
+ambiguity (the same name in multiple namespaces) errors out so the snapshot
+can't accidentally target the wrong service. Pass --namespace or use the
+"<namespace>/<name>" form to disambiguate.
 
 Metadata comes from the same target_info/traces_target_info union used by
 "gcx appo11y services list". RED numbers are computed from span metrics
@@ -35,10 +37,10 @@ gcx appo11y services get <service> [--namespace ns] [flags]
 
 ```
 
-  # Bare service name, default 5m window
+  # Bare name — namespace is resolved from target_info
   gcx appo11y services get checkoutservice
 
-  # Namespaced service, longer window
+  # Same service, explicit namespace (skips the lookup)
   gcx appo11y services get payments/checkoutservice --window 1h
 
   # JSON for scripting
--- a/docs/reference/cli/gcx_appo11y_services_get.md
+++ b/docs/reference/cli/gcx_appo11y_services_get.md
@@ -1,0 +1,82 @@
+## gcx appo11y services get
+
+Inspect a single Application Observability service: metadata + RED snapshot.
+
+### Synopsis
+
+Show metadata and a rate/errors/duration snapshot for one service.
+
+The argument is either the bare service name (matching the OTel service.name
+resource attribute) or the canonical "<namespace>/<name>" form. When multiple
+namespaces have a service with the same name, either supply the namespace in
+the argument or pass --namespace.
+
+Metadata comes from the same target_info/traces_target_info union used by
+"gcx appo11y services list". RED numbers are computed from span metrics
+over --window (default 5m), restricted to inbound spans (SERVER + CONSUMER).
+
+The span-metric family is selected by --metrics-mode:
+  auto   probe each family for this service and pick the one with data
+         (default — prefers v3 > tempo > otel when a stack double-emits)
+  v3     traces_span_metrics_calls_total / _duration_seconds_bucket
+         (modern Tempo / OTel Collector ≥ 1.0.9)
+  tempo  traces_spanmetrics_calls_total  / _latency_bucket
+         (legacy Tempo metrics-generator, Beyla)
+  otel   calls_total / duration_seconds_bucket
+         (bare OTel Collector spanmetrics connector)
+The resolved mode is reported in the snapshot output so you can confirm
+which family produced the numbers.
+
+```
+gcx appo11y services get <service> [--namespace ns] [flags]
+```
+
+### Examples
+
+```
+
+  # Bare service name, default 5m window
+  gcx appo11y services get checkoutservice
+
+  # Namespaced service, longer window
+  gcx appo11y services get payments/checkoutservice --window 1h
+
+  # JSON for scripting
+  gcx appo11y services get checkoutservice -o json
+
+  # Restrict to server-side traffic only
+  gcx appo11y services get checkoutservice --kind server
+
+  # Stack still on the legacy Tempo metrics-generator
+  gcx appo11y services get checkoutservice --metrics-mode tempo
+```
+
+### Options
+
+```
+  -d, --datasource string     Prometheus datasource UID (defaults to datasources.prometheus in config or auto-discovery)
+  -h, --help                  help for get
+      --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+      --kind string           Span kinds to include: inbound (server+consumer), server, consumer, all, or a comma list of SPAN_KIND_* literals (default "inbound")
+      --metrics-mode string   Span-metrics family: auto (default, probes the stack), v3 (traces_span_metrics_*), tempo (traces_spanmetrics_*), or otel (bare calls_total + duration_seconds_bucket) (default "auto")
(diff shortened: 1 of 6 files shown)


internal/providers/appo11y/services/query.go
```diff
@@ -364,3 +366,300 @@ func mergeServiceSets(display, baseline, graph []Service) []Service {
...
+	}
+	return expr.String(), nil
+}
+
+// jobLabel returns the PromQL `job` label value for a (namespace, name)
+// pair, matching the `<namespace>/<service>` encoding target_info uses
+// throughout this package.
+func jobLabel(namespace, name string) string {
+	if namespace == "" {
+		return name
```
nandadev, line 488: Will the bare service name example work if `checkoutservice` belongs to the namespace `payments` (and not in any other namespace)?
```
  # Bare service name, default 5m window
  gcx appo11y services get checkoutservice
```

nandadev (dismissed):

internal/providers/appo11y/services/query.go
cedricziel: Good catch — it wouldn't have. The bare-name path was building a literal `job="checkoutservice"` filter, so a service that actually lives in `payments/checkoutservice` returned no data silently.

Fixed in b67d7f94: when the user passes a bare name (no `--namespace`, no `<ns>/<name>` form), `services get` now looks the service up in `target_info` first and uses the resolved namespace for every downstream query. Ambiguous results (the same name in multiple namespaces) error out with the list of candidates so the wrong service can't be silently targeted. Live-verified: `gcx appo11y services get adservice` now produces the same numbers as `gcx appo11y services get oteldemo01/adservice`.

nandadev (dismissed):

internal/providers/appo11y/services/query.go
```diff
@@ -364,3 +366,300 @@ func mergeServiceSets(display, baseline, graph []Service) []Service {
...
+//
+// We keep `service` + `service_namespace` out of the selector on purpose:
+// not every metric family emits them. Newer stacks emit both, but the
+// legacy Tempo `traces_spanmetrics_*` family and the OTel Collector
+// variant only emit `job`. Filtering on `job` alone keeps the query
+// portable across every metrics-mode this command supports.
+func scopedSpanMetric(metric, namespace, name string, kinds []string, window string) *promql.VectorExprBuilder {
+	return promql.Vector(metric).
+		Label("job", escapePromqlValue(jobLabel(namespace, name))).
+		LabelMatchRegexp("span_kind", spanKindRegex(kinds)).
```
radiohead, line 672: nit – we could add PromQL validation for `kinds` here. Not a blocker.

internal/providers/appo11y/services/get.go
```diff
@@ -0,0 +1,526 @@
...
+}
+
+func (o *getOpts) Validate() error {
+	if err := o.IO.Validate(); err != nil {
+		return err
+	}
+	if strings.TrimSpace(o.Window) == "" {
+		return errors.New("--window must not be empty")
+	}
+	if _, err := time.ParseDuration(o.Window); err != nil {
```
radiohead: issue – `time.ParseDuration` does not align with Prometheus time parsing logic. This will allow things like `300u` (which Prometheus will reject) but prohibit `1d` (which Prometheus supports).

internal/providers/appo11y/services/get.go
```diff
@@ -0,0 +1,526 @@
...
+	o.IO.BindFlags(flags)
+
+	flags.StringVarP(&o.Datasource, "datasource", "d", "", "Prometheus datasource UID (defaults to datasources.prometheus in config or auto-discovery)")
+	flags.StringVarP(&o.Namespace, "namespace", "n", "", "Service namespace (only needed when the argument is the bare service name and multiple namespaces are in play)")
+	flags.StringVar(&o.Window, "window", defaultRedWindow, "Rate/quantile window (e.g. 1m, 5m, 1h)")
+	flags.StringVar(&o.Kind, "kind", "inbound", "Span kinds to include: inbound (server+consumer), server, consumer, all, or a comma list of SPAN_KIND_* literals")
+	flags.StringVar(&o.MetricsMode, "metrics-mode", metricsModeAuto, "Span-metrics family: auto (default, probes the stack), v3 (traces_span_metrics_*), tempo (traces_spanmetrics_*), or otel (bare calls_total + duration_seconds_bucket)")
+}
+
+func (o *getOpts) Validate() error {
```
radiohead: issue – Could you wrap the errors here in `fail.NewCommandUsageError` so it aligns with other commands' validation?

internal/providers/appo11y/services/get.go
```diff
@@ -0,0 +1,526 @@
...
+				return fmt.Errorf("metrics-mode auto-detect failed: %w", err)
+			}
+		}
+
+		detail, err := fetchServiceDetail(ctx, client, datasourceUID, namespace, name, opts.Window, kinds, mode)
+		if err != nil {
+			return err
+		}
+		if !detail.Service.Instrumented && !detail.RED.HasTraffic {
+			emitNoDataHint(cmd.ErrOrStderr(), namespace, name)
```
radiohead, line 255: nit – `get $NON_EXISTENT_SERVICE` exits with `0`, could you change the exit code to `1`, to align with other commands' "entity not-found" semantics?

docs/reference/cli/gcx_appo11y_services_get.md
```diff
@@ -0,0 +1,82 @@
...
+
+The span-metric family is selected by --metrics-mode:
+  auto   probe each family for this service and pick the one with data
+         (default — prefers v3 > tempo > otel when a stack double-emits)
+  v3     traces_span_metrics_calls_total / _duration_seconds_bucket
+         (modern Tempo / OTel Collector ≥ 1.0.9)
+  tempo  traces_spanmetrics_calls_total  / _latency_bucket
+         (legacy Tempo metrics-generator, Beyla)
+  otel   calls_total / duration_seconds_bucket
+         (bare OTel Collector spanmetrics connector)
```
edvard-falkskar: The metric names and modes are correct but the descriptions are of compared to the App O11y UI. [Reference](https://github.com/grafana/app-observability-plugin/blob/main/plugin/src/constants/metricNames.ts)

```suggestion
  v3     traces_span_metrics_calls_total / _duration_seconds_bucket
         (OTel Collector >= 0.109, Grafana Alloy >= 1.5.0)
  tempo  traces_spanmetrics_calls_total  / _latency_bucket
         (Tempo metrics-generator — Grafana Cloud default — and Beyla)
  otel   calls_total / duration_seconds_bucket
         (OTel Collector 0.94–0.108, Alloy 1.0–1.4.3, Grafana Agent >= 0.40)
```

docs/reference/cli/gcx_appo11y_services_get.md
```diff
@@ -0,0 +1,84 @@
...
+  gcx appo11y services get payments/checkoutservice --window 1h
+
+  # JSON for scripting
+  gcx appo11y services get checkoutservice -o json
+
+  # Restrict to server-side traffic only
+  gcx appo11y services get checkoutservice --kind server
+
+  # Stack still on the legacy Tempo metrics-generator
+  gcx appo11y services get checkoutservice --metrics-mode tempo
```
edvard-falkskar, line 53: Shouldn't call this legacy as it's the default metrics source, the example should focus on why you would want to explicitly specify a specific metrics mode

```suggestion
  # Stack double-emits both families; force the Tempo metrics-generator numbers
  gcx appo11y services get checkoutservice --metrics-mode tempo
```

docs/reference/cli/gcx_appo11y_services_get.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+### Options
+
+```
+  -d, --datasource string     Prometheus datasource UID (defaults to datasources.prometheus in config or auto-discovery)
+  -h, --help                  help for get
+      --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+      --kind string           Span kinds to include: inbound (server+consumer), server, consumer, all, or a comma list of SPAN_KIND_* literals (default "inbound")
+      --metrics-mode string   Span-metrics family: auto (default, probes the stack), v3 (traces_span_metrics_*), tempo (traces_spanmetrics_*), or otel (bare calls_total + duration_seconds_bucket) (default "auto")
```
edvard-falkskar: Nitpick, but it states that auto is the default two times

internal/providers/appo11y/services/get.go
```diff
@@ -0,0 +1,587 @@
...
+  # Restrict to server-side traffic only
+  gcx appo11y services get checkoutservice --kind server
+
+  # Stack still on the legacy Tempo metrics-generator
+  gcx appo11y services get checkoutservice --metrics-mode tempo`,
+		Args: cobra.ExactArgs(1),
+		RunE: runGet(opts),
+		Annotations: map[string]string{
+			agent.AnnotationTokenCost: "small",
+			agent.AnnotationLLMHint:   `Per-service RED snapshot from Tempo/OTel span metrics: rate (req/s), error rate, error percent, p50/p95/p99 latency (seconds) over --window (default 5m), scoped to inbound spans (SERVER+CONSUMER). --metrics-mode picks the family: v3 (default, traces_span_metrics_*), tempo (legacy traces_spanmetrics_*), otel (bare calls_total/duration_seconds_bucket). Pairs with 'gcx appo11y services list' to drill into a single row. Examples: gcx appo11y services get <name> -o json; gcx appo11y services get <ns>/<name> --window 1h -o json; gcx appo11y services get <name> --metrics-mode tempo -o json`,
```
edvard-falkskar: In the other places `auto` is the default metrics-mode so should be here also for consistency. And the same metrics mode description changes as in the previous places.

```suggestion
			agent.AnnotationLLMHint: `Per-service RED snapshot from Tempo/OTel span metrics: rate (req/s), error rate, error percent, p50/p95/p99 latency (seconds) over --window (default 5m), scoped to inbound spans (SERVER+CONSUMER). --metrics-mode picks the family: auto (default, probes the stack), v3 (traces_span_metrics_*, OTel Collector >= 0.109/Alloy >= 1.5), tempo (traces_spanmetrics_*, Tempo metrics-generator — Grafana Cloud default — and Beyla), otel (bare calls_total/duration_seconds_bucket, older Collector/Alloy/Agent). Pairs with 'gcx appo11y services list' to drill into a single row. Examples: gcx appo11y services get <name> -o json; gcx appo11y services get <ns>/<name> --window 1h -o json; gcx appo11y services get <name> --metrics-mode tempo -o json`,
```

radiohead (commented):
Overall LGTM, some minor issues. I've left a few in-place comments and here's a few general ones:

1. `get SVC -otable / -owide` outputs 2 tables, this is a bit of an issue. Could you please rethink the output so that it fits into 1 table?
2. `--window` is inconsistent with other range-querying commands, which use `--since`, could you rename the flag?

internal/providers/appo11y/services/get.go
cedricziel: Fixed in d27ade24: switched to `model.ParseDuration` from `github.com/prometheus/common/model`, which is the canonical PromQL duration parser. Live-verified that `--since 1d` (rejected by `time.ParseDuration`) now works, while bogus inputs like `300u` are caught at validate time.

internal/providers/appo11y/services/get.go
cedricziel: Fixed in d27ade24: option validation now wraps via `fail.NewCommandUsageError(cmd, ...)`. Invalid `--since`, `--kind`, and `--metrics-mode` values, plus the namespace/arg conflict, all render with the standard usage block now.

internal/providers/appo11y/services/get.go
cedricziel: Fixed in d27ade24: when the service has no telemetry (uninstrumented + no traffic) the command now returns a non-zero exit code while still emitting the snapshot, so callers can branch on $? and agents/scripts still see the empty shape. Verified: `gcx appo11y services get no-such-service; echo $?` → 1.

internal/providers/appo11y/services/query.go
cedricziel: `resolveSpanKinds` already rejects anything that isn't an alias (`inbound`/`server`/`consumer`/`all`) or a literal starting with `SPAN_KIND_`. The values then flow into a fixed-shape PromQL regex (`span_kind=~"<val>|<val>"`), so they can't break the query syntax. Happy to add a stricter literal allow-list (SERVER|CLIENT|PRODUCER|CONSUMER|INTERNAL|UNSPECIFIED) if you'd prefer — let me know.

docs/reference/cli/gcx_appo11y_services_get.md
cedricziel: Applied your suggested wording verbatim in d27ade24 — Long help, the per-mode descriptions, and the agent hint all use the version-pinned descriptions now.

docs/reference/cli/gcx_appo11y_services_get.md
cedricziel: Applied your suggested example in d27ade24 — dropped the "legacy" framing for the `--metrics-mode tempo` example and reframed it as forcing a specific family when the stack double-emits.

docs/reference/cli/gcx_appo11y_services_get.md
cedricziel: Fixed in d27ade24 — dropped the duplicate "default" wording. `auto` is now only shown as the default via the cobra-rendered `(default "auto")` plus one mention in the Long help.

internal/providers/appo11y/services/get.go
cedricziel: Applied your suggested annotation in d27ade24 with one small tweak: kept the `--since` reference (was `--window` previously) to match the rename radiohead asked for in the same review.

cedricziel:
Both top-level points from @radiohead's review addressed:

**1. `-o table` / `-o wide` output** — `ca05ae61` already merged the two bordered tables into a single kubectl-describe-style block (single `tabwriter`, no box borders, latency under a sub-heading, blank line between identity and RED). For reference:

```
Name:         frontend
Namespace:    oteldemo01
Language:     nodejs
Status:       instrumented
Environment:  oteldemo01

Window:        1h
Metrics mode:  v3
Span kinds:    SPAN_KIND_SERVER|SPAN_KIND_CONSUMER
Rate:          5.166 req/s
Errors:        0.000 req/s (0.00%)
Latency:
  p50:  10.64ms
  p95:  202.94ms
  p99:  383.29ms
```

(If you'd prefer a styled `style.NewTable` rendering with one bordered FIELD/VALUE table instead of the describe layout, I can swap to that — happy to iterate.)

**2. `--window` → `--since`** — renamed in `d27ade24` to match `--since` in the synth / slo / datasources query commands. The JSON field stays `window` since it's literally the PromQL `[5m]` window passed to `rate()`/`histogram_quantile`, but the user-facing flag is consistent now.

wardbekker (approved):

PR #847, opened 2026-06-17
oauth flow for `gcx cloud login`

dafydd-t:
Part of https://github.com/grafana/gcx-internal/issues/13

Followup to https://github.com/grafana/gcx/pull/698 with fewer features: no refresh tokens or org selection

## Summary

Adds `gcx cloud login`, which authenticates against the Grafana Cloud platform API (grafana.com) for managing Cloud resources like stacks and access policies. This is distinct from `gcx login`, which targets a single Grafana stack instance.

- Interactive browser OAuth2 (PKCE) by default; `--cloud-token` for non-interactive use (CI/scripts)
- Two endpoints, configured independently, each defaulting to `https://grafana.com`:
  - `oauth-url` (`GRAFANA_CLOUD_OAUTH_URL`): used only by the login flow
  - `api-url` (`GRAFANA_CLOUD_API_URL`): used by every client that talks to the Cloud API
- Strict separation, no cross-fallback: production sets neither; an environment whose OAuth host and API host differ can set both. URLs missing a scheme are normalized to `https://`.
- Hardens the proxy token-refresh path so the response body (which can echo tokens) is no longer included in error messages.

## How to test

build with `mise run build`, then validate against an internal instance: follow instructions in https://github.com/grafana/deployment_tools/pull/632434 (dev)

Then run `gcx cloud stacks list --org <org>` to use it (give yourself stacks:read permissions at least)

## blockers

- [x] Wait for https://github.com/grafana/deployment_tools/pull/632405 to be merged

### DIFF
--- a/cmd/gcx/cloud/command.go
+++ b/cmd/gcx/cloud/command.go
@@ -181,8 +181,12 @@ func saveCloudConfig(ctx context.Context, configOpts *cmdconfig.Options, cloud *
 		cfg.SetContext(contextName, true, config.Context{})
 	}
 	curCtx := cfg.Contexts[contextName]
-	// replace the entire cloud config - clears any stale OAuth fields
-	// when switching from OAuth to SA token or vice versa
+	// Replace the entire cloud config to clear any stale OAuth/token fields
+	// when switching from OAuth to SA token or vice versa, but preserve the
+	// non-auth Stack selection so re-authenticating doesn't drop it.
+	if curCtx.Cloud != nil {
+		cloud.Stack = curCtx.Cloud.Stack
+	}
 	curCtx.Cloud = cloud
 
 	if err := config.Write(ctx, source, cfg); err != nil {
--- a/cmd/gcx/cloud/command.go
+++ b/cmd/gcx/cloud/command.go
@@ -57,12 +57,9 @@ directly via --cloud-token.
 
 Two endpoints can be configured independently, both defaulting to
 https://grafana.com: --oauth-url is used only for the login flow here, while
---api-url is used by every command that talks to the Grafana Cloud API. In
-production neither needs setting; against a dev environment where the two
-differ (e.g. a private API endpoint), set both.`,
+--api-url is used by every command that talks to the Grafana Cloud API.`,
 		Example: `  gcx cloud login
-  gcx cloud login --cloud-token glsa_abc123
-  gcx cloud login --oauth-url https://grafana.example.com --api-url https://api.internal.example.com`,
+  gcx cloud login --cloud-token glsa_abc123`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			// Endpoint URLs are sticky across re-auth: when not passed
 			// explicitly, carry over whatever the current context already has so
--- a/cmd/gcx/cloud/command.go
+++ b/cmd/gcx/cloud/command.go
@@ -3,6 +3,16 @@
 package cloud
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	cmdconfig "github.com/grafana/gcx/cmd/gcx/config"
+	"github.com/grafana/gcx/internal/auth"
+	"github.com/grafana/gcx/internal/config"
+	"github.com/grafana/gcx/internal/gcxerrors"
 	"github.com/grafana/gcx/internal/providers/stacks"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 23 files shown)


dafydd-t:
@claude review

dafydd-t:
addressed in 857fb8f:

1. **redirect guard** - accepted. `exchangeGCOMToken` now uses a scoped client with a `CheckRedirect` running `ValidateEndpointURL` on each hop, mirroring `exchangeCodeForToken`
2. **`CreateStackRequest.Org`** - accepted, dropped (no writer, out of scope)
3. **redundant `transport` local** - accepted, inlined into the struct literal
4. **`CloudConfig.TokenExpiresAt`** - accepted, dropped the field + its write (no reader, refresh is explicitly out of scope for this PR). regenerated config reference
5. **`ResolveOAuthURL`** - accepted, dropped it + its test (no non-test caller; login reads `Cloud.OAuthUrl` directly with the flag-precedence carry-over)
6. **dead nil branch in `saveCloudConfig`** - accepted, removed (`ConfigSource()` never returns nil)
7. **`OAuthUrl`/`APIUrl` naming** - deferring. `APIUrl` predates this PR so keeping `OAuthUrl` consistent; will fix both together in a follow-up

dafydd-t:
@claude review

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+By default, opens a browser for interactive OAuth2 authentication.
+
+For non-interactive use (CI/CD, scripts), pass a Cloud Access Policy token
+directly via --cloud-token.
+
+Two endpoints can be configured independently, both defaulting to
+https://grafana.com: --oauth-url is used only for the login flow here, while
+--api-url is used by every command that talks to the Grafana Cloud API. In
+production neither needs setting; against a dev environment where the two
+differ (e.g. a private API endpoint), set both.`,
```
dafydd-t: remove this clarification, it is not useful to most users

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+		Example: `  gcx cloud login
+  gcx cloud login --cloud-token glsa_abc123
+  gcx cloud login --oauth-url https://grafana.example.com --api-url https://api.internal.example.com`,
```
dafydd-t: remove this example, it is not useful to most users

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
 
 	cmd.AddCommand(stacks.NewCommand())
+	cmd.AddCommand(loginCmd())
 
 	return cmd
 }
+
+const defaultClientID = "gcx"
+
+func loginCmd() *cobra.Command {
```
radiohead, line 69: nit – could you wire the flags via an ops struct + `Validate()` for consistency's sake?

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+			if !cmd.Flags().Changed("oauth-url") && cur != nil && cur.Cloud != nil && cur.Cloud.OAuthUrl != "" {
+				oauthURL = cur.Cloud.OAuthUrl
+			}
+			if !cmd.Flags().Changed("api-url") && cur != nil && cur.Cloud != nil && cur.Cloud.APIUrl != "" {
+				apiURL = cur.Cloud.APIUrl
+			}
+			// Normalize whichever values won (flag, carry-over, or default)
+			// so a bare host (e.g. "grafana.example.com") gets an https:// scheme
+			// before it reaches the OAuth flow or is saved to config.
+			oauthURL = config.NormalizeCloudURL(oauthURL)
```
radiohead: question – this accepts `http://` (though we do warn in that case) – are there any use-cases for it, or should we just stick to `https://` only?

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+				"Check your config file syntax: gcx config edit",
+				"Or reset with: rm ~/.config/gcx/config.yaml && gcx cloud login",
+			},
+		}
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		cfg = config.Config{}
+	}
+
+	contextName := cfg.CurrentContext
```
radiohead: issue – we don't respect the `--context` flag bound by `cmdconfig.Options` (same in `currentCloudContext`).

I've checked by running `gcx cloud login --context foo` – it prints `Token saved to context "default"`.

Sounds like something we should fix?

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+
+	cloud := &config.CloudConfig{
+		Token:    result.AccessToken,
+		OAuthUrl: oauthURL,
+		APIUrl:   apiURL,
+	}
+	return saveCloudConfig(ctx, configOpts, cloud)
+}
+
+func saveCloudConfig(ctx context.Context, configOpts *cmdconfig.Options, cloud *config.CloudConfig) error {
```
radiohead: nit – could you move this to e.g. `internal/config`, so that we don't have business logic in `cmd/`?

cmd/gcx/cloud/command.go
```diff
@@ -15,6 +24,181 @@ func Command() *cobra.Command {
...
+		contextName = config.DefaultContextName
+	}
+
+	if !cfg.HasContext(contextName) {
+		cfg.SetContext(contextName, true, config.Context{})
+	}
+	curCtx := cfg.Contexts[contextName]
+	// replace the entire cloud config - clears any stale OAuth fields
+	// when switching from OAuth to SA token or vice versa
+	curCtx.Cloud = cloud
```
radiohead: issue – this will wipe out `cloud.stack` from the config. Could we use e.g. `login.mergeCloudInto` (ofc make it exported first) here instead to make sure stack value is preserved?

internal/auth/gcom.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	w := opts.Writer
+	if w == nil {
+		w = os.Stderr
+	}
+	return &GCOMFlow{opts: opts, writer: w}
+}
+
+// Run executes the GCOM OAuth2 PKCE flow.
+func (f *GCOMFlow) Run(ctx context.Context) (*GCOMResult, error) {
```
radiohead, line 65: nit – should we also `ValidateEndpointURL(f.opts.GCOMURL)` here?

internal/auth/gcom.go
```diff
@@ -0,0 +1,278 @@
...
+	case result := <-resultCh:
+		return result, nil
+	case err := <-errCh:
+		return nil, err
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+}
+
+func (f *GCOMFlow) startGCOMCallbackServer(ctx context.Context, listener net.Listener, expectedState, codeVerifier, redirectURI string, resultCh chan<- *GCOMResult, errCh chan<- error) *http.Server {
```
radiohead, line 135: nit – `startGCOMCallbackServer` is near-identical to `startCallbackServer` in `flow.go`, would be nice to refactor a bit to drop duplication.

internal/auth/gcom.go
```diff
@@ -0,0 +1,278 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, tokenURL, bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create token request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{
```
radiohead: issue (non-blocking) – could we use the shared client factory from `httputils` with good defaults? We'd need to add support for passing `CheckRedirect` via opts but it sounds like a small enough fix.

cmd/gcx/cloud/command.go
dafydd-t: done in 708aca84 - dropped the clarification

cmd/gcx/cloud/command.go
dafydd-t: done in 708aca84 - removed the example

wardbekker:
🐑 I pushed 4 additive commits to this branch (they'll collapse in the squash merge). Rebased onto latest `main` and force-with-lease pushed. Behaviour-preserving except where noted.

### What changed

**1. `fix(cloud)`: preserve `Stack` across `cloud login`**
`saveCloudConfig` replaced the entire `CloudConfig` to clear stale OAuth/token fields when switching between OAuth and SA-token auth — but that also dropped a persisted non-auth `Cloud.Stack` slug on every re-login. It now carries the existing `Stack` over, so re-authenticating no longer wipes a configured stack. Added a white-box test (`command_test.go`) covering this.

**2. `refactor(auth)`: drop unused GCOM result fields and options**
`GCOMResult.ExpiresIn` / `UserID` / `Info.Name` and `GCOMOptions.Port` / `BindAddress` were populated or exposed but never read by any caller. Removed them (and trimmed the matching wire-struct fields). This also removes the latent inconsistency between the configurable `BindAddress` and the hardcoded `127.0.0.1` redirect URI — the flow now consistently binds loopback.

**3. `fix(auth)`: restore `deeplink.Open` browser call**
A rebase-time autofix had rewritten the browser-open call in the GCOM flow to an undefined `openBrowser` helper and dropped the `deeplink` import, which broke the build. Restored `deeplink.Open(authURL)`, matching the sibling proxy flow in `flow.go`.

### Verification
Local gate all green: build, `go test` (auth + cloud), `golangci-lint` (0 issues), and `reference-drift` (no docs drift — none of these changes touch a command/flag/config/env surface). CI is running on the new head.

---
*Pushed by shepherd-pr — happy to revert any of these if you'd prefer to handle them yourself.*

radiohead (commented):
Overall LGTM, a few nitpicks (non-blocking). Two real problems – the config overwrite and the lack of `--context` support.

cmd/gcx/cloud/command.go
dafydd-t: yeah good idea

cmd/gcx/cloud/command.go
dafydd-t: eek sorry. yes definitely

dafydd-t:
Note to self: i also dont think we need to set 

```
      oauth-url: https://grafana.com
      api-url: https://grafana.com
```

in the config after the oauth flow (check this happens)

dafydd-t:
Also note: when loggin in with `gcx login` and using a cloud AP token, I still see:

```
Optional: add a Grafana Cloud API token to enable Cloud management features.
Logged in to https://xxxxx.grafana.net
  Context:     xxxxx
  Auth method: oauth
  Version:     13.2.0-28607514261
  Grafana Cloud: yes
  Stack:       xxxxx
```

this is after it was previously logged in with oauth

maybe that oauth method is for the stack login, but its somewhat confusing to not see any data for grafana.com here? maybe out of scope

dafydd-t:
note to self: the client is not currently requesting all the scopes that are needed (check the hints in the gcx login flow)

cmd/gcx/cloud/command.go
dafydd-t: yeah sure, thanks

cmd/gcx/cloud/command.go
dafydd-t: local dev is an example of allowing both, so I might leave it

cmd/gcx/cloud/command.go
dafydd-t: I think we re-set stack below, but I've refactored it to use mergeCloudInto instead 👍

internal/auth/gcom.go
dafydd-t: the existing validate endpoint method only tests for `.net` domains. I created a separate `validateGCOMURL` function to validat the gcom endpoints

internal/auth/gcom.go
dafydd-t: done

internal/auth/gcom.go
dafydd-t: Added `CheckRedirect` to the clientOpts, then used NewClient 👍

dafydd-t:
> Note to self: i also dont think we need to set
> 
> ```
>       oauth-url: https://grafana.com
>       api-url: https://grafana.com
> ```
> 
> in the config after the oauth flow (check this happens)

didn't address this, to reduce scope bloat

dafydd-t:
> note to self: the client is not currently requesting all the scopes that are needed (check the hints in the gcx login flow)

needs  some changes server side first. Reduced requested scopes to only those that are a) currently valid and b) required

radiohead (approved):

PR #863, opened 2026-06-18
feat(frontend): add apps overview KPI snapshot

cedricziel:
## What

Adds `gcx frontend apps overview <app>` — a per-app Frontend Observability (Faro/RUM) KPI snapshot. Resolves an app **name or numeric id**, then queries Loki for:

- **Page loads** and **exception count** (with errors as a % of loads)
- The five **Core Web Vitals at p75** (LCP, INP, CLS, FCP, TTFB), each rated good / needs-improvement / poor against Google's thresholds
- **Top exceptions**

…over `--since` (default `1h`). Supports `table`/`wide`/`json`/`yaml` output and includes agent annotations (token cost + LLM hint). This is the frontend counterpart to `gcx appo11y services get` (backend RED snapshot).

```
gcx frontend apps overview faro-shop-demo --since 15m
gcx frontend apps overview 153 -o json
```

## How it works

The LogQL faithfully mirrors the Frontend Observability plugin's own overview queries (`app-o11y-kwl` `src/scenes/components/apps/overview/queries.ts`): instant queries keyed on the **`app_id`** label, p75 via `logfmt | unwrap` over the Faro `kind="measurement"` stream, page loads proxied by the presence of ` ttfb=`, errors/top-exceptions from `kind="exception"`. Implemented in-package (`internal/providers/faro/`) to reuse the Faro client for name→id resolution without a circular import.

## Constitution compliance

- **CLI grammar (§64–67):** `overview` nests under the `apps` resource type as an extension verb (alongside `list`/`get`/…), **not** as a top-level `frontend` command. ✅
- Format-agnostic fetch, all output via codecs, stdout/stderr split, `providers.ConfigLoader`, query client bypasses the k8s dynamic client, errgroup, Options pattern, table-driven tests, lowercase error messages. ✅
- Minor: `overview.go` imports `cmd/gcx/fail` (an `internal/`→`cmd/` import) — mirrors the merged `appo11y services get`; flagged for reviewer awareness.

## Testing

- Unit tests (table-driven) for query builders, CWV rating thresholds, scalar/top-error parsing, and formatting. `go test ./internal/providers/faro/...` ✅ (incl. `-race`)
- Lint clean for the package (`golangci-lint … ./internal/providers/faro/...` → 0 issues).
- Reference docs regenerated.
- **Verified end-to-end** against a live dev stack (`appo11ydev01`): correctly resolves `faro-shop-demo`→id 153, fires valid LogQL, renders the snapshot, emits the no-data hint, exits non-zero on empty.

## Caveats / open items

- **Needs live RUM to validate the numbers.** The dev stack used for verification has no Faro RUM in its Loki yet (`app_id` label absent until data flows), so the snapshot rendered empty. Once RUM is flowing, vital units (ms) / label names should be sanity-checked — any drift is a one-file change in `overview_query.go`.
- Draft pending that validation. Follow-ups planned: `web-vitals`, `errors`, `pages` (per `page_id`) — already sketched in the same plugin query file.
- Pre-existing, unrelated: `aio11y/eval/*` test failures (agent-mode message wording) exist on `main` too; `mkdocs` not installed locally (CI covers the docs build).

### DIFF
--- a/docs/reference/cli/gcx_frontend_apps.md
+++ b/docs/reference/cli/gcx_frontend_apps.md
@@ -28,6 +28,7 @@ Manage Frontend Observability apps.
 * [gcx frontend apps delete](gcx_frontend_apps_delete.md)	 - Delete a Frontend Observability app.
 * [gcx frontend apps get](gcx_frontend_apps_get.md)	 - Get a Frontend Observability app by slug-id or name.
 * [gcx frontend apps list](gcx_frontend_apps_list.md)	 - List Frontend Observability apps.
+* [gcx frontend apps overview](gcx_frontend_apps_overview.md)	 - Show a Frontend Observability KPI snapshot for one app: page loads, errors, and Core Web Vitals.
 * [gcx frontend apps remove-sourcemap](gcx_frontend_apps_remove-sourcemap.md)	 - Remove sourcemap bundles from a Frontend Observability app.
 * [gcx frontend apps show-sourcemaps](gcx_frontend_apps_show-sourcemaps.md)	 - Show sourcemaps for a Frontend Observability app.
 * [gcx frontend apps update](gcx_frontend_apps_update.md)	 - Update a Frontend Observability app from a file.
--- a/docs/reference/cli/gcx_frontend_apps_overview.md
+++ b/docs/reference/cli/gcx_frontend_apps_overview.md
@@ -1,0 +1,69 @@
+## gcx frontend apps overview
+
+Show a Frontend Observability KPI snapshot for one app: page loads, errors, and Core Web Vitals.
+
+### Synopsis
+
+Show the headline KPIs for one Frontend Observability (Faro) app.
+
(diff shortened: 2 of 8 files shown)


domasx2:
Page-loads/vitals silently read as "no data" whenever Loki returns a multi-point series for the "instant" query

loki.MetricQuerySample carries `Value` (vector) or `Values` (matrix). sampleValue() only reads `s.Value`, returning (0,false) for a matrix. This stack's Loki datasource returned a matrix (2-3 timestamps, "Step: 1m0s") on every repeated identical instant-query once >~60s of data existed, so overview reported pageLoads=0/all vitals "-"/exit 1 for an app verified (direct Loki query, same instant) to have 6 page loads and 1 exception. Only correct when the response happened to collapse to one timestamp.

_Reply posted by Claude on Domas's behalf._

domasx2 (changes requested):
end to end test turned up issue with erroneously reporting no data for time range series

PR #864, opened 2026-06-19
fix(config): lazy keychain resolution and eliminate redundant config loads

### LINKED ISSUE #858: config loading is slow due to redundant keychain lookups
## Problem

Commands that load config are slow — `gcx config use-context` takes ~2s on my gcx setup with ~10 contexts . The root cause is lots of OS keychain lookups during config loading.

### Why it's slow

`config.Load()` calls `resolveSentinels()` which walks **every context** and resolves **every keychain sentinel** via `store.Get()` (backed by `go-keyring` / macOS `SecItemCopyMatching`). Each lookup takes ~15ms.

With a config containing ~10 contexts and ~20 keychain-backed secrets, a single `Load()` costs ~300ms in keychain I/O alone.

Several commands then load the config multiple times per invocation. For example, `use-context` loads 3 times:

1. `LoadLayered()` in the command body (for interactive picker / context validation)
2. `LoadForWrite()` internally calls `LoadLayered()` again to discover sources
3. `LoadForWrite()` then calls `Load()` a third time on the resolved source

This produces ~60+ keychain lookups for a single command, totalling ~1s of keychain I/O on top of normal startup costs.

### Fix options

Two fixes that stack:

1. **Eliminate redundant loads** — `LoadForWrite` loads the config via `LoadLayered`, discards it, then loads the same file again. For the single-source case it should reuse what it already has. Commands like `use-context` could also pass an already-loaded config to `LoadForWrite` instead of loading from scratch.

2. **Lazy sentinel resolution** — `resolveSentinels` resolves all contexts eagerly. Most commands only need the current context's secrets. Deferring resolution to access time would cut lookups from ~20 to ~2-3 per load.

### Measured impact

```
# 22 keychain lookups via go-keyring
22 keyring lookups: 323ms

# full command
time gcx config use-context dev
0.52s user 0.46s system — 1.97s wall
```

dafydd-t:
Closes #858

## Summary

We should see snappier commands when users have lots of contexts, due to all config loading behaviour now being quicker. Particularly for macOS users where gcx uses keychain to store tokens

- `resolveSentinels` now resolves only the current context's keychain entries during `Load()`; non-current contexts keep their raw sentinel strings until explicitly resolved via `Config.ResolveContext()`
- `LoadForWrite` reuses the config already loaded by `LoadLayered` for zero/single-source cases instead of discarding it and calling `Load()` again

## Why

Each macOS keychain lookup (`SecItemCopyMatching` via go-keyring) takes ~15ms. With ~20 sentinels across ~10 contexts, a single `Load()` was spending ~300ms on keychain I/O. Commands like `use-context` called `Load()` 3 times, producing ~60 keychain lookups and ~1s of overhead.

## Before/After

```
# before (3 loads × ~20 sentinel lookups)
$ time gcx config use-context dev
0.52s user 0.46s system — 1.97s wall

# after (2 loads × ~2 sentinel lookups)
$ time gcx config use-context dev
0.12s user 0.13s system — 0.50s wall
```


internal/config/loader.go
```diff
@@ -360,7 +360,15 @@ func Load(ctx context.Context, source Source, overrides ...Override) (Config, er
 
 	log := logging.FromContext(ctx)
 	store := keychainStoreFn()
-	config.keychainFields, config.keychainPreserve = resolveSentinels(&config, store, log)
+	config.keychainStore = store
+
+	// Only resolve sentinels for the current context eagerly. Other contexts
+	// are resolved on demand via Config.ResolveContext to avoid redundant
+	// keychain lookups (each ~15ms on macOS).
```
dafydd-t:
```suggestion
	// keychain lookups.
```

dafydd-t:
@claude review

dafydd-t:
addressed in 171a8252:

- **bug — `LoadLayered` multi-source path skips post-override resolve**: fixed. `LoadLayered` now calls `merged.ResolveContext(merged.CurrentContext)` after the override loop. this also covers the case where the effective current-context (after merge) was current in no layer. added a regression test (`TestLoadLayered_OverrideResolvesSentinelsForSelectedContext`) exercising the two-layer merge path with a `--context` override.
- **test coverage gap**: added `TestLoad_LazyResolvesOnlyCurrentContext`, `TestLoad_OverrideSwitchingContextResolvesSentinels`, `TestResolveContext_NoOps` (nil store / missing context / already-resolved), and the LoadLayered test above.
- **ResolveContext nil-check placement**: lifted the nil checks above the loops.
- **comment polish**: accepted the `// keychain lookups.` suggestion (also applied via UI), clarified the defensive `case 0` in `LoadForWrite`.
- **MergeConfigs observation**: added a doc comment explaining keychain fields carry over from base and why that's safe.
- **stale architecture doc**: updated `docs/architecture/config-system.md` step 5 to describe the lazy/current-context-only resolution model, `Config.ResolveContext`, and the `keychainStore` field.

note: couldn't run `mise run lint` locally (env has golangci-lint built with go1.25 vs project's go1.26 + a vendoring quirk), but `go vet`, `go build`, and the full `go test` suite pass. relying on CI lint.

dafydd-t:
@claude review

javiermolinar (approved):
nice

wardbekker (approved):
looks good. only the claude reviews left to address

PR #868, opened 2026-06-19
perf(config): lazy-open and memoize the OS keychain probe

wardbekker:
## Context

Fix 2 of 4 targeting bottlenecks on the code path shared by **most** gcx commands.

Every config `Load()` called `credentials.Open()`, which probes the keychain backend with a syscall — a `/usr/bin/security` subprocess on macOS (~tens of ms). Worse:

- a **layered** load (system → user → local) repaid that probe **once per layer**, and
- the probe ran **even when the current context had no keychain-backed secrets** (e.g. OAuth-proxy contexts, env-var auth, anonymous).

## Change

Two changes, both behavior-preserving:

1. **Memoize `credentials.Open()`** via `sync.Once` in `defaultKeychainStore`, so the probe runs at most **once per process** instead of once per `Load`/layer.
2. **`lazyStore`** wrapper that defers opening the backend until the first `Get`/`Set`/`Delete`. Since `resolveSentinelsForContext` only calls `Get` for fields that actually hold a sentinel and `migratePlaintextSecrets` only calls `Set` when plaintext secrets are present, a config with **no keychain-backed secrets never probes the keychain at all**.

`Config.keychainStore` remains non-nil (the lazy wrapper), so on-demand `ResolveContext` for non-current contexts still works and only opens the backend if it actually finds a sentinel.

## Verification

- New tests: `TestLoad_NoSecretsDoesNotOpenKeychain` (asserts 0 opens) and `TestLoad_SentinelOpensKeychain` (asserts open + correct resolution).
- `go test ./internal/config/...` — pass
- `golangci-lint run ./internal/config/...` — 0 issues

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -32,11 +33,23 @@ import (
 //nolint:gochecknoglobals // test injection seam for the keychain backend.
 var keychainStoreFn = defaultKeychainStore
 
+// openStoreOnce memoizes the OS-keychain probe (credentials.Open) for the
+// lifetime of the process. Open probes the backend with a syscall (a
+// /usr/bin/security subprocess on macOS); without memoization every Load — and
+// every layer of a layered load — repaid that probe.
+//
+//nolint:gochecknoglobals // process-wide memoization of the keychain probe.
+var (
+	openStoreOnce sync.Once
+	openedStore   credentials.Store
+)
+
 func defaultKeychainStore() credentials.Store {
 	if testing.Testing() {
 		return testingNoopStore{}
 	}
-	return credentials.Open()
+	openStoreOnce.Do(func() { openedStore = credentials.Open() })
+	return openedStore
 }
 
 type testingNoopStore struct{}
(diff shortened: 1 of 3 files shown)


internal/config/loader.go
```diff
@@ -32,11 +33,23 @@ import (
 //nolint:gochecknoglobals // test injection seam for the keychain backend.
 var keychainStoreFn = defaultKeychainStore
 
+// openStoreOnce memoizes the OS-keychain probe (credentials.Open) for the
+// lifetime of the process. Open probes the backend with a syscall (a
+// /usr/bin/security subprocess on macOS); without memoization every Load — and
+// every layer of a layered load — repaid that probe.
+//
+//nolint:gochecknoglobals // process-wide memoization of the keychain probe.
+var (
```
javiermolinar, line 42: Do we need this? having only the lazystore seems enough. The less globals the better

wardbekker: Good question — they cover different scopes, and the lazyStore alone isn't enough.

A fresh `lazyStore` is created **per `Load()` call** (`loader.go:378`), so its `sync.Once` only dedups the probe *within a single load*. But the common path is layered: `LoadLayered` calls `Load()` once per layer — system → user → local (`loader.go:479`) — and `Write()` calls `keychainStoreFn()` directly, bypassing the lazyStore entirely (`loader.go:425`).

Without the process-global `sync.Once`, each layer's lazyStore would open the backend independently → `credentials.Open()` runs once *per layer* again. That's exactly the regression this PR set out to fix ("a layered load repaid that probe once per layer").

So the two work together:
- **lazyStore** → skip the probe *entirely* when the context has no keychain-backed secrets.
- **process-global `once`** → when a probe *is* needed, run it at most once across all layers + the Write path.

Dropping the globals would re-introduce the per-layer probe for the has-secrets case. Happy to add a code comment making this split explicit if it'd help readers.

radiohead (approved):
IMHO this makes the code a bit too difficult to follow, but functionally it seems to be fine.

PR #869, opened 2026-06-19
perf(config): cut redundant config loads on write and diagnostics paths

wardbekker:
## Context

Fix 3 of 4 targeting bottlenecks on the code path shared by **most** gcx commands.

Two places did a full layered config load (read every layer + merge + resolve the current context's keychain sentinels) when far less was needed:

1. **`loadDiagnosticsConfig`** runs on **every** invocation (`main.go`, before the command's own config load) just to read one boolean. It called `LoadLayered`, which parses every context, resolves keychain sentinels, and even **auto-creates a config file** as a side effect.
2. **`LoadForWrite`'s `--file` branch** called `LoadLayered` (load + merge + per-layer keychain resolution) only to discover which file corresponds to the requested layer — then loaded that file *again*.

## Change

- Add **`LoadDiagnostics`**: reads only the `diagnostics` block from each layer, merging with the same precedence as `MergeConfigs`. No context resolution, **no keychain probe**, no config auto-creation. `loadDiagnosticsConfig` now uses it.
- **`LoadForWrite` `--file` branch** uses `DiscoverSources` (path lookup only) instead of a full `LoadLayered`, preserving the `GCX_CONFIG` explicit-file bypass exactly.

> `readDiagnostics` decodes into the full `Config` (the YAML codec rejects unknown fields, so a partial struct won't parse a real config) but skips Load's keychain resolution / migration / auto-create.

## Verification

- New tests: `TestLoadDiagnostics_{NoConfigReturnsNil,ReadsEnabledFlag,LayersLocalOverUser}`.
- Existing `TestLoadForWrite_fileType_*` continue to pass (cover the `--file` change behaviorally).
- `go test ./internal/config/... ./cmd/gcx/...` — pass
- `golangci-lint run ./internal/config/... ./cmd/gcx/...` — 0 issues

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -523,6 +523,13 @@ func LoadForWrite(ctx context.Context, explicitFile, fileType string) (Config, S
 				return cfg, src, err
 			}
 		}
+		// Fresh system (no config files yet): preserve LoadLayered's auto-create.
+		// LoadLayered only ever created the user layer, so --file user creates and
+		// returns it; other layer types have nothing to auto-create and still error.
+		if fileType == "user" && len(sources) == 0 {
+			cfg, err := Load(ctx, StandardLocation())
+			return cfg, StandardLocation(), err
+		}
 		return Config{}, nil, fmt.Errorf("no %s config file found", fileType)
 	}
 
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -504,11 +504,19 @@ func LoadForWrite(ctx context.Context, explicitFile, fileType string) (Config, S
 	}
 
 	if fileType != "" {
-		layered, err := LoadLayered(ctx, "")
+		// Selecting a layer by --file only needs the discovered source paths, not
+		// a full layered load+merge (which would parse every layer and resolve its
+		// keychain sentinels just to discard all but one). Honor the explicit-file
+		// env bypass exactly as LoadLayered does: when set, layering is bypassed
+		// and there is no named layer to select.
+		if os.Getenv(ConfigFileEnvVar) != "" {
+			return Config{}, nil, fmt.Errorf("no %s config file found", fileType)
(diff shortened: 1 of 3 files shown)


internal/config/loader.go
```diff
@@ -504,11 +504,19 @@ func LoadForWrite(ctx context.Context, explicitFile, fileType string) (Config, S
...
-		layered, err := LoadLayered(ctx, "")
+		// Selecting a layer by --file only needs the discovered source paths, not
+		// a full layered load+merge (which would parse every layer and resolve its
+		// keychain sentinels just to discard all but one). Honor the explicit-file
+		// env bypass exactly as LoadLayered does: when set, layering is bypassed
+		// and there is no named layer to select.
+		if os.Getenv(ConfigFileEnvVar) != "" {
+			return Config{}, nil, fmt.Errorf("no %s config file found", fileType)
+		}
+		sources, err := DiscoverSources()
```
javiermolinar, line 515: Before it created a load file in a fresh system. Now it returns a no file error. Is this intended?

wardbekker: Good catch — that was an unintended regression, now fixed in 0992c9d7.

The old `--file` branch went through `LoadLayered`, which auto-created the user config when no config files existed at all. Swapping to `DiscoverSources` dropped that, so `config set --file user` on a fresh system started returning `no user config file found`.

Restored narrowly: when no sources are discovered and the requested layer is `user`, it now creates it via `StandardLocation` (matching `LoadLayered`, which only ever auto-created the user layer). Non-user layers (`local`, `system`) still error on a fresh system — that was the behavior before too. Added `TestLoadForWrite_fileType_user_freshSystem_autoCreates` and `..._nonUser_freshSystem_errors` to lock both paths in.

radiohead (approved):

PR #870, opened 2026-06-19
fix(aio11y): preserve min_idle_seconds for conversation rules

soniaAguilarPeiron:
## Problem

Creating a conversation-scope eval rule via `gcx` failed even though the equivalent direct API call succeeded:

```bash
gcx aio11y rules create -f conversation-rule.yaml --agent
# Backend: min_idle_seconds is required when selector is "conversation"
```

`gcx` was **silently dropping** `min_idle_seconds`. The `RuleDefinition` struct had no corresponding field, so the value was lost when parsing YAML/JSON and never serialized into the create/update request body. The backend then rejected the (incomplete) conversation rule.

## Fix

Add the missing field to `RuleDefinition` (`internal/providers/aio11y/eval/types.go`):

```go
MinIdleSeconds *int `json:"min_idle_seconds,omitempty"`
```

- **Pointer + `omitempty`** so the field is omitted for non-conversation rules (no behavior change) and round-trips for conversation rules.
- The entire rules CRUD flow marshals/unmarshals this single struct — the HTTP client (`client.go`), the generic `TypedCRUD` adapter, and `specToUnstructured` (where `min_idle_seconds` is **not** in `ruleStripFields`). So one field fixes **create, get, list, and update** at once.

No client-side validation of the `selector`/`min_idle_seconds` coupling is added — consistent with the provider's pattern of delegating spec validation to the backend (selectors are unvalidated free-form strings in gcx).

## Note on the feature toggle

Conversation rules are gated by a **backend feature toggle**. This change is orthogonal to the toggle: it only ensures `gcx` serializes the field correctly. With the toggle off the backend rejects the rule regardless (as it did before this fix); with the toggle on, `gcx` now works end-to-end instead of failing on its own.

## Tests

Regression tests at three layers:
- `read_test.go` — YAML `selector: conversation` + `min_idle_seconds: 10` parses into the struct; generation rules carry `nil`.
- `types_test.go` — `json.Marshal` includes `"min_idle_seconds":10` for conversation rules and omits it otherwise.
- `client_test.go` — the **outgoing HTTP request body** received by the server carries `min_idle_seconds` (the exact path the bug broke).

Verified these fail without the fix (build error / field undefined) and pass with it.

### DIFF
--- a/internal/providers/aio11y/eval/rules/client_test.go
+++ b/internal/providers/aio11y/eval/rules/client_test.go
@@ -134,14 +134,15 @@ func TestClient_Create_ConversationRule(t *testing.T) {
 		// otherwise the backend rejects conversation rules. This is the bug
 		// being guarded: gcx previously dropped the field entirely.
 		body, err := io.ReadAll(r.Body)
-		require.NoError(t, err)
+		assert.NoError(t, err)
 		assert.Contains(t, string(body), `"min_idle_seconds":10`)
 
 		var def eval.RuleDefinition
-		require.NoError(t, json.Unmarshal(body, &def))
+		assert.NoError(t, json.Unmarshal(body, &def))
 		assert.Equal(t, "conversation", def.Selector)
-		require.NotNil(t, def.MinIdleSeconds)
-		assert.Equal(t, 10, *def.MinIdleSeconds)
+		if assert.NotNil(t, def.MinIdleSeconds) {
+			assert.Equal(t, 10, *def.MinIdleSeconds)
+		}
 
 		w.WriteHeader(http.StatusCreated)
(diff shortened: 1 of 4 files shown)


soniaAguilarPeiron:
bugbot review

alexander-akhmetov (dismissed):

alexander-akhmetov (approved):

wardbekker (approved):
Verified end-to-end: `min_idle_seconds` is now carried on `RuleDefinition` (pointer + omitempty) and round-trips through all CRUD paths — confirmed it is not in `ruleStripFields`. Regression tests cover the YAML parse, JSON marshal, and outgoing HTTP request body. CI green. LGTM.

PR #871, opened 2026-06-19
feat: default pyroscope metrics step from datasource

simonswine:
## Summary

When `--step` is omitted from `gcx profiles metrics`, the query step now defaults to the Pyroscope datasource's configured `minStep` (read from `jsonData.minStep`), falling back to 15s if unset or invalid — matching Grafana's own Pyroscope backend behaviour.

Previously, omitting `--step` sent `stepSeconds=0` to the API, which could produce unexpected results.

## Changes

- **`internal/datasources/query/pyroscope_config.go`** — new `GetPyroscopeConfig` helper that fetches a Pyroscope datasource by UID and parses `jsonData.minStep` (default 15s).
- **`internal/datasources/pyroscope/series.go`** — extracts `resolveMetricsStepSeconds` to centralise step resolution: `--top` full-range → explicit `--step` → datasource minStep → 15s fallback. Updates `--step` flag description to document the default.
- Tests for both new units.

## Testing

```bash
go test ./internal/datasources/pyroscope/... ./internal/datasources/query/...
```

### DIFF
--- a/internal/datasources/pyroscope/series.go
+++ b/internal/datasources/pyroscope/series.go
@@ -150,7 +150,12 @@ Datasource is resolved from -d flag or datasources.pyroscope in your context.`,
 				groupBy = []string{"service_name"}
 			}
 
-			stepSeconds, start, end, err := resolveMetricsStepSeconds(ctx, cfg, datasourceUID, opts.Top, start, end, step)
+			// --top mode queries the full range to get one bucket per series.
+			if opts.Top && (start.IsZero() || end.IsZero()) {
+				start, end = pyroscope.DefaultTimeRange(start, end)
+			}
+
+			stepSeconds, err := resolveMetricsStepSeconds(ctx, cfg, datasourceUID, opts.Top, start, end, step)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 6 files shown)


wardbekker:
## Shepherd Status — PR #871

**State:** All CI green pre-shepherd; one tightening commit added; rebased on latest `main` (961f00e0). Awaiting CI + review approval.

### Sensibility assessment
- **Problem:** real — omitting `--step` previously sent `stepSeconds=0` to the Pyroscope API. Defaulting to the datasource `minStep` matches Grafana's backend.
- **Approach:** sound — `GetPyroscopeMinStep` mirrors the existing `GetInfluxDBConfig`/`GetDatasourceType` pattern in `internal/datasources/query`.

### Tightening applied (additive commit)
- `refactor(pyroscope): simplify step resolution` — collapsed `resolveMetricsStepSeconds` to return `(float64, error)` (the `--top` full-range defaulting now lives in the caller, which owns the start/end the request uses), and replaced the single-field `PyroscopeConfig` struct with `GetPyroscopeMinStep` returning `time.Duration` directly. Behavior-preserving; tests updated and passing.

### Verification
- `go build`, `go test`, and `golangci-lint` all clean for the changed packages.

---
*Updated by shepherd-pr*

internal/datasources/pyroscope/series.go
```diff
@@ -200,6 +199,23 @@ Datasource is resolved from -d flag or datasources.pyroscope in your context.`,
 	return cmd
 }
 
+func resolveMetricsStepSeconds(ctx context.Context, cfg internalconfig.NamespacedRESTConfig, datasourceUID string, top bool, start, end time.Time, step time.Duration) (float64, error) {
+	switch {
+	case top:
+		// One bucket per series across the (already-defaulted) full range.
+		return end.Sub(start).Seconds(), nil
```
radiohead, line 206: Small nit – IIUC this will override the passed `--step` value silently. Usually if we override something like that, it'd be a good idea to return an error instead, something along the lines of `--top is exclusive with --step`.

internal/datasources/pyroscope/series.go
```diff
@@ -200,6 +199,23 @@ Datasource is resolved from -d flag or datasources.pyroscope in your context.`,
...
+	case step > 0:
+		// Explicit --step wins over the datasource minStep.
+		return step.Seconds(), nil
+	default:
+		minStep, err := dsquery.GetPyroscopeMinStep(ctx, cfg, datasourceUID)
```
radiohead, line 211: This makes the invocations without the step quite a bit slower, since we re-fetch all the time. Could we cache the resolved step in the default datasources config fields?

I am OK with doing that in a follow-up, since that'd require a few changes.

radiohead (approved):
LGTM in general, a couple of nits which could be addressed in follow-ups.

PR #877, opened 2026-06-24
feat(irm): notification compliance rules + doctor

yuri-tceretian:
## What

Adds an OnCall **notification-compliance** surface to gcx — the "safety net" idea: admins define a minimal required notification setup for the org, and anyone can check whether a user is actually reachable when paged.

### New commands

| Command | Purpose |
|---|---|
| `gcx irm oncall compliance-rules get` | Show the org's compliance ruleset |
| `gcx irm oncall compliance-rules set` | Create/update the ruleset — flags **or** `-i` interactive checkboxes (pre-checked from current) |
| `gcx irm oncall compliance-rules evaluate` | Audit **all** org users vs the rules |
| `gcx irm doctor` | Check **your own** (or `--user <id>`) compliance |

## How it talks to the backend

The compliance API currently lives **only on the OnCall public API** (`/api/v1/oncall_compliance/` + `/evaluate/`, auth via `Authorization` token + `X-Grafana-Instance-ID`), **not** the plugin-proxy surface the rest of the IRM client uses. Until it's exposed through the proxy, these commands support a **TEST-ONLY direct transport**:

- flags `--oncall-url` / `--oncall-token` / `--instance-id` (env: `ONCALL_URL` / `ONCALL_TOKEN` / `INSTANCE_ID`)
- without `--oncall-url`, `get`/`set` fall back to the plugin-proxy client; `evaluate`/`doctor` require it (evaluate is public-API-only today)
- `doctor` resolves the current user via the proxy's `GetCurrentUser`, then looks that user up in `evaluate`

This transport is clearly marked temporary in the code; the permanent wiring (expose via plugin proxy vs. a real public-API client config) is still open.

## Output

- **text** — friendly channel/rule labels (`Slack`, `SMS`, `Phone call`, …) and evaluation results joined with user **email/username**
- **json/yaml** — raw enum values preserved (`notify_by_slack`, `phone`, …) for scripting

## Try it locally

```bash
go build -buildvcs=false -o bin/gcx ./cmd/gcx/

export ONCALL_URL="http://<oncall-engine>:8084"
export ONCALL_TOKEN="<oncall-api-token>"
export INSTANCE_ID="<grafana-instance-id>"

./bin/gcx irm oncall compliance-rules set --required-channels phone,slack \
  --default notify_by_slack --important notify_by_sms,notify_by_slack
./bin/gcx irm oncall compliance-rules get
./bin/gcx irm oncall compliance-rules evaluate
./bin/gcx irm doctor --user <oncall-user-id>      # or run against a configured context for the current user
```

## Notes / open questions

- **Transport is test-only** — needs a decision before this is production-ready (proxy exposure would drop all the `--oncall-*` flags and the second auth path).
- `doctor` pairs the proxy `GetCurrentUser().PK` with the public-API `evaluate` `user_id`; they match today (same OnCall user PK).
- Enum label map (`complianceLabels`) and selectable option lists are in `oncall_compliance_commands.go` — adjust if the backend's valid values differ.
- Tests cover the body mapping, validation guard, evaluation lookup, and label rendering. Reference docs regenerated.

### DIFF
--- a/docs/reference/cli/gcx_irm_doctor.md
+++ b/docs/reference/cli/gcx_irm_doctor.md
@@ -1,0 +1,37 @@
+## gcx irm doctor
+
+Check whether your OnCall notification setup meets the org's compliance rules.
+
+```
+gcx irm doctor [flags]
+```
+
+### Options
+
+```
+  -h, --help                  help for doctor
+      --instance-id string    TEST-ONLY: X-Grafana-Instance-ID for direct transport. Defaults to $INSTANCE_ID.
+      --jq string             jq expression to apply to JSON output. Mutually exclusive with --json.
+      --json string           Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+      --oncall-token string   TEST-ONLY: OnCall API token for direct transport. Defaults to $ONCALL_TOKEN.
+      --oncall-url string     TEST-ONLY: OnCall engine base URL (e.g. http://host:8084); bypasses the plugin proxy. Defaults to $ONCALL_URL.
+  -o, --output string         Output format. One of: agents, json, text, yaml (default "text")
+      --user string           OnCall user ID to check (defaults to the current user)
+```
+
+### Options inherited from parent commands
+
+```
+      --agent                       Enable agent mode (JSON output, no color). Auto-detected from CLAUDECODE, CLAUDE_CODE, CURSOR_AGENT, GITHUB_COPILOT, AMAZON_Q, or GCX_AGENT_MODE env vars.
+      --config string               Path to the configuration file to use
+      --context string              Name of the context to use (overrides current-context in config)
+      --insecure-log-http-payload   Log full HTTP request/response bodies including raw credentials, authorization tokens, cookies, and OAuth refresh tokens. Do not ship these logs.
+      --no-color                    Disable color output
+      --no-truncate                 Disable table column truncation (auto-enabled when stdout is piped)
+  -v, --verbose count               Verbose mode. Multiple -v options increase the verbosity (maximum: 3).
+```
+
+### SEE ALSO
+
+* [gcx irm](gcx_irm.md)	 - Manage Grafana IRM (OnCall + Incidents)
+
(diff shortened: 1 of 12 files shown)


wardbekker:
## Shepherd review — PR #877

Took a pass over this with an eye toward mergeability. The feature itself is real and in scope (OnCall notification-compliance is a genuine gap), and the command surface follows the `$AREA $NOUN $VERB` grammar nicely. Two CI checks are red but trivial; one architectural question is worth settling before this lands.

### CI (mechanical — easy fixes)
- **Tests** — `TestConsistency_AllLeafCommandsHaveTokenCost` fails: the 4 new leaf commands (`irm doctor`, `compliance-rules get/set/evaluate`) are missing `agent.token_cost` annotations.
- **Linters** — 11 issues: 1 `canonicalheader` (`X-Grafana-Instance-ID` → `X-Grafana-Instance-Id`), 2 `errchkjson` in the test, 7 `gochecknoglobals` (option/label/color vars), 1 `nestif`.

### Architecture — the one to decide
The PR ships a **test-only direct transport** to the OnCall public API, and the client is constructed as a bare `http.Client`:

```go
// internal/providers/irm/oncall_compliance_commands.go:336
httpClient: &http.Client{Timeout: 30 * time.Second},
```

This bypasses `httputils.NewClient`/`NewDefaultClient`, which CONSTITUTION.md requires for all production HTTP clients (it's what wires in the retry transport, `User-Agent`, and `X-Grafana-Caller-Id`). The "test-only" framing is honest, but the bare client is in the production binary's code path regardless, so it reads as a hard-invariant violation rather than something scoped to tests.

Related, smaller points:
- `oncall_doctor_command.go:62` — `doctor` hard-requires `--oncall-url`, so it only works via the test-only path even when the plugin proxy is available.
- `publicComplianceClient` (`oncall_compliance_commands.go:343–453`) duplicates ~95 lines of request/response handling that largely mirrors `OnCallClient`.

### Question for @yuri-tceretian
The cleanest resolution to all of the above is the plugin-proxy exposure you mention in the PR description — once `/oncall_compliance/` is reachable through the proxy, the second auth path and all the `--oncall-*` flags disappear.

**What's the timeline for exposing the compliance endpoints via the plugin proxy?**

- If it's **near-term**, it may be worth waiting and landing this against the proxy surface directly (avoids shipping then immediately removing the test-only transport).
- If it's **further out** and you'd like this usable now, the path forward is: route the public client through `httputils.NewClient` to satisfy the invariant, fix the CI items, and clearly gate/document the direct transport as interim. Happy to do that work once you confirm the direction.

No code changes pushed yet — holding for your call on the above.

---
*Posted by shepherd-pr (no commits made; awaiting decision)*

PR #880, opened 2026-06-24
feat(kg): add entity & relationship write commands (gated, experimental)

patmagee:
## Problem

`gcx kg` is read-only for entities and write-capable only for *configuration* (`prom-rules`, `model-rules`, `suppressions`). The KG (Asserts) backend added a write surface for **custom, API-origin** entities and relationships, plus a `domain` query param on entity reads. There was no way to create/update/delete those objects from gcx, nor to read them back filtered by domain.

## Approach

Expose the write API through `gcx kg`, following existing kg conventions:

- `gcx kg entities create | delete`
- `gcx kg relationships create | delete`
- read-side `--domain` filter on `gcx kg entities inspect`

Targets the K8s-style write API at `/apis/kg.grafana.com/v1alpha1/namespaces/<stacks-N>/{entities,relationships}` on the existing `kg.Client` (status-aware request distinguishes 201-created from 200-updated; deletes encode identity, incl. scope, as query params).

**Feature-gated.** The backend endpoints are themselves experimental (`kg.write.api.enabled`, default off) and roll out via backend-enterprise independently of gcx. The commands are gated behind `providers.kg.write-api-enabled` (config or `GRAFANA_PROVIDER_KG_WRITE_API_ENABLED=true`), **checked before any network call**, returning a guiding "how to enable" error when off. Commands stay **registered** (not conditionally added) so they remain discoverable and appear in the generated CLI reference.

**UX details**
- Input: flags **or** `-f` file (mutually exclusive). `--scope`/`--property`/`--from-scope`/`--to-scope` use kubectl-style `key=value` via pflag `StringToString`; relationship endpoints use a compact `domain/Type/name` ref token. `--ttl` takes a duration (omitted = never expire; negative = never; `0` = expire immediately).
- Client-side validation mirrors backend `KgPatterns` (identifier/domain slugs; writable domain rejects the reserved `kg`) and fails before any HTTP request.
- Deletes use `ConfirmDestructive` + `--force`; a 404 hints at a scope/ref identity mismatch.

## Alternatives considered

- **Conditional command registration** (omit commands when the flag is off) — rejected: config isn't resolvable at command-tree build time (the context is threaded in `PersistentPreRun`), it would drop the commands from the generated reference docs, and gcx has no precedent for it. Run-time gating matches the existing paradigm (e.g. the assistant provider's capability detection).
- **Resource-adapter (`push`/`pull`) integration** — out of scope; these ephemeral, TTL'd, MERGE-keyed objects fit imperative verbs, not declarative reconciliation.
- **Custom `key=value` parsing** — replaced with pflag `StringToString` to match `--dimensions`/`--var` already used elsewhere in gcx.

## Out of scope

- `gcx resources push`/`pull` integration for entities/relationships.
- Enabling the flag in ops/prod (dev-only today, in deployment_tools).

## Testing

- httptest client coverage: 201/200/404/409/422/5xx; path/query-param/body shaping.
- Command-level: gate-disabled issues **no** HTTP; decline-skips-delete vs `--force`; client-side validation failures never reach the wire; flag parsing; create 201-vs-200 reporting.
- `mise run lint` clean; full suite green except one **pre-existing, unrelated** failure (`TestConfigLoader_LoadGrafanaConfig_ContextOverrideBeforeEnvVars`, reproduces on `main`).

## Deployment dependency (why E2E is pending)

End-to-end is blocked until the backend-enterprise gateway routes (#13196) reach a stack. On the dev cells the api-server (`v0.207.0`) already contains `KgWriteController` and `kg.write.api.enabled: true` is set, but the deployed `cloud-backend-gateway` pin (`master-0bc0d5b1`) predates #13196, so the route currently returns a gateway 404. A `cloud_backend_gateway` image bump in `deployment_tools` is needed (tracked separately). gcx is gated precisely so it can ship ahead of that rollout.

## Notes

AI-assisted: authored with Claude Code (subagent-driven execution, TDD per functional unit); human-reviewed.


internal/providers/kg/client_test.go
```diff
@@ -652,3 +663,130 @@ func TestClient_DeleteModelRules(t *testing.T) {
...
+					return
+				}
+				w.WriteHeader(tt.status)
+				writeJSON(w, kg.EntityWriteResponse{Domain: "myapp", Type: "Service", Name: "checkout"})
+			}))
+			defer server.Close()
+			client := newTestClient(t, server)
+			resp, created, err := client.UpsertEntity(t.Context(), kg.EntityWriteRequest{
+				Domain: "myapp", Type: "Service", Name: "checkout", TTLSeconds: new(int64(-1)),
+			})
```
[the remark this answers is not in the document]
patmagee: `new` accepts a value (not just a type) as of Go 1.26, and this repo builds on go1.26.2. `new(int64(-1))` allocates an `int64`, initializes it to -1, and returns `*int64` — valid and intentional for the -1 sentinel. The package compiles and `TestClient_UpsertEntity` / `TestClient_UpsertRelationship` pass. Ref: https://go.dev/ref/spec#Allocation. No change.

internal/providers/kg/client_test.go
```diff
@@ -652,3 +663,130 @@ func TestClient_DeleteModelRules(t *testing.T) {
...
+			writeJSON(w, kg.RelationshipWriteResponse{Type: "CALLS"})
+		}))
+		defer server.Close()
+		client := newTestClient(t, server)
+		resp, err := client.UpsertRelationship(t.Context(), kg.RelationshipWriteRequest{
+			Domain: "myapp", Type: "CALLS",
+			From:       kg.EntityRef{Domain: "myapp", Type: "Service", Name: "checkout"},
+			To:         kg.EntityRef{Domain: "myapp", Type: "Service", Name: "cart"},
+			TTLSeconds: new(int64(-1)),
+		})
```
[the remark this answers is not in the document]
patmagee: `new` accepts a value (not just a type) as of Go 1.26, and this repo builds on go1.26.2. `new(int64(-1))` allocates an `int64`, initializes it to -1, and returns `*int64` — valid and intentional for the -1 sentinel. The package compiles and `TestClient_UpsertEntity` / `TestClient_UpsertRelationship` pass. Ref: https://go.dev/ref/spec#Allocation. No change.

internal/providers/kg/client_test.go
```diff
@@ -652,3 +663,130 @@ func TestClient_DeleteModelRules(t *testing.T) {
...
+	})
+	t.Run("404 from/to not found", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusNotFound)
+		}))
+		defer server.Close()
+		client := newTestClient(t, server)
+		_, err := client.UpsertRelationship(t.Context(), kg.RelationshipWriteRequest{Domain: "myapp", Type: "CALLS", TTLSeconds: new(int64(-1))})
+		var apiErr *kg.APIError
+		require.ErrorAs(t, err, &apiErr)
```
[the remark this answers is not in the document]
patmagee: `new` accepts a value (not just a type) as of Go 1.26, and this repo builds on go1.26.2. `new(int64(-1))` allocates an `int64`, initializes it to -1, and returns `*int64` — valid and intentional for the -1 sentinel. The package compiles and `TestClient_UpsertEntity` / `TestClient_UpsertRelationship` pass. Ref: https://go.dev/ref/spec#Allocation. No change.

internal/providers/kg/write_commands.go
```diff
@@ -0,0 +1,571 @@
...
+		to        string
+		toScope   map[string]string
+		force     bool
+	)
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "Delete a custom relationship.",
+		Long: `Delete an API-origin edge of the given type between the from/to entities.
+The endpoint refs (incl. scope) must match the values used at create.`,
+		Example: `  gcx kg relationships delete --type CALLS \
```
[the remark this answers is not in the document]
patmagee: Checked against the KG write OpenAPI ([kg-write.yaml@1c1d121](https://github.com/grafana/asserts-adi/blob/1c1d1216e38f756acbeb475664c6fe69ef0d2a0c/inference-engine/api-server/openapi/kg-write.yaml)): the delete endpoint `DELETE /relationships/{type}` takes `RelationshipDeleteRequestDto`, which has only `from` and `to` — each an `EntityRefDto` carrying its own `domain`. There is no top-level edge domain in the delete request; the backend identifies the edge by `(type, from, to)`. `DeleteRelationship` already sends `from.domain`/`to.domain` as query params, so it matches the spec. The top-level `domain` on create designates the owning writable domain (authz/namespacing of the API-origin edge), not an extra delete-identity axis. No change.

docs/reference/cli/gcx_kg_relationships_create.md
```diff
@@ -0,0 +1,55 @@
...
+```
+  gcx kg relationships create --type CALLS --domain myapp \
+    --from myapp/Service/checkout --to myapp/Service/cart --to-scope env=prod --ttl 1h
+  gcx kg relationships create -f rel.yaml
+```
+
+### Options
+
+```
+      --domain string               Writable domain slug for the edge (required)
```
egovova11: I think this should be more specific, i.e. `represents a specific application domain like IRM`

internal/providers/kg/client.go
```diff
@@ -731,3 +779,88 @@ func (c *Client) GetRule(ctx context.Context, name string) (*Rule, error) {
...
 	return &f, nil
 }
+
+// ---------------------------------------------------------------------------
+// KG Write API: relationships
+// ---------------------------------------------------------------------------
+
+// UpsertRelationship creates or updates a custom (API-origin) edge between two
+// existing entities. Both endpoints must already exist (404 otherwise).
+func (c *Client) UpsertRelationship(ctx context.Context, req RelationshipWriteRequest) (*RelationshipWriteResponse, error) {
```
xujiaxj, line 791: I see you have `write_commands`, `write_flags`, so it makes sense to move these methods into a separate `write_client.go`?

xujiaxj, line 791: Do we need to validate `namespace` so it's not malformed? (We don't seem to url-escape it, so assuming it's url-safe?)

internal/agent/command_annotations.go
```diff
@@ -285,37 +285,41 @@ var commandAnnotations = map[string]annotation{
...
+	"gcx kg relabel-rules get":    {Cost: "small", Hint: "fetch prologue, epilogue, or generated Mimir relabel rules [--type generated]"},
+	"gcx kg prom-rules create":    {Cost: "small"},
+	"gcx kg prom-rules delete":    {Cost: "small"},
+	"gcx kg prom-rules get":       {Cost: "small"},
+	"gcx kg prom-rules list":      {Cost: "small"},
+	"gcx kg status":               {Cost: "small"},
+	"gcx kg suppressions create":  {Cost: "small"},
+	"gcx kg suppressions delete":  {Cost: "small"},
+	"gcx kg suppressions list":    {Cost: "small"},
+	"gcx kg entities create":      {Cost: "small", Hint: "--domain <d> --type <Type> --name <n> [--scope k=v] [--property k=v] [--ttl 1h] | upsert a custom (API-origin) entity; requires the KG write API (GRAFANA_PROVIDER_KG_WRITE_API_ENABLED=true)"},
```
xujiaxj: Claude suggests "llm_hint that signals they're gated/experimental (e.g. "requires providers.kg.write-api-enabled"), so an agent doesn't repeatedly invoke them only to hit the gate error."

wardbekker:
## Shepherd Status (iteration 2)

### Merge Readiness
✅ Conflict resolved — branch rebuilt on latest `main` (merged `#886` "apply timerange to entity scopes lookup"). `mergeable: MERGEABLE`. Blocked only on **review approval**.

### CI Checks
Re-running on the merge commit (`d1f63a9b`); previous commit was all-green.

### Conflict Resolution (commit `d1f63a9b`)
`main` (#886) extracted a `timeRangeParams()` helper and added a time window to `ListEntityScopes(ctx, startMs, endMs)`. This PR independently added a `domain` param to `GetEntityInfo`/`LookupEntity`. Resolved as the **union**: kept the `timeRangeParams` helper, kept the `domain` parameter, and updated both readers to call the helper. All `ListEntityScopes` call sites pass the time window. Full gate green locally (lint 0 issues, tests, build, docs).

### Auto-Fixes Applied (additive commits)
- `c61e276c` — **docs(kg):** `--domain` help clarified as a specific application domain (e.g. `irm`); CLI reference regenerated. _Addresses @egovova11._
- `44af8ced` — **docs(agent):** `entities delete` / `relationships create|delete` llm_hints now signal the `GRAFANA_PROVIDER_KG_WRITE_API_ENABLED` gate, matching `entities create`. _Addresses @xujiaxj + claude[bot]._

### Reviewer Comments — Triage
- ✅ `new(int64(-1))` ×3 — valid Go 1.26; `go vet` passes. Author rebuttal stands.
- ✅ `relationships delete` domain — matches KG write OpenAPI. Author rebuttal stands.
- ✅ `--domain` domain-blind fallback — fixed in `af1c2896`.
- ✅ namespace URL-safety — `authlib.CloudNamespaceFormatter` output is already URL-safe.
- 💬 `write_client.go` split (@xujiaxj) — proposed leaving for this PR (writes already live in `client.go`); better as a dedicated follow-up.

---
*Updated by shepherd-pr*

internal/providers/kg/client.go
wardbekker: Leaning toward leaving this as-is for now: `client.go` already keeps write methods (`UpsertSuppression`, `DeleteRule`, `DeleteModelRules`) alongside reads, so the new entity/relationship writes follow the existing pattern. Splitting only these would scatter KG writes across two files; a real `write_client.go` would mean moving *all* of them — better as a separate refactor than bundled into this PR. Happy to do that follow-up if we want it as a convention.

wardbekker:
On the `write_client.go` suggestion ([thread](https://github.com/grafana/gcx/pull/880#discussion_r3497545469)): proposing we keep it in `client.go` for this PR. Write methods already live there next to reads (`UpsertSuppression`, `DeleteRule`, …), so a partial split would scatter KG writes across two files. If we want the read/write separation as a convention, it is cleaner as a dedicated follow-up that moves everything at once.

xujiaxj (approved):

internal/providers/kg/client.go
xujiaxj: Good poin!

radiohead (approved):

PR #891, opened 2026-06-30
fix(query): fall back to /api/ds/query on any non-200 from the k8s query API

dafydd-t:
## Summary

- The datasource query clients try Grafana's k8s query API (`/apis/query.grafana.app/.../query`) first, then fall back to the legacy `/api/ds/query`. The fallback only fired on `404`, so a user whose Grafana role can't `create` the `query` resource under `query.grafana.app` (e.g. Viewer) got a hard `403` — even though `/api/ds/query` would have served them.
- Broadens the fallback to trigger on **any non-200** response, not a `404`/`403` allowlist. The k8s query API is not enabled everywhere yet and can fail in other ways (e.g. a `500` from a rolling-out query service, or `501`); the legacy path is the universally-available one, so on any failure we fall through to it.
- Queries are idempotent reads, so the retry is always safe. Transient `429`/`5xx` are still absorbed by the retry transport below this layer, so a fallback only happens on a *persistent* non-200.
- Applies across the shared `grafanaquery` transport (athena, clickhouse, influxdb, infinity) and the bespoke loki, prometheus, and cloudwatch clients.
- Tests: 404/403/500 fallback coverage on the shared transport plus the bespoke clients; the loki error-envelope test now reflects that a `400` triggers the fallback too.

## Why

A Viewer-role user hit:

```
query failed with status 403: {"kind":"Status",...,"message":"query.query.grafana.app \"name\" is forbidden:
User \"<user>\" cannot create resource \"query\" in API group \"query.grafana.app\" in the namespace \"<ns>\"...",
"reason":"Forbidden","code":403}
```

The 403 is a Kubernetes apiserver RBAC artifact: the new query API maps a POST to a `create` verb on the `query` resource, which the Viewer role lacks. The legacy `/api/ds/query` enforces the real datasource-query permission (`datasources:query`) and serves these users fine. Rather than allowlisting specific status codes, we treat any non-200 from the not-yet-ubiquitous k8s path as a signal to use the proven legacy endpoint — which grants no access the user wouldn't already have via `/api/ds/query`.

### DIFF
--- a/internal/query/cloudwatch/client.go
+++ b/internal/query/cloudwatch/client.go
@@ -90,8 +90,9 @@ func (c *Client) Query(ctx context.Context, dsUID string, req QueryRequest) (*Qu
 		return nil, err
 	}
 
-	// Fall back to legacy /api/ds/query if K8s query API is not available.
-	if statusCode == http.StatusNotFound {
+	// Fall back to legacy /api/ds/query if the K8s query API is unavailable
+	// (404) or forbidden for the user's role (403, e.g. Viewer).
+	if statusCode == http.StatusNotFound || statusCode == http.StatusForbidden {
 		apiPath = "/api/ds/query"
 		respBody, statusCode, err = c.post(ctx, apiPath, body)
 		if err != nil {
--- a/internal/query/cloudwatch/client.go
+++ b/internal/query/cloudwatch/client.go
@@ -90,9 +90,10 @@ func (c *Client) Query(ctx context.Context, dsUID string, req QueryRequest) (*Qu
 		return nil, err
 	}
 
-	// Fall back to legacy /api/ds/query if the K8s query API is unavailable
-	// (404) or forbidden for the user's role (403, e.g. Viewer).
-	if statusCode == http.StatusNotFound || statusCode == http.StatusForbidden {
+	// Fall back to the legacy /api/ds/query endpoint on any non-200 response.
+	// The K8s query API is not enabled everywhere and can fail in ways beyond
+	// 404 (e.g. 403 for a Viewer role); /api/ds/query is the universal path.
+	if statusCode != http.StatusOK {
 		apiPath = "/api/ds/query"
 		respBody, statusCode, err = c.post(ctx, apiPath, body)
 		if err != nil {
--- a/internal/query/cloudwatch/client.go
+++ b/internal/query/cloudwatch/client.go
@@ -90,9 +90,8 @@ func (c *Client) Query(ctx context.Context, dsUID string, req QueryRequest) (*Qu
 		return nil, err
 	}
 
-	// Fall back to the legacy /api/ds/query endpoint on any non-200 response.
-	// The K8s query API is not enabled everywhere and can fail in ways beyond
-	// 404 (e.g. 403 for a Viewer role); /api/ds/query is the universal path.
+	// The K8s query API is not enabled everywhere and can return a non-200
+	// (e.g. 403 for some roles); fall back to the legacy /api/ds/query.
 	if statusCode != http.StatusOK {
 		apiPath = "/api/ds/query"
 		respBody, statusCode, err = c.post(ctx, apiPath, body)
--- a/internal/query/grafanaquery/client.go
+++ b/internal/query/grafanaquery/client.go
@@ -45,15 +45,18 @@ func NewClientWithHTTPClient(cfg config.NamespacedRESTConfig, httpClient *http.C
 }
 
 // Execute posts body to Grafana's K8s datasource query API, falls back to the
-// legacy /api/ds/query endpoint on 404, and converts non-200 responses to a
-// typed query API error for datasource and operation.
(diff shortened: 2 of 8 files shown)


internal/query/cloudwatch/client.go
```diff
@@ -90,8 +90,9 @@ func (c *Client) Query(ctx context.Context, dsUID string, req QueryRequest) (*Qu
 		return nil, err
 	}
 
-	// Fall back to legacy /api/ds/query if K8s query API is not available.
-	if statusCode == http.StatusNotFound {
+	// Fall back to legacy /api/ds/query if the K8s query API is unavailable
+	// (404) or forbidden for the user's role (403, e.g. Viewer).
```
dafydd-t:
```suggestion
	// (404) or forbidden for the user's role (403).
```

internal/query/grafanaquery/client.go
```diff
@@ -45,15 +45,18 @@ func NewClientWithHTTPClient(cfg config.NamespacedRESTConfig, httpClient *http.C
...
+// datasource and operation.
 func (c *Client) Execute(ctx context.Context, body []byte, datasource, operation string) ([]byte, error) {
 	statusCode, respBody, err := c.post(ctx, c.k8sQueryPath(), body)
 	if err != nil {
 		return nil, err
 	}
 
-	if statusCode == http.StatusNotFound {
+	// The K8s query API may be unavailable (404) or forbidden for the user's
+	// role (403, e.g. Viewer); fall back to the universally-available endpoint.
```
dafydd-t:
```suggestion
	// role (403); fall back to the universally-available endpoint.
```

internal/query/loki/client.go
```diff
@@ -91,8 +91,9 @@ func (c *Client) Query(ctx context.Context, datasourceUID string, req QueryReque
 		return nil, fmt.Errorf("failed to read response: %w", err)
 	}
 
-	// Fall back to legacy /api/ds/query if K8s query API doesn't exist.
-	if resp.StatusCode == http.StatusNotFound {
+	// Fall back to legacy /api/ds/query if the K8s query API is unavailable
+	// (404) or forbidden for the user's role (403, e.g. Viewer).
```
dafydd-t:
```suggestion
	// (404) or forbidden for the user's role (403).
```

internal/query/loki/client.go
```diff
@@ -190,8 +191,9 @@ func (c *Client) MetricQuery(ctx context.Context, datasourceUID string, req Quer
...
```
dafydd-t:
```suggestion
	// (404) or forbidden for the user's role (403).
```

internal/query/prometheus/client.go
```diff
@@ -100,8 +100,9 @@ func (c *Client) Query(ctx context.Context, datasourceUID string, req QueryReque
 		return nil, fmt.Errorf("failed to read response: %w", err)
 	}
 
-	// Fall back to legacy /api/ds/query if K8s query API doesn't exist.
-	if resp.StatusCode == http.StatusNotFound {
+	// Fall back to legacy /api/ds/query if the K8s query API is unavailable
+	// (404) or forbidden for the user's role (403, e.g. Viewer).
```
dafydd-t:
```suggestion
	// (404) or forbidden for the user's role (403).
```

dafydd-t (commented):
some nits

internal/query/cloudwatch/client.go
dafydd-t: done in 88f14c7 - dropped the 'Viewer' example. the comment now covers any non-200 (not just 404/403) since the fallback was broadened in this PR, but the verbose parenthetical is gone as suggested

internal/query/grafanaquery/client.go
dafydd-t: done in 88f14c7 - dropped the 'Viewer' example. the comment now covers any non-200 (not just 404/403) since the fallback was broadened in this PR, but the verbose parenthetical is gone as suggested

internal/query/loki/client.go
dafydd-t: done in 88f14c7 - dropped the 'Viewer' example. the comment now covers any non-200 (not just 404/403) since the fallback was broadened in this PR, but the verbose parenthetical is gone as suggested

internal/query/loki/client.go
dafydd-t: done in 88f14c7 - dropped the 'Viewer' example. the comment now covers any non-200 (not just 404/403) since the fallback was broadened in this PR, but the verbose parenthetical is gone as suggested

internal/query/prometheus/client.go
dafydd-t: done in 88f14c7 - dropped the 'Viewer' example. the comment now covers any non-200 (not just 404/403) since the fallback was broadened in this PR, but the verbose parenthetical is gone as suggested

radiohead (approved):

PR #893, opened 2026-07-02
refactor(datasources): dedupe config-loading boilerplate

wardbekker:
## Summary
- 35 call sites across the prometheus, loki, tempo, pyroscope, cloudwatch, clickhouse, influxdb, infinity, and athena datasource commands duplicated the same block: load the full gcx config for the current context (warn and fall back to auto-discovery on failure), then load the Grafana REST config.
- Extracted this into `internal/datasources/query.LoadContextAndConfig` (plus a small `ConfigLoader` interface, mirroring the existing `DatasourceSaver` interface pattern) and updated every call site to use it, deleting the inlined block. Callers still invoke whichever resolver they used before (`ResolveAndSaveDatasource` or `ResolveValidateAndSaveDatasource`) — only the config-loading prefix was deduplicated, so this is a pure refactor with no behavior change.

## Test plan
- [x] go test -race -count=1 ./...
- [x] mise run lint
- [x] GCX_AGENT_MODE=false mise run reference (no diff — pure internal refactor)

### DIFF
--- a/internal/datasources/athena/describe_table.go
+++ b/internal/datasources/athena/describe_table.go
@@ -3,15 +3,12 @@ package athena
 import (
 	"encoding/json"
 	"fmt"
-	"log/slog"
 
 	"github.com/grafana/gcx/internal/agent"
-	internalconfig "github.com/grafana/gcx/internal/config"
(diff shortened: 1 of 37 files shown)


javiermolinar (approved):
awesome

PR #896, opened 2026-07-02
refactor(providers): dedupe HTTP client construction and error parsing

wardbekker:
## Summary
- Provider `client.go` files across the cloud-provider tier (`internal/providers/*`) hand-rolled the same `rest.HTTPClientFor` boilerplate and reimplemented an identical `handleErrorResponse` (read body, extract JSON `{"error"/"message"/"msg"}`, fall back to raw body, then status-only). Extracted both into shared helpers on `internal/providers`: `NewHTTPClient`, `ParseErrorBody`/`ParseErrorBytes`, and a single `ErrorResponse` type.
- Migrated `NewClient`-boilerplate call sites: `alert`, `faro`, `kg`, `slo/definitions`, `slo/reports`, `dashboards/search`, `aio11y/aio11yhttp`, `appo11y/overrides`, `appo11y/settings`, `k6/resource_adapter.go`, `irm` (incidents + oncall), `synth/provider.go`.
- Migrated error-parsing call sites and deleted the duplicated `ErrorResponse` struct variants (`alert`, `slo/definitions`, `slo/reports`, `irm/incidents`) plus `irm/oncall_client.go`'s divergent variant, normalized `aio11yhttp.HandleErrorResponse`'s body, and `appo11y`'s two `checkStatus` fallbacks. Deleted `internal/providers/synth/smcfg/http.go` entirely, repointing its three callers at the shared parser.
- Left alone (different error contracts, not simple duplicates): `kg`'s `readError`/`APIError`, `logs/adaptive` and `traces/adaptive`'s typed `APIError` (fail-package integration via `APIServiceName`/`IsServerError`), `k6/proxy_client.go` and `internal/providers/fleet/client.go` (per-call-site formatted errors, no generic duplication), and `internal/fleet`'s existing shared base client (the pattern this PR generalizes for the rest of the provider tier).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -count=1 ./...` (only pre-existing, environment-induced failures remain — `CLAUDECODE=1` triggers agent-mode confirmation gating in unrelated destructive-command tests; reproduced identically on an unmodified sibling worktree)
- [x] `mise run lint` (0 issues)
- [x] `GCX_AGENT_MODE=false mise run reference` (no diff)
- [x] `GCX_AGENT_MODE=false mise run all` (lint + tests + build + docs, exit 0)
- [x] `go build -buildvcs=false -o bin/gcx ./cmd/gcx/`


internal/providers/httperror.go
```diff
@@ -0,0 +1,71 @@
...
+	default:
+		return ""
+	}
+}
+
+// ParseErrorBody reads a non-2xx HTTP response body and returns a descriptive
+// error: the extracted JSON error message when the body unmarshals into
+// ErrorResponse, otherwise the raw body text, otherwise a status-only
+// message. It does not close resp.Body; callers remain responsible for that.
+func ParseErrorBody(resp *http.Response) error {
```
dafydd-t:
```suggestion
func HandleErrorResponse(resp *http.Response) error {
```

internal/providers/httperror.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("request failed with status %d (could not read body: %w)", resp.StatusCode, err)
+	}
+	return ParseErrorBytes(resp.StatusCode, body)
+}
+
+// ParseErrorBytes builds a descriptive error from an already-read non-2xx
+// status code and body. Used by clients that read a proxied response into
+// memory as []byte before this point (e.g. dual-mode datasource-proxy
+// transports), so there is no *http.Response to read from directly.
+func ParseErrorBytes(statusCode int, body []byte) error {
```
dafydd-t:
```suggestion
func FormatError(statusCode int, body []byte) error {
```

or something - these are not really parsing

internal/providers/httperror.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// maxErrorBodyBytes caps how much of a non-2xx response body ParseErrorBody
+// reads into memory. Error bodies are expected to be a short JSON message or
+// a short text blob; this guards against an unbounded read from a
+// misbehaving proxy or server.
+const maxErrorBodyBytes = 1 << 20 // 1 MiB
```
dafydd-t, line 12: we could probably make this a lot smaller right?

internal/providers/httpclient.go
```diff
@@ -0,0 +1,21 @@
...
+// config. This is the boilerplate shared by provider HTTP clients: call
+// rest.HTTPClientFor and wrap any error with context. Provider packages
+// should use this instead of hand-rolling the same rest.HTTPClientFor call.
+func NewHTTPClient(cfg config.NamespacedRESTConfig) (*http.Client, error) {
+	httpClient, err := rest.HTTPClientFor(&cfg.Config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP client: %w", err)
+	}
+	return httpClient, nil
+}
```
dafydd-t: I don't personally think this is worth it - we're swapping

```
rest.HTTPClientFor(&cfg.Config)
```

with 

```
providers.NewHTTPClient(cfg)
```

and with this new function, we're handling errors twice. its a lot of overhead for no use really

internal/providers/slo/definitions/client_test.go
```diff
@@ -125,7 +126,7 @@ func TestClient_Get(t *testing.T) {
 			uuid: "uuid-missing",
 			handler: func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(http.StatusNotFound)
-				writeJSON(w, definitions.ErrorResponse{Code: 404, Error: "SLO not found"})
```
dafydd-t, line 128: Aren't these codes relevant? have we lost them here?

dafydd-t (changes requested):
Not so sure about this one - see my comments!

internal/providers/httpclient.go
wardbekker: Agreed — dropped `NewHTTPClient` entirely and deleted `httpclient.go`. The 13 constructors now call `rest.HTTPClientFor(&cfg.Config)` directly again, with their original context-specific error messages restored (faro/kg/oncall/search). Only the error-parsing dedupe (the part with real duplication across ~14 clients) is kept.

internal/providers/httperror.go
wardbekker: Done — renamed to `HandleErrorResponse`.

internal/providers/httperror.go
wardbekker: Done — renamed to `FormatError(statusCode int, body []byte)`.

internal/providers/httperror.go
wardbekker: Tightened `httperror.go` (72→64 lines): idiomatic first-non-empty loop for the message selector and trimmed the doc comments. Behavior unchanged.

internal/providers/slo/definitions/client_test.go
wardbekker: Good catch to check — but nothing is lost. The old `handleErrorResponse` only ever formatted the HTTP status code (`resp.StatusCode`) plus the body's `error` message; it never read the body's `code` field. So the shared `ErrorResponse` omitting `Code` is behavior-preserving — the tests assert on the HTTP status (`"401"` etc.), which still comes through. The `Code` field was dead data in the old struct.

wardbekker:

Addressed @dafydd-t's review (commit `eb0247ee`). All local gates green (`GCX_AGENT_MODE=false mise run all`: lint 0 issues, tests pass, build + docs OK).

### Changes
- **Dropped `NewHTTPClient`** — deleted `httpclient.go`; the 13 constructors call `rest.HTTPClientFor(&cfg.Config)` directly again, with original context-specific error messages restored. Kept only the error-parsing dedupe (real duplication across ~14 clients).
- **Renamed error helpers** — `ParseErrorBody`→`HandleErrorResponse`, `ParseErrorBytes`→`FormatError(statusCode, body)` across all 50 call sites.
- **Tightened `httperror.go`** (72→64 lines), behavior unchanged.

### Review threads
| Comment | Resolution |
|---|---|
| `NewHTTPClient` not worth it | ✅ Dropped |
| rename `HandleErrorResponse` | ✅ Done |
| rename `FormatError` | ✅ Done |
| "make it smaller" | ✅ Tightened |
| SLO `Code` lost? | ✅ No loss — `Code` was never read by the old error formatter (dead field); tests assert on HTTP status |

Re-requesting review from @dafydd-t.

---
*Updated by shepherd-pr*

radiohead (approved):

dafydd-t (approved):

PR #901, opened 2026-07-02
feat(errors): did-you-mean suggestions and agent corrections for typo'd commands and flags

wardbekker:
## Problem

Agents often guess commands and flags that do not exist, such as `gcx dashbords` and `--outpt`. The current recovery signal has these problems:

- An unknown command lists all available commands. It does not give a close match.
- An unknown flag returns exit code 1. Usage errors must return exit code 2.
- The agent JSON error envelope does not contain a correction that an agent can run.

### Suggestions for unknown commands

Mistyped subcommands now use fuzzy matching. The new `internal/suggest` package uses Levenshtein distance and prefix matching. It checks command names and aliases. The correction keeps trailing arguments. For example, `gcx resources gt dashboards` suggests `gcx resources get dashboards`.

### Suggestions for unknown flags

The root `FlagErrorFunc` receives the command that failed. It matches the typed pflag `NotExistError` against local and inherited flags. It wraps the result in a `UsageError`. Unknown flags now return exit code 2 and suggest the closest flag. The `--log-http-payload` rename shim remains unchanged. `agentlog` classifies these errors as `usage_error`.

Suggested commands use shared POSIX shell quoting. They redact values for sensitive flags, including `--token` and `--cloud-token`.

### Machine-actionable corrections

The agent JSON envelope has a new optional `error.corrections` array. Each entry contains a complete corrected invocation. It also contains the target command `llm_hint` or the corrected flag help text.

```json
{"error": {"summary": "Invalid command usage", "exitCode": 2,
  "corrections": [{"command": "gcx resources get dashboards --output json",
                   "hint": "Output format. One of: agents, json, text, wide, yaml"}]}}
```

An agent can run `corrections[0].command` without a second discovery request.

## Compliance

- `docs/design/errors.md` documents the new `corrections` field.
- `docs/architecture/project-structure.md` lists the new `internal/suggest` and `internal/shellquote` utility packages.
- The change does not add a command or flag.

## Testing

- Table-driven tests cover matching, error envelopes, command typos, aliases, trailing arguments, flag typos, equals syntax, inherited flags, shorthand flags, sensitive value redaction, and shared shell quoting.
- `GCX_AGENT_MODE=false mise run all` passes.
- Manual agent-mode checks cover command typos, flag typos, and sensitive value redaction.

### DIFF
--- a/cmd/gcx/root/flagerror.go
+++ b/cmd/gcx/root/flagerror.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/grafana/gcx/cmd/gcx/fail"
 	"github.com/grafana/gcx/internal/gcxerrors"
+	"github.com/grafana/gcx/internal/shellquote"
 	"github.com/grafana/gcx/internal/suggest"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -32,7 +33,8 @@ func flagUsageError(cmd *cobra.Command, err error, invocationArgs []string) erro
 			suggestions = append(suggestions, fmt.Sprintf("Did you mean '%s'?", candidate))
 			correction := gcxerrors.Correction{Hint: flagUsageHint(cmd, candidate)}
 			if corrected, ok := substituteFlag(invocationArgs, unknown, candidate); ok {
-				correction.Command = shellJoin(append([]string{cmd.Root().Name()}, corrected...))
+				safe := redactSensitiveValues(corrected)
+				correction.Command = shellquote.Join(append([]string{cmd.Root().Name()}, safe...))
 				corrections = append(corrections, correction)
 			}
 		}
@@ -106,22 +108,46 @@ func substituteFlag(args []string, unknown, candidate string) ([]string, bool) {
 	return corrected, true
 }
 
-// shellJoin joins argv tokens into a shell command string, single-quoting any
-// token a POSIX shell would not pass through verbatim, so emitted corrections
-// stay runnable as-is.
-func shellJoin(tokens []string) string {
-	quoted := make([]string, len(tokens))
-	for i, token := range tokens {
-		quoted[i] = shellQuote(token)
-	}
-	return strings.Join(quoted, " ")
+// sensitiveFlagNames holds long-flag names whose values must never be
+// echoed back verbatim in a suggested correction, since they commonly carry
+// secrets such as auth tokens, passwords, or API keys (e.g. `gcx login
+// --token <value>`).
+var sensitiveFlagNames = map[string]struct{}{ //nolint:gochecknoglobals // constant-like lookup table; no mutable state.
+	"token":         {},
+	"password":      {},
+	"secret":        {},
+	"client-secret": {},
+	"api-key":       {},
 }
 
-const shellSafeChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-./=:@%+,"
+// redactSensitiveValues returns a copy of args with the value of any
+// sensitive flag (see sensitiveFlagNames) replaced by "<redacted>". Other
+// flags and positionals are left untouched so the correction stays runnable
+// as typed, minus the secret.
+func redactSensitiveValues(args []string) []string {
+	redacted := make([]string, len(args))
+	copy(redacted, args)
 
-func shellQuote(token string) string {
-	if token != "" && strings.Trim(token, shellSafeChars) == "" {
-		return token
+	redactNext := false
+	for i, arg := range redacted {
+		if redactNext {
+			redacted[i] = "<redacted>"
+			redactNext = false
+			continue
+		}
+		name, isLongFlag := strings.CutPrefix(arg, "--")
+		if !isLongFlag {
+			continue
+		}
+		if flag, value, hasEquals := strings.Cut(name, "="); hasEquals {
+			if _, sensitive := sensitiveFlagNames[flag]; sensitive && value != "" {
+				redacted[i] = "--" + flag + "=<redacted>"
+			}
+			continue
+		}
+		if _, sensitive := sensitiveFlagNames[name]; sensitive {
+			redactNext = true
+		}
 	}
-	return "'" + strings.ReplaceAll(token, "'", `'\''`) + "'"
+	return redacted
 }
--- a/cmd/gcx/root/flagerror.go
+++ b/cmd/gcx/root/flagerror.go
@@ -32,7 +32,7 @@ func flagUsageError(cmd *cobra.Command, err error, invocationArgs []string) erro
 			suggestions = append(suggestions, fmt.Sprintf("Did you mean '%s'?", candidate))
 			correction := gcxerrors.Correction{Hint: flagUsageHint(cmd, candidate)}
 			if corrected, ok := substituteFlag(invocationArgs, unknown, candidate); ok {
-				correction.Command = strings.Join(append([]string{cmd.Root().Name()}, corrected...), " ")
+				correction.Command = shellJoin(append([]string{cmd.Root().Name()}, corrected...))
 				corrections = append(corrections, correction)
 			}
 		}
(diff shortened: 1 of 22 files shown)


wardbekker:
## Shepherd status (iteration 1)

### Rebase
The branch was 58 commits behind `main`, and GitHub reported a merge conflict.
I rebased the 5 commits onto `main`. The pull request is now `MERGEABLE`.

The conflict was in `AGENTS.md` only. `main` replaced the full `internal/`
package list in `AGENTS.md` with a pointer to
`docs/architecture/project-structure.md`. I kept the version of `main`. The
entries for `internal/shellquote` and `internal/suggest` stay in
`docs/architecture/project-structure.md`, which this pull request already
updates. No code conflict occurred.

New head: `c4275175`.

### Local quality gate
| Check | Result |
|-------|--------|
| `go build ./...` | PASS |
| `go test ./...` | PASS |
| `golangci-lint` | PASS (0 issues) |
| Commit signatures | PASS (5 of 5) |

### Items that need a decision from the author
I did not change any code. Two points need a decision before a merge. I
reported the detail to the author in the session.

---
*Updated by shepherd-pr*

cmd/gcx/root/flagerror.go
```diff
@@ -0,0 +1,127 @@
...
+	corrections := []gcxerrors.Correction{}
+	// Fuzzy matching only makes sense for long flags; a mistyped shorthand
+	// is a single character away from most valid shorthands.
+	if notExist.GetSpecifiedShortnames() == "" {
+		unknown := "--" + notExist.GetSpecifiedName()
+		for _, candidate := range suggest.Candidates(unknown, availableFlagNames(cmd)) {
+			suggestions = append(suggestions, fmt.Sprintf("Did you mean '%s'?", candidate))
+			correction := gcxerrors.Correction{Hint: flagUsageHint(cmd, candidate)}
+			if corrected, ok := substituteFlag(invocationArgs, unknown, candidate); ok {
+				correction.Command = shellJoin(append([]string{cmd.Root().Name()}, corrected...))
```
radiohead: issue – this will echo back secret values (e.g. tokens in `gcx login --token`). We could reuse e.g. `agentlog.StripArgValues` or something like that to string values (at least the sensitive ones).

cmd/gcx/root/flagerror.go
```diff
@@ -0,0 +1,127 @@
...
+	corrected := make([]string, len(args))
+	copy(corrected, args)
+	corrected[matched] = candidate + strings.TrimPrefix(args[matched], unknown)
+	return corrected, true
+}
+
+// shellJoin joins argv tokens into a shell command string, single-quoting any
+// token a POSIX shell would not pass through verbatim, so emitted corrections
+// stay runnable as-is.
+func shellJoin(tokens []string) string {
```
radiohead: nit – we already have a few duplicate instances shell-quoting code (`internal/output/pagination.go`, `internal/providers/instrumentation/helm/format.go`), would be nice to refactor this to a shared helper(s).

radiohead (commented):
LGTM for the most part, just one concern about leaking sensitive data in suggestions.

cmd/gcx/root/flagerror.go
wardbekker: Fixed in 196b0e26: unknown-flag corrections now redact known-sensitive flag values (token, password, secret, client-secret, api-key) before echoing the corrected invocation back, so a typo elsewhere on the line no longer leaks e.g. `--token <value>` verbatim. Covered by new tests in `cmd/gcx/root/flagerror_test.go`.

cmd/gcx/root/flagerror.go
wardbekker: Fixed in 3c1814a9: extracted the shared shell-quoting logic into a new `internal/shellquote` package (`Quote`/`Escape`/`Join`) and switched `internal/output/pagination.go`, `internal/providers/instrumentation/helm/format.go`, and this file to use it — no more duplicate implementations. Behavior is preserved (helm keeps its always-quote semantics via `Escape`; the other two keep quote-only-when-needed via `Quote`/`Join`), verified by the existing golden tests plus new unit tests in `internal/shellquote`.

cmd/gcx/root/flagerror.go
wardbekker: Verified again after the rebase. Commit 05928c0e adds coverage for the real `--cloud-token` flag. The correction now redacts both `--token` and `--cloud-token` values. The exact `GCX_AGENT_MODE=false mise run all` gate passes.

cmd/gcx/root/validation.go
```diff
@@ -54,8 +58,22 @@ func ValidateArgs(rootCmd *cobra.Command, args []string) error {
...
 
-	commandPath := strings.TrimSpace(cmd.CommandPath())
+	candidates := subcommandCandidates(cmd, positionals[0])
+
 	suggestions := []string{}
+	corrections := []gcxerrors.Correction{}
+	for _, sub := range candidates {
+		// CommandPath is a space-separated path, not a single token, so it
+		// joins as-is; only the subcommand name and user tokens get quoted.
+		invocation := cmd.CommandPath() + " " + shellquote.Join(append([]string{sub.Name()}, positionals[1:]...))
```
radiohead, line 68: **required** — `ValidateArgs` builds the correction from parsed positional arguments, so accepted flags disappear. For example, `gcx --context staging resources puhs file.yaml` emits corrections without `--context staging`, which can send a retried mutation to the default stack; a target leaf flag such as `--output` also makes parent parsing stop before any command correction is produced. Please preserve the original argv and replace only the mistyped command token. Is there another execution layer that restores every explicit flag before an agent runs this correction?

internal/suggest/suggest.go
```diff
@@ -0,0 +1,75 @@
...
+			continue
+		}
+
+		seen[lower] = true
+		matches = append(matches, scored{token: candidate, distance: distance, order: i})
+	}
+
+	sort.SliceStable(matches, func(i, j int) bool {
+		if matches[i].distance != matches[j].distance {
+			return matches[i].distance < matches[j].distance
```
radiohead, line 58: **required** — Plain Levenshtein gives `puhs` the same distance from `pull` and `push`, so vocabulary order places `pull` in `corrections[0]`. The documented contract calls this list ranked and ready to run, which can make an agent execute the wrong operation. Please recognize adjacent transpositions when ranking, or otherwise ensure `push` wins this case. Should consumers avoid selecting the first correction despite the current ranked contract?

go.mod
```diff
@@ -42,6 +42,7 @@ require (
 	charm.land/huh/v2 v2.0.3
 	charm.land/lipgloss/v2 v2.0.6
 	github.com/NimbleMarkets/ntcharts/v2 v2.2.0
+	github.com/agnivade/levenshtein v1.2.1
```
radiohead, line 45: **required** — This promotes `agnivade/levenshtein` to a direct production dependency for one `ComputeDistance` call. `CONSTITUTION.md` requires a local implementation when the needed API fits within roughly 100 lines, and this helper is below that threshold. Please inline the distance implementation or obtain an explicit maintainer waiver. Is there a constraint that makes the local implementation materially larger or riskier?

radiohead (changes requested):
**Intent.** Add focused command and flag corrections to agent error output. Token and cloud token redaction are verified at this head.

Comments below are prefixed **required** (3).

**Fix summary.** Preserve the complete original invocation when correcting a command, rank adjacent transpositions ahead of unrelated equal distance matches, and replace the one call dependency with a small local implementation.

**Verdict.** Request changes because the current corrections can target the wrong context or operation, and the direct dependency conflicts with the project constitution.

---
**Generated by Codex**

cmd/gcx/root/validation.go
wardbekker: Fixed in 7eb9ef3. Command corrections now replace only the mistyped token in the original argv, preserving explicit context, flags, and positional arguments. Added coverage for flags on both sides of the command; the focused root tests pass.

internal/suggest/suggest.go
wardbekker: Fixed in 7eb9ef3. The local edit distance now counts adjacent transpositions as one edit, so puhs ranks push before pull. The focused suggestion tests pass.

go.mod
wardbekker: Fixed in 7eb9ef3. Removed the direct levenshtein dependency and replaced the one call with a local edit distance implementation in internal/suggest. go.mod is unchanged by the PR.

radiohead (approved):

PR #908, opened 2026-07-03
fix(logs): distinguish structured metadata from indexed stream labels

### LINKED ISSUE #907: gcx logs query merges structured metadata into stream labels, yielding invalid LogQL
## TLDR

`gcx logs query` flattens Loki's three kinds of per-line keys — indexed stream labels, structured metadata, and query-time parsed labels — into one undifferentiated `stream` map. So structured metadata like Loki's auto-added `detected_level` sits in `stream` right next to real indexed labels like `job` and `service_name`, with nothing marking which is which. Anyone (or any agent) who then drops a structured-metadata key into a `{...}` selector gets a silently-empty result, because it isn't indexed. Grafana's own dataframe already hands us the answer in a `labelTypes` field (`I`/`S`/`P` per key) and we throw it away.

This is the "grafana promotes labels to fields and gcx can't tell them apart" problem in practice.

## What's wrong

Structured metadata must be filtered *after* a pipe (`{job="x"} | detected_level="error"`), never inside `{}`. gcx gives no signal of which keys are which, so the distinction is invisible until a query silently returns nothing.

Proven against a live Grafana Cloud stack (anonymized):

`gcx logs query -d <loki-uid> '{service_name=~".+"}' -o json` returns a stream like:

```
"stream": {
  "cluster": "...", "job": "...", "service_name": "...",
  "detected_level": "info"   <- structured metadata, flat among indexed labels
}
```

Behaviour that proves the distinction is load-bearing:

- `gcx logs query '{detected_level="info"}'` -> `success`, **0 streams** (it's not in the index)
- `gcx logs query '{service_name=~".+"} | detected_level="info"'` -> **1 stream**, filtered correctly

Two secondary problems fall out of the same merge:

1. Streams are grouped by the *full* merged label map, so per-line structured metadata (e.g. a unique `execution_id` per line) explodes one logical stream into one entry per line.
2. `gcx logs labels` is actually fine (it proxies Loki's `/labels`, which is indexed-only), but the shipped agent skills claim things that push agents toward the trap anyway — a two-category "stream vs extracted" model that omits structured metadata entirely, plus examples that put `level`/`detected_level` inside `{}`.

## Root cause

gcx queries Loki via Grafana's datasource query API and reshapes the returned dataframe. The dataframe carries a `labelTypes` field mapping every key to `I` (indexed), `S` (structured metadata), or `P` (parsed) — present by default, no special request header needed. `convertLogLines` in `internal/query/loki/client.go` never looks at it; it runs the whole merged `labels` map into `StreamEntry.Stream`. There is zero structured-metadata awareness anywhere in the codebase.

## The fix

Parse the `labelTypes` field and split each line's keys into their three categories. Concretely:

- `StreamEntry.Stream` becomes **indexed labels only**, and streams group by indexed labels (fixes the per-line explosion too).
- Each log entry carries its own `structuredMetadata` and `parsed` maps, since both are per-line and can differ between lines in the same stream. This means the `gcx logs query` JSON `values` element changes from a `[ts, line]` tuple to a `{timestamp, line, structuredMetadata?, parsed?}` object. **This is a deliberate breaking change to the `logs query` output shape** — it's the whole point, and it matches how gcx already models rich per-item data (named structs, like tempo's typed attributes) rather than positional tuples. Blast radius is contained to the loki query path, the graph converter, and the table/raw formatters.
- Table output: `STREAM` column stays indexed-only; structured metadata + parsed labels surface in `DETAILS` (and `detected_level` fills `LEVEL` when the body carried none).
- No type info (older Grafana without `labelTypes`) -> everything defaults to indexed, so nothing is lost or moved unexpectedly.
- Harden the bundled agent skills: replace the two-category model with the three-category one, correct the examples that put non-indexed keys in `{}`, and point agents at the new `structuredMetadata`/`parsed` output so they build valid LogQL.

## Acceptance criteria

- [ ] `labelTypes` parsed; keys split into indexed / structured metadata / parsed
- [ ] `stream` contains indexed labels only; streams grouped by indexed labels
- [ ] per-line `structuredMetadata` and `parsed` in `logs query -o json`
- [ ] graceful fallback to indexed-only when `labelTypes` is absent
- [ ] table/wide/raw formatters and the graph converter updated for the new shape
- [ ] structured metadata visible (not merged into `STREAM`) in table output
- [ ] skills teach the three-category model, no examples put non-indexed keys in `{}`
- [ ] tests cover the split, indexed-only grouping, and the no-`labelTypes` fallback
- [ ] CHANGELOG note flagging the `logs query` output-shape change

Related: #425 (drove the label commands in), #426 (fixed null stream labels) — neither touched label-type classification.

rknightion:
## What

Fixes #907 — `gcx logs query` merged Loki's three kinds of per-line keys (indexed stream labels, structured metadata, query-time parsed labels) into one undifferentiated `stream` map. So structured metadata like Loki's auto-added `detected_level` sat flat among real indexed labels, and anyone who then put it in a `{...}` selector got a silently-empty result because it isn't indexed.

Grafana's Loki datasource already hands us a `labelTypes` field (`I`/`S`/`P` per key) in the dataframe — we were parsing right past it. This wires it up.

## Changes

- **Split by label type** (`internal/query/loki`): parse `labelTypes`; `StreamEntry.Stream` now holds indexed labels only, and streams group by indexed labels (which also fixes per-line structured metadata such as a unique `execution_id` exploding one stream into one entry per line). Each log entry carries its own `structuredMetadata` and `parsed` maps.
- **Table output**: `STREAM` stays indexed-only; structured metadata + parsed labels surface in `DETAILS`, and `detected_level` fills `LEVEL` when the body carried none.
- **Graceful fallback**: no `labelTypes` (older Grafana) → everything defaults to indexed, so nothing is lost or moved unexpectedly.
- **Skills hardening**: the bundled agent skills taught a two-category "stream vs extracted" model that omitted structured metadata and showed examples putting non-indexed keys inside `{}`. Replaced with the three-category model, corrected the examples, and pointed agents at the new `structuredMetadata`/`parsed` output so they build valid LogQL.

## Breaking change

The `gcx logs query` JSON/YAML output shape changed. Each element of a stream's `values` is now an object `{timestamp, line, structuredMetadata?, parsed?}` instead of a `[timestamp, line]` tuple, and `stream` contains indexed labels only. This matches how gcx already models rich per-item data (named structs, like tempo's typed attributes) rather than positional tuples. Blast radius is contained to the loki query path, the graph converter, and the table/raw formatters.

## Verification

- Full `go test ./...` green; `go vet` and `golangci-lint run` clean; `validate-skills` passes.
- Verified live against a Grafana Cloud stack: `gcx logs query -o json` now returns `stream` with indexed labels only and `detected_level` correctly under each entry's `structuredMetadata`; `{detected_level="info"}` returns 0 streams while `{...} | detected_level="info"` returns results, confirming the distinction is real.

### DIFF
--- a/claude-plugin/skills/debug-with-grafana/SKILL.md
+++ b/claude-plugin/skills/debug-with-grafana/SKILL.md
@@ -240,17 +240,36 @@ gcx logs query -d <loki-uid> \
     "resultType": "streams",
     "result": [
       {
-        "stream": {"job": "<service-name>", "level": "<level>"},
-        "values": [["<ns-timestamp>", "<log-line>"], ...]
+        "stream": {"job": "<service-name>"},
+        "values": [
+          {
+            "timestamp": "<ns-timestamp>",
+            "line": "<log-line>",
+            "structuredMetadata": {"detected_level": "error"},
+            "parsed": {"status": "500"}
+          }
+        ]
       }
     ]
   }
 }
 ```
+`stream` holds indexed labels only (valid inside `{...}`). `structuredMetadata`
+(per-line, e.g. `detected_level`) and `parsed` (from `| json` / `| logfmt`) are
+present only when the line has them, and are usable only after a pipe — not in a
+`{}` selector.
 
 > **LogQL pitfall**: Loki requires at least one non-empty label matcher in the
 > stream selector. `{}` and `{} |~ "pattern"` will be rejected. Always include
 > at least one label, e.g., `{job=~".+"}` as a catch-all.
+>
+> **Label-kind pitfall**: only *indexed* labels are valid inside `{...}`. Loki's
+> structured metadata (e.g. its auto-added `detected_level`) and parsed fields
+> are NOT — `{detected_level="error"}` returns zero streams silently. Filter
+> them after a pipe: `{job="app"} | detected_level="error"`. In `gcx logs query`
+> output, indexed labels are in the `stream` map / `STREAM` column; structured
+> metadata and parsed labels are in per-entry `structuredMetadata` / `parsed`
+> (`-o json`) or `DETAILS` (`-o table`).
 
 Look for:
 - Repeated error messages pointing to a specific code path or dependency
--- a/claude-plugin/skills/debug-with-grafana/references/error-recovery.md
+++ b/claude-plugin/skills/debug-with-grafana/references/error-recovery.md
@@ -127,6 +127,7 @@ Or for a range query:
 - The time range (`--from` / `--to`) falls outside the retention period or before the service was instrumented.
 - The label filters in the query are too restrictive (e.g., a `job` label value that does not exist).
 - The datasource is healthy but the service being queried is down and emitting no data.
+- **Loki-specific**: a structured-metadata or parsed key was placed inside the `{...}` selector. Only indexed stream labels are valid there. `{detected_level="error"}` matches nothing — Loki's `detected_level` is structured metadata, so it must go after a pipe: `{job="app"} | detected_level="error"`. Parsed fields need the parser stage first: `{job="app"} | json | status="500"`.
 
 ### Corrective Action
 
--- a/claude-plugin/skills/debug-with-grafana/references/query-patterns.md
+++ b/claude-plugin/skills/debug-with-grafana/references/query-patterns.md
@@ -178,7 +178,11 @@ Basic log filtering:
 # All logs from a job
 gcx logs query -d <loki-uid> '{job="varlogs"}'
 
-# Multiple labels (AND)
+# Multiple labels (AND) — every key inside {} must be an INDEXED label.
+# `level` works here only if it's indexed on your stack; Loki's auto-added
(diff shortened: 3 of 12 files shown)


wardbekker:

### CI Checks
| Check | Status |
|-------|--------|
| Linters | ✅ pass |
| Tests | ✅ pass |
| Documentation | ✅ pass |
| zizmor / TruffleHog / semgrep | ✅ pass |

### Sensibility Assessment
Problem **real** · Approach **sound** · Code **tight** → **SHIP**. No rework needed.

### Auto-Fixes Applied
None required — CI green on arrival, no review comments to triage.

### Pending
- **Awaiting review approval.** All checks pass; `mergeStateStatus` is `BLOCKED` pending a required approving review (`@wardbekker` requested).

---
*Updated by shepherd-pr*

wardbekker (approved):
Approved. CI green, code reviewed — correctly distinguishes Loki structured metadata / parsed labels from indexed stream labels via labelTypes, with safe fallback when the field is absent.

PR #914, opened 2026-07-05
feat(config): add exec-credential auth method

### LINKED ISSUE #913: Add exec-credential auth for Grafana behind an identity-aware proxy
## Problem

gcx authenticates with a service account token, basic auth, mTLS, or the Cloud OAuth flow. None of these work when Grafana sits behind an identity-aware proxy (Cloudflare Access, Google IAP, oauth2-proxy, Pomerium) that terminates auth at the edge.

The proxy rejects a Grafana token, because that token is for Grafana, not for the proxy's identity provider. The only credential it accepts is a short-lived token from that IdP, which has to be refreshed. gcx has no way to plug in an external token source, so these instances can't be reached over their real hostname.

## Proposal

Add `auth-method: exec`. gcx runs a configured helper that prints a bearer token, sends it as `Authorization: Bearer`, and caches and refreshes it. This is the kubeconfig credential-plugin mechanism, which client-go (already a dependency) implements via `rest.Config.ExecProvider`.

```yaml
contexts:
  behind-proxy:
    grafana:
      server: https://grafana.example.net
      auth-method: exec
      exec:
        command: my-token-helper
        args: [--audience, grafana]
        install-hint: "go install github.com/example/my-token-helper@latest"
```

The helper prints a `client.authentication.k8s.io/v1` ExecCredential. client-go owns the run, cache, and refresh loop, so gcx only adds config plumbing. The helper stays out of tree, since it's whatever the deployment needs: an OIDC wrapper, `cloudflared access token`, `gcloud auth print-identity-token`.

## Why exec

It fits the "config is kubeconfig" invariant, bakes in no IdP opinion, and works for any proxy that accepts a bearer. A gcx-native OIDC client would be larger and would tie gcx to one auth model.

## Alternatives

Point gcx at a bypass path (a tailnet or internal ingress that skips the proxy). Works, but leaves the public SSO hostname unusable from the CLI and depends on that bypass staying in place.

I have a working implementation and will open a PR.

amckinley:
## What

Adds `auth-method: exec`, which lets gcx authenticate to a Grafana instance behind an identity-aware proxy (Cloudflare Access, Google IAP, oauth2-proxy, Pomerium) that terminates auth at the edge. gcx runs a configured helper command, sends the token it prints as `Authorization: Bearer`, and caches and refreshes it.

Closes #913.

## Motivation

gcx supports service account tokens, basic auth, mTLS, and the Cloud OAuth flow. None of them work when a proxy verifies its own identity provider's token at the edge: a Grafana token is rejected there because it isn't for the proxy's IdP, and the token the proxy does accept is short-lived and needs refreshing. There's no way to plug in an external token source today, so these instances are unreachable over their real hostname.

## Approach

This is the kubeconfig [credential-plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins) mechanism, which client-go (already a dependency) implements via `rest.Config.ExecProvider`. The helper prints a `client.authentication.k8s.io/v1` ExecCredential; client-go owns the run, parse, in-memory cache, and refresh-on-expiry loop, so gcx only adds config plumbing. The exec authenticator composes with gcx's existing transport wrappers (retry, logging, caller-id).

The helper stays out of tree, since it's whatever the deployment needs: an OIDC wrapper, `cloudflared access token`, `gcloud auth print-identity-token`.

```yaml
contexts:
  behind-proxy:
    grafana:
      server: https://grafana.example.net
      auth-method: exec
      exec:
        command: my-token-helper        # required
        args: [--audience, grafana]     # optional
        env: [{name: FOO, value: bar}]  # optional; unioned with process env
        install-hint: "go install ..."  # optional; shown if command not found
        interactive-mode: IfAvailable   # optional; Never | IfAvailable | Always
```

## Scope

The change is confined to the exec method. The existing oauth/token/basic auth arms in `NewNamespacedRESTConfig` are unchanged. The exec arm is placed first in the auth switch and gated on `InferredAuthMethod() == "exec"`, so an explicit `auth-method: exec` is honored even if stale credential fields from a prior login or a merged config layer linger; the OAuth-pairing validation is skipped for exec contexts for the same reason. `exec.env` values are redacted in `config view`. No secret is written to the config file — the credential is fetched at runtime.

## Verification

- `mise run lint`, `mise run tests`, and `mise run reference-drift` all pass.
- Exercised end-to-end against a self-hosted Grafana behind an identity-aware proxy: `gcx dashboards list` authenticates over the proxied hostname, the helper runs once and the token is reused on subsequent calls, and role projection through the proxy resolves correctly.


wardbekker:
## 🐑 Shepherd Status

Automated verification pass on this PR. **No code changes made** — reporting only.

### Sensibility assessment
| Axis | Verdict | Note |
|------|---------|------|
| Problem | ✅ real | No existing auth arm accepts an externally-minted, refreshable token; proxy-fronted Grafana is genuinely unreachable today (#913). |
| Approach | ✅ sound | Reuses client-go `rest.Config.ExecProvider` (kubeconfig credential-plugin); config-storage / token-attachment split matches the documented auth-system layering. |
| Tightness | ✅ tight | Thin mapping onto client-go, no reinvention, scope confined to exec auth, no back-compat cruft. Comments explain non-obvious *why* (redactor recursion, switch ordering) and are worth keeping. |

### Local quality gate (in worktree, `GCX_AGENT_MODE=false`)
| Check | Status |
|-------|--------|
| `mise run lint` | ✅ 0 issues |
| `go test ./...` | ✅ pass |
| `go build` | ✅ ok |
| `mise run reference-drift` | ✅ no drift |

### CI
| Check | Status |
|-------|--------|
| semgrep-cloud-platform/scan | ✅ pass |

### Merge readiness
`MERGEABLE` · `BLOCKED` — waiting on a **human review approval** (no reviewer has weighed in yet). Nothing else outstanding.

---
*Posted by shepherd-pr*

jamiegs:
I've got Grafana sitting behind Okta OIDC auth in an AWS ALB. 

To be able to use gcx I would I need to be able to set a `AWSELBAuthSessionCookie-0=<token>` cookie. 

Would it be possible to add support for setting and sending cookies with every request as well?

amckinley:
@jamiegs fwiw I hacked around the current limitations when I realized that the `token` field in the gcx config that's supposed to be used for grafana cloud access is passed through transparently, regardless of its contents. My working config is:

```
    grafana:
      server: https://grafana.internal-domain-name.net
      token: keychain:gcx:v2:blahblah
      auth-method: token
      org-id: 1
 ```

PR #920, opened 2026-07-07
feat(coreapi): shared HTTP client for core Grafana /api/* providers

wardbekker:
Part **1 of 5** of the stack that splits #918 into reviewable parts.

## What

This PR adds `internal/coreapi`. This package provides a base HTTP client for legacy Grafana core `/api/*` REST endpoints.

- `DoJSON` and `DoJSONNotFound` marshal a request body, run the request, check the status, and decode the JSON response.
- `DoStatus` and `DoStatusNotFound` run the same flow for endpoints that have no useful response body.
- `HandleErrorResponse` uses the shared provider error handler. It limits the response body to 1 MiB and supports the common Grafana error fields.

The client uses `rest.HTTPClientFor`. This keeps the configured authentication, logging, and payload dump transport.

## Why a stack

#918 ported four independent provider packages on top of this helper. These packages support annotations, organization users, public dashboards, and permissions. The full change has 64 files and 5,738 added lines. The stack keeps each review small.

## Stack

1. **`coreapi` (this PR)** → `main`
2. annotations → this PR
3. organization users → annotations
4. public dashboards → organization users
5. permissions → public dashboards

## Test plan

- [x] `GCX_AGENT_MODE=false mise run lint`
- [x] `GCX_AGENT_MODE=false mise exec -- go test ./internal/coreapi/...`
- [x] `GCX_AGENT_MODE=false mise exec -- go test ./...`
- [x] `GCX_AGENT_MODE=false mise run docs`
- [x] `GCX_AGENT_MODE=false mise run all`

### DIFF
--- a/internal/coreapi/client.go
+++ b/internal/coreapi/client.go
@@ -58,13 +58,12 @@ func (c *Client) DoRequest(ctx context.Context, method, path string, body io.Rea
 
 // DoJSON executes an HTTP request against the core API, optionally marshaling
 // body as the JSON request payload, and decodes a JSON response into Resp. Pass
-// a nil body for requests with no request body (e.g. GET/DELETE) — in that case
-// Req can be instantiated as `any`.
+// a nil body for requests with no request body (e.g. GET/DELETE).
 //
 // Any response status not present in okStatuses is treated as an error via
 // HandleErrorResponse.
-func DoJSON[Req, Resp any](ctx context.Context, c *Client, method, path string, body *Req, okStatuses ...int) (Resp, error) {
-	return doJSON[Req, Resp](ctx, c, method, path, body, nil, okStatuses)
+func DoJSON[Resp any](ctx context.Context, c *Client, method, path string, body any, okStatuses ...int) (Resp, error) {
+	return doJSON[Resp](ctx, c, method, path, body, nil, okStatuses)
 }
 
 // DoJSONNotFound behaves like DoJSON, but returns notFoundErr instead of the
@@ -72,11 +71,11 @@ func DoJSON[Req, Resp any](ctx context.Context, c *Client, method, path string, 
 // typically pass a sentinel wrapped with request-specific context, e.g.
 // fmt.Errorf("%s: %w", id, ErrNotFound), so that errors.Is still matches the
 // resource-specific sentinel.
-func DoJSONNotFound[Req, Resp any](ctx context.Context, c *Client, method, path string, body *Req, notFoundErr error, okStatuses ...int) (Resp, error) {
-	return doJSON[Req, Resp](ctx, c, method, path, body, notFoundErr, okStatuses)
+func DoJSONNotFound[Resp any](ctx context.Context, c *Client, method, path string, body any, notFoundErr error, okStatuses ...int) (Resp, error) {
+	return doJSON[Resp](ctx, c, method, path, body, notFoundErr, okStatuses)
 }
 
-func doJSON[Req, Resp any](ctx context.Context, c *Client, method, path string, body *Req, notFoundErr error, okStatuses []int) (Resp, error) {
+func doJSON[Resp any](ctx context.Context, c *Client, method, path string, body any, notFoundErr error, okStatuses []int) (Resp, error) {
 	var zero Resp
 
 	var reader io.Reader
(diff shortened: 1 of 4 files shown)


internal/coreapi/client.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// NewClient creates a new core API client from a Grafana REST config.
+func NewClient(cfg config.NamespacedRESTConfig) (*Client, error) {
+	httpClient, err := rest.HTTPClientFor(&cfg.Config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP client: %w", err)
+	}
+	return &Client{restConfig: cfg, httpClient: httpClient}, nil
+}
```
dafydd-t, line 38: can we use the httputils client here? or at least embed it

internal/coreapi/client.go
```diff
@@ -0,0 +1,192 @@
...
+// DoJSON executes an HTTP request against the core API, optionally marshaling
+// body as the JSON request payload, and decodes a JSON response into Resp. Pass
+// a nil body for requests with no request body (e.g. GET/DELETE) — in that case
+// Req can be instantiated as `any`.
+//
+// Any response status not present in okStatuses is treated as an error via
+// HandleErrorResponse.
+func DoJSON[Req, Resp any](ctx context.Context, c *Client, method, path string, body *Req, okStatuses ...int) (Resp, error) {
+	return doJSON[Req, Resp](ctx, c, method, path, body, nil, okStatuses)
+}
```
dafydd-t, line 67: I don't really want to spread the use of generics purely for the benefit of the DRY principle, i'd prefer the duplication

internal/coreapi/client.go
dafydd-t: Maybe we could use this existing grafana client? https://github.com/grafana/gcx/blob/7941f050d60cd0badcf27f41c9de92f5a324be5a/internal/grafana/client.go#L99

wardbekker:

### CI Checks
All required checks passed on the previous head; re-running on `d49d008e` (rebase onto latest main + one refactor commit).

### Auto-Fixes Applied
- `refactor(coreapi): drop unnecessary Req type parameter in favor of body any` — removes the awkward `[any, T]` instantiations at nil-body call sites; `Resp` is kept where needed for decoding. Partially addresses the generics feedback by shrinking the generic surface.

### Pending Review
- Reviewer feedback on client reuse (httputils / internal/grafana) and generics is being discussed with the author — responses to follow on the review threads.

### Escalations
- None auto-fixable; see review threads.

---
*Updated by shepherd-pr*

wardbekker:
@dafydd-t thanks for the review! A couple of quick notes before I rework anything:

- **Client reuse**: `httputils` is a plain `*http.Client` (logging/retry only), no Grafana host/auth/TLS resolution — not a drop-in here. `internal/grafana/client.go` is currently only used for `/api/health` version checks and has no coverage for one of coreapi's planned consumers (public dashboards). The `coreapi.NewClient` shape actually mirrors the same `rest.HTTPClientFor(&cfg.Config)` pattern already used in ~29 other files across providers (aio11y, alert, faro, irm, kg, k6, slo, synth, etc.), so it's following existing convention rather than introducing a new one.
- **Generics**: the `DoJSON`/`DoStatus` shape here is the same pattern from `internal/providers/aio11y/aio11yhttp/client.go`, merged in #895 — which you approved. Happy to align on reverting both if you'd rather we walk that back too, but wanted to flag the precedent first.

Let me know if you still want this reworked given the above, or if this unblocks an approval.

internal/coreapi/client.go
wardbekker: Neither quite fits, unfortunately: `httputils` deliberately carries no Grafana auth, and `internal/grafana` wraps the generated OpenAPI client, so it can't hit arbitrary `/api/*` paths. `coreapi` uses `rest.HTTPClientFor` to inherit the configured auth/logging transport — same pattern as `assistanthttp` and `aio11yhttp`. Agreed we have too many base clients though; consolidating those is good follow-up (à la #896), but I'd rather not block the stack on it.

internal/coreapi/client.go
wardbekker: This matches what's already merged in `aio11yhttp` (#895) — same generic `DoJSON`/`DoStatus` set. The alternative is copying the marshal/execute/decode boilerplate into all four stacked providers. I did trim it in d49d008e: `Req` is gone (`body any`), so only the response type is generic and `DoStatus` has no type params at all now.

wardbekker:
@dafydd-t re-requesting review — both threads have replies and the generics surface was trimmed in d49d008e (`Req` type param dropped, `DoStatus` has no type params). All checks are green. This unblocks the stacked providers in #921, which is now conflict-free and carries the annotations/org/public-dashboards/permissions work.

dafydd-t:
@claude review

dafydd-t (changes requested):

PR #925, opened 2026-07-07
perf(resources): build discovery registry once, cap fan-out errgroups

wardbekker:
## Changes

Two small resources-pipeline fixes surfaced while profiling read commands.

### 1. Discovery registry built twice per command
`FetchResources` builds a discovery registry, then `NewDefaultPuller` built a **second** one internally. HTTP is disk-cached (10-min TTL) so there was no double network hit, but the in-process index build + `adapter.RegisterAll` ran twice on every resource command. Added `NewDefaultPullerWithRegistry` and reuse the already-built registry; `NewDefaultPuller` now delegates to it.

### 2. Two uncapped fan-out errgroups
Both contradicted the documented default concurrency of 10:
- `NamespacedClient.GetMultiple` (one GET per comma-separated name) had `SetLimit` commented out — a connection storm on `get x/a,b,c,...`.
- `Puller.Pull` (one List per resource type on a bare `get`) had no limit.

Both now `SetLimit(10)`, matching the existing `adapter/router.go` precedent.

## Tests

Existing resources-pipeline tests are limit-agnostic and pass unchanged (results are written by index, so ordering is preserved). Full suite green.

### DIFF
--- a/internal/resources/dynamic/namespaced_client.go
+++ b/internal/resources/dynamic/namespaced_client.go
@@ -15,6 +15,11 @@ import (
 	"k8s.io/client-go/tools/pager"
 )
 
+// maxConcurrentGetRequests bounds the per-name fan-out in GetMultiple, matching
+// the fan-out cap used elsewhere in the resources pipeline (adapter router,
+// puller).
+const maxConcurrentGetRequests = 10
+
 // NamespacedClient is a dynamic client with a namespace and a discovery registry.
 type NamespacedClient struct {
 	namespace string
@@ -94,9 +99,7 @@ func (c *NamespacedClient) GetMultiple(
 	ctx context.Context, desc resources.Descriptor, names []string, opts metav1.GetOptions,
 ) ([]unstructured.Unstructured, error) {
 	g, ctx := errgroup.WithContext(ctx)
-
-	// TODO: consider using a limit
-	// g.SetLimit(maxConcurrentGetRequests)
+	g.SetLimit(maxConcurrentGetRequests)
 
 	res := make([]unstructured.Unstructured, len(names))
 
--- a/internal/resources/dynamic/namespaced_client.go
+++ b/internal/resources/dynamic/namespaced_client.go
@@ -15,33 +15,52 @@ import (
 	"k8s.io/client-go/tools/pager"
 )
 
-// maxConcurrentGetRequests bounds the per-name fan-out in GetMultiple, matching
-// the fan-out cap used elsewhere in the resources pipeline (adapter router,
-// puller).
-const maxConcurrentGetRequests = 10
+// defaultMaxConcurrentGetRequests bounds the per-name fan-out in GetMultiple by
+// default, matching the fan-out cap used elsewhere in the resources pipeline
+// (adapter router, puller). Override via WithMaxConcurrentGetRequests.
+const defaultMaxConcurrentGetRequests = 10
 
 // NamespacedClient is a dynamic client with a namespace and a discovery registry.
 type NamespacedClient struct {
-	namespace string
-	client    dynamic.Interface
+	namespace                string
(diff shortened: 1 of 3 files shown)


internal/resources/dynamic/namespaced_client.go
```diff
@@ -15,6 +15,11 @@ import (
 	"k8s.io/client-go/tools/pager"
 )
 
+// maxConcurrentGetRequests bounds the per-name fan-out in GetMultiple, matching
+// the fan-out cap used elsewhere in the resources pipeline (adapter router,
+// puller).
+const maxConcurrentGetRequests = 10
```
radiohead: Small nit – this should be part of the client config – default to 10 but allow overrides.

internal/resources/remote/puller.go
```diff
@@ -115,6 +127,7 @@ func (p *Puller) Pull(ctx context.Context, req PullRequest) (*OperationSummary,
 	logger.Debug("Pulling resources")
 
 	errg, ctx := errgroup.WithContext(ctx)
+	errg.SetLimit(maxConcurrentListRequests)
```
radiohead: Same here – could use a config field instead of a const.

radiohead (dismissed):
LGTM, one nit about making the limit configurable.

internal/resources/dynamic/namespaced_client.go
wardbekker: Addressed in 8617027: `maxConcurrentGetRequests` is now a `NamespacedClient` field (default 10, unchanged) with a `WithMaxConcurrentGetRequests(n int)` functional-option override, matching the existing options pattern in `internal/config/loader.go`.

internal/resources/remote/puller.go
wardbekker: Addressed in 8617027: `maxConcurrentListRequests` is now a `Puller` field (default 10, unchanged) with a `WithMaxConcurrentListRequests(n int)` functional-option override, same pattern as the `NamespacedClient` fix above.

wardbekker:

### Sensibility Assessment (Stage 1.5)
Problem: **real** (high confidence) — double registry build + two uncapped fan-out errgroups confirmed in code.
Approach: **sound** (high confidence) — `NewDefaultPullerWithRegistry` matches existing `WithX` constructor conventions; `SetLimit(10)` matches `adapter/router.go` precedent.
Tightness: **loose** → tightened — `NewDefaultPuller` had lost its only caller within this PR; removed as dead code per repo convention (no back-compat shims kept around).
Overall: tightened, then proceeded.

### Auto-Fixes Applied (commit 8617027)
- `internal/resources/dynamic/namespaced_client.go`: `maxConcurrentGetRequests` is now a `NamespacedClient` field (default 10, unchanged) with a `WithMaxConcurrentGetRequests(n int)` functional-option override — addresses @radiohead's nit.
- `internal/resources/remote/puller.go`: `maxConcurrentListRequests` is now a `Puller` field (default 10, unchanged) with a `WithMaxConcurrentListRequests(n int)` functional-option override — addresses @radiohead's nit.
- Removed `NewDefaultPuller` (dead code — its only caller now calls `NewDefaultPullerWithRegistry` directly).
- Rebased onto latest `main` (branch was 14 commits behind, including the module-mode lint config change in #884).

### CI Checks
All required checks pass on the latest commit (8617027): CI (Tests/Linters/Documentation), zizmor, TruffleHog, semgrep.

### Pending Review
Both inline nits from @radiohead are code-fixed and replied to inline. **The prior APPROVED review was dismissed by the new commits** — `reviewDecision` is now empty and `mergeStateStatus` is `BLOCKED` pending a fresh review.

---
*Updated by shepherd-pr*

dafydd-t (approved):

PR #928, opened 2026-07-07
perf(config): check contexts concurrently in gcx config check

wardbekker:
> **Stacked on #927** (`perf/skip-bootdata-when-stackid-set`) — it depends on that PR's `Validate(ctx)` signature change. Base will retarget to `main` once #927 merges.

## Problem

`gcx config check` iterated contexts **serially**, and each context did two **sequential** network round-trips (discovery connectivity probe + `GetVersion`). Cost scaled as K contexts × 2 round-trips — **~3.75s across 8 contexts**.

## Change

Run contexts under a bounded `errgroup` (`SetLimit(10)`), each writing into its own `bytes.Buffer`. Buffers are flushed in **sorted-name order**, so output is now deterministic (the prior serial loop ranged a map, so ordering was already non-deterministic). Errors are collected per-index and goroutines return `nil`, so one failing context never cancels the others — preserving the prior "check every context, return the last error" behavior. `checkContext` now takes `(ctx, io.Writer)` instead of `*cobra.Command`.

Concurrency-safe: each goroutine builds its own `restCfg`; token-refresh writes are already `flock`-guarded across contexts.

## Result (live, 8 contexts)

| | Before | After |
|---|---|---|
| `config check` | ~3.75s | **~1.1s** (bounded by the slowest single context) |

Output verified byte-identical across repeated runs.

### DIFF
--- a/cmd/gcx/config/command.go
+++ b/cmd/gcx/config/command.go
@@ -27,6 +27,7 @@ import (
 	"github.com/grafana/gcx/internal/terminal"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/sync/errgroup"
 )
 
 type Options struct {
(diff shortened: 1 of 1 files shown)


dafydd-t:
> Buffers are flushed in sorted-name order, so output is now deterministic 

🎉

dafydd-t (dismissed):

wardbekker:

### Actions Taken
- Rebased onto latest `main` (twice — base advanced mid-run); commit `164cc3f2` from stacked PR #927 dropped as already merged.
- Sensibility assessment: **SHIP** (problem real, approach sound, code tight).
- Simplifier pass: unified writer name in `checkContext` (dropped `stdout := w` alias) — `a9d6f5b6`.
- Full local gate (lint + tests + build + docs) passed.

### CI Checks
Re-running after push `a9d6f5b6`; previous run on the rebase was fully green.

### Pending Review
- Awaiting 2 qualifying approvals (prior approval was dismissed by the rebase push).

### Escalations
- None.

---
*Updated by shepherd-pr*

awsome-o (dismissed):

annanay25 (dismissed):
i would add a test for concurrency but non-blocking!

annanay25 (approved):
lgtm

PR #930, opened 2026-07-07
feat(fleet)!: route fleet and instrumentation via collector-app proxy

radiohead:
## What

`gcx fleet` and `gcx instrumentation` now reach the Grafana Fleet Management API through the `grafana-collector-app` app plugin proxy at `cfg.Host` (`/api/plugin-proxy/grafana-collector-app/fleet-management-api/...`), built with `rest.HTTPClientFor(&cfg.Config)`, instead of calling Fleet Management directly over Basic auth. Config resolution switches from `LoadCloudConfig` to `LoadGrafanaConfig` for both command groups, and the Cloud/Fleet Management access-policy token is no longer required to use either command group.

Several pieces of the old direct-FM transport are removed as a result:

- Client-side `X-Prom-*`/`X-Scope-OrgID` header decoration, the `PromHeaders` type, and `PromHeadersFromStack` are deleted — the proxy injects these headers server-side.
- `fleet.Client.DoRequestWithHeaders` is removed in favor of a single `DoRequest` method; typed method signatures on the instrumentation client (`SetupK8sDiscovery`, `RunK8sDiscovery`, `RunK8sMonitoring`, `IsNamespaceDiscovered`) drop their `promHeaders` parameter.
- Instrumentation now sources backend datasource URLs (Mimir/Loki/Tempo/Pyroscope) from the collector-app's Viewer-role instance-metadata proxy route (`/api/plugin-proxy/grafana-collector-app/grafanacom-api/instances/`) instead of a separate Cloud API call.
- HTTP 401/403 responses from the proxy are mapped to Grafana auth/RBAC guidance — re-authenticate with `gcx login`, or acquire the Viewer/Grafana Admin role — instead of Cloud access-policy-scope hints.

Typed method surfaces, Connect service/method path suffixes, and the JSON wire format are unchanged. `CONSTITUTION.md`'s external-API carve-out no longer lists Fleet, since fleet and instrumentation traffic now targets `cfg.Host` like the other plugin-proxied providers. The CLI reference, `docs/design/errors.md`, and related architecture docs are updated to describe the new transport and the Viewer/Grafana Admin role split.

## Why

Fleet Management was previously reached directly with Basic auth, requiring a separate Cloud access-policy token (`fleet-management:read`/`fleet-management:write`) on top of the Grafana credential every other command needs. That diverged from the pattern the rest of the plugin-proxied providers follow — reaching `cfg.Host` through the shared bearer transport `rest.HTTPClientFor` already injects — and meant a user authenticated to Grafana via OAuth still could not use `gcx fleet` or `gcx instrumentation` without also minting and scoping a Cloud token.

Routing both command groups through the `grafana-collector-app` proxy removes that redundant credential requirement and lets role enforcement happen where it is observable and actionable: a missing Viewer or Grafana Admin role produces a clear re-authentication or role-request suggestion, rather than an opaque Cloud-scope error. An alternative hybrid transport that kept the Cloud token as a fallback path was considered and rejected in favor of a single, consistent credential path across both command groups.

## How it was tested

- [x] `GCX_AGENT_MODE=false go test -race -count=1 ./...` passed across the full module
- [x] Scoped `golangci-lint` run against the changed packages (`internal/fleet`, `internal/providers/fleet`, `internal/providers/instrumentation`, `cmd/gcx/instrumentation/...`, `cmd/gcx/fail`) reported no issues
- [x] `mise run reference` regenerated the CLI reference docs; `docs/reference/cli/gcx_fleet.md` and `docs/reference/cli/gcx_instrumentation.md` now document the Viewer/Grafana Admin role split and no longer reference a Cloud token
- [x] Passed the `/review-spec` five-check gate (verdict PASS after a 3-round fix loop)

BREAKING CHANGE: `gcx fleet` and `gcx instrumentation` now authenticate through the `grafana-collector-app` plugin proxy instead of calling Fleet Management directly with Basic auth. `internal/providers/instrumentation.PromHeaders`, `PromHeadersFromStack`, and `fleet.Client.DoRequestWithHeaders` are removed. A context carrying only a Cloud/Fleet-Management access-policy token (no Grafana credential) no longer authenticates; reads require the Viewer role and mutations require the Grafana Admin role.

---
**Generated by Claude Code**

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -128,6 +128,8 @@ Action-verb command tree for Grafana Cloud's Instrumentation Hub. Backed by flee
 
 Uses `internal/providers/instrumentation/` (provider, types, output codecs, RMW helper, helm formatter, enumeration helper) and `internal/fleet/` (shared base HTTP client, also used by the fleet provider). See ADR-018 for the design.
 
+**Transport.** Both `fleet` and `instrumentation` reach the Fleet Management API through the `grafana-collector-app` app plugin proxy at `cfg.Host` (`/api/plugin-proxy/grafana-collector-app/fleet-management-api/…` + the unchanged Connect service/method path), built with `rest.HTTPClientFor(&cfg.Config)`. The Grafana bearer credential (OAuth included) is injected by the k8s round-tripper; the proxy injects the FM credential and the `X-Prom-*`/`X-Scope-OrgID` headers server-side. `instrumentation` sources backend datasource URLs from the collector-app's Viewer-role instance-metadata proxy route (`/api/plugin-proxy/grafana-collector-app/grafanacom-api/instances/`, decoded into `cloud.StackInfo`). No Cloud access-policy token is required; reads need the Viewer role and mutations the Grafana Admin role. See ADR-021.
+
 ### 7. Configuration
 
 kubectl-inspired context-based multi-environment configuration.
@@ -163,7 +165,7 @@ Multiple auth mechanisms for different tiers.
 | **Basic auth** | Legacy Grafana instances | Username/password in `rest.Config` |
 | **Adaptive auth** | Signal provider adaptive telemetry APIs | `internal/auth/adaptive/` — GCOM-cached Basic auth shared across signal providers |
 
-**Precedence:** Token > OAuth > user/password. Explicit flags override env vars override config file. `httputils.NewDefaultClient(ctx)` must be used for APIs outside the Grafana server (k6 Cloud, Synth, Fleet) — the k8s transport injects the Grafana bearer token on every request, which conflicts with product-specific auth.
+**Precedence:** Token > OAuth > user/password. Explicit flags override env vars override config file. `httputils.NewDefaultClient(ctx)` must be used for APIs outside the Grafana server (k6 Cloud, OnCall) — the k8s transport injects the Grafana bearer token on every request, which conflicts with product-specific auth. `fleet`/`instrumentation` are **not** in that set: they reach Fleet Management through the `grafana-collector-app` plugin proxy at `cfg.Host` via `rest.HTTPClientFor()` (ADR-021), so the Grafana bearer is the correct credential and no Cloud access-policy token is required.
 
 **Deep-dive:** [client-api-layer.md](docs/architecture/client-api-layer.md), [config-system.md](docs/architecture/config-system.md).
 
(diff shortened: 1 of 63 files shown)


wardbekker:
## Shepherd Status (iteration 2) — ⚠️ live-test escalation

### CI Checks
All green (Linters, Tests, Documentation, semgrep, TruffleHog, zizmor) on the current head.

### Maintenance Applied
- Rebased onto latest `main` (CHANGELOG + login scope-hint conflicts resolved)
- `b144aea` `refactor(fleet)`: single-emission of the 401/403 role guidance, shared `fleet.StatusError` constructor, trimmed comments (net −71 lines, behavior-preserving)

### ⚠️ Escalation: OAuth sessions are denied by the collector-app proxy routes (live-tested)
Extensive live testing against two production Grafana Cloud stacks (Grafana 13.2.0, grafana-collector-app 4.22.1, caller has org role **Admin** and holds both `grafana-collector-app:read` and `grafana-collector-app:admin` RBAC actions) shows **every** fleet/instrumentation command fails:

```
gcx --context <my-context> fleet pipelines list
→ fleet: list pipelines: status 403: {"message":"plugin proxy route access denied"}
```

A/B isolation on the identical route (`POST /api/plugin-proxy/grafana-collector-app/fleet-management-api/pipeline.v1.PipelineService/ListPipelines`, same stack):
- **Service-account token → HTTP 200** (real pipeline data)
- **gcx OAuth (grafana.com) token → HTTP 403** (`plugin proxy route access denied`)

Diagnosis: collector-app is frontend-only, so its proxy routes are gated per-route on plugin-declared RBAC actions (`reqAction: grafana-collector-app:read|admin`). Grafana Cloud's scope→permission mapping for external OAuth sessions (gcx requests `grafana-api:read/write/delete`) does not cover app-plugin custom actions, so OAuth callers are denied regardless of their actual role. The providers cited as precedent (slo, irm, k6, …) work with OAuth because they are **backend** plugins reached via `/api/plugins/<id>/resources/...` and declare no proxy routes — the plugin-proxy reqAction gate is unique to this transport.

Consequences as merged:
1. The ADR-021 headline benefit ("any Grafana credential — OAuth included — now works") does not hold on current production stacks; OAuth is gcx's primary login flow.
2. Contexts that today work via a Cloud access-policy token would **regress** (old direct-FM path removed, new path 403s), leaving no working transport for fleet/instrumentation except Grafana service-account-token auth.

Possible directions (author's call): a collector-app/Grafana-side fix to admit extended-JWT sessions on these routes (or drop `reqAction` in favor of `reqRole`); an additional OAuth scope gcx could request; or retaining the direct-FM transport until the platform side lands.

### Pending Review
- 0/2 approvals. Recommend **holding merge** until the escalation above is resolved.

---
*Updated by shepherd-pr*

radiohead:
Superseded by #1261

PR #931, opened 2026-07-07
feat(adapter): add declarative provider registration machinery

radiohead:
## What

Adds the reusable provider-registration machinery: `adapter.Resource[T]`, capability interfaces, shared registration builders, and command factories. Schema, identity metadata, and examples flow through the shared adapter construction paths. Removes unused `TypedRegistration[T]` and updates its remaining authoring references. Missing client capabilities return `ErrUnsupported`, including clients that can list but cannot get.

This is the first PR in a two-part stack. It contains the machinery, adapter tests, and design documentation. SLO and OnCall remain on their existing registration paths here; their migration is isolated in the dependent PR.

## Why

Separating the registration API from provider adoption lets reviewers evaluate the machinery independently. The child PR then shows the concrete reduction in provider wiring without repeating the machinery diff.

## How it was tested

- [x] `GOFLAGS=-mod=readonly GCX_AGENT_MODE=false mise run all` passed on this branch independently: lint, full tests, build, skill validation, and documentation.
- [x] Race-enabled adapter, SLO, and OnCall tests passed with the providers still using their original registration paths.
- [x] Generated references remained unchanged.

## Related

- Followed by #1293 for SLO and OnCall adoption.

<hr/>

#### Stack

* **PR #931** 👈
  * **PR #1293**

---
**Generated by Codex**


wardbekker:

### Branch
Rebased onto latest `main` (was CONFLICTING). Three collisions resolved:
- `internal/providers/irm/oncall_adapter.go` — kept #898's shared `*CRUDOpts()` helpers (dual-CRUD invariant), re-expressed on the new `adapter.BuildRegistration`/`adapter.With*` API; local `crudOption` is now an alias of `adapter.CRUDOption[T, OnCallAPI]`.
- `internal/providers/slo/definitions/client.go` — kept both #896's `providers.HandleErrorResponse` and the new adapter import.
- Dry-run semantics — aligned `Resource[T]`'s non-`Validator` dry-run path with #938: returns `ErrDryRunUnverified` (reported as skipped) instead of silently succeeding; test + docs updated.

### CI Checks
All checks passed on the rebase push; the tightening push is running now.

### Auto-Fixes Applied
- `fix(adapter)`: dry-run seam reconciliation with #938 — 6ae76059
- `refactor(adapter)`: dropped unused speculative surface (`Cols[T]`/`Resource.Columns`, `Named`/`IDNamed` — no consumer in this PR; spec/ADR/guide docs record the deferral), added compile-time capability-interface guards on the SLO client so signature drift fails the build instead of silently demoting verbs to `ErrUnsupported`, thinned spec-ID/import-cycle comment noise — 9b31476f
- `fix(lint)`: recvcheck suppression on the test fixture (same pattern as domain types) — 2ab60803

### Pending Review
- Awaiting review approval — no reviews submitted yet.

### Escalations
- None.

---
*Updated by shepherd-pr*

radiohead:
@claude please review this pull request

internal/resources/adapter/resource.go
```diff
@@ -0,0 +1,214 @@
...
+// This is where Schema/GVK/Singular/Plural are derived, NaturalKey is folded
+// into RegisterNaturalKey, and the Factory lazily resolves ClientDeps, builds
+// the client, and runs it through the single capability-assertion seam.
+func (r Resource[T]) registration(loadDeps DepsLoader) Registration {
+	desc := r.descriptor()
+	gvk := desc.GroupVersionKind()
+
+	if r.NaturalKey != "" {
+		RegisterNaturalKey(gvk, SpecFieldKey(r.NaturalKey))
+	}
```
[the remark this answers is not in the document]
radiohead: Documented the global natural-key registration side effect on NewProvider itself, where callers see it. The comment distinguishes eager metadata registration from lazy client dependency loading without claiming the constructor can only be called from init. Added coverage that constructs a provider and exercises its registered extractor; the test restores only its own registry entry.

---
**Generated by Codex**

internal/resources/adapter/capability.go
```diff
@@ -0,0 +1,102 @@
...
+	Validate(ctx context.Context, items []*T) error
+}
+
+// ListOptions carries list-verb knobs for Lister[T].List, mirroring
+// TypedCRUD.ListFn's limit parameter and metav1.ListOptions.Limit. Room for
+// label/field selectors as the declarative model grows.
+type ListOptions struct {
+	// Limit caps the number of items returned. Zero means no limit.
+	Limit int64
+}
```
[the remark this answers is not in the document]
radiohead: Removed the forward-looking selector comment. ListOptions now documents only its current limit behavior.

---
**Generated by Codex**

radiohead:
Addressed the review with a focused follow-up:

- Added a compile-time `*adapter.Provider` compatibility guard in the existing providers test file.
- Added coverage for configured and omitted natural-key/URL metadata. The natural-key check exercises extraction from the declared spec field and restores only its own registry entry.
- Added the builder-selection table, renamed the declaration parameter to `decls`, and clarified the builder error as “failed to load … client.” The declarative path retains its distinct dependency-loading and client-construction errors because those are separate stages.
- Documented the constructor’s global natural-key registration and the existing Delete dry-run behavior; removed the speculative selector comment.
- Kept the `Sheep` case: it verifies the existing plural override rather than introducing a new capability.

No CRUD behavior changes. The two inline threads have corresponding replies. The stack was rebased onto current main, with the migration remaining isolated in #1293.

Validation: full `GOFLAGS=-mod=readonly GCX_AGENT_MODE=false mise run all` on both branches, plus race-enabled adapter, provider-core, SLO, and OnCall tests. Generated reference files remained unchanged.

---
**Generated by Codex**

internal/resources/adapter/resource.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// exampleJSON wraps item in the standard TypedObject envelope and marshals
+// it, giving Resource[T].Example a derived JSON manifest without requiring
+// authors to hand-write one (contrast SLO's hand-written SloExample()).
+func exampleJSON[T ResourceNamer](item T, desc resources.Descriptor) json.RawMessage {
+	obj := TypedObject[T]{
+		TypeMeta:   metav1.TypeMeta{APIVersion: desc.GroupVersion.String(), Kind: desc.Kind},
+		ObjectMeta: metav1.ObjectMeta{Name: item.GetResourceName()},
+		Spec:       item,
```
awsome-o: Could we keep `spec.uuid` out of the SLO example people get from `resources list-examples`? In #1293, `SloResource()` lists it in `StripFields`, but this path skips the stripping that `internal/resources/adapter/typed.go:264` applies to normal resource output.

I haven't checked whether the extra field changes a push. Happy for this to be picked up alongside #1293 if you want; it doesn't need to hold this PR up.

internal/resources/adapter/resource.go
```diff
@@ -0,0 +1,214 @@
...
+	// URLTemplate, when set, is carried on the Registration; providers.Register
+	// already folds every Registration.URLTemplate into
+	// deeplink.RegisterPattern, so no separate init() is needed for it either.
+	URLTemplate string
+
+	// Example is a representative value of T. It is wrapped in the standard
+	// {apiVersion, kind, metadata, spec} envelope and exposed via
+	// Registration.Example / ResourceAdapter.Example() — no hand-written
+	// example manifest required.
+	Example T
```
awsome-o: Could we avoid showing an empty template when a provider hasn't supplied an example? `resource.go:113` always serializes `Example`, so an omitted value can still appear in `resources list-examples` with no `metadata.name`.

SLO supplies its example in #1293, so I don't see this affecting that migration. Fine to leave for that PR or a later follow up if you think it's worth supporting.

awsome-o (dismissed):
Looks good to me.
A couple of small questions on generated examples, nothing blocking approval.
Happy for these to go alongside #1293 or into a follow up, up to you.

internal/resources/adapter/resource.go
radiohead: Fixed in the machinery: generated examples now go through TypedCRUD.ToUnstructured, so StripFields is applied exactly as it is for normal resource output. The SLO regression in #1293 checks that metadata.name stays my-slo while spec.uuid and spec.readOnly are absent. The rebuilt resources list-examples output was also checked against a local API-discovery fixture.

---
**Generated by Codex**

internal/resources/adapter/resource.go
radiohead: Fixed by making Resource.Example a *T: nil produces no template in either Registration.Example or the constructed adapter. Added configured/omitted regression cases and updated the SLO declaration and provider guide in #1293 to supply an explicit example pointer. Writable resources still require an example; read-only resources may omit it.

---
**Generated by Codex**

awsome-o (approved):
lgtm

PR #936, opened 2026-07-08
fix(skills): apply best-practices review to bundled skills

dafydd-t:
## Summary

Part of https://github.com/grafana/gcx-internal/issues/8

First step of skills evaluation is to perform this analysis of their contents, making sure they are all following best practices. Once they are all "the best version" of what they can be, we can cut/keep based on performance data or heuristic evaluation.

Review of all 22 bundled skills against [Anthropic's agent-skills best practices](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices), with fixes applied. Part of the skills evaluation work in grafana/gcx-internal#8 (full findings written up there).

The dominant defect was factual drift from the real CLI surface, not prose style:

- **32+ invalid query invocations** across 5 skills used `gcx metrics query <uid> '<expr>'` - the UID would have been sent as the PromQL expression. All fixed to `-d <uid>`.
- **Synth family asserted CLI facts that don't exist**: an invented `--dry-run` flag, a `-f` shorthand (`--force` only), wrong table columns, a made-up 50% OK/FAILING threshold (actual: sensitivity-based 95/90/75%), a YAML labels template in map form where the API needs a `{name, value}` list, and jq pipelines written against the wrong JSON shape. Verified against provider source.
- **Broken workflows**: import-dashboards was missing its render step (`go run .`) between editing builders and pushing; scaffold-project used stale `gcx config set server` keys (now `grafana.server`/`grafana.token`); gcx-observability's Phase 5 pointed at the wrong sub-agent label.
- **Structure**: gcx-observability (553 lines) and debug-with-grafana (641 lines) exceeded the 500-line body limit - both split into overview + reference files. slo-investigate's PromQL reference was never linked from its SKILL.md so it could never be read. TOCs added to all >100-line references.
- **Descriptions**: about half stated only when to trigger, not what the skill does - all now lead with a third-person capability statement, preserving trigger phrases and sibling cross-routing.
- **Trimmed**: ~250 lines of generic LogQL/PromQL/bash tutorial content, stale version pins, roadmap-speculation sections, and a bogus `gcx` entry in `allowed-tools` frontmatter (9 skills).

gcx-demo passed review clean and is untouched.

No skill directories or `name:` fields changed, so `gcx skills install/update` behaviour is unaffected. `go test ./internal/skills/... ./cmd/gcx/skills/...` passes.

## How to review this PR

It's one commit across 21 skills; reviewing per skill family is easiest. The riskiest changes are the command-syntax corrections (synth + query invocations) - each was verified against `docs/reference/cli/` or provider source, but a second pair of eyes on those is most valuable. The description rewrites are worth a skim for triggering regressions.

### DIFF
--- a/claude-plugin/skills/aio11y/SKILL.md
+++ b/claude-plugin/skills/aio11y/SKILL.md
@@ -1,13 +1,15 @@
 ---
 name: aio11y
 description: >
-  Use when the user wants to list or search AI Observability conversations, inspect generations,
+  Inspects and manages Grafana AI Observability resources via gcx:
+  conversations, generations, evaluators, rules, scores, and templates.
+  Use when the user wants to list or search conversations, inspect generations,
   manage evaluators (create, test, delete), set up evaluation rules, check scores,
   or browse evaluator templates. Trigger on phrases like "list conversations",
   "search generations", "what did the agent do", "debug LLM conversation",
   "create evaluator", "set up evaluation rule", "test evaluator", "check scores",
   "evaluate generation quality", or "set up online evaluation".
-allowed-tools: [Bash, Read, Write, Edit]
+allowed-tools: Bash, Read, Write, Edit
 ---
 
 # AI Observability
--- a/claude-plugin/skills/aio11y/SKILL.md
+++ b/claude-plugin/skills/aio11y/SKILL.md
@@ -7,12 +7,12 @@ description: >
   "search generations", "what did the agent do", "debug LLM conversation",
   "create evaluator", "set up evaluation rule", "test evaluator", "check scores",
   "evaluate generation quality", or "set up online evaluation".
-allowed-tools: [gcx, Bash, Read, Write, Edit]
+allowed-tools: [Bash, Read, Write, Edit]
 ---
 
 # AI Observability
 
-AI Observability is Grafana's AI observability platform. It records what LLM-powered applications do in production and scores the quality of their output.
+AI Observability records what LLM-powered applications do in production and scores the quality of their output.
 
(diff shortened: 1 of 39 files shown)


dafydd-t:
@claude review

dafydd-t:
pushed eb5cb674 with 5 more drift fixes found by a new automated check (branch `dafydd-t/skills-gcx-drift-check`, PR to follow) that validates every `gcx` invocation in the skills against the real command tree:

- `gcx kg health` / `gcx kg cypher` don't exist - now `kg summary` / `kg entities query` (diagnose-entity-graph)
- `resources pull --all-versions` doesn't exist - example dropped (manage-dashboards ref)
- `irm oncall alert-groups acknowledge --yes` - the flag is `--force` (oncall-triage, so it wasn't clean after all)

all verified against the built binary

dafydd-t:
addressed the review nits in c9197cb0:

1. **PR body wording** - accepted, fixed (twice now: oncall-triage subsequently got real fixes in eb5cb674, so the body no longer claims it was clean)
2. **mixed allowed-tools forms** - accepted: normalized all 12 to the comma-separated string form the Claude Code docs use
3. **debug-with-grafana example trimming** - rejected for this PR: it's under the 500-line limit and cutting working examples adds churn right before merge; queued for the second pass
4. **remaining 'Use when...' descriptions** - accepted: all 7 now lead with a third-person capability statement, trigger phrases unchanged
5. **Task* tools in gcx-observability** - rejected: TaskCreate/TaskUpdate/TaskList/TaskGet are real tools in current Claude Code harnesses (TodoWrite is the older single-tool form), so the frontmatter is correct; harness portability is a fair concern but not a bug

wardbekker (approved):

Issue #929, opened 2026-07-07
Track: alert rules API honoring dryRun (currently ignored server-side)


dafydd-t:
## Context

`gcx resources push/delete --dry-run` silently mutates alert rules (and other alerting resources) because the `rules.alerting.grafana.app` apiserver ignores the `dryRun` option and applies the write.

gcx sends the dry-run option correctly on both paths — verified with `-vvv --insecure-log-http-payload`:
- push: `POST .../alertrules?dryRun=All`
- delete: body `{"kind":"DeleteOptions","apiVersion":"v1","dryRun":["All"]}`

Both returned `201 Created` / a real delete that **persisted**, via the assistant CLI proxy (OAuth) *and* a direct service-account-token connection. Root cause is server-side (Grafana alerting legacy storage discards the options).

Upstream: grafana/grafana-enterprise#12569

gcx relies purely on server-side dry-run for adapter-less (dynamically-routed) resources — there is no client-side guard, so this class of gap surfaces as silent mutation. The `alert` provider is read-only, so `alertrule` push/delete fall through to the raw dynamic client (`internal/resources/remote/{pusher,deleter}.go`).

## Acceptance criteria

- [ ] Confirm the alerting apiserver honors `dryRun` (dry-run `push`/`delete` of an alert rule is a no-op — re-run the repro from grafana/grafana-enterprise#12569)
- [ ] Remove/relax any interim client-side guard gcx adds for alerting resources
- [ ] Regression test asserting `--dry-run` does not create/delete an alert rule

## Notes

Blast radius on the server side is all alerting `legacy_storage` resources (contact points, notification policies, templates, mute timings, recording rules), not just alert rules.

Separately worth considering: a general client-side dry-run guard for adapter-less resources so gcx never *silently* depends on server-side dry-run (track separately if we go that route).

PR #938, opened 2026-07-08
feat(resources): fail-safe client-side dry-run for APIs that ignore dryRun

dafydd-t:
Part of #929. Long-term fix is server-side (grafana-enterprise#12569).

## Problem

`resources push --dry-run` and `delete --dry-run` are pure server-side dry-run: gcx forwards `DryRun: ["All"]` and trusts the server to validate-without-persisting. That assumption is false for legacy-storage-backed resources — **all alerting resources today** (alert rules, recording rules, contact points, notification policies, templates, mute timings) are served by legacy bridges that ignore `dryRun` and apply the mutation. A dry-run push of an alert rule returns `201 Created` and the rule actually exists afterwards. Verified on a live stack over both the OAuth/proxy path and a direct service-account token, so it's the apiserver, not the client.

## Summary

- Add a **fail-safe client-side guard** (`internal/resources/remote/dryrun_guard.go`) that decorates the dynamic client. For a mutating dry-run against a resource **not** known to honor server-side dryRun, it refuses to send the request, warns once on stderr, and records the resource as **skipped** (not pushed, not failed).
- **Default-deny allowlist** (`dryrun_allowlist.go`), keyed by `GroupResource` (the unit Grafana itself keys storage mode on). Seeded only with resources that honor dryRun regardless of storage mode: `dashboards`, `folders`, `playlists`. Everything else — notably all alerting — is denied.
- **Skipped, not failed:** mirrors the puller's existing `RecordSkipped` precedent. Skips don't trip `--on-error fail` and keep exit 0; the summary line reports them separately. `validate` reports non-allowlisted resources as skipped (in text **and** structured output) rather than a false "valid".
- **Escape hatch:** `--assume-server-dry-run <resource>.<group>` (on push/delete/validate) and per-context `resources.assume-server-dry-run` config augment the allowlist for stacks known to run a resource on dual-write/unified. Both augment, never replace, the static seed. Malformed values are ignored with a warning (fail-safe fallback), never fatal.
- Provider-adapter-backed resources (SLO, synth, etc.) are unaffected — `typedAdapter` already short-circuits dry-run, and the guard wraps only the dynamic fallback.

## Before / After

`gcx resources push alertrules/my-rule --dry-run`

**Before:** `1 resources pushed, 0 errors` — and the rule now exists.
**After:** stderr `⚠ alertrules.rules.alerting.grafana.app does not support server-side dry-run; checked client-side only (not verified). No changes were sent.` and stdout `0 resources pushed, 0 errors (1 skipped: dry-run not supported server-side, not verified)` — nothing sent.

## Guarantee

Best-effort: confirms the operation is well-formed and (for delete) that the target exists. It does **not** validate the spec or server-side semantics — hence "not verified". Client-side schema validation is a tracked follow-up. Upstream root-cause analysis is tracked in grafana-enterprise#12569.

## Notes for review

Reviewed with `/simplify` and a recall-biased `/code-review` before opening. Guard construction wraps only the dynamic fallback at the shared `buildRouter` seam; the sentinel `errDryRunUnverified` is translated to `RecordSkipped` in the pusher/deleter.


internal/resources/remote/dryrun_allowlist.go
```diff
@@ -0,0 +1,77 @@
...
+// Only add things here that are guaranteed to honour dryRun: safety is decided by the
+// storage mode a request routes to, not by whether the legacy bridge honours dryRun.
+// See grafana-enterprise#12569.
+//
+//nolint:gochecknoglobals // constant lookup table.
+var staticServerDryRunAllowlist = map[schema.GroupResource]struct{}{
+	{Group: "dashboard.grafana.app", Resource: "dashboards"}: {},
+	{Group: "folder.grafana.app", Resource: "folders"}:       {},
+	{Group: "playlist.grafana.app", Resource: "playlists"}:   {},
+}
```
dafydd-t: this is where we define the APIs that we know always respect dryRun

internal/resources/remote/dryrun_guard.go
```diff
@@ -0,0 +1,174 @@
...
+// through. Only the dynamic fallback is wrapped, so the provider-adapter path (already
+// dry-run-safe via typedAdapter) is untouched.
+type dryRunGuard struct {
+	inner     adapter.DynamicClient
+	allowlist dryRunAllowlist
+	warn      io.Writer
+
+	mu        sync.Mutex
+	announced map[schema.GroupResource]struct{} // dedupe: one stderr note per GroupResource per run
+}
```
dafydd-t: this is the DynamicClient wrapper that we use to check whether the remote API is known to honour dryRun.

cmd/gcx/resources/dryrun.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/grafana/gcx/internal/config"
+	cmdio "github.com/grafana/gcx/internal/output"
+	"github.com/grafana/gcx/internal/resources/remote"
+)
+
+const (
+	assumeServerDryRunFlag  = "assume-server-dry-run"
+	assumeServerDryRunUsage = "Assert that the given resources honor server-side dry-run, augmenting the built-in allowlist. " +
+		"Repeatable or comma-separated, each value a GroupResource string (<resource>.<group>), e.g. alertrules.rules.alerting.grafana.app"
```
dafydd-t, line 15: lets users mark other APIs as respecting dryRun, so they can allow more APIs to use dryRun if they know their server has an API in a configuration that supports it

internal/config/types.go
```diff
@@ -172,6 +172,28 @@ type Context struct {
...
+
+// ResourcesConfig holds per-context settings for the `gcx resources` commands.
+type ResourcesConfig struct {
+	// AssumeServerDryRun lists GroupResource strings ("<resource>.<group>", e.g.
+	// "alertrules.rules.alerting.grafana.app") that the user asserts honor server-side
+	// dry-run on this stack. It augments the built-in allowlist so --dry-run passes those
+	// requests through instead of falling back to a best-effort client-side check. It never
+	// removes built-in entries.
+	AssumeServerDryRun []string `json:"assume-server-dry-run,omitempty" yaml:"assume-server-dry-run,omitempty"`
+}
```
dafydd-t, line 186: adding a `resources` property to config, so users can set per-context config for this `assume-server-dry-run` property in there instead of as flags.

cmd/gcx/config/command.go
```diff
@@ -131,18 +131,27 @@ func (opts *Options) LoadConfig(ctx context.Context) (config.Config, error) {
 // When OAuth proxy mode is active, it wires the OnRefresh callback to persist
 // refreshed tokens back to the config file.
 func (opts *Options) LoadGrafanaConfig(ctx context.Context) (config.NamespacedRESTConfig, error) {
+	restCfg, _, err := opts.LoadGrafanaConfigWithContext(ctx)
+	return restCfg, err
+}
+
+// LoadGrafanaConfigWithContext behaves like LoadGrafanaConfig but also returns the resolved
+// current Context, so callers can read per-context resource settings alongside the REST
+// config.
```
dafydd-t: adding `resources.assume-server-dry-run` as a config item, so we need the per-context config so we can read this.

cmd/gcx/resources/validate.go
```diff
@@ -176,6 +151,60 @@ func validateCmd(configOpts *cmdconfig.Options) *cobra.Command {
...
+	if skipped > 0 {
+		cmdio.Warning(w, "%d resources skipped (server-side dry-run unsupported, not verified)", skipped)
+	}
+	return nil
+}
+
+func encodeValidationSummary(w io.Writer, ioOpts cmdio.Options, summary *remote.OperationSummary) error {
+	printableSummary := struct {
+		Failures []map[string]string `json:"failures" yaml:"failures"`
+		Skipped  int                 `json:"skipped" yaml:"skipped"`
```
radiohead, line 188: suggestion – could we also populate this for provider-backed resources which don't run the validation (via the `ValidateFn` method)?

internal/config/merge.go
```diff
@@ -125,6 +135,30 @@ func mergeContexts(base, over *Context) *Context {
 	return &result
 }
 
+func mergeResourcesConfig(base, over *ResourcesConfig) ResourcesConfig {
```
radiohead, line 140: nit – I think we should instead follow the "last definition wins" OR "fall back to the safest option" here, WDYT?

radiohead (dismissed):
Overall LGTM, a couple of minor issues, non-blocking.

One other thing – could we add the new guard to the flow in `docs/architecture/data-flows.md`? Not critical, but would be nice to prevent docs rot.

cmd/gcx/resources/validate.go
dafydd-t: yes, good idea! I'll add it

internal/config/merge.go
dafydd-t: went with last-definition-wins 👍

wardbekker:
## Shepherd Status

**Head:** `00e32fa4` · **CI:** ✅ all checks pass · **Mergeable:** yes · **Blocked on:** review re-approval

### Change added this run
- `docs: document dry-run guard in push/delete data flows` — adds the client-side dry-run guard to the push & delete pipeline diagrams in `docs/architecture/data-flows.md`, addressing @radiohead's request to prevent docs rot. Branch was also rebased onto latest `main`.

### Review items — status
| Item | Source | Status |
|------|--------|--------|
| `merge.go` "last-definition-wins" | @radiohead | ✅ done (commit) |
| `validate.go` populate provider dry-runs via `ValidateFn` | @radiohead | ✅ done (commit) |
| Document guard in `data-flows.md` | @radiohead | ✅ done (this run) |

### Next
All CI green and all review feedback addressed. Waiting on a fresh review approval (the prior LGTM was auto-dismissed by new commits). Re-requesting review.

---
*Updated by shepherd-pr*

radiohead (approved):

PR #941, opened 2026-07-09
fix(dev): derive SDK imports for generated import code from usage

awsome-o:
## Problem

`gcx dev import` emitted Go files that never compiled. Converter output references up to nine foundation-sdk packages (`cog`, `common`, panel/query packages, nested `cog/variants`) while the generated file imported only two, relying on goimports discovery for the rest — and goimports cannot resolve them reliably (the module cache offers several `cog` candidates: foundation-sdk vs promql-builder). Result: `undefined: cog` on first build, or a silently wrong `promql-builder/go/cog` import. The goimports-failure fallback then wrote the uncompilable file anyway.

## Fix

Derive the import list from the generated code itself:

- render the template, parse the result, and collect selector roots that resolve to nothing declared in the file (AST object resolution excludes locals like `builder`/`err` and predeclared identifiers; string literals are naturally ignored)
- map each identifier to its SDK import path, with overrides for the SDK's only nested packages (`variants` → `go/cog/variants`, `plugins` → `go/cog/plugins`)
- re-render with the complete import block; goimports is now formatter-only, so its failure can no longer produce uncompilable output

Also updates the import-dashboards skill's Common Issues table to name the exact failure signature and remedy for users on older gcx binaries.

## Testing

- New table-driven unit tests for the scanner (locals/shadowing, predeclared identifiers, selector-lookalikes inside string literals, nested `variants` mapping) plus an end-to-end self-consistency test through `convertResource`: every selector root in a generated file must be satisfied by its import block.
- Live end-to-end against a seeded Grafana 12.4: imported 5 dashboards of varying complexity (including a 125-panel one) — all build **untouched** and render via `go run .`.
- Measured with paired with/without-skill agent evals (3 runs per config, deterministic graders), before vs after: compile-failure recovery loops disappeared entirely; mean task time −25%/−22%, tokens −38%/−32%, variance halved in both configurations, with pass rates unchanged (no regression). Full findings: https://github.com/grafana/gcx-internal/issues/8#issuecomment-4920296821
- `mise run gate` green (lint 0 issues, full test suite, build); reference docs regenerated with no drift.

Part of the Category B skill evaluation work in grafana/gcx-internal#8.

### DIFF
--- a/claude-plugin/skills/import-dashboards/SKILL.md
+++ b/claude-plugin/skills/import-dashboards/SKILL.md
@@ -109,4 +109,5 @@ gcx resources push
 | "no converter found" | The resource's API version has no converter; the error message names the unsupported version |
 | Auth error (401/403) | Run `gcx config check` to verify credentials |
 | Empty import | The selector matched no resources; check UID with `gcx resources get dashboards` |
-| Imported code doesn't compile | Some complex dashboards produce converter output that needs manual fixup |
+| Build fails with `undefined: cog` (or another lowercase SDK identifier) | Older gcx versions omit SDK imports from generated files. Add the missing import `github.com/grafana/grafana-foundation-sdk/go/<pkg>` for each undefined identifier — `variants` is nested at `go/cog/variants`. Newer gcx emits complete imports |
+| Imported code doesn't compile for other reasons | Complex dashboards can produce converter output that needs manual fixup: build, read the first error, fix, repeat |
--- a/cmd/gcx/dev/import.go
+++ b/cmd/gcx/dev/import.go
@@ -4,8 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"go/types"
 	"os"
 	"path/filepath"
+	"sort"
 	"text/template"
 
 	cmdconfig "github.com/grafana/gcx/cmd/gcx/config"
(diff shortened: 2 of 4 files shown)


wardbekker:
## Shepherd Status (iteration 2)

### Sensibility Assessment
Problem: real · Approach: suboptimal → fixed · Tightness: loose → fixed (tighten-then-ship, approved)

### CI Checks
| Check | Status |
|-------|--------|
| Documentation | PASS |
| Linters | PASS |
| Tests | PASS |
| zizmor | PASS |
| semgrep-cloud-platform/scan | PASS |
| TruffleHog Secret Scan | PASS |

### Auto-Fixes Applied
- Pass `imports.Options{FormatOnly: true}` to `imports.Process` so goimports can no longer reintroduce the ambiguous-import bug this PR fixes (previously relied on the derived import list making resolution a no-op)
- Removed the now-dead `sdkPkg` string return value from `resourceConverter` and its three implementations, and the corresponding test assertions
- Minor simplification (`imported += 1` → `imported++`) via code-simplifier pass
- Rebased twice onto latest `main` as it advanced (`main` moved during shepherding) and re-verified the full quality gate (lint + tests + build) each time — current head: `4592135f`

### Pending Review
- No reviews yet on this PR — `mergeStateStatus` is `BLOCKED` pending approval, not CI. Shepherd will continue polling for review activity.

### Escalations
- None

---
*Updated by shepherd-pr*

dafydd-t (approved):
let's give it a go 🚀

PR #944, opened 2026-07-09
feat(telemetry): add anonymous usage stats core library

dafydd-t:
## Summary

- new `internal/telemetry` package: 
  - add the `Event` model for usage stats
  - Add opt-out controls
  - Add persistent anonymous device ID (`$XDG_STATE_HOME/gcx/device-id`, with ephemeral fallback)
  - Add CI provider detection from env vars
- detect agents (e.g. `claude-code`)
- new `diagnostics.telemetry` config field, wired into layered merge

Nothing is activated yet and nothing is emitted. This is just laying some groundwork.

This is [stage G1](https://github.com/grafana/gcx-internal/pull/32/changes#diff-8c6b65564b71c720ea66e00b5245f6118877ae70ce09057b43f8d72325c58b62R52) of the implementation plan in grafana/gcx-internal#32 (`docs/specs/2026-07-09-usage-stats-implementation-plan.md`).

## Notes for review

- We'll re-use the Event struct in the usage-stats server. This will be the shape of the JSON schema.
- preparing for reporting missing commands: `parse_error_*` fields  are `omitempty` and only set on parse failures (populated later); a test guards that `parse_error_distance: -1` survives marshalling.

### DIFF
--- a/docs/reference/environment-variables/index.md
+++ b/docs/reference/environment-variables/index.md
@@ -1,5 +1,11 @@
 # Environment variables reference
 
+## `DO_NOT_TRACK`
+
+DoNotTrack disables anonymous usage telemetry when set to "1" or
+"true" (cross-tool convention, see https://consoledonottrack.com).
+Overridden by GCX_TELEMETRY.
+
 ## `GCX_AUTO_APPROVE`
 
 AutoApprove automatically enables the --force flag on delete operations,
@@ -10,6 +16,13 @@ enabling non-interactive operation in CI/CD pipelines.
 DisableUpdateNotifier disables the periodic notifier that reminds users
 when their installed gcx skills can be updated. Any non-empty value
 disables the notifier (NO_COLOR convention).
+
+## `GCX_TELEMETRY`
+
+Telemetry controls anonymous usage telemetry for this invocation:
+"enabled", "disabled", or "log" (print the event to stderr and send
+nothing). Takes precedence over DO_NOT_TRACK and the
+`diagnostics.telemetry` config field.
 
 ## `GRAFANA_CLOUD_API_URL`
 
--- a/docs/reference/environment-variables/index.md
+++ b/docs/reference/environment-variables/index.md
@@ -3,8 +3,8 @@
 ## `DO_NOT_TRACK`
 
 DoNotTrack disables anonymous usage telemetry when set to "1" or
-"true" (cross-tool convention, see https://consoledonottrack.com).
-Overridden by GCX_TELEMETRY.
+"true" (cross-tool DO_NOT_TRACK convention). Overridden by
+GCX_TELEMETRY.
 
 ## `GCX_AUTO_APPROVE`
 
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -11,15 +11,16 @@ import (
 	"strings"
 )
 
-// Environment variables that signal agent mode.
-var agentEnvVars = []string{ //nolint:gochecknoglobals
-	"GCX_AGENT_MODE",
-	"CLAUDECODE",
-	"CLAUDE_CODE",
-	"CURSOR_AGENT",
-	"GITHUB_COPILOT",
-	"AMAZON_Q",
-	"OPENCODE",
+// Environment variables that signal agent mode, with the harness name each
+// one identifies. GCX_AGENT_MODE is handled separately (it is an explicit
+// override, not a harness signal).
+var harnessEnvVars = []struct{ envVar, name string }{ //nolint:gochecknoglobals
+	{"CLAUDECODE", "claude-code"},
+	{"CLAUDE_CODE", "claude-code"},
+	{"CURSOR_AGENT", "cursor"},
+	{"GITHUB_COPILOT", "github-copilot"},
+	{"AMAZON_Q", "amazon-q"},
+	{"OPENCODE", "opencode"},
 }
 
 var (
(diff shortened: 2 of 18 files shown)


docs/reference/environment-variables/index.md
```diff
@@ -1,5 +1,11 @@
 # Environment variables reference
 
+## `DO_NOT_TRACK`
+
+DoNotTrack disables anonymous usage telemetry when set to "1" or
+"true" (cross-tool convention, see https://consoledonottrack.com).
+Overridden by GCX_TELEMETRY.
+
```
dafydd-t, line 8: i dont think this link is the right one

internal/agent/agent.go
```diff
@@ -89,6 +90,19 @@ func detectFromEnv() {
 	}
 }
 
+// Harness returns the name of the detected agent harness (e.g. "claude-code"),
+// or "" when no known harness env var is set. It reports the harness even when
+// agent mode was explicitly disabled via GCX_AGENT_MODE or --agent=false;
+// callers should combine it with [IsAgentMode].
+func Harness() string {
```
dafydd-t: GetAgentName maybe

internal/config/types.go
```diff
@@ -67,6 +67,11 @@ type DiagnosticsConfig struct {
 	// LogDir overrides the output directory for agent invocation log files.
 	// Default: $XDG_STATE_HOME/gcx/ (platform-specific).
 	LogDir string `json:"log-dir,omitempty" yaml:"log-dir,omitempty"`
+
+	// Telemetry controls anonymous usage telemetry: "enabled", "disabled",
+	// or "log" (print the event to stderr and send nothing). Overridden by
+	// the GCX_TELEMETRY and DO_NOT_TRACK environment variables.
```
dafydd-t:
```suggestion
	// Telemetry controls anonymous usage telemetry: "enabled", "disabled",
	// or "log" (prints to stderr). Overridden by
	// the GCX_TELEMETRY and DO_NOT_TRACK environment variables.
```

internal/telemetry/ci.go
dafydd-t, line 1: we could drop this if we want to reduce scope, but its at least self-contained to this file.

internal/telemetry/deviceid.go
```diff
@@ -0,0 +1,45 @@
...
+func DeviceIDPath() string {
+	return filepath.Join(xdg.StateHome(), "gcx", deviceIDFileName)
+}
+
+// DeviceID returns the anonymous per-install device ID and whether it is
+// persisted, creating and storing a random UUIDv4 on first use. It is
+// random, never machine-derived. When the state dir is unwritable (or the
+// ID file is corrupt and cannot be rewritten) it returns a fresh ephemeral
+// ID with persisted=false, so ephemeral IDs can be excluded from install
+// counts.
```
dafydd-t:
```suggestion
// DeviceID returns the anonymous per-install device ID and whether it is
// persisted, creating and storing a random UUIDv4 on first use. It is
// random. When the state dir is unwritable (or the
// ID file is corrupt and cannot be rewritten) it returns a fresh ephemeral
// ID with persisted=false, so ephemeral IDs can be excluded from install
// counts.
```

dafydd-t: double definition

internal/telemetry/event.go
```diff
@@ -0,0 +1,61 @@
...
+// service.name; the usage-stats receiver dispatches on it.
+const ServiceName = "gcx"
+
+// Outcome values for Event.Outcome.
+const (
+	OutcomeOK           = "ok"
+	OutcomeRuntimeError = "runtime_error"
+	OutcomeParseError   = "parse_error"
+	OutcomeHelp         = "help"
+)
```
dafydd-t, line 13: is there no way to derive these from the existing error codes?

internal/telemetry/event.go
```diff
@@ -0,0 +1,61 @@
...
+// Event is the flat wide event describing one gcx invocation. Field names
+// follow the usage-stats JSON schema (snake_case); the same fields travel as
+// OTLP span/resource attributes on the wire.
+//
+// Privacy invariant: no field may carry argument or flag values, resource
+// names, hostnames, or anything else that identifies a person, an
+// organisation, or their data. Flags holds flag NAMES only; Command is the
+// resolved command path only. The parse_error_* fields are shape-filtered
+// before they are set (see #578).
+type Event struct {
```
dafydd-t, line 24: why are we defining this json schema, we're sending traces?

docs/reference/environment-variables/index.md
dafydd-t: good catch - the domain has lapsed, it's a parked wordpress page now with no mention of DO_NOT_TRACK. dropped the link, kept the convention name

internal/agent/agent.go
dafydd-t: renamed - went with `agent.Name()` rather than `GetAgentName`: avoids the Get prefix (go convention) and the `agent.GetAgentName()` stutter at call sites. happy to switch if you'd still prefer the explicit form

internal/config/types.go
dafydd-t: applied

internal/telemetry/deviceid.go
dafydd-t: applied (dropped "never machine-derived")

internal/telemetry/ci.go
dafydd-t: kept it - `ci_provider` is in the event schema, so the U1 flattener and BigQuery columns already mirror it; dropping it now would ripple through the contract. it's ~50 lines and self-contained as you say, and `is_ci` alone can't split e.g. github actions from jenkins when we look at where agents struggle

internal/telemetry/event.go
dafydd-t: partially, and G2 derives what it can: `error_kind` comes straight from the exit code via `agentlog.KindFromExitCode`. `outcome` can't be fully derived though - `help` and `ok` both exit 0, and the parse_error/runtime_error split needs the cobra classification (unknown command/flag from `Find()`), not the numeric code: exit 2 also covers things like missing required flags on real commands. so these four constants are the event vocabulary, not a parallel error taxonomy

internal/telemetry/event.go
dafydd-t: the JSON tags aren't the wire format - the event travels as OTLP span/resource attributes (G4). they exist because the flat JSON *is* the schema at both ends of the wire: `GCX_TELEMETRY=log` prints exactly this to stderr for auditability, and usage-stats flattens the span back into this JSON for the BigQuery `report` column. the tags + inventory test pin those shared field names in one place; the G3 golden PII test snapshots the same shape

docs/reference/environment-variables/index.md
```diff
@@ -1,5 +1,11 @@
 # Environment variables reference
 
+## `DO_NOT_TRACK`
+
+DoNotTrack disables anonymous usage telemetry when set to "1" or
+"true" (cross-tool DO_NOT_TRACK convention). Overridden by
+GCX_TELEMETRY.
```
dafydd-t, line 7: I don't have a strong opinion about whether we should have this or not. open to opinions

wardbekker (approved):

PR #946, opened 2026-07-09
fix(kg): align entity/relationship deletes with collection-path API

patmagee:
## Summary
- Update `DeleteEntity` / `DeleteRelationship` to hit the KG Write collection paths (`/entities`, `/relationships`) with identity in query params, matching [asserts-adi#6513](https://github.com/grafana/asserts-adi/pull/6513).
- This allows entity names containing `/` (Tomcat no longer rejects encoded solidus in the path) and removes the need for subresource delete gateway routes.
- CLI command surface is unchanged; only the HTTP wire format for deletes changes.

## Test plan
- [x] `go test -mod=mod -race -count=1 ./internal/providers/kg/ -run 'TestClient_DeleteEntity|TestClient_DeleteRelationship|TestEntitiesDelete|TestRelationshipsDelete'`
- [ ] Confirm against a stack with the asserts-adi delete API change: create an entity named `foo/bar`, then `gcx kg entities delete` succeeds with 204

### DIFF
--- a/internal/providers/kg/client.go
+++ b/internal/providers/kg/client.go
@@ -51,11 +51,12 @@ const (
 	v2RelabelRulesPath   = v2ConfigPath + "/relabel-rules"
 
 	// KG Write API (K8s-style, namespaced). %s is the stacks-<id> namespace.
+	// Entity and relationship deletes use these collection paths with identity
+	// in the query string (type/name are not path segments), so names containing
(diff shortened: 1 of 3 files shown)


egovova11 (approved):

patmagee:
waiting on grafana/backend-enterprise#13443

jgehrett (approved):

wardbekker (approved):

Issue #940, opened 2026-07-08
gcx kg prom-rules schema — expose GET /v1/config/prom-rules/schema


xujiaxj:
## Summary

Add `gcx kg prom-rules schema` to fetch the live JSON Schema for the Prometheus rules config from the backend — mirroring the existing `gcx kg model-rules schema`.

This lets agents validate prom-rules YAML against the authoritative schema before uploading (instead of learning the shape from 400s), and unblocks richer `--dry-run` validation for prom-rules writes.

## Backend endpoint (confirmed)

`GET /v1/config/prom-rules/schema` — live in `asserts-adi` (`PromRulesConfigController.getPrometheusRulesSchema`). Returns an untyped JSON Schema, cached 1h. Same mechanism as the model-rules schema endpoint.

## Implementation

Copy the existing `model-rules schema` slice in `internal/providers/kg/` — endpoint constant + client method (`GetModelRulesSchema`) in `client.go`, and the `schema` subcommand + opts in `commands.go`. Feed the prom-rules path instead.

## Acceptance criteria

- [ ] `gcx kg prom-rules schema` returns the JSON Schema (default output `json`, supports yaml)
- [ ] Unit test mirroring the model-rules schema test
- [ ] Reference docs regenerated (`GCX_AGENT_MODE=false mise run reference`)
- [ ] Manual end-to-end test: prompt Claude Code to author a simple prom rule from the fetched schema, then run it through the full `gcx kg prom-rules` CRUD lifecycle (create → get/list → delete) against a real stack, confirming each step succeeds.

PR #947, opened 2026-07-09
feat(kg): add prom-rules schema command

kblaschke512:
Per https://github.com/grafana/gcx/issues/940

Expose GET /v1/config/prom-rules/schema via the new `gcx kg prom-rules schema` command, mirroring the existing `gcx kg model-rules schema`. Fetches the live JSON Schema for the Custom Prometheus rules config from the backend so agents can validate prom-rules YAML against the authoritative shape before uploading, instead of learning it from 400s.

- client.go: rulesSchemaPath constant + GetPromRulesSchema (GET)
- commands.go: schema subcommand (default json, supports yaml)
- client_test.go: TestClient_GetPromRulesSchema
- command_annotations.go: agent cost/hint annotation
- regenerated CLI reference docs


internal/agent/command_annotations.go
```diff
@@ -318,6 +318,7 @@ var commandAnnotations = map[string]annotation{
 	"gcx kg prom-rules delete":    {Cost: "small"},
 	"gcx kg prom-rules get":       {Cost: "small"},
 	"gcx kg prom-rules list":      {Cost: "small"},
+	"gcx kg prom-rules schema":    {Cost: "medium", Hint: "live JSON Schema for Custom Prometheus rules from backend — pipe to file for editor autocomplete, or validate prom-rules YAML before create"},
```
xujiaxj: Mmm, this cost here should really be `small`. The same for `mode-rules schema`.

xujiaxj: Mmm, validation failed after the change. Is it something we cannot manually modify?

kblaschke512: We can modify - there was a breaking changed merged between original push and second push.

Fixed with https://github.com/grafana/gcx/pull/948

jgehrett (approved):

xujiaxj (approved):

wardbekker (approved):

PR #747, opened 2026-05-19
Feat/assistant mcp servers

mbaykara:
Adds gcx assistant mcp-servers CRUD for managing Grafana Assistant MCP integrations including OAuth initiation, tenant/user scope handling, safe delete confirmation, table/structured output

```
  gcx assistant mcp-servers list

  gcx assistant mcp-servers list --output wide

  gcx assistant mcp-servers create \
    --name GitHub \
    --url https://api.githubcopilot.com/mcp

  gcx assistant mcp-servers create \
    --name SharedTools \
    --url https://mcp.example.com/mcp \
    --scope tenant \
    --header "Authorization=Bearer <token>"

  gcx assistant mcp-servers update SharedTools \
    --description "Shared internal MCP tools"
```

PR #956, opened 2026-07-10
feat(assistant): register assistant as a first-class provider

radiohead:
## What

Registers `assistant` as a first-class provider following the `slo`/`irm` template. Its command tree now mounts through the standard provider-registration loop instead of being hand-mounted outside it, and its `TypedRegistrations()` returns a single `MCPServer` adapter.

Introduces a dedicated `MCPServer` manifest domain type wired through `TypedCRUD[MCPServer]`, so MCP-server configuration for both `user` and `tenant` scope now flows through the `gcx resources get/pull/push/delete` pipeline. Identity is composite (`metadata.name = {scope}-{slug(name)}`); cross-stack matching uses the natural key `(scope, name, url)`; the server-assigned opaque ID rides as an annotation. Because scope, name, and url form this identity, `mcp-servers update` now rejects attempts to change any of them — delete and recreate the server instead.

Routes the `mcp-servers create/update/delete` command verbs through the same `TypedCRUD[MCPServer]` adapter used by the resources pipeline, and adds secret round-trip via explicit per-header write intent (overwrite, preserve-existing, remove) with `fromEnv`/`fromFile` sourcing for CI.

Three regressions are fixed along the way:

- Offset pagination — the adapter `List` now exhausts all pages instead of truncating; the human `mcp-servers list` command stays bounded to a default page with a hint.
- `Update` header handling — sends the full desired header list instead of a partial one, fixing a tenant-update header drop.
- `create` against an existing user-scoped server no longer silently strips its stored auth header, and `delete` by ID now resolves past `(scope, name)` collisions across multiple matching servers.

## Why

Assistant's command tree was mounted by hand outside the provider-registration loop, so it satisfied none of the `providers.Provider` interface. Adapter self-registration is only legal through a provider's `TypedRegistrations()`, so MCP-server configuration could not reach the `gcx resources` pipeline and could not participate in GitOps round-trips.

MCP servers are configuration — which remote MCP endpoints a stack trusts, their scope, and their auth — not command output, so they belong in the same versionable-configuration model as other typed resources rather than staying a standalone command tree.

The prerequisite client fixes (pagination, header handling) were bundled into this deliverable because they are inseparable from the adapter's correctness: an adapter `pull` that truncates, or an `Update` that drops headers, would silently corrupt GitOps state once MCP servers flow through the resources pipeline.

## How it was tested

- [x] `go build -buildvcs=false -o /tmp/gcx-verify-build ./cmd/gcx/` succeeded
- [x] `go test -count=1 ./internal/providers/assistant/... ./internal/assistant/...` passed (187 tests, 0 failures)
- [x] `mise run lint` (`golangci-lint run -c .golangci.yaml`) reported 0 issues
- [ ] CI pending until branch is pushed

## Related

- Relates to #747

---
**Generated by Claude Code**


wardbekker:

### CI Checks
All checks pass on e6c8841 (Linters, Tests, Documentation, semgrep, zizmor, TruffleHog). Previous failure was a `Validate(ctx)` signature drift vs `main`, fixed by rebase.

### Auto-Fixes Applied
- Rebase onto `main` + pass `ctx` to `Context.Validate` — 21179b8e
- Inline the unused exported `HeaderIntent`/`HeaderWrite`/`HeaderWritesForUpdate` classification into a single unexported `updateHeadersWire` (−120 lines; wire behavior unchanged and still covered by the Update payload tests) — 03aeddb4
- Drop ephemeral `docs/specs/**/plan.md`/`spec.md` planning artifacts (ADR-021 remains the durable record, matching repo convention of gardening stale specs) — 5d2289db
- Trim spec-ID comment narration (FR-xxx/AC-xxx/T-number citations) to durable constraint statements across 13 files; kept ADR-021 references — 60ccaa48
- Dedupe CLI-header→manifest-header conversion into `manifestHeaders` — 73195761
- staticcheck S1016: convert `Header`→`headerWire` by type conversion — e6c88418

### Intentionally Not Changed
- The create path performs a natural-key lookup in both the CLI pre-check (fail-on-exists semantics) and the adapter CreateFn (upsert/race guard for the push pipeline). These serve distinct layers; deduping would require bypassing TypedCRUD or widening the shared framework. Left as-is.

### Pending Review
- Awaiting reviewer approval; no review comments yet.

---
*Updated by shepherd-pr*

internal/providers/assistant/conversation.go
```diff
@@ -4,14 +4,14 @@ import (
...
 	cmdio "github.com/grafana/gcx/internal/output"
+	"github.com/grafana/gcx/internal/providers"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
-func conversationCommand(configOpts *cmdconfig.Options) *cobra.Command {
+func conversationCommand(configOpts *providers.ConfigLoader) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "conversation",
```
dafydd-t, line 16: I think this should be a plural? `gcx assistant conversation` looks weird

dafydd-t: since the verbs are `get` and `list`. I'd understand if it was like "start a conversation", but it isnt

dafydd-t: appreciate this is out of scope though

dafydd-t (dismissed):

dafydd-t (approved):

PR #958, opened 2026-07-10
feat(kg): add --dry-run to suppressions create (validate + diff)

### LINKED ISSUE #577: gcx kg supressions create should have a dry-run mode.

kblaschke512:
## What

Adds `--dry-run` to `gcx kg suppressions create`. It validates the config against the backend's server-side validator and previews what `create` would change — **without ever writing**.

```
# Human preview: unified diff on stdout, banner on stderr
gcx kg suppressions create -f suppressions.yaml --dry-run

# Structured result for agents / scripts
gcx kg suppressions create -f suppressions.yaml --dry-run -o json
```

Fixes #577. Supersedes #587 (that branch predated the CRUD refactor and only checked that the YAML *parsed*).

## Why

`create` pushed straight to the remote with no preview. Users wanted to confirm validity and see the change before applying. The prior attempt only confirmed the YAML parsed — a semantically broken config (empty name, bad `matchLabels`) reported "valid" then failed with a `422` on the real write. Parsing is not validation.

## How

- **Real validation** via the backend's validate-only endpoint `POST .../v1/config/disabled-alerts-validate` (batch; added in `asserts-adi` #5629 + gateway routes in `backend-enterprise` #12333). `200` = valid; `422` returns an `ApiError` envelope (`message` + `subErrors[]{field,message}`) which is parsed into per-field errors and rendered. On failure the command exits non-zero with the field errors on stderr and writes nothing.
- **Diff** of the current remote config vs the input, rendered as a canonical-YAML unified diff (`go-difflib`, promoted to a direct dep).
- **Output model**: diagnostic banner → stderr, result → stdout (CONSTITUTION: stdout is the result, stderr is the diagnostic). The result runs through the `internal/output` codec registry — default `text` renders the unified diff; `-o json`/`-o yaml` (and the agents format) emit a structured `SuppressionsDryRunResult` (`valid`, `changed`, `changes[]{name,action}`, `diff`) so consumers don't parse diff text.

### Diff scope (decided with team input)

The diff and change list are **scoped to the input file's entries**. `create` is upsert-only — it adds/updates named entries and never deletes — so remote entries absent from the file are ignored rather than shown as deletions the command won't perform. **System-managed fields (`managedBy`) are excluded** from both the comparison and the diff, since they're backend-populated, not user-authored; including them would flag unchanged entries as modified and show spurious removals.

## Discussion point

This introduces a **unified-diff** format that's new to the repo. The one comparable command, `datasources update --dry-run`, uses a field-level `ChangeSummary` summary rather than a unified diff. I kept the unified diff because #577 explicitly asked to "show a diff" and it reads well for small suppression documents (and agents get the structured `changes[]` regardless). Happy to align to the `ChangeSummary` convention if reviewers prefer cross-command consistency.

## Testing

- Unit: `ValidateSuppressions` (200 / 422-with-subErrors / 500); dry-run scenarios — invalid, valid+changes (text), `-o json` structured output, no-changes (`changes: []`), remote-only ignored, and system-field (`managedBy`) ignored. Assert banner→stderr, result→stdout, and that the write endpoint is never hit.
- `go test ./internal/providers/kg/... ./internal/agent/... ./cmd/gcx/root/...` pass; lint clean; reference docs regenerated.
- Manual end-to-end against a dev stack (KG enabled): valid config → diff + structured output; invalid config → HTTP 422 with per-field errors, empty stdout, exit 1; suppressions list unchanged after dry-runs (no writes).

## Follow-up (deferred)

- `-o`/`--json`/`--jq` are no-ops on a *non*-dry-run `create` today; either honor the format for the upload result or gate the flags to `--dry-run`.


xujiaxj (dismissed):
Works pretty well.

However, I found a backend bug. With this suppression
```yml
disabledAlertConfigs:
  - name: ConfigMapUpdated
```
It throws 500.
```
gcx kg suppressions create -f suppressions-example.yaml --dry-run
Error: Knowledge Graph API request failed (HTTP 500)
│
│ {"timestamp":"2026-07-10T21:31:16.250+00:00","status":500,"error":"Internal Server Error","path":"/api-server/v1/config/disabled-alerts-validate"}
│
└─
```

This is unrelated to the correctness of this PR, but a backend NPE.
```
Servlet.service() for servlet [dispatcherServlet] in context with path [/api-server] threw exception [Request processing failed: java.lang.NullPointerException: Cannot invoke "java.util.Map.containsKey(Object)" because "matchLabels" is null] with root cause
java.lang.NullPointerException: Cannot invoke "java.util.Map.containsKey(Object)" because "matchLabels" is null
	at ai.asserts.server.api.controller.config.AlertConfigValidator.validate(AlertConfigValidator.java:85) ~[!/:26.7.0.]
	at ai.asserts.server.api.controller.config.AlertConfigValidator.validate(AlertConfigValidator.java:55) ~[!/:26.7.0.]
	at ai.asserts.server.api.controller.config.DisabledAlertConfigController.validateDisabledAlertConfigs(DisabledAlertConfigController.java:162) ~[!/:26.7.0.]
```

cc: @nandadev

dafydd-t (approved):
Cool! command surface is unchanged really, so LGTM from a gcx perspective 👍

PR #965, opened 2026-07-13
feat(datasources): add support for Azure Monitor datasource

alyssajoyner:
Used the add-datasource skill to add the Azure Monitor datasource.

**New commands**:
- gcx datasources azuremonitor query
- gcx datasources azuremonitor logs
- gcx datasources azuremonitor resource-graph
- gcx datasources azuremonitor list-subscriptions
- gcx datasources azuremonitor list-resource-groups
- gcx datasources azuremonitor list-resources
- gcx datasources azuremonitor list-metrics

**Querying**
Three query types (the plugin's fourth, Application Insights Traces, is excluded — it's a KQL convenience over the same Logs API, and logs can query App Insights tables directly):

- query — metrics. No query language; built from flags: a resource (--subscription, --resource-group, --resource, --namespace) plus --metric and --aggregation. --dimensions 'ApiName=*' splits into one series per value.
- logs — KQL against a Log Analytics workspace: logs 'AppRequests | take 10' --subscription … --resource-group … --workspace …
- resource-graph — KQL against Azure's resource inventory: resource-graph 'Resources | project name, type' --subscription …

The list-* commands discover valid flag values, each feeding the next: subscriptions → resource groups → resources → metric definitions (with supported aggregations, units, and dimensions).

**Notes:**
- Queries reuse the shared grafanaquery/dataframe packages; discovery calls go through the plugin's resource proxy to Azure Resource Manager.
- One shared-code change: additive Unit field on dataframe.FieldConfig (previously-ignored JSON field).
- Datasources using "Current User" auth can't be queried with API tokens (no Azure AD token to forward) — documented in the command help.

**Testing**: unit tests for the query client and codecs; all commands, output formats, and error paths manually verified against a live Azure subscription


dafydd-t:
@claude review

internal/query/azuremonitor/client.go
```diff
@@ -0,0 +1,333 @@
...
+	return strings.Join(parts, "/")
+}
+
+// listARM fetches all pages of an Azure Resource Manager list endpoint through
+// the datasource resource proxy, following nextLink pagination.
+func (c *Client) listARM(ctx context.Context, dsUID, armPath, apiVersion, operation string) ([]json.RawMessage, error) {
+	pathAndQuery := armPath + "?api-version=" + url.QueryEscape(apiVersion)
+
+	var items []json.RawMessage
+	for range maxResourcePages {
```
dafydd-t, line 204: What happens if there are more resources than we paginate over? Should we warn? I appreciate 32 pages is probably a lot, but still worth thinking about. Your call

internal/query/azuremonitor/client.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	azm := map[string]any{
+		"resources":        []any{resource},
+		"metricNamespace":  req.MetricNamespace,
+		"metricName":       req.MetricName,
+		"aggregation":      req.Aggregation,
+		"timeGrain":        req.TimeGrain,
+		"dimensionFilters": filters,
+	}
```
dafydd-t: there are a lot of `map[string]any` types here, resulting in a `bodyMap` that gets json marshalled. Is there any benefit creating types for these? your call - its more of a question for the azure monitor domain than a gcx one

dafydd-t: We have a load of types in `types.go` in this package - is that what they are intended for?

internal/query/azuremonitor/client.go
```diff
@@ -0,0 +1,333 @@
...
+		return nil, simplifyQueryError(err)
+	}
+
+	return ParseTableResponse(respBody, operation)
+}
+
+// intervalMsFor derives the query interval from the time range so the
+// plugin's "auto" time grain picks a grain that keeps the series under the
+// Azure Monitor API's data-point cap, with a 60s floor (the smallest grain).
+func intervalMsFor(req QueryRequest) int64 {
```
dafydd-t: nit for simplicity: could we just pass in the start and end time.Time types here? It would save us from creating a `QueryRequest` sometimes

internal/query/azuremonitor/client.go
```diff
@@ -0,0 +1,333 @@
...
+		restConfig:  cfg,
+		httpClient:  httpClient,
+		queryClient: grafanaquery.NewClientWithHTTPClient(cfg, httpClient),
+	}, nil
+}
+
+// Query executes an Azure Monitor metrics query via the Grafana datasource query API.
+func (c *Client) Query(ctx context.Context, dsUID string, req QueryRequest) (*QueryResponse, error) {
+	filters := make([]any, 0, len(req.DimensionFilters))
+	for dim, val := range req.DimensionFilters {
```
dafydd-t: tiny optional nit: we iterate over a map here so the order is not guaranteed. No tests for multiple filters at the moment, but I guess it might make a flaky test if we added more in the future

docs/reference/cli/gcx_datasources_azuremonitor_list-resources.md
```diff
@@ -0,0 +1,48 @@
...
+### Options
+
+```
+  -d, --datasource string       Datasource UID (required unless datasources.azuremonitor is configured)
+  -h, --help                    help for list-resources
+      --jq string               jq expression to apply to JSON output. Mutually exclusive with --json.
+      --json string             Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+  -o, --output string           Output format. One of: agents, json, table, yaml (default "table")
+      --resource-group string   Azure resource group name (optional; lists the whole subscription when omitted)
+      --subscription string     Azure subscription ID (required)
```
dafydd-t: I see azure moniotor datasources can have a default subscrition set. Could we use that if it is set?

dafydd-t:
Added a few comments, other than that it looks OK from a gcx perspective 👍

dafydd-t:
in general, could you add your team to the https://github.com/grafana/gcx/blob/main/.github/CODEOWNERS please?

internal/query/azuremonitor/client.go
alyssajoyner: i think this is good nit! the filters are now sorted by dimension name so the order is deterministic, and I added a multi-filter test

internal/query/azuremonitor/client.go
alyssajoyner: good call! i added a warning.

docs/reference/cli/gcx_datasources_azuremonitor_list-resources.md
alyssajoyner: yes! added a check that will use the default subscription when `--subscription` isn't provided

internal/query/azuremonitor/client.go
alyssajoyner: this matches the pattern in other data sources - the reuqest bodies are `map[string]any` in the client and `types.go` holds the response types

internal/datasources/azuremonitor/logs.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return nil
+}
+
+// LogsCmd returns the `logs` subcommand for an Azure Monitor datasource.
+func LogsCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &logsOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "logs KQL",
```
awsome-o, line 64: Making this call as gcx owner so it isn't yours to guess at. Please use `azuremonitor logs query KQL` and `azuremonitor resource-graph query KQL` (see the matching comment on `resource_graph.go`), keeping `azuremonitor query` for structured metrics.

Reasoning: `command-naming.md:81` reserves `query` for executing a backend query, and the shorthand set at `:87` is closed at labels/series/metrics/metadata. I don't want to widen it here, because Logs and Resource Graph each carry materially different required inputs (a workspace identity versus a subscription list), so they earn their own target group rather than being variants of the metrics query.

To be clear, this is a newly owner-approved subfamily pattern for datasources, not an existing precedent I'm pointing you at. Everything under `datasources <kind>` is flat today, and `command-naming.md:9` is explicit that a shipped command is not automatically a precedent. I'm codifying the rule in #1139 so it outlives this thread.

Not `query-resource-graph`, since `<operation>-<subject>` is for discovery facets rather than query domains. These paths freeze for all of v1, which is why I'd rather settle it now than at merge.

internal/datasources/azuremonitor/resource_graph.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return nil
+}
+
+// ResourceGraphCmd returns the `resource-graph` subcommand for an Azure Monitor datasource.
+func ResourceGraphCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &resourceGraphOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "resource-graph KQL",
```
awsome-o, line 48: Paired with the `logs.go` comment: please make this `azuremonitor resource-graph query KQL`. Resource Graph takes a subscription list rather than the metrics query's single scope, which is what puts it in its own target group instead of being a mode.

internal/datasources/azuremonitor/subscription.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+
+	"github.com/grafana/gcx/internal/config"
+	dsquery "github.com/grafana/gcx/internal/datasources/query"
+)
+
+// resolveSubscription returns the --subscription flag value, falling back to
+// the datasource's configured default subscription when the flag is empty.
+func resolveSubscription(ctx context.Context, cfg config.NamespacedRESTConfig, datasourceUID, flagValue string) (string, error) {
+	if flagValue != "" {
```
awsome-o, line 15: `if flagValue != ""` can't tell an explicit `--subscription ""` apart from an omitted flag, so an unset `$AZ_SUB` in a script falls through to the datasource default and returns a well-formed metric series from a *different* subscription. It succeeds, which is what makes it worth catching rather than leaving.

`cmd.Flags().Changed("subscription")` separates the two cases, and it'd be worth rejecting whitespace-only values in the same pass. A regression asserting the explicit-empty case fails before any I/O would lock it.

internal/datasources/azuremonitor/list_resources.go
```diff
@@ -0,0 +1,113 @@
...
+	opts.IO.DefaultFormat("table")
+	opts.IO.BindFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.azuremonitor is configured)")
+	flags.StringVar(&opts.Subscription, "subscription", "", "Azure subscription ID (defaults to the datasource's default subscription)")
+	flags.StringVar(&opts.ResourceGroup, "resource-group", "", "Azure resource group name (optional; lists the whole subscription when omitted)")
+}
+
+func (opts *listResourcesOpts) Validate() error {
+	return opts.IO.Validate()
```
awsome-o, line 36: `Validate()` only checks IO here, so `--resource-group ""` drops the resource-group path segment and quietly lists the entire subscription instead of the one group the caller named. Same explicit-empty treatment as `--subscription`.

internal/datasources/azuremonitor/logs.go
```diff
@@ -0,0 +1,171 @@
...
+	if opts.Subscription == "" {
+		return errors.New("--subscription is required")
+	}
+	if opts.ResourceGroup == "" {
+		return errors.New("--resource-group is required")
+	}
+	if opts.Workspace == "" {
+		return errors.New("--workspace is required")
+	}
+	return nil
```
awsome-o, line 56: The KQL positional isn't validated at all. `Args: cobra.ExactArgs(1)` happily accepts an empty or whitespace-only string and `args[0]` goes straight into the request at `:134`, so `azuremonitor logs ""` reaches the backend before anything complains. Worth rejecting empty and whitespace-only here.

internal/datasources/azuremonitor/resource_graph.go
```diff
@@ -0,0 +1,142 @@
...
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.azuremonitor is configured)")
+	flags.StringArrayVar(&opts.Subscriptions, "subscription", nil, "Azure subscription ID to query (repeatable; at least one required)")
+}
+
+func (opts *resourceGraphOpts) Validate() error {
+	if err := opts.IO.Validate(); err != nil {
+		return err
+	}
+	if len(opts.Subscriptions) == 0 {
+		return errors.New("--subscription is required (repeatable)")
```
awsome-o, line 38: Two gaps here. The KQL positional isn't checked (same as `logs.go`), and this check is `len(opts.Subscriptions) == 0`, so `--subscription ""` passes with a single empty entry and an empty subscription ID reaches the request. Rejecting empty and whitespace entries locally would be clearer for both people and agents.

internal/datasources/azuremonitor/query.go
```diff
@@ -0,0 +1,215 @@
...
+	if opts.Resource == "" {
+		return errors.New("--resource is required")
+	}
+	if opts.Namespace == "" {
+		return errors.New("--namespace is required")
+	}
+	if opts.Metric == "" {
+		return errors.New("--metric is required")
+	}
+	if opts.Aggregation == "" {
```
awsome-o, line 73: Alongside the aggregation emptiness check, could we validate the actual values? Right now `--aggregation` isn't checked against the Azure enum, `--time-grain` isn't checked as ISO-8601 or `auto`, `--top` isn't checked as a positive integer, and `--top` is only meaningful together with dimensions.

All of these get past `Validate()` today and reach config loading and datasource resolution before failing. I've confirmed they get that far; I have not confirmed what Azure returns for each, so this is about failing locally with a clear message rather than a proven backend rejection. `--time-grain 5m` is the natural mistake for someone coming from CloudWatch, where `--period 5m` is legal, and `cloudwatch/query.go:79` validates that analogue client-side if you want the shape.

internal/datasources/azuremonitor/list_resources.go
```diff
@@ -0,0 +1,113 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to list resources: %w", err)
+			}
+
+			return opts.IO.Encode(cmd.OutOrStdout(), resources)
+		},
+	}
+
+	cmd.Annotations = map[string]string{
+		agent.AnnotationTokenCost: "small",
```
awsome-o, line 91: This should be `large` rather than `small`. With `maxResourcePages = 32` and ARM pages carrying up to 1000 resources each (your own comment at `internal/query/azuremonitor/client.go:22`), the successful worst case is around 32,000 resources.

The `llm_hint` on the next line already points at `--resource-group`, so only the cost annotation needs changing.

Non-blocking follow-ups I'd note but not gate on: there's no `--type` filter here, so the workspace-discovery route in the `logs` help needs `--jq` to narrow; the `*azuremonitor.TableResponse` case in the graph codec looks unreachable; and `getResource` closely mirrors the CloudWatch client's.

awsome-o (changes requested):
Thanks for this, it's a substantial piece of work and the ARM client layer is well built. The paging guard fails honestly rather than silently truncating, which I appreciate, and `internal/query/azuremonitor/client_test.go` covers the wire shape properly.

Five things to sort before merge. One is a naming decision I'm making as gcx owner so it isn't left sitting on you, the rest are input validation plus one metadata correction.

On tests: please add focused command-level regressions alongside the fixes, covering flag binding, command-owned validation firing before any loader or server I/O, and the resolved request fields. Your client tests already cover request construction, so nothing there needs duplicating and there's no need for a flag-by-output matrix.

Two things that are ours, not yours, so you don't go chasing them:

* If CI shows a `maintidx` failure in `cmd/gcx/datasources/query.go` after you rebase, that's not your diff. `main` sits one point above the threshold and the third datasource PR to land tips it over. Tracked in #1137 and we're fixing it.
* Validation errors currently exit 1 rather than 2. We're adding a usable helper for command-owned usage errors in #1138, so please don't hand-roll a wrapper for it.

The naming rule I'm applying below is being written down in #1139.

internal/datasources/azuremonitor/logs.go
ktw4071: Fixed in 55ca2795.

`logs KQL` is now `logs query KQL` — `logs` is a bare subcommand group with a single `query` child, mirroring the `versions`/`load-tests` noun-group pattern elsewhere in the repo (e.g. `internal/providers/dashboards/versions/commands.go`, `internal/providers/k6/commands.go`). Paired with the same restructure on `resource-graph` (next thread).

```
$ gcx datasources azuremonitor logs --help
Query Log Analytics workspaces with KQL

Available Commands:
  query       Query a Log Analytics workspace with KQL

$ gcx datasources azuremonitor logs query 'AppRequests | take 3' -d UID \
    --subscription SUB_ID --resource-group RG --workspace WS --since 24h
→ succeeds against a live Azure Monitor datasource, same output shape as before
```

Also had to catch up `cmd/gcx/root/testdata/output_classes.json` in a follow-up commit (245b995b) — the restructure changed which paths are leaf commands, and `TestConsistency_AllLeafCommandsHaveOutputClass` caught the fixture drift.

internal/datasources/azuremonitor/resource_graph.go
ktw4071: Fixed in 55ca2795 (same commit as the `logs` restructure above).

`resource-graph KQL` is now `resource-graph query KQL`, same shape and same reasoning.

```
$ gcx datasources azuremonitor resource-graph query 'Resources | project name, type | limit 5' \
    -d UID --subscription SUB_ID -o json
→ succeeds against a live Azure Monitor datasource
```

internal/datasources/azuremonitor/subscription.go
ktw4071: Fixed in 1f3083ec.

Added `rejectExplicitEmptyFlag(cmd, name, value)` — uses `cmd.Flags().Changed(name)` to distinguish "omitted" from "explicitly empty", and rejects whitespace-only in the same pass. Wired into all four `--subscription` call sites (`query`, `list-resources`, `list-resource-groups`, `list-metrics`).

```
$ gcx datasources azuremonitor query -d UID --subscription="" --resource-group rg ...
→ error: --subscription must not be empty

$ gcx datasources azuremonitor list-resource-groups -d UID --subscription SUB_ID -o json
→ still resolves and lists normally (unaffected):
[
  { "name": "pulumi", "location": "uksouth" },
  { "name": "DefaultResourceGroup-SUK", "location": "uksouth" },
  { "name": "NetworkWatcherRG", "location": "uksouth" },
  ...
]
```

Regression: cmd-level test per command asserting the explicit-empty case fails with the zero-value `ConfigLoader{}` before any config/datasource I/O runs, plus a direct unit test of `rejectExplicitEmptyFlag` covering the whitespace-only case.

internal/datasources/azuremonitor/list_resources.go
ktw4071: Fixed in 1f3083ec (same commit as the subscription fix above).

Same `rejectExplicitEmptyFlag` helper, applied to list-resources' optional `--resource-group`.

```
$ gcx datasources azuremonitor list-resources -d UID --subscription SUB_ID --resource-group=""
→ error: --resource-group must not be empty
```

internal/datasources/azuremonitor/logs.go
ktw4071: Fixed in 31f5b449.

Added a shared `validateKQLArg` (used by both `logs query` and `resource-graph query`, which take the same raw-KQL-positional shape) that rejects empty and whitespace-only KQL before any config/datasource I/O.

```
$ gcx datasources azuremonitor logs query "" -d UID --subscription SUB --resource-group RG --workspace WS
→ error: KQL query must not be empty
```

internal/datasources/azuremonitor/resource_graph.go
ktw4071: Split across two commits since the two gaps you flagged are independent:

- The KQL-positional half is fixed in 31f5b449 (same `validateKQLArg` as the `logs.go` thread):
  ```
  $ gcx datasources azuremonitor resource-graph query "   " -d UID --subscription SUB
  → error: KQL query must not be empty
  ```
- The `--subscription` empty-entry half is fixed in 062f1aa4: `Validate()` now rejects any empty or whitespace-only entry in the `--subscription` list, not just an empty list:
  ```
  $ gcx datasources azuremonitor resource-graph query "Resources | limit 1" -d UID \
      --subscription SUB_ID --subscription ""
  → error: --subscription entries must not be empty
  ```
  Confirmed a valid entry still works:
  ```
  $ gcx datasources azuremonitor resource-graph query "Resources | count" -d UID --subscription SUB_ID -o json
  → {"columns":[{"name":"Count","type":"number"}],"rows":[[210]]}
  ```

internal/datasources/azuremonitor/query.go
ktw4071: Fixed in 3a843fd1.

- `--aggregation` now validated against Average/Total/Maximum/Minimum/Count.
- `--time-grain` now validated as `"auto"` or an ISO 8601 duration.
- `--top` now validated as a positive integer, and rejected outright without `--dimensions`.

```
$ gcx datasources azuremonitor query -d UID ... --aggregation Sum
→ error: --aggregation must be one of Average, Total, Maximum, Minimum, or Count, got "Sum"

$ gcx datasources azuremonitor query -d UID ... --time-grain 5m
→ error: --time-grain must be "auto" or an ISO 8601 duration (e.g. PT1M, PT1H, P1D), got "5m"
   (confirmed this is exactly the CloudWatch-muscle-memory mistake you called out)

$ gcx datasources azuremonitor query -d UID ... --top 5
→ error: --top is only meaningful together with --dimensions

$ gcx datasources azuremonitor query -d UID ... --time-grain PT1H
→ succeeds (valid ISO 8601 duration)
```

Also confirmed a valid `--top` + `--dimensions` combination passes local validation and reaches the backend:
```
$ gcx datasources azuremonitor query -d UID ... --dimensions DatabaseResourceId='*' --top 3 -o json
→ {"type":"gcx.error", ..., "details":"BadRequest: Metric: cpu_percent does not support
   requested dimension combination: databaseresourceid, supported ones are: , TraceId: ..."}
```
This is Azure itself rejecting it for a metric-specific reason (this metric does not support that dimension) — good evidence the local check let a syntactically valid request through rather than overreaching.

internal/datasources/azuremonitor/list_resources.go
ktw4071: Fixed in d556e99d.

Changed the token-cost annotation from `small` to `large`.

```
$ gcx commands --flat -o json | jq 'select(.full_path == "gcx datasources azuremonitor list-resources")'
{
  "full_path": "gcx datasources azuremonitor list-resources",
  "token_cost": "large",
  "llm_hint": "gcx datasources azuremonitor list-resources -d UID --subscription SUB_ID --resource-group RG"
}
```

Left the three non-blocking follow-ups (no `--type` filter, possibly-unreachable `TableResponse` graph-codec case, `getResource` mirroring CloudWatch's client) as noted in your comment — happy to open follow-up issues for any of them if you'd like them tracked separately.

awsome-o (dismissed):
approved, but you might need to rebase and fix the CI

internal/query/azuremonitor/client.go
```diff
@@ -0,0 +1,256 @@
...
+		if page.NextLink == "" {
+			return items, nil
+		}
+		// nextLink is an absolute ARM URL (https://management.azure.com/...);
+		// re-route its path and query through the datasource proxy.
+		next, err := url.Parse(page.NextLink)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse %s nextLink: %w", operation, err)
+		}
+		pathAndQuery = strings.TrimPrefix(next.Path, "/") + "?" + next.RawQuery
```
awsome-o, line 225: Page 1 escapes its path and page 2 doesn't. `escapeNamespace` PathEscapes every segment on the way in, but the nextLink branch uses `next.Path`, which `url.Parse` has already percent-decoded, and that gets spliced straight back into a URL string in `getResource`. `next.RawQuery` is the right choice for the query half; it's only the path that loses its encoding.

`next.EscapedPath()` instead of `next.Path` keeps it. Worth fixing rather than noting because the failure is quiet: a resource name carrying a `%2F` turns into a literal separator and page 2 queries a different scope instead of erroring.

internal/datasources/azuremonitor/logs.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (opts *logsOpts) Validate() error {
+	if err := opts.IO.Validate(); err != nil {
+		return err
+	}
+	if err := opts.ValidateTimeRange(); err != nil {
+		return err
+	}
+	if opts.Subscription == "" {
```
awsome-o, line 58: Minor inconsistency with the helper you added in this round: `rejectExplicitEmptyFlag` trims, these required checks don't. So `--subscription " " --resource-group " " --workspace " "` gets past `Validate()` and builds `/subscriptions/ /resourceGroups/ /providers/.../workspaces/ ` for the backend.

Same shape in `query.go:64-73` and `list_metrics.go:43-49`. `strings.TrimSpace(...) == ""` in those seven checks closes it and matches what the fallback flags already do. `resource_graph.go` is already fine, it trims each entry.

Non-blocking.

awsome-o (approved):
Approving.

Re-checked all eight threads against 764a0f6a rather than taking the replies on faith, and they're all genuinely fixed: the paging cap errors instead of truncating, the dimension filters are sorted, the explicit-empty handling uses `Flags().Changed`, the KQL positional is validated in both places, aggregation/time-grain/top are checked locally, and `list-resources` is `large`.

The rebase was done right too. The azuremonitor redirect moved into the `redirects` map rather than the inline form coming back, which matters because the map entry shows up in `supportedKinds()` and an inline branch never would, and the `supportedKinds` test literal was updated in the same commit. CI is green.

Two non-blocking one-liners below. Follow-up is fine, they don't need to hold this.

PR #969, opened 2026-07-14
feat(datasources): add support for Postgres datasource

alyssajoyner:
Used the add-datasource skill to add the PostgreSQL datasource.

**New commands:**
- gcx datasources postgres query
- gcx datasources postgres list-tables
- gcx datasources postgres describe-table

**Querying**
query executes raw SQL (with Grafana server-side macros like $__timeFilter) and auto-appends LIMIT 100 unless the statement uses OFFSET/FETCH/RETURNING/etc. (--limit 0 disables). list-tables and describe-table discover the schema via information_schema. The generic gcx datasources query <uid> 'SQL' auto-detect path also supports postgres now (both grafana-postgresql-datasource and the legacy postgres plugin ID).

**Notes:**
- Reuses the shared grafanaquery transport and internal/query/sql response layer (same as ClickHouse/Athena) — the client is ~70 lines, no shared-code changes, no new codecs.
- Extracted the clickhouse and postgres branches of the auto-detect command into helpers — the function was at the lint complexity threshold before this change.
- Adds a postgres service to docker-compose.yml for local testing (same approach as InfluxDB in #539).


docker-compose.yml
```diff
@@ -77,13 +77,35 @@ services:
...
+      POSTGRES_PASSWORD: grafana
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U grafana -d testdb"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    networks:
+      - grafana
```
dafydd-t, line 98: maybe a followup: could we pull all the datasource dependencies into a deparate docker-compose file, so we dont have to spin them all up by default? otherwise it will get pretty cumbersome

cmd/gcx/datasources/query.go
```diff
@@ -215,3 +213,49 @@ that do not have a dedicated subcommand.`,
...
+	if step > 0 {
+		req.IntervalMs = step.Milliseconds()
+	}
+
+	resp, err := client.Query(ctx, datasourceUID, req)
+	if err != nil {
+		return nil, fmt.Errorf("query failed: %w", err)
+	}
+	return resp, nil
+}
```
dafydd-t, line 240: why do we have clickhouse related changes in here? could we pull that out to its own refactor, or is it coupled?

dafydd-t: oh i see it was a complexity lint thing. OK 👍

internal/query/postgres/types.go
```diff
@@ -0,0 +1,46 @@
...
+
+var limitBailRe = regexp.MustCompile(`(?im)(\bLIMIT\s+\d+\s+OFFSET\b|\bOFFSET\s+\d+\b|\bFETCH\s+(FIRST|NEXT)\b|\bRETURNING\b|\bFOR\s+(UPDATE|SHARE)\b|^\s*EXPLAIN\b|^\s*SHOW\b)`)
+
+// EnforceLimit ensures the SQL has a LIMIT clause within bounds.
+// If limit is 0, enforcement is disabled (pass-through).
+// If the SQL uses OFFSET, FETCH, RETURNING, row locking, or a metadata
+// statement (EXPLAIN/SHOW), it bails out (pass-through).
+func EnforceLimit(sql string, limit, maxLimit int) string {
+	return querysql.EnforceLimit(sql, limit, maxLimit, limitBailRe.MatchString)
+}
```
dafydd-t: if we cap the limit, does it warn the user at all?

internal/query/postgres/types_test.go
```diff
@@ -0,0 +1,56 @@
...
+		want  string
+	}{
+		{"appends LIMIT when missing", "SELECT 1", 100, "SELECT 1 LIMIT 100"},
+		{"caps existing LIMIT exceeding max", "SELECT 1 LIMIT 5000", 100, "SELECT 1 LIMIT 1000"},
+		{"keeps existing LIMIT if under max", "SELECT 1 LIMIT 50", 100, "SELECT 1 LIMIT 50"},
+		{"limit 0 disables enforcement", "SELECT 1", 0, "SELECT 1"},
+		{"bail on LIMIT OFFSET", "SELECT id FROM t LIMIT 10 OFFSET 20", 100, "SELECT id FROM t LIMIT 10 OFFSET 20"},
+		{"bail on OFFSET", "SELECT id FROM t OFFSET 20", 100, "SELECT id FROM t OFFSET 20"},
+		{"bail on FETCH FIRST", "SELECT id FROM t FETCH FIRST 10 ROWS ONLY", 100, "SELECT id FROM t FETCH FIRST 10 ROWS ONLY"},
+		{"bail on RETURNING", "UPDATE t SET a = 1 RETURNING id", 100, "UPDATE t SET a = 1 RETURNING id"},
```
dafydd-t: What about `UPDATE t SET a = 1` (no `RETURNING`)? is the result valid?

dafydd-t: Also, what about DELETE and INSERT commands? do they take LIMITs?

dafydd-t:
Some things that claude found:

**describe-table silently merges same-named tables across schemas**

[internal/datasources/postgres/describe_table.go:78](https://github.com/grafana/gcx/pull/969/changes#diff-4257212e6f244e7ffa8f7c7986e0bb72739d95b0f64c51e810d16fe24943d0edR78): without the `--schema` flag, the query filters only on table_name and doesn't select table_schema. Two consequences:

- if the table exists in multiple schemas, columns from all of them are interleaved in one result with nothing showing which schema each row belongs to
- system schemas aren't excluded (unlike list-tables), so e.g. describe-table tables returns information_schema/pg_catalog columns

Fix: add table_schema to the SELECT and filter table_schema NOT IN ('pg_catalog', 'information_schema').

**dotted identifiers silently return "not found"**

[ValidateIdentifier](https://github.com/grafana/gcx/pull/969/changes#diff-3521d949de413d5622a3dbbd44208ed1d47180d711c7eb12988389c668789a62R20) accepts dots, so describe-table public.orders builds table_name = 'public.orders' — which never matches, producing table "public.orders" not found. Inherited from ClickHouse, but schema-qualified names are the idiom in Postgres.

Fix (or follow-up): split on the dot into schema + table, or reject dotted names with a hint to use `--schema`.

docs/reference/cli/gcx_datasources_postgres_describe-table.md
```diff
@@ -0,0 +1,55 @@
...
+  gcx datasources postgres describe-table orders --schema public
+
+  # Output as JSON
+  gcx datasources postgres describe-table orders -o json
+```
+
+### Options
+
+```
+  -d, --datasource string   Datasource UID (required unless datasources.postgres is configured)
```
dafydd-t, line 34: should we make some of the examples use this flag, since its a common pattern?

docker-compose.yml
alyssajoyner: i can create a separate issue for this

internal/query/postgres/types.go
alyssajoyner: it does not! added now 😊

internal/query/postgres/types_test.go
alyssajoyner: great catch 🙏 we were actually appending `LIMIT 100` to everything (and you are correct `DELETE/UPDATE/INSERT` do not accept limits) - it is now fixed with updated tests

docs/reference/cli/gcx_datasources_postgres_describe-table.md
alyssajoyner: done, the lead examples for describe-table and list-tables show `-d UID` now

docker-compose.yml
alyssajoyner: https://github.com/grafana/gcx/issues/1116

internal/query/postgres/types.go
```diff
@@ -0,0 +1,66 @@
...
+// it to DML/DDL/utility statements would produce a syntax error.
+var limitStatementRe = regexp.MustCompile(`(?is)^\s*(SELECT|WITH|TABLE|VALUES)\b`)
+
+// The DML keywords catch CTE-wrapped writes (e.g. WITH ... DELETE FROM ...),
+// which start with WITH and so pass limitStatementRe. Matching them anywhere
+// means a SELECT mentioning e.g. 'DELETE' in a string literal also skips
+// enforcement — that fails safe (no LIMIT added) rather than corrupting SQL.
+// EXPLAIN/SHOW need no bail entries: limitStatementRe already excludes them,
+// and line-anchored entries would misfire on formatted multiline queries.
+var limitBailRe = regexp.MustCompile(`(?i)(\bLIMIT\s+\d+\s+OFFSET\b|\bOFFSET\s+\d+\b|\bFETCH\s+(FIRST|NEXT)\b|\bRETURNING\b|\bFOR\s+(UPDATE|SHARE)\b|\b(INSERT|UPDATE|DELETE|MERGE)\b)`)
```
awsome-o, line 44: Two problems in this one regex.

`LIMIT ALL` is valid, documented Postgres (https://www.postgresql.org/docs/current/queries-limit.html) but this only matches numeric limits, so the appender still fires:

```
in:   SELECT * FROM t LIMIT ALL
sent: SELECT * FROM t LIMIT ALL LIMIT 100
```

Lowercase behaves the same. `LIMIT ALL OFFSET 5` happens to escape via the OFFSET branch. `internal/query/athena/types.go:37` already has the `ALL` alternative if that helps.

Separately, because the limit is appended as a bare suffix, a query ending in a `--` line comment absorbs it and comes back unbounded.

Both want a regression on the emitted SQL.

internal/datasources/postgres/query.go
```diff
@@ -0,0 +1,128 @@
...
+	dsquery.SharedOpts
+
+	Datasource string
+	Limit      int
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	opts.Setup(flags, false)
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.postgres is configured)")
+	flags.IntVar(&opts.Limit, "limit", defaultLimit, "Max rows to return (0 disables enforcement)")
```
awsome-o, line 31: Two things on this flag.

A negative value isn't checked anywhere, so `--limit -5` emits `LIMIT -5` and the server rejects it. Catching that locally gives a far better message.

And the help says "Max rows to return" while it behaves as a default. With an existing `LIMIT` at or below 1000 the flag is ignored entirely, and `--limit 5000` is silently reduced to 1000 with nothing on stderr, even though an in-query `LIMIT 5000` *does* warn. Same clamp, two different levels of honesty.

I don't think adjusting the help alone closes this, because the default can still silently become unbounded. Either enforce a real maximum on emitted rows, or warn on every bypass and describe the flag as best-effort.

internal/datasources/postgres/list_tables.go
```diff
@@ -0,0 +1,101 @@
...
+
+			if err := postgres.ValidateIdentifier(opts.Schema, "schema"); err != nil {
+				return err
+			}
+
+			sql := "SELECT table_schema AS schema, table_name AS name, table_type AS type FROM information_schema.tables WHERE table_schema NOT IN ('pg_catalog', 'information_schema')"
+			if opts.Schema != "" {
+				sql += fmt.Sprintf(" AND table_schema = '%s'", postgres.EscapeSQLString(opts.Schema))
+			}
+			sql += " ORDER BY table_schema, table_name LIMIT 500"
```
awsome-o, line 77: This stops at 500 with nothing telling the caller, so a database with 600 tables reports 500 and reads as the complete inventory.

The `query` sibling in this same package already warns when it caps, so the pattern is right here. A stderr hint is enough; no `list_meta` work needed, since `docs/design/output.md` §15 is still proposed and opt-in.

internal/datasources/postgres/list_tables.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return err
+			}
+
+			if err := postgres.ValidateIdentifier(opts.Schema, "schema"); err != nil {
+				return err
+			}
+
+			sql := "SELECT table_schema AS schema, table_name AS name, table_type AS type FROM information_schema.tables WHERE table_schema NOT IN ('pg_catalog', 'information_schema')"
+			if opts.Schema != "" {
```
awsome-o, line 74: `--schema ""` is indistinguishable from omission here, so it silently returns every schema. This is the unset-shell-variable case, and it broadens scope rather than narrowing it, which is why I'd rather fix it than note it. `describe_table.go:105` has the same shape.

internal/datasources/postgres/list_tables.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return err
+			}
+
+			datasourceUID, _, err := dsquery.ResolveValidateAndSaveDatasource(ctx, loader, opts.Datasource, cfgCtx, cfg, "postgres")
+			if err != nil {
+				return err
+			}
+
+			if err := postgres.ValidateIdentifier(opts.Schema, "schema"); err != nil {
```
awsome-o, line 69: This identifier check runs after `LoadContextAndConfig` at `:59`, so validation happens after config loading and datasource resolution, which may perform Grafana I/O.

`describe_table.go` gets this right, validating at `:82` and `:85` before its loader call at `:91`. So the same bad input fails differently depending on which command you reached for. Worth aligning, with a regression asserting no I/O on the bad-input path.

Non-blocking: `--schema` matching is exact and case-sensitive, but the help doesn't say so, and a mixed-case schema silently returns nothing.

awsome-o (changes requested):
Nice work on this. The client tests are solid and the shared infrastructure reuse (grafanaquery, dataframe, dsquery) is exactly right.

The limit contract is where I'd like changes. Two paths emit SQL that Postgres rejects, and a third quietly under-reports, so these are correctness rather than polish.

On tests: please pair each fix with a focused regression on the emitted SQL, plus one asserting validation fires before loader or datasource I/O. Your existing client and transport tests cover the wire shape and don't need duplicating.

Two things that are ours, not yours:

* If a `maintidx` failure shows up in `cmd/gcx/datasources/query.go` after you rebase, that's not your diff. `main` sits one point above the threshold and the third datasource PR to land tips it over. Tracked in #1137.
* Validation errors currently exit 1 rather than 2. We're adding a reachable helper for command-owned usage errors in #1138, so please don't invent a wrapper for it here.

internal/query/postgres/types.go
ktw4071: Fixed in 13475c07.

`limitBailRe` now bails on a bare `LIMIT ALL` (not just `LIMIT ALL OFFSET`), and a new `trailingLineCommentRe` bails when the statement ends in a `--` comment.

Verified against a live Postgres datasource:
```
$ gcx datasources postgres query -d postgresql-e2e "SELECT * FROM metrics LIMIT ALL" -o json
→ succeeds, 182 rows returned (pre-fix this produced "LIMIT ALL LIMIT 100", a Postgres syntax error)

$ gcx datasources postgres query -d postgresql-e2e "SELECT * FROM metrics -- get everything" -o json
→ succeeds, 182 rows returned unbounded (pre-fix the injected LIMIT was silently swallowed into the comment)
```

Regression cases added to `types_test.go`: `LIMIT ALL`, lowercase `limit all`, `LIMIT ALL OFFSET`, trailing-comment, trailing-comment-after-semicolon, and a "comment mid-query still gets LIMIT" case to confirm we do not over-bail.

internal/datasources/postgres/query.go
ktw4071: Fixed in 09d037dd.

`--limit < 0` is now rejected locally before any query is built. Separately, I traced the "silently reduced to 1000 with nothing on stderr" bug to the shared `internal/query/sql.EnforceLimit`: it only reported `capped=true` when an *existing inline* LIMIT was clamped down, never when the *requested* default itself exceeded max. Went with your "warn on every bypass" option — both paths now report `capped=true` symmetrically.

```
$ gcx datasources postgres query -d postgresql-e2e --limit=-5 "SELECT * FROM metrics"
→ error: --limit must be >= 0, got -5

$ gcx datasources postgres query -d postgresql-e2e --limit 5000 "SELECT * FROM metrics" -o json
→ 182 rows, stderr: "⚠ LIMIT in query exceeds the maximum of 1000 and was capped; use --limit 0 to disable enforcement"

$ gcx datasources postgres query -d postgresql-e2e "SELECT * FROM metrics LIMIT 5000" -o json
→ same warning, same result — parity between the two paths confirmed
```

One thing worth double-checking with you: this is a *soft* cap — we still return up to `maxLimit` (1000) rows plus a stderr warning, we do not hard-fail or require an explicit override. That matches the "warn on every bypass" option you offered, but I want to confirm that is the behavior you want as the final answer here (vs., say, requiring `--limit 0` explicitly once a request is that far over max), rather than something we should revisit later.

internal/datasources/postgres/list_tables.go
ktw4071: Fixed in 1b19d165.

Query now requests `rowCap+1` (501) rows; if 501 come back we trim to 500 and warn on stderr, same pattern as `query`. Extracted `buildListTablesSQL` and `warnIfTruncated` as pure functions for direct unit testing.

Verified against a real schema with 504 tables across 3 schemas:
```
$ gcx datasources postgres list-tables -d postgresql-e2e -o json
→ exactly 500 rows returned
stderr: "⚠ showing the first 500 tables; more tables match — use --schema to narrow the results"
```

Same question as on the `--limit` thread: this is a stderr-hint-only fix per your note ("no list_meta work needed"), so a database with, say, 5,000 tables still just shows the first 500 + the same one-line warning. Want to keep it at a plain warning regardless of how large the overage gets, or is there already an appetite for something more structured (list_meta, pagination) here later?

internal/datasources/postgres/list_tables.go
ktw4071: Fixed in 93ddf1ae.

`--schema ""` (explicitly passed) now errors with `--schema must not be empty` in both `list-tables` and `describe-table` (same shape, both fixed). A real `--schema <value>` still filters correctly.

```
$ gcx datasources postgres list-tables -d postgresql-e2e --schema=""
→ error: --schema must not be empty

$ gcx datasources postgres describe-table metrics -d postgresql-e2e --schema=""
→ error: --schema must not be empty

$ gcx datasources postgres list-tables -d postgresql-e2e --schema public -o json
→ rows: [["public", "metrics", "BASE TABLE"]]

$ gcx datasources postgres list-tables -d postgresql-e2e --schema analytics -o json
→ rows: [["analytics", "events", "BASE TABLE"], ["analytics", "metrics", "BASE TABLE"]]
```

Also spot-checked schema-qualified describe-table against two same-named tables in different schemas (`public.metrics` vs `analytics.metrics`), which correctly returned different column sets for each.

internal/datasources/postgres/list_tables.go
ktw4071: Fixed in 147e8f63.

Moved the schema checks above `LoadContextAndConfig`/`ResolveValidateAndSaveDatasource`, matching `describe-table`'s order.

To prove the ordering (not just that the check exists), I used a deliberately nonexistent datasource UID as a trip wire — if the code ever reaches datasource resolution, that UID is guaranteed to fail with "not found":

```
$ gcx datasources postgres list-tables -d nonexistent-uid-that-would-fail-io --schema public
→ error: Datasource "nonexistent-uid-that-would-fail-io" not found
   (valid schema, so it reaches datasource resolution — confirms the trip wire works)

$ gcx datasources postgres list-tables -d nonexistent-uid-that-would-fail-io --schema "bad; DROP TABLE x"
→ error: Invalid schema — must contain only letters, numbers, and underscores
   (same bogus UID, but now fails on the schema check instead of "not found",
   proving validation short-circuits before the datasource-resolution I/O runs)
```

Also added the case-sensitivity note to `--schema`'s help text per your non-blocking comment.

awsome-o (approved):

PR #970, opened 2026-07-14
feat(datasources): add support for MySQL datasource

alyssajoyner:
NOTE: merge after https://github.com/grafana/gcx/pull/969 due to shared code.

**New commands:**
- gcx datasources mysql query
- gcx datasources mysql list-tables
- gcx datasources mysql describe-table

**Querying**
query executes raw SQL (with Grafana server-side macros like $__timeFilter). list-tables and describe-table discover the schema via information_schema, with --database to filter. The generic gcx datasources query <uid> 'SQL' auto-detect path also supports mysql now.

**Notes:**
- Same shape as PostgreSQL (#969): reuses the shared grafanaquery transport and internal/query/sql response layer — no shared-code changes, no new codecs. The wire contract is identical (both use Grafana's sqleng backend); verified live before implementation.
- LIMIT is only appended to SELECT-shaped statements (SELECT/WITH/TABLE/VALUES) — MySQL accepts LIMIT on UPDATE/DELETE and INSERT…SELECT, so DML passes through untouched. MySQL-specific bails within SELECTs: LIMIT offset,count, INTO OUTFILE, LOCK IN SHARE MODE. An explicit LIMIT above 1000 is capped with a stderr warning; --limit 0 disables.
- No docker-compose change — local testing reuses the stack's existing mysql container with a separate testdb database, so tests never touch Grafana's own backing store.


dafydd-t:

internal/query/mysql/types.go
```diff
@@ -0,0 +1,47 @@
...
+	if name == "" {
+		return nil
+	}
+	if !identifierRe.MatchString(name) {
+		return fmt.Errorf("invalid %s: must contain only letters, numbers, underscores, and dots", field)
+	}
+	return nil
+}
+
+var limitBailRe = regexp.MustCompile(`(?im)(\bLIMIT\s+\d+\s+OFFSET\b|\bLIMIT\s+\d+\s*,|\bOFFSET\s+\d+\b|\bINTO\s+(OUTFILE|DUMPFILE)\b|\bFOR\s+(UPDATE|SHARE)\b|\bLOCK\s+IN\s+SHARE\s+MODE\b|^\s*EXPLAIN\b|^\s*SHOW\b|^\s*DESC(RIBE)?\b)`)
```
dafydd-t: claude doesn't like a part of this regex:

The (?m) flag makes `^\s*DESC(RIBE)?\b` match at any line start, so a formatted query like `SELECT * FROM t ORDER BY created_at\nDESC` bails and gets no LIMIT appended — silently skipping the safety net on exactly the kind of formatted SQL agents emit. Postgres has the same latent issue with EXPLAIN/SHOW, but DESC at line start is far more plausible. Fix: drop the m flag (only the ^-anchored alternatives are affected; \s* already covers leading newlines) — worth applying to #969 too. A test case for inline ORDER BY x DESC would pin it.

dafydd-t: my question, your call: would it be easier to flip the use of the regex, to say explicitly which scenarios we _would_ add  the limit to, rather than a list of commands to bail on? it could plausibly be longer.

internal/query/mysql/client.go
dafydd-t, line 1: after #969 is merged, it would be nice to dedupe these clients, i think they are the same

internal/query/mysql/types.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+	"time"
+
+	querysql "github.com/grafana/gcx/internal/query/sql"
+)
+
+// EscapeSQLString escapes single quotes for use in SQL string literals.
+func EscapeSQLString(s string) string {
+	return strings.ReplaceAll(s, "'", "''")
+}
```
dafydd-t, line 18: maybe worth a note that we already reject backticks in `ValidateIdentifier`

internal/query/mysql/types.go
alyssajoyner: i agree that its easier to _flip_ the use of the regex - it is now an allow list. that also made the EXPLAIN/SHOW/DESC entries redundant so theyre removed. I will also apply this to #969 as well

internal/query/mysql/types.go
```diff
@@ -0,0 +1,71 @@
...
+var limitStatementRe = regexp.MustCompile(`(?is)^\s*(SELECT|WITH|TABLE|VALUES)\b`)
+
+// The DML keywords catch CTE-wrapped writes (e.g. WITH ... UPDATE ...),
+// which start with WITH and so pass limitStatementRe. Matching them anywhere
+// means a SELECT mentioning e.g. 'DELETE' in a string literal also skips
+// enforcement — that fails safe (no LIMIT added) rather than limiting a write.
+// EXPLAIN/SHOW/DESCRIBE need no bail entries: limitStatementRe already
+// excludes them, and line-anchored entries would misfire on formatted
+// queries (e.g. a multiline ORDER BY ... \nDESC).
+var limitBailRe = regexp.MustCompile(`(?i)(\bLIMIT\s+\d+\s+OFFSET\b|\bLIMIT\s+\d+\s*,|\bOFFSET\s+\d+\b|\bINTO\s+(OUTFILE|DUMPFILE)\b|\bFOR\s+(UPDATE|SHARE)\b|\bLOCK\s+IN\s+SHARE\s+MODE\b|\b(INSERT|UPDATE|DELETE|REPLACE)\b)`)
```
awsome-o, line 49: `REPLACE` is in the DML set here, but it's also MySQL's most common string function, so this silently disables limit enforcement for an ordinary read:

```
SELECT REPLACE(col,'a','b') FROM big_table   ->  no LIMIT appended, no warning
```

Someone running that against a large table gets the whole thing back. Worth noting the `\b` boundaries don't help, since `REPLACE(` has a word boundary too. This needs token or comment-aware handling keyed on statement position, and it should keep protecting genuine writes, including CTE-wrapped ones.

Could we get regressions for `SELECT REPLACE(...)` receiving a LIMIT, and for a real `REPLACE` write plus a `WITH ... REPLACE` still bailing out?

Second issue in the same place: MySQL supports both `#` and `-- ` line comments (https://dev.mysql.com/doc/refman/8.0/en/comments.html), and since the limit is appended as a plain suffix, a query ending in either swallows it and returns unbounded rows. `#` doesn't exist in Postgres, so that one is specific to this dialect.

internal/datasources/mysql/query.go
```diff
@@ -0,0 +1,128 @@
...
+	dsquery.SharedOpts
+
+	Datasource string
+	Limit      int
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	opts.Setup(flags, false)
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.mysql is configured)")
+	flags.IntVar(&opts.Limit, "limit", defaultLimit, "Max rows to return (0 disables enforcement)")
```
awsome-o, line 31: Same limit contract as the Postgres sibling. An existing `LIMIT` at or below 1000 wins over the flag and it's ignored entirely (`EnforceLimit("SELECT * FROM t LIMIT 900", 10, 1000)` returns the query untouched), and a negative value goes straight into the SQL. The help says "Max rows to return", which doesn't match either behaviour.

Worth fixing consistently with #969 so the two dialects don't diverge.

internal/datasources/mysql/list_tables.go
```diff
@@ -0,0 +1,101 @@
...
+
+			if err := mysql.ValidateIdentifier(opts.Database, "database"); err != nil {
+				return err
+			}
+
+			sql := "SELECT table_schema AS `database`, table_name AS name, table_type AS type FROM information_schema.tables WHERE table_schema NOT IN ('mysql', 'information_schema', 'performance_schema', 'sys')"
+			if opts.Database != "" {
+				sql += fmt.Sprintf(" AND table_schema = '%s'", mysql.EscapeSQLString(opts.Database))
+			}
+			sql += " ORDER BY table_schema, table_name LIMIT 500"
```
awsome-o, line 77: Caps at 500 silently, while the Long text says it lists tables and views from all non-system databases. A stderr hint is enough here; no `list_meta` work needed, since `docs/design/output.md` §15 is still proposed and opt-in.

internal/datasources/mysql/list_tables.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return err
+			}
+
+			datasourceUID, _, err := dsquery.ResolveValidateAndSaveDatasource(ctx, loader, opts.Datasource, cfgCtx, cfg, "mysql")
+			if err != nil {
+				return err
+			}
+
+			if err := mysql.ValidateIdentifier(opts.Database, "database"); err != nil {
```
awsome-o, line 69: Two things. `--database ""` reads as omitted and broadens scope to every database rather than narrowing, and this identifier check runs after `LoadContextAndConfig` at `:59`, so validation happens after config loading and datasource resolution, which may perform Grafana I/O. `describe-table` validates first, so the same bad input fails differently depending on which command you reach for.

Non-blocking: neither `--database` nor the TABLE argument documents case-sensitivity, which is worth stating given `information_schema` folds case differently from on-disk table names on Linux.

awsome-o (changes requested):
The wiring here is correct (output classes, token costs, hints, Args validators, codecs, CODEOWNERS, package map, generated reference all in order) and the client tests are the strongest part of this.

Two of the findings below are genuinely MySQL-specific rather than inherited from the Postgres shape, plus there's one logistics item about the stack.

**Stacking.** This is based on an older Postgres head. The merge-base with #969 is `c130ece5`, and #969's current head `47768e87` is not an ancestor of this branch. Because we squash-merge, once #969 lands the ancestry link disappears and merging this produces add/add conflicts on `internal/query/postgres/types.go` and `types_test.go`, where this branch holds the *older* `limitBailRe` and is missing #969's regression test. Resolving those toward this branch would quietly undo the fix and delete the test guarding it.

So after #969 lands, please rebase onto main so this becomes a MySQL-only diff, and confirm `limitBailRe` still reads `(?i)` with the multiline regression present.

On tests: focused regressions on the emitted SQL for each limit fix, and one asserting validation fires before loader I/O. Your client tests already pin the wire shape and don't need duplicating.

Two things that are ours, not yours: a `maintidx` failure in `cmd/gcx/datasources/query.go` after rebasing is not your diff (tracked in #1137), and the exit-1-on-validation-error problem is getting a shared helper in #1138, so please don't hand-roll one.

internal/query/mysql/client.go
ktw4071: You were right, they're identical — extracted the shared body into `querysql.BuildRawQueryBody` (`8ec5979b`), which both mysql and postgres now call. Left ClickHouse alone since its sqlds-based plugin sends a numeric `format` instead of `"table"` and doesn't actually fit this helper. Both dialects' existing wire-shape tests pass unchanged, and I live-verified `query`/`list-tables` against both `mysql-ds-m` and `postgres-ds-m` post-refactor to be sure nothing moved.

internal/query/mysql/types.go
ktw4071: Both issues confirmed and fixed in `7e333aed`.

REPLACE is now told apart from REPLACE(...) by position rather than `\b`: a write's REPLACE is never immediately followed by `(`, a function call's always is (`isReplaceWrite` in `types.go`). Added the regression tests you asked for — `SELECT REPLACE(...)` gets a LIMIT, a real `REPLACE INTO` write still bails, and a CTE-wrapped `WITH ... REPLACE INTO ...` write still bails too.

For the comment issue: `#` and `-- ` line comments now bail (pass through unchanged) instead of getting a dead LIMIT n folded into the comment — matches postgres's identical `13475c07` fix for `--`, extended here to also cover MySQL's `#`.

Live-verified against `mysql-ds-m` (a real 78k-row table) with `--limit 5`:
```
SELECT REPLACE(base_country, 'Name', 'Country') AS c FROM grafanadb.world_data
before: 100 rows (--limit silently ignored)
after:    5 rows (--limit honored)
```

internal/datasources/mysql/query.go
ktw4071: Fixed in `8dc4321a`, mirroring postgres's `09d037dd` exactly.

`--limit < 0` is now rejected locally before any query runs. The "silently reduced to 1000 with nothing on stderr" half was already fixed for free here — it lives in the shared `internal/query/sql.EnforceLimit`, fixed for postgres and inherited automatically by mysql's own wrapper once I merged main in.

Live-verified against `mysql-ds-m`:
```
--limit=-5   before: db query error: ... syntax error ... near '-5'
             after:  --limit must be >= 0, got -5
--limit 5000 -> warns on stderr and rewrites the SQL's LIMIT clause to 1000 (both before and after — inherited fix)
```
Note: the actual row count returned here is capped at ~100 regardless of the SQL's own LIMIT — a separate, pre-existing response cap on this environment, unrelated to this fix.

internal/datasources/mysql/list_tables.go
ktw4071: Fixed in `02fef0a8`, mirroring postgres's `1b19d165`. `list-tables` now requests 501 rows and warns on stderr when the 501st shows up, instead of silently truncating.

Only 1 non-system table exists on our test instance so the cap itself can't be triggered live — covered instead by unit tests on the extracted `buildListTablesSQL`/`warnIfTruncated` helpers, same as postgres's own test shape. If you want the full validation, I can run MySQL locally and add enough tables to actually exercise the 500+ path end-to-end — happy to do that if it'd help close this out with live evidence rather than just the unit tests.

internal/datasources/mysql/list_tables.go
ktw4071: Fixed in `02fef0a8`. `--database ""` now rejected via `cmd.Flags().Changed`, checked before `LoadContextAndConfig`/datasource resolution — matches `describe-table`'s existing validation order (and postgres's). Applied the same fix to `describe-table`'s own `--database` for consistency, since it had the identical silent-widening risk. Also added the case-sensitivity note to both commands' help text.

Live-verified against `mysql-ds-m`:
```
--database ""  before: same result as omitted (1 row — silently widened)
               after:  --database must not be empty
```

awsome-o (dismissed):
Approving. I went through all six threads against the head tree rather than the replies, and every one has a real fix behind it: `BuildRawQueryBody` is a faithful extraction of the body builder that was inline before, and both dialects call it; `REPLACE` is told apart from `REPLACE(...)` by position; `--limit` rejects negatives in `Validate()` and both `capped` branches warn; `list-tables` requests 501 rows and warns on stderr; `--database ""` is checked via `Flags().Changed` before `LoadContextAndConfig`. #969's `limitBailRe` fix survived the merge from main too, and the `maintidx` item is gone now that #1141 landed.

One thing I'd still fix, here or in a followup, whichever is easier. `trailingLineCommentRe` in `internal/query/mysql/types.go` requires a space or tab after `--`, so a statement ending in a bare `--` doesn't bail, and appending the suffix is what creates the comment that swallows the LIMIT: `SELECT * FROM big_table --` becomes `SELECT * FROM big_table -- LIMIT 100`. MySQL would have rejected the original as a double unary minus, so we turn a syntax error into an unbounded scan. `(--([ \t][^\n]*)?|#[^\n]*)$` closes it. Postgres's looser pattern already bails on this shape, so the gap came in with the MySQL tightening.

Two nits I'm not asking for. `--database mysql` (or `information_schema`, `performance_schema`, `sys`) ANDs against the system-schema exclusion and comes back empty with nothing explaining why. And the "fails safe" comments read as safe for reads when they're only safe for writes: `limitStatementRe` anchors at the start, so a query with a leading comment never gets a LIMIT at all, which is a shape LLMs emit fairly often. Both are fine to leave.

Followup for its own PR, not this one: `splitTableArg` and `warnIfTruncated` are the remaining copy-paste against postgres, differing only in the words schema/database. Worth lifting into `internal/query/sql` before a third dialect copies them a third time.

I'm also adding the missing half of our own guidance — `add-datasource` named `grafanaquery` and `dataframe` but said nothing about raw-SQL body building or LIMIT enforcement, which is why this PR and #969 each spent three rounds on the same bail logic.

Nothing here blocks. The changes-requested review from 5 Aug is what's holding the merge button, so it needs a re-review to clear.

ktw4071:
Thanks for the thorough pass. Fixed the one remaining item in `57e33324` — `trailingLineCommentRe`'s space/tab requirement made the `--` alternative optional (`(--([ \t][^\n]*)?|#[^\n]*)$`), so a bare trailing `--` now bails same as `-- text` and `# text` already did. Added `TestEnforceLimit/bail_on_bare_trailing_--` (and the after-semicolon variant), which fails against the old regex — it asserts the query comes back unchanged, and the old code corrupted it into `... -- LIMIT 100`.

One correction to the theory, found while trying to live-verify: on `mysql-ds-m`'s MySQL version, a bare trailing `--` at true end-of-statement is accepted as a valid empty comment, not rejected as a double-unary-minus syntax error — so before/after both run successfully with identical row counts (the platform's own response cap masks it either way). The defect and the fix are real regardless — proven at the `EnforceLimit` output level via the unit test — but I didn't want to claim a live reproduction I couldn't actually show.

Also caught and fixed in `6bc2c0e4`: the merge from main dropped `mysql` from two golden-list tests (`TestQueryRoutesSupportedKindsIsTheSortedUnion` and the unsupported-kind message) when it picked up `cloudmonitoring`'s addition to the same lists — the routing code itself was untouched and still dispatches mysql correctly, just the fixtures regressed. Caught by the existing tests failing post-merge.

Leaving the two nits and the `splitTableArg`/`warnIfTruncated` follow-up as you scoped them — not touching this PR for those.

awsome-o (approved):

awsome-o:
One more thing spotted on a second pass over the diff, non-blocking and cheap. `request.go` documents the split correctly on `RawQueryRequest`, but the package doc at the top of `internal/query/sql/types.go` still says the opposite:

> Dialect packages keep their own request construction, schema discovery, and LIMIT bail rules; the common response shape, table formatting, response parsing, and LIMIT clamping live here.

Once this merges, that sentence tells the next dialect author to hand-roll exactly what you just extracted — which is the failure mode that cost this PR and #969 three rounds each. You're already adding a file to that package, so it's a one-line fold-in: move request construction to the shared list, leave schema discovery and the bail rules dialect-local.

Approval stands either way.

PR #971, opened 2026-07-14
feat(datasources): add support for Elasticsearch datasource

alyssajoyner:
Used the add-datasource skill to add the Elasticsearch datasource.

**New commands:**
- gcx datasources elasticsearch query
- gcx datasources elasticsearch logs
- gcx datasources elasticsearch metrics
- gcx datasources elasticsearch list-indices
- gcx datasources elasticsearch list-fields

**Querying**

Three query modes, all taking an optional Lucene expression (omit to match all documents):
- query — matching documents as a table (raw_data): query 'app:frontend AND level:error' --since 1h --size 20
- logs — newest-first log view; plugin-internal fields (_source, sort, highlight) trimmed
- metrics — aggregations over a time histogram, optionally split into series: metrics 'level:error' --agg count --group-by app.keyword --since 6h. Supports count/avg/sum/min/max/cardinality and -o graph for terminal charts. Empty buckets are dropped (min_doc_count: 1) so tables show data, not zero-fill.

list-indices and fields discover indices and field names (with types) from the plugin's _mapping endpoint — everything needed to write the queries. The index pattern itself comes from datasource config (the ES query model has no per-query index field).

**Notes:**
- Reuses the shared grafanaquery transport and tabular response layer; metrics adds a typed series response with a graph adapter + codec entries (same pattern as InfluxDB's graph support in #539).
- Advanced aggregation types (percentiles, extended stats, top metrics, rate, pipeline aggs like moving average) are excluded — each has a distinct response shape and can be added as new --agg values later.
- No changes to resolve.go or the auto-detect command (elasticsearch is already a passthrough plugin ID) — no overlap with #969/#970.
- Wire-shape pins for the probe discoveries: intervalMs/maxDataPoints are required (their absence triggers ES "too many buckets" errors), settings values are strings, and group labels ride on the frame name.


dafydd-t:

dafydd-t:
Thanks for this.  only minor note: the expr-resolution + default-time-range + client-setup block is duplicated between runSearch in run.go and the metrics RunE - could be shared, but not worth blocking on.

cmd/gcx/root/testdata/output_classes.json
```diff
@@ -180,6 +180,11 @@
  "gcx datasources cloudwatch query": "finite",
  "gcx datasources create": "finite",
  "gcx datasources delete": "finite",
+ "gcx datasources elasticsearch fields": "finite",
```
dafydd-t:
```suggestion
 "gcx datasources elasticsearch list-fields": "finite",
```

would `list-fields` make sense as a command name? We want commands to end with a verb or verb-resource pattern

dafydd-t: can you check against the naming conventions we're encoding [here](https://github.com/grafana/gcx/pull/1023)? it will either be `list-fields` or `fields list | fields get`

alyssajoyner: renamed to list-fields 👍

internal/query/elasticsearch/types.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// FieldInfo describes a field within an index mapping.
+type FieldInfo struct {
+	Index string `json:"index"`
+	Name  string `json:"name"`
+	Type  string `json:"type"`
+}
+
+type mappingProperty struct {
```
awsome-o, line 141: `mappingProperty` has `Type` and `Properties` but no `fields`, so `flattenProperties` at `:179` walks straight past every multi-field.

Running `ParseMapping` over a mapping with `message` (text, with a `keyword` sub-field) and `host.name` (same shape) returns four fields instead of six, and `message.keyword` is simply absent.

What makes this more than a gap is that `metrics --group-by` help at `metrics.go:30` actively tells callers to use exactly those names, so the discovery command can't surface what its own sibling instructs you to type. Multi-fields are standard mapping behaviour, not an edge case: https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/multi-fields

A regression over a mapping containing a `fields` sub-mapping would lock this.

internal/datasources/elasticsearch/logs.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/grafana/gcx/internal/providers"
+	"github.com/grafana/gcx/internal/query/elasticsearch"
+	querysql "github.com/grafana/gcx/internal/query/sql"
+	"github.com/spf13/cobra"
+)
+
+// LogsCmd returns the `logs` subcommand for an Elasticsearch datasource parent.
+func LogsCmd(loader *providers.ConfigLoader) *cobra.Command {
+	return newSearchCmd(loader, searchCmdSpec{
+		use:   "logs [EXPR]",
```
awsome-o, line 16: Owner decision, so this isn't yours to litigate. Please fold this into `elasticsearch query [EXPR] --mode logs` rather than shipping a separate leaf.

These two share the Lucene expression, the time controls, the output shape and the implementation, and there's no stated rule today for choosing between them, so agent routing between them is a coin flip. A validated mode enum with a documented default of `documents` is clearer than two competing leaves. Please reject unknown modes locally.

Keep `metrics` as its own leaf. Its inputs and result schema genuinely differ. Worth noting that isn't a protocol-class distinction, since all five leaves are `finite` in `output_classes.json`.

The general rule is being written into the naming guide in #1139.

internal/datasources/elasticsearch/query.go
```diff
@@ -0,0 +1,56 @@
...
+
+  # Output as JSON, limit results
+  gcx datasources elasticsearch query -d UID 'datacenter:us-east' --size 20 -o json
+
+  # Print a Grafana Explore share link for the executed query
+  gcx datasources elasticsearch query 'level:error' --since 1h --share-link
+
+  # Continue the same search in Grafana Explore
+  gcx datasources elasticsearch query 'level:error' --since 1h --open`,
+		sizeFlag:       "size",
```
awsome-o, line 46: While you're restructuring, could we settle on one row control? This exposes the knob as `--size` while `logs.go:40` exposes the same knob as `--limit`. Please standardize on `--limit`, default 100, valid 1 to 1000.

Related, at `run.go:39-43` the clamp silently rewrites out-of-range input: `<= 0` becomes 100 and `> 1000` becomes 1000. The 1000 cap is at least stated in the flag help, so this is about rejection rather than disclosure. Rejecting with a clear message beats silently changing what the caller asked for. Note `--size 0` currently means 100 here, while the generic sibling documents `--limit 0` as no limit, so the same-shaped input means opposite things in two places.

internal/datasources/elasticsearch/list_indices.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return err
+			}
+
+			client, err := elasticsearch.NewClient(cfg)
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			indices, _, err := client.Mapping(ctx, datasourceUID, "")
```
awsome-o, line 68: This always pulls the full `_mapping` for every index and exposes no way to narrow, unlike `list-fields`. On a large cluster it hits the 4 MB response cap and the resulting error advises narrowing that isn't actually possible, so there's no workaround available to the caller.

Either accept an index pattern here, or use a cheaper enumeration endpoint.

Non-blocking: `--time-field ""` silently becomes `@timestamp` rather than being a usage error, and `docs/architecture/project-structure.md:95` names a `fields` command that should be `list-fields`.

awsome-o (changes requested):
Clean integration, and the transport and dataframe reuse is right. Every mechanical gate I ran is green: build, targeted tests, the `cmd/gcx/root` conformance suites, lint, and CLI-reference drift.

Four things before merge. One data-correctness bug, one consolidation I'm deciding as owner, one routing gap, and one unbounded discovery path.

**Generic query routing.** This didn't land on an inline comment because the file isn't in your diff, but `gcx datasources query <es-uid>` currently answers `datasource type "elasticsearch" is not supported (supported: prometheus, loki, pyroscope, influxdb, clickhouse)` while a full `datasources elasticsearch` subtree ships in this PR. That's confusing for anyone who tries the generic form first, and the PR body's note that no auto-detect changes were needed is true of `resolve.go` but not of the command.

Since your expression is a plain Lucene string, the generic `<uid> <expr>` form can carry it honestly, so please add a dispatch case in `cmd/gcx/datasources/query.go` rather than a redirect. A small routing regression alongside it would be good. If you'd rather not dispatch, an explicit redirect is acceptable, but it has to sit before `shared.ResolveExpr` so an argument-less call gets the redirect instead of "expression required". The bare "not supported" default is the one option I don't want to ship.

On tests: command-level regressions for the mode enum, the limit bounds, and generic dispatch. Your `internal/query/elasticsearch/client_test.go` already covers the wire shape, so no duplication needed there.

Ours, not yours: the exit-1-on-validation-error problem is getting a shared helper in #1138, so please don't hand-roll a wrapper. The naming rule I'm applying below is being written down in #1139.

internal/query/elasticsearch/types.go
ktw4071: Fixed in 263aec59.

Added `Fields map[string]mappingProperty` to `mappingProperty` and flatten each entry as its own dotted sibling field (`parent.subfield`), same shape as the existing nested-object handling.

Verified against a live Elasticsearch 9 datasource, rebuilding the binary from both sides of the fix for a clean comparison:

```
$ gcx datasources elasticsearch list-fields -d UID --index grafana-logs -o json   # before
[
  {"index": "grafana-logs", "name": "@timestamp", "type": "date"},
  {"index": "grafana-logs", "name": "app", "type": "text"},
  {"index": "grafana-logs", "name": "datacenter", "type": "text"},
  {"index": "grafana-logs", "name": "duration_ms", "type": "long"},
  {"index": "grafana-logs", "name": "host", "type": "text"},
  {"index": "grafana-logs", "name": "level", "type": "text"},
  {"index": "grafana-logs", "name": "logger", "type": "text"},
  {"index": "grafana-logs", "name": "message", "type": "text"},
  {"index": "grafana-logs", "name": "request_id", "type": "text"},
  {"index": "grafana-logs", "name": "type", "type": "text"},
  {"index": "grafana-logs", "name": "user_id", "type": "text"}
]
```
11 fields, zero `.keyword` entries — `message.keyword` genuinely absent, exactly as described.

```
$ gcx datasources elasticsearch list-fields -d UID --index grafana-logs -o json   # after
[
  {"index": "grafana-logs", "name": "@timestamp", "type": "date"},
  {"index": "grafana-logs", "name": "app", "type": "text"},
  {"index": "grafana-logs", "name": "app.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "datacenter", "type": "text"},
  {"index": "grafana-logs", "name": "datacenter.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "duration_ms", "type": "long"},
  {"index": "grafana-logs", "name": "host", "type": "text"},
  {"index": "grafana-logs", "name": "host.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "level", "type": "text"},
  {"index": "grafana-logs", "name": "level.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "logger", "type": "text"},
  {"index": "grafana-logs", "name": "logger.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "message", "type": "text"},
  {"index": "grafana-logs", "name": "message.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "request_id", "type": "text"},
  {"index": "grafana-logs", "name": "request_id.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "type", "type": "text"},
  {"index": "grafana-logs", "name": "type.keyword", "type": "keyword"},
  {"index": "grafana-logs", "name": "user_id", "type": "text"},
  {"index": "grafana-logs", "name": "user_id.keyword", "type": "keyword"}
]
```
20 fields now, including `message.keyword`.

One correction to be precise about scope: I also compared `metrics --group-by app.keyword` before/after, and it returned the same 6 real series (loki, faro, tempo, mimir, grafana, beyla) both times — Elasticsearch itself already accepted the raw field name regardless of whether `list-fields` could see it. The bug was specifically about discoverability, not about this aggregation being broken, which matches how you scoped it ("the discovery command can't surface what its own sibling instructs you to type").

Regression test added over a mapping with both a top-level multi-field (`message`) and a nested-object multi-field (`host.name`), matching the exact case you described.

internal/datasources/elasticsearch/logs.go
ktw4071: Fixed in dee4c512.

`logs` is gone; `query` now takes `--mode documents` (default) or `--mode logs`, rejected locally if it's anything else. `metrics` untouched.

```
$ gcx datasources elasticsearch logs -d UID --since 1h --limit 2   # before: this command existed
-> succeeded, returned rows

$ gcx datasources elasticsearch logs -d UID   # after: command is gone
-> Unknown shorthand flag
```

Same query, both modes, showing the real difference — `_type`/`highlight`/`sort` present in `documents`, gone in `logs`:

```
$ gcx datasources elasticsearch query -d UID --since 1h --limit 2 -o json
{
  "columns": [
    {"name": "@timestamp", "type": "time"}, {"name": "_id", "type": "string"},
    {"name": "_index", "type": "string"}, {"name": "_type", "type": "other"},
    {"name": "app", "type": "string"}, {"name": "datacenter", "type": "string"},
    {"name": "highlight", "type": "other"}, {"name": "host", "type": "string"},
    {"name": "metric", "type": "string"}, {"name": "sort", "type": "other"},
    {"name": "tags.app", "type": "string"}, {"name": "tags.datacenter", "type": "string"},
    {"name": "tags.host", "type": "string"}, {"name": "type", "type": "string"},
    {"name": "value", "type": "number"}
  ],
  "rows": [
    ["2026-08-19T21:51:30Z", "GSICHKAB5NCosIG0CAgK", "grafana-metrics", null, "tempo", "us0", null, "host5", "memory.usage_bytes", [1787176290000, 2021969], "tempo", "us0", "host5", "metric", 116062937912],
    ["2026-08-19T21:51:30Z", "GCICHKAB5NCosIG0CAgK", "grafana-metrics", null, "mimir", "us1", null, "host2", "cpu.utilisation", [1787176290000, 2021968], "mimir", "us1", "host2", "metric", 361]
  ]
}

$ gcx datasources elasticsearch query -d UID --mode logs --since 1h --limit 2 -o json
{
  "columns": [
    {"name": "@timestamp", "type": "time"}, {"name": "_id", "type": "string"},
    {"name": "_index", "type": "string"}, {"name": "app", "type": "string"},
    {"name": "datacenter", "type": "string"}, {"name": "host", "type": "string"},
    {"name": "id", "type": "string"}, {"name": "metric", "type": "string"},
    {"name": "tags.app", "type": "string"}, {"name": "tags.datacenter", "type": "string"},
    {"name": "tags.host", "type": "string"}, {"name": "type", "type": "string"},
    {"name": "value", "type": "number"}
  ],
  "rows": [
    ["2026-08-19T21:51:30Z", "GSICHKAB5NCosIG0CAgK", "grafana-metrics", "tempo", "us0", "host5", "grafana-metrics#GSICHKAB5NCosIG0CAgK", "memory.usage_bytes", "tempo", "us0", "host5", "metric", 116062937912],
    ["2026-08-19T21:51:30Z", "GCICHKAB5NCosIG0CAgK", "grafana-metrics", "mimir", "us1", "host2", "grafana-metrics#GCICHKAB5NCosIG0CAgK", "cpu.utilisation", "mimir", "us1", "host2", "metric", 361]
  ]
}
```

Mode validation:
```
$ gcx datasources elasticsearch query -d UID --mode bogus
-> --mode must be "documents" or "logs", got "bogus"
```

The `--limit` standardization is in the same commit — see my reply on the `query.go:46` thread for that before/after.

internal/datasources/elasticsearch/query.go
ktw4071: Fixed in dee4c512 (same commit as the mode restructure above, since you tied the two together).

Both commands' row-limit knob is now `--limit`, default 100, valid 1-1000, rejected (not clamped) outside that range. Rebuilt the binary from both sides of the fix to confirm the actual before/after behavior, not just the error path:

```
$ gcx datasources elasticsearch query -d UID --size 0 --since 1h -o json   # before
-> 100 rows returned, no error or warning -- silent clamp

$ gcx datasources elasticsearch query -d UID --size 5000 --since 1h -o json   # before
-> 1000 rows returned, no error or warning -- silent clamp

$ gcx datasources elasticsearch query -d UID --limit 0   # after
-> --limit must be between 1 and 1000, got 0

$ gcx datasources elasticsearch query -d UID --limit 5000   # after
-> --limit must be between 1 and 1000, got 5000
```

On the `--size 0` / `--limit 0` clash you flagged: Elasticsearch's search `size` has no "unlimited" concept the way the generic sibling's datasources do, so rather than bolt on a second special meaning for `0` here, I made `0` simply invalid -- that removes the ambiguity instead of trying to match it.

Also collapsed `searchCmdSpec`/`newSearchCmd`/`runSearch` in `run.go` -- that machinery existed only to share code between `query` and `logs`; with one command instead of two it wasn't needed anymore.

internal/datasources/elasticsearch/list_indices.go
ktw4071: Fixed in 406eef2c.

Added the same `--index` flag `list-fields` already exposes on this exact client method -- no new client code needed.

```
$ gcx datasources elasticsearch list-indices -d UID --index grafana-metrics   # before
-> error: Unknown flag --index

$ gcx datasources elasticsearch list-indices -d UID -o json   # before, always everything
[{"name": "grafana-logs", "fields": 11}, {"name": "grafana-metrics", "fields": 10}]

$ gcx datasources elasticsearch list-indices -d UID -o json   # after
[{"name": "grafana-logs", "fields": 20}, {"name": "grafana-metrics", "fields": 18}]

$ gcx datasources elasticsearch list-indices -d UID --index grafana-metrics -o json   # after
[{"name": "grafana-metrics", "fields": 18}]

$ gcx datasources elasticsearch list-indices -d UID -o table
INDEX            FIELDS
grafana-logs     20
grafana-metrics  18
```

(Field counts jumped between before/after because the multi-field fix landed in the same branch -- 11->20 and 10->18 reflects both fixes together, not just `--index`.)

This made `list_indices.go`/`list_fields.go` structurally near-identical (the `dupl` linter flagged it), so I extracted the shared shape into a `newMappingListCmd`/`mappingListSpec` builder in a new `mapping_list.go`.

Also fixed the `project-structure.md:95` doc typo as part of the `logs`->`query --mode` commit, since I was touching that same line anyway for the `logs` removal. Left `--time-field ""` as-is for now -- happy to take it as a follow-up if you'd like it in scope here rather than filed separately.

internal/query/elasticsearch/client.go
```diff
@@ -0,0 +1,315 @@
...
+	if bucketAggs == nil {
+		bucketAggs = []any{}
+	}
+	return map[string]any{
+		"refId":      "A",
+		"datasource": map[string]any{"type": pluginID, "uid": dsUID},
+		"query":      query,
+		"metrics":    []any{metric},
+		"bucketAggs": bucketAggs,
+		"timeField":  orDefault(timeField, DefaultTimeField),
```
awsome-o, line 139: **required** — the `timeField` we send here never gets read, so `--time-field` isn't controlling range filtering on any of these commands. I checked the plugin at `17d7c528`: `Query` in `pkg/elasticsearch/models.go:11-30` has no `timeField` field, so the key is dropped at unmarshal. `processQuery` in `pkg/elasticsearch/data_query_processor.go:20,27` builds the range filter from `e.client.GetConfiguredFields().TimeField`, which is the datasource's own setting. And `addDateHistogramAgg` in `pkg/elasticsearch/aggregation_factory.go:21-26` is the one place a per-query field survives, via `bucketAgg.Field`.

That splits two ways. On `query` and `query --mode logs` the flag does nothing at all, while `query.go:43` tells people it's the "Time field used for range filtering". On `metrics` your help text is accurate about the histogram, but `AggsQueryModel` always fills the date_histogram `field` in with `orDefault(req.TimeField, DefaultTimeField)`, so the bucket moves and the filter doesn't. On a datasource configured with `event.time` that means filtering on `event.time` and bucketing on `@timestamp` — empty series if the documents have no `@timestamp`, wrong ones if they have both. The `@timestamp` default is why none of this shows up in the tests.

For `metrics` the fix falls out of that last citation: pass `req.TimeField` straight through instead of defaulting it, and an unset flag leaves `field` empty so the plugin drops the configured field into both the filter and the bucket. For `query` I'd just remove the flag rather than document a no-op. But if I've misread the plugin, or you're aiming at a release that does parse `timeField`, tell me and I'll withdraw it.

internal/datasources/elasticsearch/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+	flags.StringVar(&opts.GroupBy, "group-by", "", "Split series by this field's terms (use .keyword for text fields)")
+	flags.IntVar(&opts.GroupSize, "group-size", 10, "Max number of series when using --group-by")
+	flags.StringVar(&opts.TimeField, "time-field", elasticsearch.DefaultTimeField, "Time field for the date histogram")
+}
+
+func (opts *metricsOpts) Validate() error {
+	if err := opts.SharedOpts.Validate(); err != nil {
+		return err
+	}
+	return elasticsearch.ValidateAgg(opts.Agg, opts.Field)
```
awsome-o, line 42: **recommended** — `--group-size` gets to `AggsQueryModel` unvalidated, and then `client.go:90-92` quietly rewrites anything `<= 0` back to 10. So `metrics --group-by app.keyword --group-size 0` hands back 10 series and says nothing about it, and `--group-size -5` does the same. That's the silent clamp we just took out of `--limit`, still sitting on the other row-count knob in the same package, which means the same bad input fails two different ways depending on which leaf someone reached for. Upper bound is up to you; the lower one is the part I'd fix.

```suggestion
	if opts.GroupSize < 1 {
		return fmt.Errorf("--group-size must be at least 1, got %d", opts.GroupSize)
	}
	return elasticsearch.ValidateAgg(opts.Agg, opts.Field)
```

awsome-o (commented):
Second look at this. Everything I asked for last round is in, and the fixes are the real thing rather than paperwork. Multi-fields come out as `parent.subfield` now, and the test even covers one sitting under an object. `logs` is gone and lives as `query --mode logs`, with the mode enum rejected in `Validate()` before any I/O happens. `--size` is `--limit` with 1-1000 bounds and a real error instead of that silent reset to 100. `list-indices` takes `--index` and hands it to `Client.Mapping`. I re-ran the gates on 5c08d2c3 and they're green: build, both elasticsearch packages, `internal/datasources/query`, `cmd/gcx/datasources` and the `cmd/gcx/root` conformance suites, on go1.26.5. I'm closing those four threads.

One new thing though, and it's a real bug rather than a taste call, so I'd rather you saw it before this lands. `--time-field` doesn't do what its help says. Detail is on the inline comment; the gist is that the plugin's `Query` struct has no `timeField` field at all, so the key we send gets dropped when it unmarshals. The range filter always comes from whatever time field the datasource itself is configured with, and the only thing that honours a per-query field is the date histogram. So on `query` and `--mode logs` the flag does nothing, and on `metrics` it moves the bucket field without moving the filter field. Point that at a datasource configured with `event.time` and you end up filtering on `event.time` while bucketing on `@timestamp`: empty series if the documents have no `@timestamp`, quietly wrong ones if they have both. The `@timestamp` default hides all of it, which is why nothing in the tests caught it. If I've read the plugin wrong, or you're targeting a release that does parse `timeField`, say so and I'll drop it.

On generic query routing, I'm taking that back as a merge requirement, and I owe you two corrections. First, the guidance is clearer than I gave it credit for. `add-datasource/SKILL.md` says outright that adding neither a dispatch nor a redirect is also a choice, so long as you make it deliberately, and `integrate-with-gcx` calls reaching the generic command a judgement rather than a checkbox. Both of those arrived with #1141, which is in your head commit, so they weren't in view when I wrote the original comment. Second, I said a dispatch was about 18 lines with no codec work, and that was wrong: you'd have to decide what the generic `--limit` means here, where 0 means "no limit" for Loki but would send `size: 0` and get you nothing, and `-o graph` would land on the `*querysql.QueryResponse` case in the generic graph codec and tell an Elasticsearch user that graph output isn't supported for SQL datasources.

So it's your call: a redirect placed before `shared.ResolveExpr`, a full dispatch with those two handled, or leave it unrouted. If it's the last one I'll file a single issue covering elasticsearch, tempo, athena and infinity together, since they're all in the same spot. All I'd ask is that you say which one and why, here in the thread, because the guidance does ask for it to be a deliberate call rather than an omission. Not blocking either way.

One more with no line to hang it on: the description is out of date. It still lists `gcx datasources elasticsearch logs` as a shipped command and shows `--size 20`.

Comments below are prefixed **required** (1), **recommended** (2) and **nit** (1). Only the required one needs to happen before merge.

Two things from earlier rounds I'm dropping. `--step` no-opping on `query` is a shared-infrastructure question, not yours: `SharedOpts.Setup` hands `--step` to every datasource query command, postgres and clickhouse included. And `--agg count --field x` putting a field on the count metric, which I'd half-raised: the plugin ignores fields for count, and the client test doesn't actually establish that passing one is wrong.

internal/datasources/elasticsearch/mapping_list.go
```diff
@@ -0,0 +1,106 @@
...
+	cmd.Annotations = map[string]string{
+		agent.AnnotationTokenCost: spec.tokenCost,
+		agent.AnnotationLLMHint:   spec.llmHint,
+	}
+
+	opts.IO.RegisterCustomCodec("table", &mappingListTableCodec{spec: spec})
+	opts.IO.DefaultFormat("table")
+	opts.IO.BindFlags(cmd.Flags())
+	cmd.Flags().StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.elasticsearch is configured)")
+	cmd.Flags().StringVar(&opts.Index, "index", "", "Restrict to this index or index pattern")
```
awsome-o, line 93: **recommended** — an explicit `--index ""` widens the request from one index to all of them, which is the opposite of why we added the flag. `list-indices --index "$IDX"` with `IDX` unset pulls the whole cluster's `_mapping` and can run into the 4MB cap, and nothing tells the caller their scope got ignored. `value != ""` can't tell that apart from an omitted flag, but `cmd.Flags().Changed("index")` can. Rejecting explicit empty in `Validate()` covers it, and omitted keeps meaning all indices like the help says.

internal/datasources/elasticsearch/query.go
```diff
@@ -0,0 +1,162 @@
...
+		Short: "Search documents in an Elasticsearch datasource",
+		Long: `Search documents in an Elasticsearch datasource with a Lucene query.
+
+EXPR is a Lucene query string (e.g. 'app:frontend AND level:error'); omit it to
+match all documents in the time range. The index pattern comes from the
+datasource configuration.
+
+--mode documents (default) returns raw source documents. --mode logs returns
+the same documents newest-first with plugin-internal fields (_source, sort,
+highlight) omitted, matching how Grafana Explore's Logs view reads them.
```
awsome-o: **nit** — `isLogsInternalField` at `internal/query/elasticsearch/types.go:54` drops four fields, not three: `_source`, `sort`, `highlight` and `_type`. This sentence, the comment at `query.go:22`, the generated page under `docs/reference/cli/` and the PR body all say three, so anyone reading them expects a `_type` column and won't get one.

```suggestion
the same documents newest-first with plugin-internal fields (_source, sort,
highlight, _type) omitted, matching how Grafana Explore's Logs view reads them.
```

internal/query/elasticsearch/client.go
ktw4071: Confirmed — and this splits into two independent problems, addressed in two commits.

**`query` / `query --mode logs`**: these never build a date_histogram bucket, so `--time-field` only ever fed the top-level `timeField` key you identified as dead (no field to unmarshal into). It did nothing, while `query.go:43` claimed it controlled range filtering. Fixed in `3f200c3b` — removed the flag entirely rather than keep documenting behavior it never had.

**`metrics`**: this one does have a live per-request field — the date_histogram bucket's own `field` (`addDateHistogramAgg`/`bucketAgg.Field`). Our bug: `AggsQueryModel` defaulted an omitted `--time-field` to a hardcoded `"@timestamp"`, while the actual range filter always comes from the datasource's own configured `timeField` setting — two independently-sourced values that silently disagree whenever a datasource isn't configured with `@timestamp`. Fixed in `1c3b417f` — the bucket now gets `req.TimeField` unmodified (no default), and the flag's own default changed from `"@timestamp"` to `""`, so an omitted flag reaches the plugin empty and the plugin substitutes its configured field into both the filter and the bucket.

**Wire-level diff** (`metrics --since 1h --share-link`, `--time-field` omitted):
```
before (1ef135f9): bucketAggs[1].field = "@timestamp"   (hardcoded default)
after:              bucketAggs[1].field = ""            (plugin substitutes its configured field)
```

**Reproducing the actual symptom.** `elasticsearch9-ds-m` happens to be configured with `timeField: "@timestamp"` — same as our old hardcoded default — so the bug was invisible on this datasource and in every test fixture we have. To show what the hardcoded default does when it *doesn't* match the datasource's real setting, I forced the mismatch directly:
```
$ gcx datasources elasticsearch metrics -d elasticsearch9-ds-m --agg count --time-field "@timestamp" --since 1h -o json
{"series": [{"name": "Count", ...}]}     # 360 buckets, 28800 total docs — real field, real data

$ gcx datasources elasticsearch metrics -d elasticsearch9-ds-m --agg count --time-field "bogus.does.not.exist" --since 1h -o json
{"series": null}                          # exit code 0 — no error, just silently empty
```
That second command is exactly the failure you described: bucket field disagrees with the field the plugin filters on → no error, just a silently empty series. Before this fix, any datasource configured with something other than `@timestamp` would hit this on the ordinary, flagless `gcx datasources elasticsearch metrics` invocation, every time.

Also added `TestAggregations/explicit_time_field_overrides_the_histogram_bucket` to pin that an explicit `--time-field` still overrides the bucket correctly, so the override path can't silently regress.

internal/datasources/elasticsearch/mapping_list.go
ktw4071: Done in `ea99b61d` — `--index ""` now rejected via `cmd.Flags().Changed("index")`, before any config/datasource I/O runs.

Verified against `elasticsearch9-ds-m`:
```
$ gcx datasources elasticsearch list-indices -d elasticsearch9-ds-m
INDEX            FIELDS
grafana-logs     20
grafana-metrics  18
```
That's omitted `--index` — still means "all indices" exactly as documented, both before and after this fix.

```
$ gcx datasources elasticsearch list-indices -d elasticsearch9-ds-m --index ""
before: INDEX            FIELDS            (identical to the omitted case above — silently widened)
        grafana-logs     20
        grafana-metrics  18
after:  {"error":{"summary":"--index must not be empty"}}
```

And a real, narrowing value still works unaffected:
```
$ gcx datasources elasticsearch list-indices -d elasticsearch9-ds-m --index "grafana-logs"
INDEX         FIELDS
grafana-logs  20
```
`list-fields` shares the same builder and inherits the fix — verified that too.

awsome-o (approved):
Approving. I went through the five fixes at `ea99b61d` rather than going off the commit messages, and re-ran the gates locally: build, the full test suite, `golangci-lint` on both changed packages after a cache clean, and `mise run reference` with no drift. CI is green here too.

The time-field fix is the right shape. Passing `req.TimeField` through unmodified and dropping the flag's own default means an omitted flag lands at the plugin empty, so the bucket and the range filter both come from the datasource's configured field and can't drift apart. Taking the flag off `query` instead of documenting the no-op was the right call. And thanks for the forced-mismatch repro with `bogus.does.not.exist` — that's the clearest demonstration of the symptom there is, and it explains why no fixture caught it.

Two things still open, neither of which needs another review round.

The description is out of date. It still lists `gcx datasources elasticsearch logs` as a shipped command, still shows `--size 20`, still describes three query modes with `logs` as its own, and still lists three trimmed fields instead of four. Worth fixing before you merge, since this is the record for a new command surface.

Routing: I withdrew it as a requirement, but `add-datasource/SKILL.md` asks for it to be a deliberate call rather than an omission, so a sentence in that thread saying which of the three you went with would close it out. If it's "leave it unrouted", I'll file the issue covering elasticsearch, tempo, athena and infinity together.

Two cosmetic leftovers, genuinely take-or-leave: the comment at `query.go:21-22` still lists three trimmed fields, and `SearchRequest.TimeField` is now written by nobody, so it's dead along with the top-level `timeField` key it feeds.

Nice work on this one. Four rounds and it got better each time.

PR #974, opened 2026-07-15
feat(datasources): add support for Google Cloud Monitoring datasource

alyssajoyner:
Used the add-datasource skill to add the Google Cloud Monitoring datasource.

**New commands:**
- gcx datasources cloudmonitoring query
- gcx datasources cloudmonitoring list-projects
- gcx datasources cloudmonitoring list-metrics

**Querying**

query is structured (no expression language): a project + metric type, with GCM's native aggregation vocabulary — --reducer (cross-series) and --aligner (per-series), --group-by to split series by label (repeatable), --filter key=value (repeatable), and --alignment-period (auto by default):

gcx datasources cloudmonitoring query -d UID --project my-project \
  --metric compute.googleapis.com/instance/cpu/utilization \
  --reducer REDUCE_MEAN --group-by resource.label.instance_name --since 1h

Output: table/wide/json/yaml/graph. list-projects and list-metrics discover flag values; list-metrics --service compute.googleapis.com narrows server-side (recommended — unfiltered listings page through every descriptor in the project and can be slow).

**Notes:**
- Kind is cloudmonitoring; NormalizeKind maps the plugin's pre-rebrand stackdriver ID to it.
- Reuses the shared grafanaquery transport; time-series parsing/graph adapter follow the CloudWatch/Azure Monitor pattern.
- GCP error envelopes are simplified on both error paths — a bad metric shows NOT_FOUND: Cannot find metric(s)… instead of a raw JSON dump.
- Wire-shape pins: filter triplets joined with "AND", the resource proxy's embedded GCP API paths, and the plugin's query failed:  prefix on error envelopes (all discovered by live probing).
- Scope: timeSeriesList queries only — the plugin's MQL (deprecated), PromQL (use a Prometheus datasource), and SLO query types are excluded and additive later.
- Merge note: trivial adjacent-line overlaps with #971 (codecs.go) and #969/#970 (resolve.go) — whichever lands last resolves in seconds.


dafydd-t:

dafydd-t:
reviewed this - looks solid. two minor notes, neither blocking:

- `--filter` is a `stringToString` map, so with multiple filters the triplet order in the request body is nondeterministic across runs. AND is commutative so it's harmless, but it also means equality-only and one value per key - fine for the initial scope, just something to keep in mind if comparator support gets asked for
- in `parseFrame`, a non-nil value that fails `toFloat64` silently drops its timestamp by truncating the slice - correct, but subtle enough that a short comment there wouldn't hurt

internal/datasources/query/resolve.go
```diff
@@ -191,6 +191,8 @@ func NormalizeKind(pluginID string) string {
 		return "infinity"
 	case "synthetic-monitoring-datasource":
 		return "synthetic-monitoring"
+	case "stackdriver":
```
awsome-o, line 194: This teaches `NormalizeKind` about the stackdriver plugin ID, but `cmd/gcx/datasources/query.go` gains neither a dispatch case nor a redirect, so both plausible generic invocations dead-end without ever naming the command the caller actually wants:

```
gcx datasources query <uid>            ->  "expression is required..."
gcx datasources query <uid> '<expr>'   ->  'datasource type "cloudmonitoring" is not supported'
```

A GCM query is structured (project, metric type, filters, group-bys, aligner), so no single expression string represents it. That makes a CloudWatch-style redirect the right answer here rather than generic dispatch. It needs to sit **before** `shared.ResolveExpr` so the argument-less call gets the redirect too, not the expression error. `cmd/gcx/datasources/query.go:72` has the pattern.

Worth a small routing regression covering both invocations.

internal/datasources/cloudmonitoring/query.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (opts *queryOpts) Validate() error {
+	if err := opts.IO.Validate(); err != nil {
+		return err
+	}
+	if err := opts.ValidateTimeRange(); err != nil {
+		return err
+	}
+	if opts.Project == "" {
```
awsome-o, line 53: `--project` and `--metric` are covered for exact-empty here, good.

Still unchecked: `--reducer` and `--aligner` against the enums their own help text lists, `--alignment-period` as a duration, and the `--group-by` entries. Whitespace-only values pass all of these too, since the checks are `== ""` comparisons.

Today these get past `Validate()` and reach config loading and datasource resolution before anything fails, so the caller gets a config or backend error instead of a clear local one. `cloudwatch/query.go:66` rejects the analogous shapes client-side and has tests pinning it.

internal/datasources/cloudmonitoring/query.go
```diff
@@ -0,0 +1,182 @@
...
+	opts.SetupTimeFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.cloudmonitoring is configured)")
+	flags.StringVar(&opts.Project, "project", "", "GCP project ID (required)")
+	flags.StringVar(&opts.Metric, "metric", "", "Metric type, e.g. compute.googleapis.com/instance/cpu/utilization (required)")
+	flags.StringVar(&opts.Reducer, "reducer", "REDUCE_NONE", "Cross-series reducer: REDUCE_NONE, REDUCE_MEAN, REDUCE_SUM, REDUCE_MIN, REDUCE_MAX, REDUCE_COUNT, ...")
+	flags.StringVar(&opts.Aligner, "aligner", "ALIGN_MEAN", "Per-series aligner: ALIGN_MEAN, ALIGN_SUM, ALIGN_MIN, ALIGN_MAX, ALIGN_RATE, ALIGN_DELTA, ...")
+	flags.StringVar(&opts.AlignmentPeriod, "alignment-period", "", `Alignment period, e.g. +60s (default: auto-fit the time range)`)
+	flags.StringArrayVar(&opts.GroupBys, "group-by", nil, "Label to split series by, e.g. resource.label.instance_name (repeatable)")
+	flags.StringToStringVar(&opts.Filters, "filter", nil, "Label filter key=value (repeatable, e.g. --filter resource.label.zone=us-east1-b)")
```
awsome-o, line 43: Two things here.

On validation, I checked what pflag actually does rather than assuming, and it's narrower than I first thought. `--filter foo` is already rejected by pflag with "must be formatted as key=value", and so is `--filter ""`. But `--filter =value` parses to an empty key and `--filter key=` parses to an empty value, and both reach the request unchecked. Those two are worth rejecting locally.

On the help text, could it say that filters are equality-only, AND-combined across repeats, and case-sensitive? (https://docs.cloud.google.com/monitoring/api/v3/filters) The concrete cost is that something like `zone=~us-east.*` looks like it should work, becomes a literal equality match, and returns `No data` with exit 0, so the caller reasonably concludes they have nothing in that zone.

Non-blocking: the 8 MB descriptor cap surfaces as a bare read error without suggesting `--service`, which is the documented narrowing flag; and when everything comes back empty you print `No data` with no pointers, while the CloudWatch sibling emits a hint in the same situation.

awsome-o (changes requested):
The wiring here is in good shape: output classes, token costs and hints, CODEOWNERS, package map and generated reference are all correct and drift-free, and `internal/query/cloudmonitoring/client_test.go` covers request construction properly against a capture server.

Three things before merge: the generic query path has nowhere to send people, several structured inputs aren't validated, and the filter semantics need documenting.

On tests: command-level regressions for the redirect and for each validation that fires before loader I/O. The client tests already cover the wire shape and don't need duplicating, and there's no need for a flag-by-output matrix.

One I looked at and am **not** asking you to change: `parseFrame` drops frames whose value column isn't numeric and reads only the first numeric column. As far as I can tell the current plugin emits one frame per time series with exactly a time and a numeric value field, so nothing reaches those branches in practice. A short comment recording that assumption would be worth having, but it isn't a blocker.

Ours, not yours: validation errors exiting 1 rather than 2 is getting a shared helper in #1138, so please don't hand-roll a wrapper.

internal/datasources/query/resolve.go
ktw4071: Fixed in e4e38984.

Registered a `structuredQueryRedirect` entry for `cloudmonitoring` in `cmd/gcx/datasources/query_routes.go`, same pattern as the existing `azuremonitor`/`cloudwatch` entries. The ordering concern (redirect must beat `shared.ResolveExpr`) is already satisfied by the current `queryRoutes` architecture -- this was just the missing table entry, no logic reordering needed.

Added `TestGenericQueryCloudMonitoringShortCircuit`, mirroring the existing CloudWatch short-circuit test, covering both invocation shapes against a datasource reporting the raw `stackdriver` plugin ID.

Verified against a live Google Cloud Monitoring datasource (`gcm-ds-m` is a GCM-type datasource UID in our dev Grafana stack, passed as the bare `<uid>` argument to the generic `gcx datasources query` command -- not the typed `cloudmonitoring` subcommand, since that generic-vs-redirect path is exactly what this fix touches):

```
$ gcx datasources query gcm-ds-m
Error: Expression is required...   (before)

$ gcx datasources query gcm-ds-m "some-expr"
Error: datasource type "cloudmonitoring" is not supported (supported: azuremonitor, clickhouse, cloudwatch, influxdb, loki, postgres, prometheus, pyroscope)   (before)

$ gcx datasources query gcm-ds-m
Error: Google Cloud Monitoring queries are structured (project, metric type, reducer, aligner, filters, group-bys); the generic `gcx datasources query <uid> <expr>` form can't carry them — use `gcx datasources cloudmonitoring query --project ... --metric ...` instead   (after)

$ gcx datasources query gcm-ds-m "some-expr"
Error: same redirect   (after — identical for both invocation shapes)
```

internal/datasources/cloudmonitoring/query.go
ktw4071: Fixed in 4df9bc9f, with error-message wording improved in c45444ca based on feedback in review discussion -- the original messages said a value was invalid but not what valid looks like, so I added the actual accepted enum lists (same pattern as azuremonitor's `--aggregation`).

For `--reducer`/`--aligner`, an invalid value was always going to fail -- the fix is that it now fails locally and immediately, with the valid options listed, instead of after a wasted round-trip to Google's API:

```
$ gcx datasources cloudmonitoring query -d UID ... --reducer REDUCE_AVERAGE   # before
-> reaches the backend: INVALID_ARGUMENT: Invalid value at aggregation.cross_series_reducer, "REDUCE_AVERAGE"

$ gcx datasources cloudmonitoring query -d UID ... --reducer REDUCE_AVERAGE   # after
-> --reducer "REDUCE_AVERAGE" is not a valid cross-series reducer, must be one of:
   REDUCE_NONE, REDUCE_MEAN, REDUCE_MIN, REDUCE_MAX, REDUCE_SUM, REDUCE_STDDEV, REDUCE_COUNT,
   REDUCE_COUNT_TRUE, REDUCE_COUNT_FALSE, REDUCE_FRACTION_TRUE, REDUCE_PERCENTILE_99,
   REDUCE_PERCENTILE_95, REDUCE_PERCENTILE_50, REDUCE_PERCENTILE_05

$ gcx datasources cloudmonitoring query -d UID ... --aligner ALIGN_AVERAGE   # after
-> --aligner "ALIGN_AVERAGE" is not a valid per-series aligner, must be one of:
   ALIGN_NONE, ALIGN_DELTA, ALIGN_RATE, ALIGN_INTERPOLATE, ALIGN_NEXT_OLDER, ALIGN_MIN, ALIGN_MAX,
   ALIGN_MEAN, ALIGN_COUNT, ALIGN_SUM, ALIGN_STDDEV, ALIGN_COUNT_TRUE, ALIGN_COUNT_FALSE,
   ALIGN_FRACTION_TRUE, ALIGN_PERCENTILE_99, ALIGN_PERCENTILE_95, ALIGN_PERCENTILE_50,
   ALIGN_PERCENTILE_05, ALIGN_PERCENT_CHANGE
```

For `--alignment-period`/`--group-by`, the difference is more serious than error-timing -- before the fix, malformed input was not rejected at all, it was silently ignored and the query still ran and returned real data:

```
$ gcx datasources cloudmonitoring query -d UID ... --alignment-period 60s   # before (CloudWatch-style mistake)
-> succeeds silently: 2 real frames, 655/662 points -- the malformed period was just dropped, no error, no data change

$ gcx datasources cloudmonitoring query -d UID ... --group-by "   "   # before
-> succeeds silently: same 2 frames, 655/662 points, ungrouped -- the whitespace entry was just dropped

$ gcx datasources cloudmonitoring query -d UID ... --alignment-period 60s   # after
-> --alignment-period "60s" must be empty (auto-fit) or match "+<seconds>s", e.g. +60s

$ gcx datasources cloudmonitoring query -d UID ... --group-by "   "   # after
-> --group-by entries must not be empty

$ gcx datasources cloudmonitoring query -d UID ... --aligner ALIGN_NONE --alignment-period +300s   # after, valid input
-> succeeds, 2 real frames -- unchanged for correct input
```

internal/datasources/cloudmonitoring/query.go
ktw4071: Fixed in 6711d745.

Rejects an empty key or empty value in `--filter` locally. Also updated the flag help and the command's `Long` description to state filters are exact-match, case-sensitive, and AND-combined across repeats -- no regex/wildcard.

**Empty key/value:**
```
$ gcx datasources cloudmonitoring query -d UID ... --filter "=us-east1-b"           # before
-> reaches the backend: INVALID_ARGUMENT: ... syntax error at line 1, column 63, token '='

$ gcx datasources cloudmonitoring query -d UID ... --filter "resource.label.zone="  # before
-> reaches the backend: INVALID_ARGUMENT: The supplied filter does not specify a valid
   combination of metric and monitored resource descriptors.

$ gcx datasources cloudmonitoring query -d UID ... --filter "foo"                   # before/after, unchanged
-> Invalid argument "foo" for "--filter" flag: foo must be formatted as key=value
   (pflag's own check -- already correct, shown for contrast)

$ gcx datasources cloudmonitoring query -d UID ... --filter "=us-east1-b"           # after
-> --filter =us-east1-b: key must not be empty

$ gcx datasources cloudmonitoring query -d UID ... --filter "resource.label.zone="  # after
-> --filter resource.label.zone=: value must not be empty
```

**The regex/wildcard footgun** -- reproduced with a label this metric actually has (`resource.label.location`, value `global`), and traced precisely rather than just asserted: pflag's key=value parser splits on the first `=` only, so `resource.label.location=~glob.*` parses to `key="resource.label.location"`, `value="~glob.*"`. `BuildQueryModel` (`internal/query/cloudmonitoring/client.go:66`) always hardcodes the operator to `=` -- there is no branch anywhere that recognizes `=~` and swaps in a regex operator. So the filter clause sent to Google is a literal equality check against the 8-character string `~glob.*`, not a regex evaluation of `glob.*`. There is no regex support here at all, on either side of the wire:

```
$ gcx datasources cloudmonitoring query -d UID ... --filter "resource.label.location=global" -o json
-> succeeds, 2 real frames (zonal_dns_ready: 655 points, zonal_dns_risky: 662 points)

$ gcx datasources cloudmonitoring query -d UID ... --filter "resource.label.location=~glob.*" -o table
-> No data
   (exit code: 0)
```
Same underlying data, only the filter value changed from a real equality match to something that looks like a regex -- and it silently returns nothing rather than erroring, exactly as you described.

**Both non-blocking items confirmed real, left out of scope for this commit:**
- The 8 MB cap message (`internal/httputils/read.go:21`, shared across many clients) reads `"response body exceeds 8 MB limit; try narrowing your query or adding filters"` -- no mention of `--service`. Could not trigger the actual cap against our test project (`list-metrics` unfiltered returned 8,987 descriptors and still fit under 8MB), but the generic wording is confirmed via the shared helper's own test (`internal/httputils/read_test.go:35-44`).
- The bare `"No data"` output above, with no stderr hint, is the same gap you flagged -- confirmed live in the wildcard-footgun run right above.

Happy to take either of these as a follow-up if you'd like them tracked here rather than filed separately.

internal/datasources/cloudmonitoring/query.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (opts *queryOpts) Validate() error {
+	if err := opts.IO.Validate(); err != nil {
+		return err
+	}
+	if err := opts.ValidateTimeRange(); err != nil {
+		return err
+	}
+	if opts.Project == "" {
```
awsome-o: The TrimSpace fix landed for --group-by, but --project and --metric are still bare `== ""` checks, so `--project "$SOMETHING_UNSET "` or `--metric " "` still gets through Validate, loads config, resolves the datasource and POSTs before Google comes back with INVALID_ARGUMENT. That's the round-trip the original comment was about, and these two are the required flags.

What convinced me it's worth one more push: explore.go already TrimSpaces both of these and returns an empty URL for whitespace-only, so with --share-link the Explore path calls the input invalid while the query path calls it fine. Same two calls needed in list_metrics.go for --project, and --service is the worse one over there since `--service "   "` builds `metric.type = starts_with("   ")` and hands back an empty listing with exit 0.

internal/datasources/cloudmonitoring/query.go
```diff
@@ -0,0 +1,288 @@
...
+	opts.IO.BindFlags(flags)
+	opts.SetupTimeFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.cloudmonitoring is configured)")
+	flags.StringVar(&opts.Project, "project", "", "GCP project ID (required)")
+	flags.StringVar(&opts.Metric, "metric", "", "Metric type, e.g. compute.googleapis.com/instance/cpu/utilization (required)")
+	flags.StringVar(&opts.Reducer, "reducer", "REDUCE_NONE", "Cross-series reducer: REDUCE_NONE, REDUCE_MEAN, REDUCE_SUM, REDUCE_MIN, REDUCE_MAX, REDUCE_COUNT, ...")
+	flags.StringVar(&opts.Aligner, "aligner", "ALIGN_MEAN", "Per-series aligner: ALIGN_MEAN, ALIGN_SUM, ALIGN_MIN, ALIGN_MAX, ALIGN_RATE, ALIGN_DELTA, ...")
+	flags.StringVar(&opts.AlignmentPeriod, "alignment-period", "", `Alignment period, e.g. +60s (default: auto-fit the time range)`)
+	flags.StringArrayVar(&opts.GroupBys, "group-by", nil, "Label to split series by, e.g. resource.label.instance_name (repeatable)")
```
awsome-o: --group-by does nothing unless a reducer is set, and neither this help string nor the Long says so. GCP's Aggregation reference is explicit about groupByFields: if cross_series_reducer isn't defined, the field is ignored. Your own probe in the thread above shows it, same 2 frames, 655 and 662 points, ungrouped, no error.

Since the default is REDUCE_NONE, anyone who passes only --group-by gets well-formed data back and concludes the label doesn't vary. Same silent-wrong-answer shape as the `=~` filter footgun we documented last round. Every Example in the command pairs the two flags, so you clearly know the rule, it's just nowhere the user can see it. Either reject `len(GroupBys) > 0 && Reducer == "REDUCE_NONE"` in Validate, or put it in the flag help and the Long.

internal/query/cloudmonitoring/types.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// MetricDescriptor represents a Cloud Monitoring metric descriptor.
+type MetricDescriptor struct {
+	Type        string `json:"type"`
+	DisplayName string `json:"displayName"`
+	MetricKind  string `json:"metricKind"`
+	ValueType   string `json:"valueType"`
+	Unit        string `json:"unit,omitempty"`
+	Service     string `json:"service,omitempty"`
```
awsome-o: Service is never populated and never read. Grepping `.Service` across both cloudmonitoring packages only finds listMetricsOpts.Service, the flag; FormatMetricDescriptors renders Type/Kind/Value/Unit; and the GCP metricDescriptors response has no `service` key to fill it from anyway.

Because it's omitempty it silently never appears, which is a bit worse than appearing empty, since it's a declared key in the JSON contract that someone could write a jq expression against. I'd just drop the field.

internal/query/cloudmonitoring/types.go
```diff
@@ -0,0 +1,206 @@
...
+		if frame, ok := parseFrame(df); ok {
+			resp.Frames = append(resp.Frames, frame)
+		}
+	}
+	return resp, nil
+}
+
+// parseFrame extracts the first time/value column pair with its labels and
+// datasource-provided display name.
+func parseFrame(df dataframe.Frame) (Frame, bool) {
```
awsome-o, line 92: Small leftover from last round: still no note that parseFrame takes only the first numeric column and drops frames that don't have one. The doc comment describes what it extracts, and the inline comment down at 133 is about the timestamp/value length pairing, which is a different point. One sentence naming the frame shape you're assuming closes it.

internal/datasources/cloudmonitoring/list_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to list metrics: %w", err)
+			}
+
+			return opts.IO.Encode(cmd.OutOrStdout(), descriptors)
+		},
+	}
+
+	cmd.Annotations = map[string]string{
+		agent.AnnotationTokenCost: "medium",
```
awsome-o: You measured 8987 descriptors unfiltered, which reads more like `large` than `medium` to me. The llm_hint already teaches --service as the narrowing flag so the guidance itself is right, and cloudwatch and azure both claim `small` for their list-metrics, so this is already the most honest one in the family. One token, take it or leave it.

awsome-o (dismissed):
Approving. I re-checked the three threads from last round against 407053ef and they're all genuinely fixed, and the tests are the right shape: they drive Execute against a zero-value loader, so a validation regression fails the test instead of quietly turning into a network call.

The redirect works. TestGenericQueryCloudMonitoringShortCircuit covers both invocation shapes, and the ordering claim holds, datasource type resolution really does run before ResolveExpr. The --filter empty key/value fix is right and I re-derived the pflag behaviour you described in the thread: bare `foo` and `""` are pflag's to reject, `=v` and `k=` are yours. Reducer, aligner and alignment-period validation are all there with tests.

Two things I'd push once more, neither blocking. The whitespace-only case for --project and --metric is the one slice of my earlier comment that didn't get done, and --group-by silently doing nothing under the default reducer isn't written down anywhere a user would see it. Both inline.

I also went looking for over-engineering and mostly came up empty, which is meant as a compliment. The IsZero/Since time fallback is byte-identical to azuremonitor and cloudwatch, the `query failed: %w` wrap is what all ten sibling query commands do, and the result.Error branch is live because a query-level failure arrives inside a 200. I nearly flagged the hardcoded intervalMs of 60000 since azuremonitor derives it from the range, but cloud-monitoring-auto computes the alignment period from the query duration rather than intervalMs, so it doesn't change the default path. The only real one is a dead struct field.

On the two you raised yourself: the 8MB cap message not naming --service lives in shared httputils, so that's a cross-cutting change, and the bare `No data` with no stderr hint where cloudwatch emits one. Both fine to leave here, worth follow-up issues instead of holding this up.

internal/datasources/cloudmonitoring/query.go
ktw4071: Both gaps fixed in `ce45539d`.

`--project`/`--metric` now reject a whitespace-only value locally (`strings.TrimSpace(...) == ""`) instead of round-tripping to Google. Same fix applied to `list-metrics`' `--project`. For `--service` (optional, so empty must still mean "no filter" when *omitted*), added the `cmd.Flags().Changed`-gated explicit-empty rejection used elsewhere this round for postgres/mysql — an explicitly-passed empty or whitespace value now errors instead of silently building `starts_with("   ")`.

Live-verified against `gcm-ds-m`:
```
--project "   " --metric ...  before: INVALID_ARGUMENT: Invalid project ID.
                               after:  --project is required
--metric "   "                before: NOT_FOUND: Cannot find metric(s) that match type = "   "
                               after:  --metric is required
list-metrics --service "   "  before: exit 0, 0 descriptors (silently empty)
                               after:  --service must not be empty
```

internal/datasources/cloudmonitoring/query.go
ktw4071: Went with reject rather than document-only, since you framed it as the same silent-wrong-answer shape as the `=~` footgun — `Validate()` now rejects `len(GroupBys) > 0 && Reducer == "REDUCE_NONE"`, and the flag help/Long also spell out the requirement as defense in depth.

Live-verified against `gcm-ds-m` — proved it was a true no-op, not just undocumented, by comparing against the exact same query with `--group-by` omitted entirely:
```
--group-by resource.label.instance_id, no --reducer, before this fix:
  3 frames — byte-identical (same instance labels, same order) to the same query with --group-by omitted
after this fix:
  --group-by has no effect while --reducer is REDUCE_NONE (the default); pass a cross-series reducer such as REDUCE_MEAN or REDUCE_SUM so the split actually applies
```

internal/query/cloudmonitoring/types.go
ktw4071: Dropped in `8b664165`, per your grep — confirmed nothing populates or reads it.

internal/query/cloudmonitoring/types.go
ktw4071: Added in `8b664165`: "It assumes exactly one numeric column per frame — the shape the Cloud Monitoring plugin emits for these queries — and drops any frame that has none."

internal/datasources/cloudmonitoring/list_metrics.go
ktw4071: Changed to `large` in `ce45539d`. Reproduced your 8987-descriptor count exactly on our test project, so `large` it is.

awsome-o (approved):

PR #976, opened 2026-07-15
fix(profiles): include trace IDs in exemplars

simonswine:
## Summary

- decode and propagate Pyroscope exemplar `traceId` values
- include trace IDs in structured profile and span exemplar output
- show the span table `TRACE ID` column only when at least one value is present, preserving compatibility with older servers
- update generated CLI reference documentation

Ports the profilecli behavior from https://github.com/grafana/pyroscope/pull/5353.

## Testing

- `mise run lint`
- `go test ./...`
- `GCX_AGENT_MODE=false mise run reference`
- `mise run build`
- `mise run docs` not run locally because `mkdocs` is unavailable; CI will perform the docs build


aleks-p (approved):
LGTM

alsoba13 (approved):
lgtm

dafydd-t (approved):

PR #979, opened 2026-07-15
feat(profiles): add span and trace selectors

simonswine:
## Summary

- add repeatable `--span-id` and `--trace-id` flags to Pyroscope profile queries
- route span-filtered flamegraph queries through `SelectMergeSpanProfile`
- forward trace selectors through both `SelectMergeStacktraces` JSON and `SelectMergeProfile` protobuf requests
- validate span/trace ID formats and profilecli-compatible selector exclusions
- update command examples, architecture docs, and generated CLI references

`--span-selector` is intentionally rejected with `-o pprof`: the span RPC returns a Pyroscope tree rather than a pprof profile, and gcx does not currently carry the server-side tree conversion code. Trace selectors support pprof output.

## Testing

- `GCX_AGENT_MODE=false mise run all` passes lint, all Go tests, linter rules, skill validation, reference generation, and build
- local MkDocs rendering was unavailable because `mkdocs` is not installed; CI will perform the documentation build

### DIFF
--- a/docs/architecture/cli-layer.md
+++ b/docs/architecture/cli-layer.md
@@ -82,7 +82,7 @@ gcx (root)
 │   └── adaptive             Adaptive Traces (policies, recommendations)
 │
 ├── profiles                 [internal/providers/profiles/provider.go] (registered via providers.Register)
-│   ├── query                [DATASOURCE_UID] EXPR --profile-type TYPE [--from] [--to] [--since] [--max-nodes] [--profile-id UUID]... [--stacktrace-selector FN]... [-o]
+│   ├── query                [DATASOURCE_UID] EXPR --profile-type TYPE [--from] [--to] [--since] [--max-nodes] [--profile-id UUID]... [--span-selector ID]... [--trace-id ID]... [--stacktrace-selector FN]... [-o]
 │   ├── labels               [--datasource/-d UID] [--label/-l NAME]
 │   ├── profile-types        [--datasource/-d UID]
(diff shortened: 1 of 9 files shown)


docs/architecture/cli-layer.md
```diff
@@ -82,7 +82,7 @@ gcx (root)
 │   └── adaptive             Adaptive Traces (policies, recommendations)
 │
 ├── profiles                 [internal/providers/profiles/provider.go] (registered via providers.Register)
-│   ├── query                [DATASOURCE_UID] EXPR --profile-type TYPE [--from] [--to] [--since] [--max-nodes] [--profile-id UUID]... [--stacktrace-selector FN]... [-o]
+│   ├── query                [DATASOURCE_UID] EXPR --profile-type TYPE [--from] [--to] [--since] [--max-nodes] [--profile-id UUID]... [--span-selector ID]... [--trace-id ID]... [--stacktrace-selector FN]... [-o]
```
aleks-p: Can we use `--span-id` for consistency? We call it `SpanId` later on anyway. 

They are all called `-selector` in the backend, but I think keeping them short here makes sense. We can keep `--stacktrace-selector` as is, since this one is also more complex than a single id.

simonswine: I was tempted to do that, but then I opted for profilecli aligment 😂 

Let's rename it!

simonswine: Done in ffa8dbd

dafydd-t (approved):
LGTM from the gcx side 👍

aleks-p (approved):

PR #981, opened 2026-07-15
fix(alert): make GMA alert rules modifiable via the resources tier

wardbekker:
## Problem

A customer reported they **cannot modify alerts with gcx**. Root cause analysis found two defects:

1. The alert provider registered read-only typed adapters (`ListFn`/`GetFn` only) for `AlertRule`/`AlertRuleGroup` under the gcx-synthetic `alerting.ext.grafana.app` group. The backing API is the Prometheus-compatible **status** endpoint (`/api/prometheus/grafana/api/v1/rules`), which cannot support writes even in principle — so any push/delete routed to these adapters failed with `ErrUnsupported`.
2. Bare `gcx resources push alertrules` resolved to that read-only adapter instead of the fully writable native `rules.alerting.grafana.app/v0alpha1` group, because `RegistryIndex.filterCandidates` committed to the **first** candidate group when no group was specified (acknowledged in an in-code TODO). Reproduced live: bare `alertrules` returned `alerting.ext` status-shaped objects that cannot be pushed back.

## Fix

- **Remove the `alerting.ext` adapter registrations** (`TypedRegistrations() → nil`, dashboards-provider precedent). Alert rule CRUD is served by the K8s dynamic tier via `rules.alerting.grafana.app`. The `gcx alert rules|groups` commands are unaffected — they use the raw status client, not the adapter.
- **Fix first-group-wins resolution**: when no group is given, pick the first candidate group that actually serves the requested (or preferred) version; deletes the TODO in `registry_index.go`.
- **Help text** on `gcx alert rules|groups` now says they inspect state/health and points modification at `gcx resources pull/push alertrules`.
- **Docs**: migration-gap-analysis corrected (contact-points/mute-timings/policies/templates exist; rule writes documented), architecture docs no longer cite the alert adapter, gcx skill gains the pull → edit → push workflow.

## Compliance notes

- Restores CONSTITUTION compliance: a provider-only status surface must not mimic adapter CRUD verbs.
- `gcx alert rules list/get` keep their `list/get` verbs — renaming to `show/describe` would be a breaking CLI change requiring separate approval.
- Alerting deliberately remains **excluded** from the server-dry-run allowlist (unchanged behavior; guard verified still firing).

## Verification (live stack)

- Discovery: `rules.alerting.grafana.app/v0alpha1` serves `alertrules` with create/delete/get/list/patch/update (no `alertrulegroups` resource exists — agent metadata untouched).
- Bare `gcx resources get alertrules` now resolves **only** to the native group.
[...]

PR #982, opened 2026-07-15
feat(alert): add ruler subtree for datasource-managed rule writes

wardbekker:
## Problem

Part 2 of the "customers cannot modify alerts with gcx" fix (part 1: #981, GMA rules). gcx had **zero write support for datasource-managed (Mimir/Loki ruler) rules** — no ruler config API client existed anywhere; only rule *state* was readable via `alert instances list --datasource`. The old CLI's `recording-rules prometheus/loki` commands were never ported (known gap in the migration analysis).

## What's new

A `gcx alert ruler` subtree (mimirtool mental model, `--datasource <uid>` required on every command):

```
gcx alert ruler namespaces list   --datasource <uid>
gcx alert ruler namespaces delete NAMESPACE --datasource <uid> [--force]
gcx alert ruler groups list       --datasource <uid> [--namespace NS]
gcx alert ruler groups get        NAMESPACE GROUP --datasource <uid>     # YAML, round-trips into upsert
gcx alert ruler groups upsert     NAMESPACE -f FILE --datasource <uid> [--dry-run]
gcx alert ruler groups delete     NAMESPACE GROUP --datasource <uid> [--force]
```

### Implementation notes

- **`RulerClient`** against Grafana's per-datasource proxy `/api/ruler/{dsUID}/api/v1/rules`. Responses are YAML (Lotex contract), but **rule-group POST bodies must be JSON** — Grafana's handler binds the incoming body as JSON before re-serializing toward the ruler; a YAML body gets 400 "bad request data". Discovered and verified in live testing against both backends.
- **Subtype resolution** from the datasource plugin type via the existing `query.GetDatasourceType`/`NormalizeKind` (prometheus-flavored → `?subtype=mimir`, loki → none); non-ruler types get a usage error listing supported kinds.
- **Empty-tenant quirk**: the Loki ruler answers 404 "no rule groups found" when the tenant has no rules; the list path maps that to an empty result instead of an error (gets/deletes still surface 404 as not-found).
- **`upsert -f`** accepts a standard Prometheus rules file (`groups:` list — mimirtool-compatible) or a bare single group. Client-side validation before any HTTP: exactly one of record/alert, durations, recording-rule constraints, PromQL parse via the already-vendored `promql/parser` (skipped for LogQL). `--dry-run` stops after validation.
- **Safety**: deletes prompt unless `--force` (agent mode requires `--force`); upsert is idempotent create-or-update; partial upsert failures enumerate per group and exit non-zero.
- Shared `doBody` transport extracted so the provisioning (JSON) and ruler (YAML) clients use one HTTP path; 404 error headlines show the logical API path rather than internal proxy prefixes.

## Verification (live stack, both backends)

Two independent deep-test passes (Mimir datasource + Loki datasource), full matrices green:

- Empty-state listing (incl. the Loki 404-on-empty mapping), `-o json` variants.
- Upsert from `groups:` rules file and bare group; read-back with correct namespace/interval/rule counts; get→edit→upsert round-trip with **upsert semantics confirmed** (replaced, not duplicated); LogQL expr round-trips byte-identical.
- Recording rule wired end-to-end: `metrics query` returned samples from the applied rule within minutes.
- Validation rejects invalid PromQL/structure/durations client-side with zero HTTP mutations; PromQL parsing correctly skipped for Loki.
- Deletes (group + namespace), not-found error paths, missing `--datasource`, non-ruler datasource (tempo) rejection, non-TTY delete without `--force` does not delete.
- All scratch namespaces removed; both tenants verified empty afterwards.

`GCX_AGENT_MODE=false mise run all` green (lint, race tests, build, docs); skillsdrift and command-annotation consistency tests pass.

## Compliance

- CONSTITUTION grammar: nested sub-area (`alert ruler`) with precedent (`irm oncall`, `metrics adaptive`); CRUD verbs only. The mutation verb was originally `apply` and is now `upsert`, per the naming conventions guide (#1023): `upsert` is the direct create-or-update verb (precedent: `alert templates upsert`, `kg entities upsert`; the templates `apply` alias was removed in #1010 for the same reason).
- No adapter registration: ruler groups are not K8s-tier resources.
- Docs: reference pages generated; migration-gap-analysis row for `recording-rules` updated; gcx skill gains the ruler mapping.


wardbekker:
## Shepherd Status

### CI Checks
All required checks passing on `44ac3ad0` (rebased onto latest main; the conflict in `internal/agent/command_skills.go` was the agento11y rename racing the new `gcx alert ruler` override entry).

### Fixes Applied (`9e1dce07`)
- Error formatting: dropped the colon from the no-ruler-API error message.
- `-d` shorthand for `--datasource` on every ruler command; `-n` for the `groups list` namespace filter; `--datasource` help reworded to "UID of the Mimir or Loki datasource used as ruler (required)".
- Grammar consistency: `apply` now takes the target namespace as a positional (`gcx alert ruler groups apply NAMESPACE -f FILE`), matching `get`/`delete` — per the CONSTITUTION rule that positionals are the subject and flags are modifiers. The `groups list --namespace` flag stays a flag since it is a filter there.
- Consistent "datasource-managed (Mimir/Loki ruler)" wording on the `ruler`, `namespaces`, and `groups` Shorts; the `ruler` long help now leads with the Mimir/Loki-only requirement and drops the Grafana-managed version-compat digression.
- Deletes resolve the datasource before prompting, so an unsupported datasource type fails before the confirmation.
- `validate()` → `Validate()` was already fixed in an earlier commit.
- CLI reference regenerated.

### Pending
- Re-review requested from @dafydd-t.

---
*Updated by shepherd-pr*

dafydd-t:
`rules.alerting.grafana.app/v0alpha1` is enabled by default.

This has been true since v13.0.0. However, we claim gcx is compatible with 12+: https://github.com/grafana/gcx#compatibility

So the quickest fix, is to annotate this command somehow in its help docs, and document the fact that its only compatible with G13+

More generally, we could add a mehcanism for declaring requirements (some combination of min/max version, feature toggles) on each route, so we can show the user what state their grafana stack needs to be in to use that given command. I think this is one for later.

wardbekker:
Addressed in 7f2ec850 (quick fix, as suggested):

- `gcx alert ruler --help` now spells out that Grafana-managed rule *writes* go via `gcx resources pull/push alertrules` and require **Grafana 13+** (where `rules.alerting.grafana.app` is enabled by default; on 12 it needs an explicit feature toggle). Reference docs regenerated.
- README compatibility matrix: the `alert` row now carries a ‡ footnote scoping the 12+ claim — reads and datasource-managed ruler writes work on 12+, K8s-tier Grafana-managed writes need 13+.
- `docs/reference/migration-gap-analysis.md` row annotated likewise.

The general per-route requirements mechanism (min/max version, feature toggles, surfaced in help/preflight) is tracked as a follow-up in #989.

wardbekker:
Follow-up in e573e88f: restructured the README compatibility section around explicit version tiers instead of a blanket "12+" —

| Grafana | Support level |
|---|---|
| Cloud | Full support |
| 13+ (OSS/Enterprise) | Full support of self-hosted features (all app-platform API groups on by default) |
| 12.x | Most features supported; app-platform groups not yet default-enabled in 12 need feature toggles (footnoted per route) |
| < 12 | Unsupported (exit code 6) |

The feature matrix keeps per-deployment columns (OSS/Enterprise/Cloud/BYOC) with the version claims moved out of the column headers, so future default-enablement flips during the app-platform migration only touch the tier table / footnotes. Per-command requirement declarations remain tracked in #989.

wardbekker:
Toggle named in 46d634ad. Verified against the Grafana source rather than memory:

- **12.x** (`v12.4.5`): the rules app is gated on `featuremgmt.FlagKubernetesAlertingRules` (`pkg/registry/apps/apps.go`), defined in `featuremgmt/registry.go` as `kubernetesAlertingRules` — stage *experimental*, default `false`, `RequiresRestart: true`. Unified Alerting must also be enabled (default).
- **main / 13+**: the gate and the toggle are gone from the registry entirely — the app registers unconditionally, matching enabled-by-default since 13.0.0.

The README footnote now carries the `[feature_toggles]` ini snippet, and the ruler command help + migration-gap row name the toggle explicitly.

docs/reference/cli/gcx_alert_ruler_groups_apply.md
```diff
@@ -0,0 +1,40 @@
...
+bare rule group. Applying a group replaces the group with the same name.
+
+```
+gcx alert ruler groups apply [flags]
+```
+
+### Options
+
+```
+      --datasource string   Datasource UID of the Mimir/Loki ruler (required)
```
dafydd-t: I think we should support `-d` here too, like datasources commands:

```
gcx metrics labels -h | grep datasource
List all labels or get values for a specific label from a Prometheus datasource.
  # List all labels (use datasource UID, not name)
  -d, --datasource string   Datasource UID (required unless default-prometheus-datasource is configured)
```

docs/reference/cli/gcx_alert_ruler_groups_apply.md
```diff
@@ -0,0 +1,40 @@
...
+      --dry-run             Parse and validate only; send nothing to the ruler
+  -f, --filename string     File containing rule groups (Prometheus rules file or a single group; YAML/JSON, use - for stdin)
+  -h, --help                help for apply
+      --namespace string    Ruler namespace to apply the groups to (required)
```
dafydd-t: similarly, i think we could accept `-n` here

docs/reference/cli/gcx_alert_ruler.md
```diff
@@ -0,0 +1,41 @@
...
+Manage alerting and recording rules stored in a Mimir or Loki ruler,
+via Grafana's per-datasource ruler proxy.
+
+These are datasource-managed rules, distinct from Grafana-managed alert rules.
+Grafana-managed rules are read via 'gcx alert rules' and written via
+'gcx resources pull/push alertrules' — the write path requires Grafana 13+,
+where the rules.alerting.grafana.app API is enabled by default (on Grafana 12
+it requires the experimental 'kubernetesAlertingRules' feature toggle and a
+restart). Every ruler command requires --datasource with the UID of a
+Prometheus-flavored or Loki datasource.
```
dafydd-t: ∂o we need all this context about the Grafana-managed alert rules version compatibility in here?

docs/reference/cli/gcx_alert_ruler_groups.md
```diff
@@ -0,0 +1,30 @@
+## gcx alert ruler groups
+
+Manage ruler rule groups.
```
dafydd-t:
```suggestion
Manage datasource-managed (Mimir/Loki ruler) rule groups.
```


like above?

dafydd-t: Maybe we should drop the `Mimir/Loki` part from the top level description? your call. Either way, it should probably be consistent all the way down

docs/reference/cli/gcx_alert_ruler_groups_apply.md
dafydd-t:
```suggestion
      --datasource string   UID of the of the Mimir or Loki datasource used as ruler (required)
```

maybe something like this?

docs/reference/cli/gcx_alert_ruler_groups_delete.md
```diff
@@ -0,0 +1,32 @@
+## gcx alert ruler groups delete
+
+Delete a ruler rule group.
+
+```
+gcx alert ruler groups delete NAMESPACE GROUP [flags]
```
dafydd-t, line 6: for apply, we have required flags:

```
gcx alert ruler groups apply [flags]
```

but here we have a mixture of required positional arg (namespace), and required flag (--datasource). can we be consistent?

dafydd-t: same applies to the other verbs

docs/reference/cli/gcx_alert_ruler.md
```diff
@@ -0,0 +1,41 @@
+## gcx alert ruler
+
+Manage datasource-managed (Mimir/Loki ruler) rule groups.
```
dafydd-t: do we need to advertise more that these commands only work against mimir and loki type datasources?

internal/providers/alert/ruler_client.go
```diff
@@ -0,0 +1,132 @@
...
+// Managed Prometheus, ...) map to "mimir" like plain Prometheus.
+func rulerSubtypeForDatasourceType(dsType string) (string, error) {
+	switch query.NormalizeKind(dsType) {
+	case "prometheus":
+		return "mimir", nil
+	case "loki":
+		return "", nil
+	default:
+		return "", fmt.Errorf("datasource type %q has no ruler API (supported: prometheus-flavored, loki)", dsType)
+	}
```
dafydd-t, line 60: this colon breaks the formatting on an error:

```
gcx alert ruler groups list --datasource grafanacloud-traces
Error: Datasource type "tempo" has no ruler API (supported
│
│ prometheus-flavored, loki)
│
└─
```

dafydd-t (changes requested):
mostly docs comments, and an error formatting issue

internal/providers/alert/ruler_commands.go
```diff
@@ -0,0 +1,501 @@
...
+// rulerOpts carries the flags shared by every ruler command.
+type rulerOpts struct {
+	Datasource string
+}
+
+func (o *rulerOpts) setup(flags *pflag.FlagSet) {
+	flags.StringVar(&o.Datasource, "datasource", "", "Datasource UID of the Mimir/Loki ruler (required)")
+}
+
+func (o *rulerOpts) validate() error {
```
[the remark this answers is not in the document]
wardbekker: Fixed — all four ruler opts types (`rulerOpts`, `rulerNamespacesDeleteOpts`, `rulerGroupsApplyOpts`, `rulerGroupsDeleteOpts`) now expose an exported `Validate()`.

docs/reference/cli/gcx_alert_ruler_groups_apply.md
wardbekker: Added — the flag is now `-d, --datasource`, matching the datasource commands.

docs/reference/cli/gcx_alert_ruler_groups_apply.md
wardbekker: Handled a bit differently: namespace became a positional arg (`apply NAMESPACE`) so that `get`, `apply`, and `delete` all take it the same way. Happy to add `-n` as well if you'd still prefer the flag form.

docs/reference/cli/gcx_alert_ruler.md
wardbekker: Trimmed — the version-compatibility detail moved out of the command help and into the README compatibility matrix, which now breaks support down by version tier.

docs/reference/cli/gcx_alert_ruler_groups.md
wardbekker: Applied verbatim.

docs/reference/cli/gcx_alert_ruler_groups_apply.md
wardbekker: Applied (minus the duplicated "of the"): `UID of the Mimir or Loki datasource used as ruler (required)`.

docs/reference/cli/gcx_alert_ruler_groups_delete.md
wardbekker: Made consistent — `get`, `apply`, and `delete` all now take a positional `NAMESPACE` plus the `--datasource` flag.

docs/reference/cli/gcx_alert_ruler.md
wardbekker: Now called out explicitly in the ruler help: "These commands work only against Mimir (or another Prometheus-flavored ruler) and Loki datasources: every command requires --datasource with the UID of such a datasource."

internal/providers/alert/ruler_client.go
wardbekker: Fixed — the message is now `datasource type %q has no ruler API (supported types are prometheus-flavored and loki)`, no colon.

wardbekker:
@dafydd-t all 8 of your inline comments plus the Claude bot's `validate()`/`Validate()` note are addressed (each thread replied to and resolved). Two worth calling out:

- **`-n` for namespace on apply** — handled differently than suggested: namespace became a positional arg (`apply NAMESPACE`), which also settles your `apply`/`delete` inconsistency comment. `get`, `apply`, and `delete` now all take positional `NAMESPACE` + `--datasource`. Say the word if you'd rather have the `-n` flag.
- **Version-compat context** — pulled out of the ruler help and into the README compatibility matrix, now structured by version tier.

Rebased onto latest `main` (clean, no conflicts). Lint clean, full suite green, no reference-doc drift. Ready for another look.

dafydd-t:
I renamed `gcx alert ruler groups apply` to `gcx alert ruler groups upsert` to follow our naming conventions, otherwise we're good!

dafydd-t (approved):
wont have any effect, but LGTM

annanay25 (approved):

PR #986, opened 2026-07-15
fix(login): request full cloud scopes in login cloud followup

dafydd-t:
## Summary

- `gcx login`'s cloud followup requested only `stacks:read/write/delete` (via `auth.DefaultGCOMScopes`), while `gcx cloud login` hardcoded the full 8-scope set as its `--scope` default. Tokens minted by the followup then lacked the signal write scopes (needed for minting the SM token) and fleet-management scopes.
- Expands `DefaultGCOMScopes()` to the full set and points `gcx cloud login`'s `--scope` default at it, so both paths share one definition.
- New tests pin the scope set (`internal/auth`) and assert the flag default equals `DefaultGCOMScopes()` (`cmd/gcx/cloud`), guarding against re-divergence.

No reference-doc changes needed: the rendered `--scope` default value is unchanged.

Noticed while reviewing, left out of scope: `gcx cloud login --scope ""` passes `Validate` (len 1 slice with an empty string) and fails server-side instead of at the CLI — pre-existing, happy to fix separately if we care.


dafydd-t:
@claude review

wardbekker (approved):

PR #987, opened 2026-07-16
fix(output): descend into single-key list envelopes for --json discovery and selection

wardbekker:
## Summary
- `gcx datasources list --json list` printed only the wrapper key `datasources` instead of the item-level field names (reported in Slack), and `--json uid,name` returned all-null values.
- Root cause: the `--json` discovery path (`marshalToSampleMap`) and selection path (`FieldSelectCodec.Encode`) special-cased only the k8s `items` wrapper key; envelopes keyed otherwise (`datasources`, `results`, `types`) were treated as single objects.
- Fix: generalize both paths via a shared `singleKeyItems` helper — an object with exactly one key holding an array of objects is treated as a list envelope. Discovery samples the first item (reflection fallback on the wrapper's sole slice field for empty lists); selection extracts per item and preserves the wrapper key.

Before / after:

```
$ gcx datasources list --json list
datasources                          # before
access default name readOnly type uid url   # after (one per line)

$ gcx datasources list --json uid,name
{"name": null, "uid": null}          # before
{"datasources": [{"uid": "...", "name": "..."}, ...]}   # after
```

## Changes
- `internal/output/field_select.go` — new `singleKeyItems` helper; `FieldSelectCodec.Encode` descends into single-key envelopes preserving the wrapper key; doc comment updated.
- `internal/output/format.go` — object branch of `marshalToSampleMap` extracted into `sampleFromObject`, which now also handles single-key envelopes; `reflectSingleSliceField` + `nullFieldMap` helpers for empty/nil envelopes.
- Tests: table-driven coverage in `internal/output` (selection, discovery, empty/nil envelope, scalar-array/map-value/multi-key non-envelopes) and end-to-end `--json list` / `--json uid,name` in `cmd/gcx/datasources/list_test.go`.

The k8s `items` path (with pagination metadata) is unchanged; bare slices, scalar arrays, map values, and multi-key objects keep their previous behavior. Also fixes `datasources health` (`results`) and `datasources schema` (`types`), which share the envelope shape.

**Behavior note:** selecting the wrapper key itself (`--json datasources`) previously echoed the whole array; it now yields per-item nulls like any non-item field — use `--jq '.datasources'` for that.

## Test Plan
- [x] Tests pass locally (`GCX_AGENT_MODE=false mise run all`)
- [x] New tests cover the change (failing-first TDD)
- [x] Lints clean
- [x] Verified live against a stack: discovery lists item fields; selection returns real values; `providers list --json list` (bare slice) and k8s `items` paths unchanged
- [ ] Architecture docs: n/a (no new packages, no command/flag surface change)

### DIFF
--- a/cmd/gcx/datasources/list_test.go
+++ b/cmd/gcx/datasources/list_test.go
@@ -166,6 +166,37 @@ func TestListDefaultReturnsAllDatasources(t *testing.T) {
 	assert.Len(t, result.Datasources, 60)
 }
 
+func TestListJSONFieldDiscoveryAndSelection(t *testing.T) {
+	server := newDatasourceListServer(t, 2)
+	defer server.Close()
+
+	configFile := newConfigFileForServer(t, server.URL)
+
+	t.Run("--json list discovers item fields", func(t *testing.T) {
+		stdout, err := executeDatasourceCommand(t, []string{"datasources", "list", "--config", configFile, "--json", "list"})
+		require.NoError(t, err)
+
+		for _, field := range []string{"uid", "name", "type", "url"} {
+			assert.Contains(t, stdout, field, "item field %q must be discovered", field)
+		}
+		assert.NotContains(t, stdout, "datasources", "wrapper key must not be listed")
+	})
+
+	t.Run("--json uid,name selects per item", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


dafydd-t (approved):
Thanks!

PR #994, opened 2026-07-17
docs(adr): command operation semantics and pre-GA naming convergence (revised: rubric + proposed registry)

awsome-o:
## Status (2026-07-20)

Synced with Dafydd on Slack: this PR is alignment only and we won't block any implementation on it. It stays open as the record of the conventions and decisions while the actual work proceeds. His review answers from 2026-07-20 are recorded in the decision table and in §1a of the rollout plan. The rollout order changed to match: inventory/classification, the pilot renames and per-provider owner outreach start immediately in parallel, the registry + lexical CI PR goes last so it blocks nothing, and the output format / global `--limit` work moves to its own sub-issue(s) under #387.

## What

This is a docs-only `proposed` decision PR, the naming workstream of #387, revised 2026-07-17 after maintainer feedback. It proposes a behavior-based naming rubric, an allowed-operation registry with a lexical CI check, and a clean v1 migration policy with complete migration dispositions.

It does not implement the registry, the inventory/classification, any command renames, the output-format or global `--limit` work (now separate sub-issue(s) of #387), any datasource-persistence refactor, or a machine-readable command contract. The full executable per-command contract is explicitly deferred.

### What changed since the first draft (maintainer feedback, 2026-07-17)

The universal machine-readable per-command contract (11-field tuples, bootstrap inference, migration census, shared resolver, versioned surface JSON) is deferred. The semantic model became a review rubric used to classify the surface as OK / suggested fix / human review, LLM-assisted but human-reviewed and source-cited.

Enforcement is an allowed-operation registry (each operation registered once, with a written definition) plus lexical CI that validates the terminal tokens of runnable commands, including hidden runnable commands, built-ins, and aliases attached to runnable commands. Group aliases are path prefixes: inventoried and dispositioned in the classification worksheet, never operation-validated. CI checks spelling; semantic fit stays human review.

Already confirmed by the maintainer and recorded: behavior determines meaning; clean-start v1 with a migration guide (D9); `profile-types` to `list-profile-types` at both mounts (D4c, respelled from `list-types` in pilot-PR review 2026-07-21); `wide` only when it adds useful columns. The 2026-07-21 KG owner decision is also recorded: `upsert` is the direct create-or-update workflow for one or more explicitly supplied subjects, processed independently and potentially non-atomically; cardinality does not distinguish it from `push`.

Migration completeness: every changed or removed primary command and alias resolves to `old → new`, `old → removed; use <replacement>`, or `old → removed; no replacement`. Group aliases map once as a prefix (`gcx k6 test …` maps to `gcx k6 load-tests …`).

### Files

ADR 024 (rubric + registry + naming-convergence gates), the rollout plan (D1–D14 dispositions, reordered rollout sequence, the two-part classification worksheet spec, migration dispositions), CONSTITUTION / DESIGN / naming.md §9.7 / provider-checklist (including the `wide`-codec policy correction), ADR 023 notes, UX-consistency plan corrections, and the July-1 audit imported with a historical/non-authoritative label.

### Open approval questions, status after the 2026-07-20 review

1. Registry + lexical CI (D2/D10): answered yes, with the caveat that v1 consistency of current commands is the priority. The registry PR moves last.
2. Keep `upsert`, remove misleading aliases (D6): answered yes.
3. Addressability refinement of `list-<subject>` (D10): still open. The original wording didn't land ("I don't know what this means"), so ADR §8 now opens with a plain-language restatement and the question is re-asked in the review thread.
4. ADR 023 ratification bundled or separate (D4a/b): answered "whatever gets the rename done fastest", so it's bundled housekeeping and never blocks a rename.
5. Output-format + global `--limit` tracking (D13): answered, separate sub-issue(s) of #387.
6. Initial registry contents: rescoped, seeded from the converged post-migration surface and shipped as the final PR.

Instrumentation `clusters apps` routes to its provider batch. That's an owner decision, not a prerequisite.


docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+| D14 | ~~Reads become pure~~ — **withdrawn as a decision.** Reduced to a factual caveat (ADR §4): several datasource reads persist auto-discovered datasource UIDs into config today (`ResolveAndSaveDatasource`), and documentation, help text, and classification must describe that behavior truthfully. Any pure-resolver/configuration refactor is **separate, out-of-scope work** — not designed, decided, or gated here. | Withdrawn (not a decision) | Affected today: e.g. `profiles profile-types`, `metrics labels` paths through `internal/datasources/query/resolve.go` |
+
+### 1a. Open approval questions (explicit — none of these are decided)
+
+Presenting this plan does not depend on resolving these, but none of them
+may be treated as already approved:
+
+1. Adopt the allowed-operation registry + lexical CI as the proposed
+   first implementation? (D2/D10)
```
dafydd-t: yes, but the most important thing for v1 is that the current commands are consistent.

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+2. Keep `upsert` and remove its misleading `create`/`update`/`apply`
+   aliases? (D6)
+3. Accept the addressability-based refinement of the `list-<subject>`
+   rule? (D10 — the generic allowance follows the maintainer's
+   suggestion; the two conditions are this proposal's)
+4. Ratify ADR 023 bundled with this decision, or separately? (D4a/b)
+5. How does #387 track the output-format-consistency and global
+   `--limit` work relative to this naming workstream? (D13)
```
dafydd-t: can we do it separately? sub-issue?

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+6. Exact initial operation-registry contents and definitions (separately
+   approved diff).
```
dafydd-t: I guess we need to have changed all the commands to be consistent before we can make this?

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+   and Cobra built-ins) and an **alias record** for every alias on every
+   node, runnable or not (§3). Classify each command as **OK / suggested
+   fix / human review** against the ADR rubric (LLM-assisted,
+   human-reviewed). Every *suggested fix* cites the observed source
+   behavior it is based on.
+5. **Owner approval**: provider/product owners approve renames, removals,
+   ratified exceptions, every *human review* disposition, every alias
+   disposition, and any *suggested fix* resolved as `keep`.
+   Instrumentation's `clusters apps` questions route to the
+   Beyla/Instrumentation Hub owners as part of their batch (§3b).
```
dafydd-t: I would crack on with these ASAP, I think we can work on the first 3 steps in parallel. we don't have much time to get these changes through

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+   scoped exceptions; a temporary exact-path deviation list that must be
+   empty before v1.0.0).
+7. **Pilot**: the three adaptive `show → list` renames (D12) — clean
+   renames per D9 (no forwarders), updating skills, annotations, docs,
+   and generated references atomically. The `profile-types → list-types`
+   rename (D4c) rides the same early wave.
+8. **Provider-sized migration batches in parallel, all before GA**: each
+   command record resolves keep / ratify / rename / remove; every alias
+   record resolves keep-synonym / remove; each batch records its
+   migration dispositions (§4 gate 6).
```
dafydd-t: put the registry PR last; we don't want to block on it.

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+
+### 1a. Open approval questions (explicit — none of these are decided)
+
+Presenting this plan does not depend on resolving these, but none of them
+may be treated as already approved:
+
+1. Adopt the allowed-operation registry + lexical CI as the proposed
+   first implementation? (D2/D10)
+2. Keep `upsert` and remove its misleading `create`/`update`/`apply`
+   aliases? (D6)
```
dafydd-t: yes

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+3. Accept the addressability-based refinement of the `list-<subject>`
+   rule? (D10 — the generic allowance follows the maintainer's
+   suggestion; the two conditions are this proposal's)
```
dafydd-t: I don't know what this means

docs/plans/2026-07-16-command-operation-contract-rollout.md
```diff
@@ -0,0 +1,386 @@
...
+4. Ratify ADR 023 bundled with this decision, or separately? (D4a/b)
```
dafydd-t: whatever gets the rename done fastest 😅

awsome-o:
Thanks for the review! I've folded your answers into the decision table and the rollout plan, and per our Slack chat we're treating this PR as alignment only. Nothing blocks on it, it just stays open as the record of the conventions while we get on with the actual work.

So the plan from here: the worksheet listing every command that doesn't fit the conventions is ready and will go up as its own PR. The first renames (profile-types to list-types, plus the three adaptive show to list ones) go out right after. From the worksheet we cut provider sized batches, open a PR per provider and message each team about theirs. The registry and lexical CI PR goes last like you asked so it never blocks anything, and the output format / --limit work gets its own sub-issue under #387 with separate PRs.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Yeah fair, that sentence was way too dense. What it's trying to say: it's the rule for picking between `gcx <area> things list` and `gcx <area> list-things`.

If a thing has its own ID, meaning you can fetch exactly one of them by that ID, it gets its own command group: `k6 load-tests list`, `k6 load-tests get <id>`.

`list-things` is only for listings where that can't work. Either the values have no ID at all (`cloudwatch list-regions`, there's no such thing as `region get`), or the list hangs off the parent's ID (`alert-groups list-alerts <group-id>`, the ID you pass belongs to the group, not to an alert).

The "refinement" part is just this: "the subject only has a list command today" doesn't count as a reason to write `list-things`. If we allowed that, every one-verb subject would become a flat `list-x` leaf, and the moment it grows a `get` we'd have to either rename it (breaking change) or live with a mixed shape. So the test is always "does it have its own ID", never "how many verbs does it have right now". Without that rule, `list-<anything>` is legal everywhere and we're back to the leaf token sprawl we're trying to kill.

Updated

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Done, moved it to the end of the rollout. Nothing waits on it now.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Agreed. Reordered the plan so the inventory/classification, the pilot renames and the owner outreach all start right away in parallel, and the ADR paperwork trails behind. The worksheet and the first renames are already built, they just need to be cut into PRs. And per Slack, nothing blocks on this ADR, it stays open as the record.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: That's how the plan reads now: getting the current surface consistent is the v1 priority, the registry is just the guard rail we add at the end.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Recorded in D6. `upsert` stays, and the `create`/`update`/`apply` aliases go away in the alert batch.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Fastest is bundling it, so that's what the plan says now. Either way no rename waits on it.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Yes, will do. I'll file it as a sub-issue of #387 and the output format / `--limit` work gets its own PRs, separate from the naming batches.

docs/plans/2026-07-16-command-operation-contract-rollout.md
awsome-o: Pretty much. The registry contents get written from whatever the surface looks like after the migration batches land, which is also why that PR now goes last.

docs/plans/2026-07-16-command-operation-contract-rollout.md
dafydd-t: >The "refinement" part is just this: "the subject only has a list command today" doesn't count as a reason to write list-things. If we allowed that, every one-verb subject would become a flat list-x leaf, and the moment it grows a get we'd have to either rename it (breaking change) or live with a mixed shape. So the test is always "does it have its own ID", never "how many verbs does it have right now". Without that rule, list-<anything> is legal everywhere and we're back to the leaf token sprawl we're trying to kill.



I think this is OK, lets see how many commands we'd have to change to implement this

awsome-o: Ran the numbers. Nothing that exists today changes: there are 12 `list-*` commands in the tree and all of them already fit the rule. The impact goes the other way: 26 commands are candidates to become a `list-<subject>` spelling, spread across 11 providers: agento11y 7, datasources 5, assistant 3, irm 3, k6 2, and one each in appo11y, cloud, dashboards, dev, frontend and resources. That's on top of the two profile-types -> list-types mounts already in flight.

A few examples so it's concrete: `agento11y agents versions` -> `agents list-versions <agent-name>`, `influxdb measurements` -> `list-measurements`, `athena describe-table` -> `list-columns` (it only returns column names, the current name oversells it).

Treat 26 as the ceiling rather than a commitment. Each one gets a final keep-or-rename call in its provider batch since a few will likely stay put (`dashboards versions list` is arguably fine as is). The first two batches, agento11y and datasources, cover 12 of the 26 and are being prepared as draft PRs now.

awsome-o:
Rebased onto current main and closed the remaining wording gaps. The naming guide now matches the owner-approved multi-subject upsert behavior, the rollout records the completed alert alias removal, the cross-signal ADR no longer calls an unmerged #1009 shipped, and the remaining active plan references use list-profile-types. The PR description now records the KG owner decision too.

PR #996, opened 2026-07-17
fix(irm): decode expanded webhook integration filters

d0ugal:
## Summary

- accept both integration ID strings and expanded `{id, name}` objects in Cloud IRM webhook responses
- normalize expanded references to IDs so manifests and create/update requests retain the portable string-array format
- add regression coverage for legacy, expanded, mixed, empty, null, invalid, and marshal-round-trip cases

## Why

The Cloud IRM internal webhook endpoint can expand `integration_filter` entries into objects. The existing `[]string` response model fails to decode those webhooks, causing `gcx irm oncall webhooks list` to fail before producing output.

## Reproduction

Before this fix, listing webhooks against a stack that returns expanded integration references fails immediately:

```console
$ gcx irm oncall webhooks list
Error: Irm: decode webhook

json: cannot unmarshal object into Go struct field Webhook.integration_filter of type string
```

## Test plan

- [x] `go test ./internal/providers/irm/...`
- [x] `go test ./...`
- [x] `golangci-lint run -c .golangci.yaml`
- [x] `GCX_AGENT_MODE=false mise run --jobs=1 reference-drift`
- [x] build and verify `gcx irm oncall webhooks list -o json` against a Grafana Cloud IRM stack
- [ ] local MkDocs render (the local Python environment does not have `mkdocs-material`; CI will run the documentation build)

### DIFF
--- a/internal/providers/irm/oncall_client_test.go
+++ b/internal/providers/irm/oncall_client_test.go
@@ -489,6 +489,87 @@ func TestListWebhookPresets(t *testing.T) {
 	}
 }
 
+func TestListWebhooksIntegrationFilterShapes(t *testing.T) {
+	t.Parallel()
+
+	client := newTestOnCallClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		wantPath := irm.BasePath + "/webhooks/"
+		if r.URL.Path != wantPath {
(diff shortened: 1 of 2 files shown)


annanay25 (approved):
lgtm!

olegbespalov (approved):

PR #999, opened 2026-07-18
docs(readme): clarify gcx works with OSS and Enterprise, not just Cloud

wardbekker:
## Summary

The README states OSS/Enterprise/Cloud support in the tagline and compatibility matrix, but several sections skewed Cloud-first. This tightens the deployment story:

- **Quick Start**: reorder the query examples so the works-on-any-Grafana example comes first; label the `grafanacloud-usage` example as Grafana Cloud only
- **What is gcx?**: state explicitly that gcx targets Cloud, Enterprise, and OSS, with core features on any Grafana 12+ and Cloud adding product commands on top
- **The Agentic Workflow**: flag that the Assistant-based example requires Grafana Cloud, linking to the compatibility matrix and the works-anywhere workflows below it

Docs-only change; `GCX_AGENT_MODE=false mise run all` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Query production. Investigate alerts. Let the Assistant root-cause issues. Ship 
 
 ## What is gcx?
 
-gcx is a CLI for Grafana. It gives you and your AI coding agent structured access to your Grafana instance: dashboards, alerts, SLOs, metrics, logs, traces, and more.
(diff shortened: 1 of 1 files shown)


annanay25 (approved):
lgtm.

dafydd-t (approved):

Issue #997, opened 2026-07-17
Complete v2 investigation create/read parity


CadenMG:
## Goal

Complete v2 parity for **creating and reading** Lodestone investigations.

## Current coverage

Gcx already supports the full v2 create body: instruction, title, teams, and profile ID. It also supports list, snapshot-based get, and read-only chat-derived views.

## Missing pieces

- [x] **Lossless list output.** Preserve the full v2 summary and collection total. Gcx currently drops `chatId`, description, variant, labels, owner, active-loop count, token usage, completion quality, degradation reason, agent/progress data, richer source metadata, and `total`—including from JSON/YAML output.
- [ ] **Remaining list inputs.** Add `--chat-id`, `--include-archived`, and `--archived-only`. Reconsider defaulting `--include-legacy` to true for a v2-focused command.
- [x] **Expose both identifiers.** `get` internally resolves the investigation row but discards its `chatId`. Read output should expose both `investigationId` and backing `chatId`.
- [x] **Structured evidence.** Add `gcx assistant investigations evidence <id>` backed by `GET /api/v2/investigations/{id}/evidence`. The chat-derived `tools` view is not the canonical citation index.
- [ ] **Optional wait ergonomics.** Consider `create --wait` or `get --watch` for polling until a terminal state.
- [ ] Cleanup v1 Investigation code paths

## Out of scope

Track management and A2A interaction separately: loops, team-scope replacement, Mermaid repair, follow-up/conclude, reconnect, interrupts, `/ask`, images, live events, and cleanup of the orphaned `regenerate-report` command.

PR #988, opened 2026-07-16
feat(output): standardize list truncation with shared, agent-legible metadata

wardbekker:
## Summary
- Adds a shared, pagination-aware list-output helper (`internal/output/listmeta.go`): truncation is never silent — a truncated page carries `list_meta` (`{truncated, returned, total?, continue}`) in the structured payload itself, plus a standardized `hint: showing first N of M: <cmd> --limit 0` on stderr.
- Migrates the two reference commands: `datasources list` (cheaply-complete source — silent client-side slice removed, `--limit` now defaults to 0/full set, observed total reported) and `irm oncall alert-groups list` (paginated source — server more-signal on the rich path, over-fetch-by-one on the SA-token fallback, bespoke doubled-limit hint removed).
- Documents the contract as `docs/design/output.md` §14 (List Truncation Contract) so remaining list commands can migrate incrementally.

## Design
Two source shapes, three constructors — total is only ever asserted when actually observed:

| Source shape | Helper | `total` |
|---|---|---|
| Cheaply complete (no server-side limit; full set already fetched) | `TruncateCompleteList` | observed |
| Paginated, API reports more-pages | `PagedListMeta` | omitted (unknown) |
| Paginated, no more-pages signal | request `limit+1`, then `TruncatePagedList` | omitted (unknown) |

`list_meta` is absent when the output is the complete set, so its presence *is* the truncation signal — an agent reading stdout cannot mistake a page for the whole set. `--limit 0` = unlimited, uniformly, and every hint suggests it. In agent mode the stderr hint switches to the existing JSONL `{"class":"hint"}` form.

Not included: the k8s per-plugin-group fan-out parallelization called out as a separate perf fix — it already landed on main (`internal/datasources/k8stransport.go`, bounded errgroup, limit 10).

## Changes
- `internal/output/listmeta.go` — `ListMeta`, `TruncateCompleteList`, `TruncatePagedList`, `PagedListMeta`, `EmitListTruncationHint` (+ table-driven tests)
- `cmd/gcx/datasources/list.go` — helper migration, `--limit` default 50→0
- `internal/providers/irm/oncall_commands_extra.go` — rich + legacy list paths migrated; `emitAlertGroupListLimitHint` removed
- `docs/design/output.md` — §14 List Truncation Contract
- `docs/reference/cli/gcx_datasources_list.md` — regenerated
- `CLAUDE.md` — package map entry

## Test Plan
[...]

PR #1001, opened 2026-07-18
feat(assistant): preserve full v2 investigation summaries and total in list output

CadenMG:
First slice of #997: make `gcx assistant investigations list` lossless on v2 stacks.

## Problem

The v2 list response was decoded into the v1 `InvestigationSummary` struct, silently dropping most of the payload (`chatId`, description, labels, owner, token usage, quality/degradation, agent/progress data, …) and the collection `total` — from every output format, including json/yaml.

## Changes

- New `LodestoneList` / `LodestoneInvestigationSummary` types mirroring the server schema field-for-field (including omitempty placement), so json/yaml output is lossless and carries `total`. Deprecated always-null `confidence` omitted; v1 unchanged.
- Table/wide columns unchanged for both API versions; `wide` falls back to `ownerUserId` when `source.userId` is absent.
- New `output.ListEnvelope` opt-in (`ListItemsKey() string`) so `--json` field selection/discovery and the agents spill preview descend into multi-key envelopes like `{investigations, total}` instead of operating on the wrapper. Explicit opt-in rather than a heuristic, to avoid misclassifying detail objects with a nested array. Documented in `docs/design/output.md` §1.5 and `docs/architecture/cli-layer.md`; also covers #988's `{items, list_meta}` shape.

Note: v2 list json/yaml output changes from a bare array to `{"investigations": [...], "total": N}`.

## Testing

Unit tests pin the lossless decode, the required-vs-omitempty split, the envelope codec paths, and the spill preview. E2E-verified against a local mock of the plugin API: table/wide, json/yaml, `--json` selection + discovery, `--jq`, agent mode, and spill. `GCX_AGENT_MODE=false mise run all` passes.

Remaining #997 bullets follow as separate PRs.


alexsniffin (approved):
LGTM

annanay25 (approved):
should we keep v1 at all?

PR #1003, opened 2026-07-20
feat(skills): bundle agento11y-eval-starter in gcx

soniaAguilarPeiron:
Consolidate the Agent Observability skill family so they all install the same way.

## Problem
The `agento11y` skills were split across two distribution regimes:
- **3 in gcx** (`agento11y`, `agento11y-instrument`, `agento11y-prod-setup`) — installable via `/plugin install gcx@gcx-marketplace` or `gcx agent skills install <name>`.
- **2 in grafana/sigil-sdk** (`agento11y-eval-starter`, `agento11y-experiments`) — **loose files with no install mechanism** (llms.txt routes to them but says "check out the repo").

This asymmetry blocks a clean install story in docs and the UI.

## This PR
Moves **`agento11y-eval-starter`** into the gcx bundle so it installs like the others. The move is clean: the skill has **zero `gcx` invocations** (drift test unaffected), no sigil-sdk-relative paths, and its generated runner imports the `sigil-sdk` pip package (not a local path).

- Copied verbatim, plus a temporary note that its sibling `agento11y-experiments` still ships in sigil-sdk (its consolidation is pending a separate decision — no change here).
- **Bonus:** bringing eval-starter into the bundle resolves the previously-dangling handoff references to it from `agento11y-instrument` (see #998) and `agento11y-prod-setup` — they now resolve within one bundle.
- Also fixes `claude-plugin/README.md` (was missing the `agento11y-prod-setup` row) and the stale bundled-skill count in the root README example.

`agento11y-experiments` is intentionally **not** moved here — pending a decision on whether it should lockstep with the Python SDK.

## Not mapped in command_skills.go
eval-starter is offline (reads code, scaffolds a suite), not `gcx`-command-driven, so it stays unmapped — `TestCommandSkills_AllSkillsExistInBundle` only requires mapped skills to exist, so this is fine.

## Testing
- `go build ./...`, skills-drift (`TestSkillsGcxInvocationsMatchCommandTree`), and command-skills tests pass.
- `gcx agent skills list` shows `agento11y-eval-starter`; `gcx agent skills install agento11y-eval-starter --dry-run` works.

## Follow-up
A grafana/sigil-sdk PR (merged after this) removes the original loose file and repoints `llms.txt` Path C at the gcx install command. (The onboarding-wizard copy of llms.txt needs the same edit.)

### DIFF
--- a/claude-plugin/skills/agento11y-eval-starter/SKILL.md
+++ b/claude-plugin/skills/agento11y-eval-starter/SKILL.md
@@ -1,0 +1,364 @@
+---
+name: agento11y-eval-starter
+description: >
+  Use early in an AI-agent project — before ship, before real traffic — to decide which
+  evaluations to set up and to scaffold a starter experiment. Reads the agent's own code (system
+  prompt, tools, task), recommends specific evaluators with reasons that cite real lines, and writes
+  a labeled draft test suite as an Agent Observability suite YAML. It also assesses how runnable the agent is: for
+  an easily-invoked agent it generates a runner stub (run_experiment.py) with one hole to fill and
+  can optionally run it (only with permission, only against the endpoint the developer configured);
+  for agents that need a harness or a full runtime it points to the existing eval infra instead of
+  emitting a runner that can't call the agent. It never creates tenant-level evaluators, rules, or
+  guards.
+---
+
+# agento11y eval starter
+
+Help a developer who has an AI agent but no evaluation set up yet. The hard part isn't
+running an experiment — it's knowing **what to evaluate** and having **cases to test
+against** before there's any traffic. Answer both, grounded in the agent's actual code.
+
+Always produce:
+
+1. A ranked, justified **evaluator recommendation** for this agent.
+2. A starter **suite YAML** the developer reviews and extends.
+
+Then, depending on how runnable the agent is (Step 1):
+
+3. For an easily-invoked agent, a **runner stub** (`run_experiment.py`) that wires the suite
+   to the SDK with one hole to fill — and optionally run it (Step 6), only with permission.
+   For an agent that needs a harness or full runtime, point to the existing eval infra instead
+   of a runner that can't actually call it.
+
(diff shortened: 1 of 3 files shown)


claude-plugin/skills/agento11y-eval-starter/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+  prompt, tools, task), recommends specific evaluators with reasons that cite real lines, and writes
+  a labeled draft test suite as a Sigil suite YAML. It also assesses how runnable the agent is: for
+  an easily-invoked agent it generates a runner stub (run_experiment.py) with one hole to fill and
+  can optionally run it (only with permission, only against the endpoint the developer configured);
+  for agents that need a harness or a full runtime it points to the existing eval infra instead of
+  emitting a runner that can't call the agent. It never creates tenant-level evaluators, rules, or
+  guards.
+---
+
+# Sigil eval starter
```
dafydd-t:
```suggestion
# agento11y eval starter
```

maybe check elsewhere for this?

dafydd-t: lots of hits for `sigil` in this diff

soniaAguilarPeiron: goooooood catch!!!!! thanks!!!

soniaAguilarPeiron: Fixed — rebranded the visible `Sigil` mentions to `Agent Observability` (title + prose: tenant, Rules, surfaces, SDK, experiment, product). Left the real technical names as-is (`sigil_sdk`, `sigil.experiment()`/`Evaluator`/`TestSuite`, the Go `sigil` package, the grafana/sigil-sdk repo) since renaming those would break code. Thanks for the catch! 🙏

dafydd-t (approved):

PR #1004, opened 2026-07-20
feat(assistant): add investigations list-profiles command

CadenMG:
Part of #997 — closes the **Profile discovery** bullet only.

Adds `gcx assistant investigations list-profiles` (v2-only, gated like pause/resume), backed by `GET /api/v2/investigations/profiles`, so users can discover valid values for `create --profile-id`. Named per the naming conventions guide (#1023): no single-profile GET exists, so the `list-things` form applies rather than a `profiles` noun group.

- Default `table` output (ID / NAME / DEFAULT / MAX STEPS / TOOLS; `wide` adds DESCRIPTION); `json`/`yaml` emit the lossless `{"profiles": [...]}` envelope.
- `LodestoneProfile` mirrors the server's required-vs-omitempty split exactly (`description` is the only optional field), pinned by a serialization test — same issue class as the #1001 review finding.
- Help text documents the `assistant.lodestone-allow-profile-selection` tenant flag (without it: default profile only, create rejects non-default IDs); `create --profile-id` cross-references the command.
- Agent `token_cost` annotation + regenerated CLI reference docs.

**Verified:** `GCX_AGENT_MODE=false mise run all` green; httptest client + codec tests; sandboxed mock-server e2e (table/wide/json/yaml/`--json` selection + discovery); live against a Cloud stack (returns the single default profile as expected without the flag).


alexsniffin (dismissed):
LGTM

CadenMG:
Heads up for re-review: renamed the command `profiles` → `list-profiles` in 9704347c to conform to the naming conventions guide (#1023) — a noun group requires a fetchable single resource, and the profiles API has no single-profile GET, so the `list-things` form applies (precedent: `cloudwatch list-regions`). Same behavior otherwise; annotation key, help cross-reference, and reference docs updated to match.

Issue #972, opened 2026-07-14
gcx kg entities correlate — resolve affected entities from alert labels


xujiaxj:
## Summary
Add `gcx kg entities correlate`: given one or more firing alerts (labels and/or a PromQL expr) and a time window, return the affected KG entities and the surrounding assertion graph. This fills a real gap — every current entity read path (`list`, `inspect`, `query`) requires you to *already know* the entity type+name or write Cypher. `correlate` is the "I have an alert, which entity does it hang off?" entry point, using the same correlation logic the product uses to attach alerts to entities.

## Backend
- **Endpoint:** `POST /v1/alert-inspection` — `AlertInspectionController.inspect` (asserts-adi).
- **Request** (`AlertEntityLookupRequestDto`): `alertLabels: List<Map<String,String>>` (multiple alerts), optional `query` (PromQL), `timeCriteria{start,end}` (epoch ms), `config` (has a server default — omit).
- **Resolution precedence** (`AlertEntityLookupService.inspect`): scope-enriched labels → raw labels → **fallback to `query`**, which is executed as a live **instant query** and its result series' labels are correlated. So labels win; `query` is the fallback.
- **Response** (`AlertEntityLookupResponseDto`): an assertion **graph** (`data`, `type: graph`), not a single entity.
- **Empty result is `200` with empty `data`, not `204`** — detect "no correlation" from the payload, exit 0, print a clean message.
- **No backend change** (endpoint already exists) — but confirm it's exposed through the **api-gateway** route (raw-passthrough test against a real stack).

## Command design
```bash
# inline labels
gcx kg entities correlate --alert-labels 'alertname=ErrorRatioBreach,job=cart' --since 1h
# Alertmanager JSON via -f/stdin (envelope {alerts:[…]} or array [{labels:…}] auto-detected)
gcx kg entities correlate -f am-webhook.json --since 1h
# alert rule expr (fallback)
gcx kg entities correlate --query 'rate(http_errors[5m]) > 0.1' --since 1h
```

| Flag | Maps to |
|---|---|
| `--alert-labels 'k=v,k=v'` (repeatable) | one entry each in `alertLabels` |
| `-f/--file` (+ stdin) | `alertLabels` ← `.alerts[].labels` / bare `[].labels` (matches kg's `-f` convention) |
| `--query` | `query` (PromQL fallback) |
| `--since` / `--from`/`--to` | `timeCriteria` (required; default 1h) |

[...]

PR #1005, opened 2026-07-20
feat(kg): add entities correlate to resolve entities from alert labels

kblaschke512:
## Summary

Adds `gcx kg entities correlate`: given one or more firing alerts (their label sets) and a time window, resolve the affected Knowledge Graph entities via `POST /v1/alert-inspection` — the same correlation the product uses to attach alerts to entities (backs the IRM Alert Groups "Impact" tab).

This is the **"I have an alert, which entity is it?"** entry point. Every existing entity read path (`list`, `inspect`, `query`) requires you to already know the entity type/name or write Cypher; `correlate` fills that gap.

Implements #972.

## Usage

```bash
# inline labels (one --alert-labels per alert; repeatable)
gcx kg entities correlate --alert-labels 'alertname=ErrorRatioBreach,job=cart' --since 1h

# Alertmanager webhook JSON via -f / stdin (envelope {"alerts":[...]} or bare array, auto-detected)
gcx kg entities correlate -f am-webhook.json --since 1h
cat am-webhook.json | gcx kg entities correlate -f - --since 6h
```

- At least one of `--alert-labels` / `-f` is required; time window defaults to 1h.
- Table output shows each entity plus its connected-entity impact counts; `-o json|yaml` return the full graph envelope.
- An empty correlation is a `200` with empty `data.entities` (not `204`), so it prints a notice and **exits 0** rather than erroring.

## Design notes

- **Label-only for v1.** The `--query` PromQL fallback is deferred: raw-passthrough probing showed the backend fallback returns HTTP 500 on every form tested, while the label path worked every time (matches the backend owner's "results not guaranteed" caveat in the issue thread). The flag is not exposed.
- **Parent help rewritten** into a routing guide across `correlate` / `list` / `inspect` / `query`.
- Reuses the existing `GraphEntity` type for the response entities.

## Testing

- Unit tests: request mapping, Alertmanager envelope + bare-array detection, empty-200 → exit 0, JSON output, no-input validation.
- `mise run gate` (lint + tests + build) green after merging `main`.
- Manual end-to-end against two dev stacks:
  - `assertsdemo01` (Asserts enabled): real correlation → `Service ad` with impact counts `Namespace=1, Pod=1, Service=4`; all input/output paths verified.
  - `kevinblaschke` (Asserts disabled): confirmed the api-gateway route is exposed on a vanilla stack and the command degrades cleanly (empty-200 → exit 0).

## Follow-ups (not in this PR)

- **`--query` fallback** — blocked on backend hardening in asserts-adi; revisit exposing the flag once robust.
- **asserts-eval gcx-commands suite** — adding a subcommand changes the `gcx kg entities` command surface; snapshots + new eval cases need updating in that repo (handoff drafted separately).


internal/providers/kg/commands.go
```diff
@@ -2309,6 +2315,205 @@ func (c *CypherTableCodec) Decode(_ io.Reader, _ any) error {
...
+				TimeCriteria: &TimeCriteria{Start: startMs, End: endMs},
+			})
+			if err != nil {
+				return err
+			}
+			if len(resp.Data.Entities) == 0 {
+				fmt.Fprintln(cmd.ErrOrStderr(), "no entities correlated for the given alert(s)")
+			}
+			return ioOpts.IO.Encode(cmd.OutOrStdout(), resp)
+		},
```
xujiaxj, line 2717: We can simplify the output a little bit. It now has an envelope that makes an agent doing more work when parsing the output
```
gcx kg entities correlate --alert-labels 'alertname=Asserts: api-server-latency - Error Budget Burn Rate is High,__alert_rule_uid__=dfh3ynmstbfghe,asserts_env=prod-ca-east-0,grafana_folder=Grafana SLO,grafana_slo_provenance=asserts,grafana_slo_severity=warning,grafana_slo_uuid=j3zavgokdp354hreordgq,namespace=asserts,service_name=api-server,team_name=asserts' --since 1h)
{"class":"hint","summary":"use --json list / --json field1,field2 for field selection, or --jq '\u003cexpr\u003e' for transformation (group_by, filter, count) — no external parsing needed"}
{
  "type": "graph",
  "timeCriteria": {
    "start": 1784569296971,
    "end": 1784572896971
  },
  "data": {
    "entities": [
      {
        "id": 249876,
        "type": "Service",
        "name": "api-server",
        "active": true,
        "scope": {
          "env": "prod-ca-east-0",
          "namespace": "asserts"
        },
        "properties": {

        },
        "connectedEntityTypes": {
        },
        "assertion": {
        },
        "connectedAssertion": {},
        "assertionCount": 6
      }
    ],
    "edges": [],
    "pageNum": 0,
    "lastPage": true,
    "searchResultsMaxLimitHit": false
  }
}
```

We only need the `data.entities` field, just like the result from `entities list`:
```
gcx kg entities list --type Service --env prod-ca-east-0 --namespace asserts --property name=api-server -o json)
{"class":"hint","summary":"use --json list / --json field1,field2 for field selection, or --jq '\u003cexpr\u003e' for transformation (group_by, filter, count) — no external parsing needed"}
[
  {
    "id": 249876,
    "name": "api-server",
    "type": "Service",
    "active": true,
    "scope": {
      "env": "prod-ca-east-0",
      "namespace": "asserts"
    },
    "properties": {

    },
    "assertion": {
    },
    "connectedAssertion": {
      "amend": false
    }
  }
]
```

xujiaxj (dismissed):

xujiaxj (dismissed):

xujiaxj (dismissed):

docs/reference/cli/gcx_kg_entities.md
```diff
@@ -7,10 +7,23 @@ Manage Knowledge Graph entities.
...
+  list       Default for listing and basic lookups: an entity's identity, scope,
+             and properties. Filter to one known entity with
+             '--property name=<name>'. Cheap — use this for plain lookups.
+  correlate  You have a firing alert (its labels) but not the entity → find
+             which entity the alert hangs off. The "I have an alert, which
+             entity is it?" entry point.
+  inspect    Root-cause analysis only — heavy: insight timeline + related
+             entities. Don't use it just to read an entity's properties (use
+             'list'); reach for it only when you need the RCA view.
+  query      Arbitrary Cypher over the graph.
```
annanay25, line 26: i'm not sure we need this? running `gcx kg entities` should print the subtree under it.

annanay25 (approved):
lgtm!

PR #1006, opened 2026-07-20
feat(instrumentation): otel checker explain

maryliag:
Update to the newest version of Otel Checker, which includes the new `explain` command. Add the command on GCX as well.
Example of the existing check command, with the new explain_id column:
Screenshot 2026-07-20 at 12 50 28 PM

Example of the explain command:
Screenshot 2026-07-20 at 12 34 44 PM

Example of list-explanations command:
Screenshot 2026-07-28 at 12 41 51 PM


maryliag:
PR updated with latest feedback to move the list command on its own

zeitlinger:
Nice — the `explain_id` in the results makes the runtime handoff I'm proposing in internal-docs#14 actually implementable, so this is a prerequisite for it.

One thing I'd like to pin down while this is open: are explain IDs intended to be stable across otel-checker versions? I'm planning to persist an ID across a handoff (static finding → runtime diagnosis), and possibly write one into a saved OATS case. If IDs can be renumbered or retired between releases, both of those break silently. Worth documenting the intent either way — even "best effort, no guarantee" is useful to know.

maryliag:
The IDs are best effort, for example, I had some checks specific for Beyla, but they're no longer valid so I had to remove them. Is possible that a similar situation will happen again

dafydd-t:
@claude review

cmd/gcx/instrumentation/explain/command.go
```diff
@@ -0,0 +1,147 @@
...
+func (*docTextCodec) Format() format.Format { return "text" }
+
+func (*docTextCodec) Encode(w io.Writer, v any) error {
+	doc, ok := v.(DocView)
+	if !ok {
+		return fmt.Errorf("docTextCodec: expected DocView, got %T", v)
+	}
+	source := fmt.Sprintf("# %s\n\n%s", doc.Title, doc.Body)
+	if isTerminalWriter(w) {
+		if out, err := glamour.Render(source, "dark"); err == nil {
```
dafydd-t: `glamour.Render`  defaults to an 80 char column width, do you want to set it to 0 (no wrap) to make sure any links don't break, or make it easier to copy & paste stuff? since the doc output has a lot of stuff in it. There's an example [here](https://github.com/grafana/gcx/blob/main/internal/style/help.go#L20) that we use for rendering help docs, in case there are URLs in it

cmd/gcx/instrumentation/explain/command.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (*docTextCodec) Format() format.Format { return "text" }
+
+func (*docTextCodec) Encode(w io.Writer, v any) error {
+	doc, ok := v.(DocView)
+	if !ok {
+		return fmt.Errorf("docTextCodec: expected DocView, got %T", v)
+	}
+	source := fmt.Sprintf("# %s\n\n%s", doc.Title, doc.Body)
+	if isTerminalWriter(w) {
```
dafydd-t: Can we also check `style.IsStylingEnabled()` here? we don't want fancy output if we are asking not to have it

dafydd-t:
response to claude review:


> 4. glamour.Render(source, "dark") is hard-coded to the dark style (command.go:124)

We do this elsewhere, so no need to address here

> 6. explain <id> JSON output emits raw markdown in body

Seems OK to me, as is

The rest of the nits are for you to decide to address or not, but I will approve without them.

cmd/gcx/instrumentation/explain/command.go
maryliag: updated

cmd/gcx/instrumentation/explain/command.go
maryliag: done!

maryliag:
Addressed all inline comments and the items 1, 2, 3 and 5 from Claude (ignoring 4 and 6 per Dafydd's comments)

dafydd-t (approved):

PR #1007, opened 2026-07-20
feat(assistant): expose chatId alongside investigationId in get

CadenMG:
Part of #997 — the **"Expose both identifiers"** bullet.

On v2 stacks, `get` resolves both identifiers but encoded only the snapshot, which carries `investigationId` and not the backing `chatId`. The resolved `chatId` is now injected into the output map — only when absent, so a future server-provided field wins. Output is otherwise the raw snapshot; the v1 fallback (resolve 404 → legacy detail) is unchanged. Help text updated, CLI reference regenerated.

**Tests:** command-level (httptest fake stack) — both identifiers present, server `chatId` not clobbered, nil snapshot doesn't panic, v1 fallback verbatim. Full suite passes; live-verified on a v2 stack. Lint shows only two pre-existing staticcheck findings in `scripts/` (also fail on a clean checkout).


annanay25 (approved):
lgtm, thanks!

PR #1008, opened 2026-07-20
feat(assistant): add investigations evidence command

CadenMG:
Part of #997 - Adds `gcx assistant investigations list-evidence <id>`, backed by `GET /api/v2/investigations/{investigationId}/evidence` — the canonical citation index for an investigation report, mapping each cited panel ID to the tool and query that produced it. Distinct from the chat-derived `tools` command, which is not canonical.

```
$ gcx assistant investigations list-evidence <id>
PANEL  TOOL        QUERY                                     EPOCH  TIME
p5     prometheus  sum by (http_response_status_code) (r...  3      17:48:53Z
p7     loki        {namespace="myapp", container="collec...  4      17:49:09Z
```

## Implementation

- Types mirror the server's `evidenceItem` JSON tags exactly (`toolUseId` is the only `omitempty`), pinned by lossless round-trip + required-vs-omitempty tests.
- v2-only verb gated with `requireV2` + `resolveID`, like pause/resume. Empty evidence (no stored session) renders as an empty table — the server returns `{"evidence": []}`, not 404.
- Default `table` (query truncated at 40 runes, whitespace flattened); `wide` adds TOOL USE ID + full query; `json`/`yaml` lossless.
- Token-cost annotation added; reference docs regenerated.

## Testing

- httptest client tests + table-driven codec tests; full `-race` suite passes.
- Live-verified against a Cloud stack: table/wide/json output, `--json` field selection, and not-found error path.


docs/reference/cli/gcx_assistant_investigations_list-evidence.md
```diff
@@ -0,0 +1,37 @@
+## gcx assistant investigations list-evidence
+
+List panel evidence for a v2 investigation.
+
+### Synopsis
+
+List panel evidence for a v2 investigation — the canonical mapping from report citation keys (panel IDs like p3) to the tool and query that produced each panel. For raw chat-derived tool calls, use `gcx assistant investigations list-tool-calls` instead.
+
+```
+gcx assistant investigations list-evidence <id> [flags]
```
annanay25: can we name the parameter `<investigation-id>` instead? it was a bit confusing where we are referencing evidence by ID.

annanay25 (dismissed):
lgtm, just added a comment on parameter naming.

annanay25 (approved):

PR #1009, opened 2026-07-20
feat(profiles)!: rename profile-types to list-profile-types at both mounts

awsome-o:
This is the first rename from the #387 naming workstream. Dafydd decided the rename on 2026-07-17, and after simonswine's review feedback both maintainers agreed on the `list-profile-types` spelling (2026-07-21) — it matches the `--profile-type` flag, the Pyroscope API (`QuerierService/ProfileTypes`) and profilecli vocabulary, and stays unambiguous if Pyroscope grows other type-like facets later. Since #994 is an alignment record that won't merge, this PR is self-contained: it carries the rename plus the one CONSTITUTION.md clause change it needs.

Migration mapping:

| old | new |
|---|---|
| `gcx profiles profile-types` | `gcx profiles list-profile-types` |
| `gcx datasources pyroscope profile-types` | `gcx datasources pyroscope list-profile-types` |

Both mounts share one command builder, so the rename itself is a single Use-string change plus every place the old path appeared in active text: the help examples, the --profile-type flag help that points at it, the agent llm_hint annotations, the path-keyed cost registry, the error-help suggestion in cmd/gcx/fail, the two bundled dashboard skills, the cli-layer tree diagram, the cross-signal ADR surface diagram, the migration gap analysis row, and the regenerated CLI reference. Per review feedback the command file is also renamed to list_profile_types.go (builder: ListProfileTypesCmd). CHANGELOG and the 2026-04-14 UX plan doc keep the old name on purpose, they're historical records.

Per the pre-v1 clean-start policy there is no deprecated alias and no forwarder. Instead of an alias, the old spelling now gets a clear error (also from review feedback): the command declares SuggestFor, and the unknown-command usage error surfaces it, so `gcx profiles profile-types` fails with exit 2 plus "Did you mean 'gcx profiles list-profile-types'?" in the suggestions. Behavior, flags, positional args, output schema and exit codes of the command itself are all unchanged.

One thing reviewers should look at closely: this PR amends CONSTITUTION.md. The Provider Architecture clause said provider-only resources must use alternative verbs and "never `get`, `list`, `create`, `update`, `delete`", which read literally would have made this rename a documented violation. The amendment narrows the prohibition to the bare verbs and allows an <operation>-<subject> compound that honestly describes the behavior (list-profile-types, list-tables for genuine enumerations). That matches what the tree already ships (clickhouse and athena list-tables, the cloudwatch list-metrics family, and the sub-resource clause already blesses alert-groups list-alerts), and it's the smallest change that makes the decided rename legal. It deliberately does not import the operation-semantics ADR from #994.

While touching the command I also added one sentence to its help text: when gcx auto-discovers the datasource from your Grafana Cloud stack, the discovered datasource UID may be saved to your config. That's existing ResolveAndSaveDatasource behavior, not a change; the help just says so now. The same behavior exists in the other datasource commands — a family-wide disclosure is a recorded follow-up, this PR doesn't expand it.

Note for #1026: that PR adds time-range flags to this same command; whichever lands second rebases, and I'm happy to help with that rebase either way.

Verified locally: built the binary and ran both new paths (help renders with the disclosure at both mounts), confirmed both old paths fail with exit 2 and the did-you-mean suggestion (in both the styled and agent-mode JSON error output), gofmt clean, lint clean, go test -race -count=1 ./... passes, reference-drift passes. mkdocs isn't installed on my machine so the docs build runs in CI.

### DIFF
--- a/docs/reference/cli/gcx_datasources_pyroscope_query.md
+++ b/docs/reference/cli/gcx_datasources_pyroscope_query.md
@@ -75,7 +75,7 @@ gcx datasources pyroscope query [EXPR] [flags]
       --pprof-overwrite               Overwrite the output file if it already exists (only with -o pprof)
       --pprof-path string             Destination path for pprof binary output (only with -o pprof; default: profile-YYYY-MM-DD-HHMMSS.pb.gz)
       --profile-id strings            Drill down to specific profile UUIDs from exemplar queries (repeatable)
-      --profile-type string           Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds'); use 'gcx profiles profile-types' to list available (required)
+      --profile-type string           Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds'); use 'gcx profiles list-profile-types' to list available (required)
       --since string                  Duration before --to, or now if omitted (e.g., 30m, 6h, 7d); mutually exclusive with --from
       --span-id strings               Only query profiles with these 16-character hex span IDs (repeatable; unavailable with -o pprof)
       --stacktrace-selector strings   Only query locations with these function names, starting from the root (repeatable)
--- a/internal/datasources/pyroscope/list_profile_types.go
+++ b/internal/datasources/pyroscope/list_profile_types.go
@@ -32,19 +32,23 @@ func (opts *profileTypesOpts) Validate() error {
 	return opts.IO.Validate()
 }
 
-func ProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
+func ListProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
 	opts := &profileTypesOpts{}
 
 	cmd := &cobra.Command{
-		Use:   "profile-types",
-		Short: "List available profile types",
-		Long:  "List available profile types from a Pyroscope datasource.",
+		Use:        "list-profile-types",
+		SuggestFor: []string{"profile-types"},
+		Short:      "List available profile types",
+		Long: "List available profile types from a Pyroscope datasource.\n\n" +
+			"If gcx auto-discovers the datasource from your Grafana Cloud stack, " +
+			"the discovered datasource UID may be saved to your gcx configuration " +
+			"for future commands.",
 		Example: `
 	# List profile types (use datasource UID, not name)
-	gcx datasources pyroscope profile-types -d UID
+	gcx datasources pyroscope list-profile-types -d UID
 
 	# Output as JSON
-	gcx datasources pyroscope profile-types -d UID -o json`,
+	gcx datasources pyroscope list-profile-types -d UID -o json`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if err := opts.Validate(); err != nil {
 				return err
@@ -81,7 +85,7 @@ func ProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
 
 	cmd.Annotations = map[string]string{
 		agent.AnnotationTokenCost: "small",
-		agent.AnnotationLLMHint:   "gcx datasources pyroscope profile-types -d UID -o json",
+		agent.AnnotationLLMHint:   "gcx datasources pyroscope list-profile-types -d UID -o json",
 	}
 
 	opts.setup(cmd.Flags())
--- a/CONSTITUTION.md
+++ b/CONSTITUTION.md
@@ -123,10 +123,14 @@ agent mode detection, behavior changes, and opt-out mechanisms.
   does not obey standard list/get/create/update/delete semantics (e.g.,
   composite keys, scope-required lookups, query-only endpoints), do not
   register it as an adapter. Keep it in the provider command tree only, but
-  use alternative verbs (`show`, `describe`, `search`) — never `get`, `list`,
-  `create`, `update`, `delete`. This avoids user confusion: adapter verbs
-  (`resources get`) and provider verbs should not overlap for resources that
-  behave differently across the two paths.
+  use alternative verbs (`show`, `describe`, `search`) or an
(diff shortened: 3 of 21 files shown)


internal/datasources/pyroscope/list_profile_types.go
```diff
@@ -36,15 +36,18 @@ func ProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
...
 
 	cmd := &cobra.Command{
-		Use:   "profile-types",
+		Use:   "list-types",
 		Short: "List available profile types",
-		Long:  "List available profile types from a Pyroscope datasource.",
+		Long: "List available profile types from a Pyroscope datasource.\n\n" +
+			"If gcx auto-discovers the datasource from your Grafana Cloud stack, " +
+			"the discovered datasource UID may be saved to your gcx configuration " +
+			"for future commands.",
```
simonswine, line 45: Is this message about the datasource UID helpful here? It probably would need to be on any command

docs/reference/cli/gcx_datasources_pyroscope_query.md
```diff
@@ -75,7 +75,7 @@ gcx datasources pyroscope query [EXPR] [flags]
       --pprof-overwrite               Overwrite the output file if it already exists (only with -o pprof)
       --pprof-path string             Destination path for pprof binary output (only with -o pprof; default: profile-YYYY-MM-DD-HHMMSS.pb.gz)
       --profile-id strings            Drill down to specific profile UUIDs from exemplar queries (repeatable)
-      --profile-type string           Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds'); use 'gcx profiles profile-types' to list available (required)
+      --profile-type string           Profile type ID (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds'); use 'gcx profiles list-types' to list available (required)
```
simonswine: I think this is my only concern with this PR, this flag is no longer consistent with its related list command.

I would prefer to name the command  `gcx profiles list-profile-types`, as this is along with how we call it in the API and the profilecli tool.

Going forward we might have other types like query-type, so I rather no want to rename this flag here to `type`.

internal/datasources/pyroscope/list_profile_types.go
```diff
@@ -36,15 +36,18 @@ func ProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
 	opts := &profileTypesOpts{}
 
 	cmd := &cobra.Command{
```
simonswine, line 38: Maybe worth defining an alias for profile-types or a clear error:

```
profiles-types is gone use $new_command
```

internal/datasources/pyroscope/list_profile_types.go
```diff
@@ -36,15 +36,18 @@ func ProfileTypesCmd(loader *providers.ConfigLoader) *cobra.Command {
 	opts := &profileTypesOpts{}
```
simonswine, line 36: The whole file probably should be renamed to the new command name too,

profile_types.go -> list_(profile_)types.go

simonswine (dismissed):
A couple of thoughts, no strong feelings about it though.

docs/reference/cli/gcx_datasources_pyroscope_query.md
awsome-o: Done — renamed to list-profile-types at both mounts (Dafydd also ack'd this spelling in Slack). The --profile-type flag stays as is, so flag and command vocabulary line up again. Good catch on the future type-like facets, that also avoids clashing with the resources list-types rename coming in #1019.

internal/datasources/pyroscope/list_profile_types.go
awsome-o: Went with the clear error half: the command now declares SuggestFor for the old name, and the unknown-command error includes "Did you mean 'gcx profiles list-profile-types'?" while still failing with exit 2. Skipped the alias on purpose — pre-v1 we're doing clean renames with no aliases/forwarders so the surface converges before GA, and an alias here would set the wrong precedent for the rest of the #387 renames.

internal/datasources/pyroscope/list_profile_types.go
awsome-o: Done — profile_types.go is now list_profile_types.go and the builder is ListProfileTypesCmd. Kept the ProfileTypes* names in internal/query/pyroscope since those are bound to the API path and wire format.

internal/datasources/pyroscope/list_profile_types.go
awsome-o: You're right that it applies much more widely — the same save happens in every datasource command that goes through the shared resolver, roughly 40 of them. Doing that family-wide (probably a shared help snippet appended wherever a command passes a DatasourceSaver) is a recorded follow-up; kept the one-liner here since this PR was already touching this command's help text and the behavior is real.

awsome-o:
@simonswine all four points addressed: renamed to list-profile-types per your suggestion (Dafydd ack'd in Slack), file renamed to list_profile_types.go, and the old spelling now fails with a "Did you mean 'gcx profiles list-profile-types'?" suggestion instead of an alias (details in the thread replies). Force-pushed as a clean 2-commit history, all gates green. PTAL when you get a chance — Dafydd wants to merge today.

awsome-o:
@dafydd-t #1009 is ready for another look. It is rebased on current main and the final command is list-profile-types at both mounts. The old path fails with exit 2 and a clear did-you-mean message, and all CI checks are green. Simon’s concrete naming, file-name, and clear-error comments are implemented. The only thread left open is the broader question about documenting datasource-UID persistence consistently across the command family.

internal/datasources/pyroscope/list_profile_types.go
dafydd-t: generally would probably be better in references, or in the `gcx` skill or somewhere. food for thought

dafydd-t (approved):

Issue #890, opened 2026-06-30
Review gcx config format


dafydd-t:
currently, the `cloud` section of config lives under `contexts`, and users have to set a token for each context, which is annoying.

We should allow a 'global' cloud auth config, so by default, all contexts will attempt to use the same cloud auth config. We will maintain cloud.token type auth config per-stack, to allow stack-scoped CAPs in grfaana.com

We also need to allow multiple cloud "contexts", for non-public environments.

a suggestion from @radiohead on structure:

```
stacks:
  raintank-ops:
    grafana:
      server: https://ops.grafana-ops.net
      oauth-token: "**REDACTED**"
      oauth-refresh-token: "**REDACTED**"
    providers:
      synthetic-monitoring:
       url: https://synthetic-monitoring-api.grafana-ops.net
cloud:
  raintank:
     token: "**REDACTED**"
     api-url: https://grafana-ops.com

contexts:
  ops:
    stack: raintank-ops
    cloud: raintank
```

dafydd-t:
From a discussion with @radiohead , about cloud config specifically

- ServiceAccount auth: when you get an access policy from  grafana.com, you choose a realm for the policy: the whole org, or particular stacks in the org
- OAuth: when the user logs in, they do not choose an org: the token gives them access perform their allowed actions on all their orgs

So these two auth methods operate at different levels, somewhat. We suggest we can encapsulate these differences with a structure like the following:

```yaml
cloud:
  dafyddt: # OAuth-ed cloud  config entry
    oauth-token: "**REDACTED**"
    orgs:  # OAuth auth provides access to all the user's orgs
     - daf1
     - daf2
  daft: # Access Policy config entry, org-based realm
    token: "**REDACTED**"
    orgs:
     - daf1  # Access Policy auth allows only one org
  daft-some-stacks-only: # Access Policy auth, realm is two stacks (not the whole org)
    token: "**REDACTED**"
    orgs:
     - daf1
    stacks: # stacks config only needed when particular stacks are chosen for the AP realm
     - dafstack1
     - dafstack2
```

dafydd-t:
Consolidating the discussion above into a full proposed format, plus answers to the open questions. Verified the auto-discovery story against the gcom codebase.

## Proposed format

```yaml
current-context: ops

resources:                    # global: applies to all stacks
  assume-server-dry-run:
    - alertrules.rules.alerting.grafana.app

stacks:
  raintank-ops:
    grafana:
[...]

Issue #949, opened 2026-07-09
cloud stacks create: stack silently lands in the account's default org — no way to see or choose the target org


awsome-o:
## What happened

Running stack creation with an OAuth token from `gcx cloud login` (experimental flow):

```
gcx cloud login                       # OAuth PKCE, account belongs to multiple orgs
gcx cloud stacks create --name "gcx eval" --slug gcxeval --region us
```

The stack was created successfully — but in an unexpected org. My account is a member of two orgs (my personal org `<my-org>` and a teammate's org `<other-org>`); the stack landed in `<other-org>` because GCOM resolved the account's *default org* server-side. gcx gave no indication of which org it was targeting, before or after.

## Why

- `CreateStackRequest` (internal/cloud/gcom.go) sends only `{name, slug, region, ...}` to `POST /api/instances` — no org field.
- The [public Cloud API](https://grafana.com/docs/grafana-cloud/developer-resources/api-reference/cloud-api/) accepts no `org`/`orgSlug`/`orgId` parameter on this endpoint either — the org is derived from the credential. For Cloud Access Policy tokens that's unambiguous (the token is realm-scoped), but for the OAuth login flow the token is account-wide and GCOM falls back to the default org.
- `gcx cloud stacks list` *requires* `--org`, so the asymmetry is surprising: you must name an org to list, but cannot name one to create.

## Expected

At minimum, `create` should resolve and print the target org (and include it in `--dry-run` output) so the user can abort. Ideally:

1. Add `--org` — worth checking whether the org-scoped path used by list (`/api/orgs/{slug}/instances`) also accepts POST.
2. If the API genuinely can't target an org, fail or warn when the account has >1 org membership and `--org` wasn't given.
3. Mention the default-org behaviour in the command help / `gcx cloud login` docs.

## Impact

A user can unknowingly provision billable infrastructure in someone else's org. Cleanup is also awkward: the OAuth grant in my session came back without `stacks:delete` (requested but not granted), so the misplaced stack couldn't be removed via gcx.

gcx: built from main. Found while dogfooding gcx for the bundled-skills evaluation work.

PR #1012, opened 2026-07-20
feat(config)!: split config into stacks, cloud entries, and contexts

dafydd-t:
Part of #890. This implements the format proposed in https://github.com/grafana/gcx/issues/890#issuecomment-5004314096. The full design and failure model are in `docs/plans/2026-07-20-config-format-rework.md`.

This keeps the original config format work and adds the security, migration, and recovery changes found during GA review.

## Summary

1. This versions the config file as `v1`.

2. It adds named `stacks` and named `cloud` entries that can be shared across thin `contexts`. Contexts keep per-context settings such as datasource defaults.

3. Single-source legacy configs migrate on load. Layered legacy configs are checked and converted in memory first, then migrated one selected layer at a time. Migration writes an exact backup, verifies the converted connection independently, and gives exact recovery commands if it cannot finish safely.

4. Legacy keychain entries needed for rollback are kept. New keychain references are bound to the owning config file, entry, secret field, and destination, so copying a reference into another file or pointing it at another server does not grant access to the credential.

5. Legacy `default-*-datasource` values move into the context `datasources` field. `gcx config set` now uses literal paths such as `stacks.<name>...`; former context-relative paths return a clear error.

6. `GRAFANA_CLOUD_TOKEN`, `GRAFANA_CLOUD_API_URL`, and `GRAFANA_CLOUD_OAUTH_URL` create an in-memory Cloud entry for the current command. `GRAFANA_CLOUD_STACK` only overrides the stack slug and does not change the config file.

7. A context does not need a Cloud reference. Commands that need Cloud credentials fail with recovery guidance when none is configured.

8. Future config versions are rejected before migration, backup creation, keychain access, or network use. `gcx config edit` remains available as a raw recovery path even when the normal loader cannot decode a file.

## Before

```yaml
contexts:
  prod:
    grafana:
      server: https://myorg.grafana.net
      token: "..."
    cloud:
      token: "..."       # repeated in every context
      stack: myorg
  dev:
    grafana:
      server: https://myorg-dev.grafana.net
      token: "..."
    cloud:
      token: "..."       # same token again
current-context: prod
```

## After

```yaml
version: 1
stacks:
  prod:
    slug: myorg
    grafana: {server: "https://myorg.grafana.net", token: "..."}
  dev:
    grafana: {server: "https://myorg-dev.grafana.net", token: "..."}
cloud:
  grafana-com:
    token: "..."         # one entry, shared
contexts:
  prod: {stack: prod, cloud: grafana-com}
  dev: {stack: dev, cloud: grafana-com}
current-context: prod
```

## How to review this PR

1. `docs: add config format rework plan` explains the target shape, decisions, migration model, and resolved questions.

2. `feat(config): split config into stacks, cloud entries, and contexts` contains the original format change. A useful reading order is `internal/config/types.go`, `internal/config/migrate.go`, `internal/config/keychain.go`, then the consumers under `internal/login`, `internal/providers`, and `cmd/gcx`.

3. The following config and documentation commits refine layering, Cloud entries, OAuth fields, migration guidance, and command paths.

4. `feat(config): harden credential routing and migration` contains the source-bound keychain trust model and the first security review fixes.

5. `fix(config): stabilize credential routing and migration for GA` covers safe write ownership, wrong-account prevention, migration recovery, rejected credentials, and regression tests.

6. `fix(config): close remaining GA readiness gaps` completes destination-bound credential reuse, removes unsafe mixed-layer guidance, gives `config check` a real failure status, and makes fallback diagnostics unconditional.

7. `fix(config): close final GA readiness gaps` normalizes schemeless servers, detaches runtime environment overlays, makes fresh-config TLS/proxy recovery actionable, consolidates layered migration guidance, and accepts blank issuer expiry metadata without weakening OAuth locking or generation CAS.

8. `fix(config): surface plaintext keychain fallback` makes the already-supported no-keychain fallback visible at default verbosity only after a durable config commit, without exposing the credential or duplicating the warning under verbose logging.

## GA stabilization

1. `--config` now travels through one central command context. Resource reads, writes, deletes, lazy provider adapters, and provider write-back all use the same selected file. This closes the case where a command displayed or deleted resources from a different account than the one named by `--config`.

2. Keychain references now include source, owner, field, destination, and generation checks. New writes are transactional. A known unavailable keychain can fall back to plaintext only for a new credential and with a request-scoped warning that is visible at default verbosity only after the durable config commit. Rotations, deletions, oversized secrets, and unexpected backend errors still fail closed.

3. Login and provider write-back update a file only when the owner is unambiguous. Layered configurations require `--config` when ownership cannot be proved. Environment values and merged system, user, or repository values are not flattened into a target file.

4. Login changes to shared Cloud credentials use copy-on-write, so logging in to one context does not silently change every context that shared the old entry. Cloud Access Policy and OAuth credentials stay distinct, and OAuth metadata is retained when it is available.

5. `grafana.auth-method` is authoritative. Partial, rejected, or blank selected credentials fail before a network request instead of falling through to anonymous access, empty Basic auth, or a stale credential from another method.

6. Migration performs a side-effect-free layered preflight, compares against an independently decoded legacy baseline, handles duplicate names correctly, writes durable mode `0600` backups, and leaves exact repair commands when migration cannot proceed.

7. `gcx config edit` selects and opens the raw file without decoding it. It can repair malformed YAML, unsupported future versions, interrupted migrations, and semantic failures that correctly block normal commands.

8. Credential reuse now enforces the same complete server, proxy, and TLS destination binding as ordinary config loading. Copied configs and runtime destination changes fail before credential use.

9. Schemeless login destinations normalize before context selection and override decisions. Runtime `GRAFANA_*` overlays detach the selected stack, so an override cannot mutate or affect sibling contexts that share persisted entries.

10. Runtime-only TLS/proxy bearer login remains fail closed when it cannot persist the destination safely, but now provides exact `config set` commands that can initialize a fresh explicit config.

11. Layered migration reports one consolidated warning with every exact migration and recovery command. Blank OAuth expiry metadata is treated as optional while refresh locking, persistence, generation CAS, and rotation safety remain fail closed.

## Tested

1. `GCX_AGENT_MODE=false mise run all` passed on the final rebased head (`a8656a07`). This includes lint, all normal Go tests, the production build, reference regeneration and drift checks, and the complete MkDocs build.

2. `go test -race -count=1 ./internal/auth ./internal/config ./internal/login ./cmd/gcx/config ./cmd/gcx/login` passed on that exact head.

3. The full `go test -race -count=1 ./...` suite passed during the stabilization run before the conflict-free rebase; range-diff confirmed the PR patches remained equivalent.

4. Built-binary E2E, with isolated home and XDG directories plus real local HTTP/TLS servers, covered named and unnamed schemeless TLS login, shared-stack environment isolation, fresh-config TLS recovery, and layered legacy migration. Persisted source files remained byte-identical in the non-migration cases, and the migration diagnostic was emitted once with exact commands.

5. OAuth regressions cover blank issuer expiries, cross-invocation persistence, refresh locking, owner disappearance, generation CAS, and rotated-token persistence failure.

6. A disposable-stack canary passed against a real non-production Grafana stack and the macOS keychain using isolated config copies. Legacy and v1 plaintext credentials migrated to bound keychain references; a fresh process authenticated and listed real resources; env-token login persisted and was reusable; copied-config and destination-override attempts failed before credential use; and headless env-only auth remained non-persistent. The original config stayed byte-identical and temporary keychain entries were removed.

7. A real browser-based instance OAuth canary passed on the exact final head with a read-only grant. Login completed through the real proxy with valid RFC3339 expiry metadata; two consecutive refreshes were forced with isolated discovery caches; each refresh persisted a new credential generation; separate fresh processes reloaded both generations; and a protected resource read succeeded after the second persisted refresh. The server-side OAuth connection and all isolated local artifacts were removed afterward.

8. The raw recovery path was exercised with a config version that normal loading rejects. `gcx config edit` still opened it successfully.

9. Fresh PR CI passed on `906a881a`, including Tests, Linters, Documentation/reference drift, preview build/deploy/link checks, and the security scans.

10. On `906a881a`, `GCX_AGENT_MODE=false mise run gate` passed (full normal Go suite, all 44 linter-rule tests, lint, and build). Reference generation produced no tracked diff, the pinned MkDocs build passed separately, and the focused new warning tests passed normally and with the race detector. An independent code review found no substantive GA issue.

11. A network-disabled Alpine canary with no Secret Service backend exercised the pushed Linux binary: the warning appeared exactly once on stderr, the secret never appeared in output, the deliberate plaintext fallback remained mode `0600`, default config view redacted it, and no keychain sentinel was written.

12. A temporary stack-scoped Cloud Access Policy with only `stacks:read` exercised a live Cloud stack read on the immediately preceding PR head (the final commit changes only local warning delivery and tests): exit zero, expected stack match, nonzero ID, byte-identical config, and no credential in output. The token and policy were revoked immediately afterward.

The three deliberately fail-closed OAuth refresh behaviors remain unchanged. The config rework moves CAP storage and selection into named Cloud entries; downstream GCOM CAP transport semantics were not changed.

## Remaining before public GA

1. ADR-022 needs explicit owner ratification.

2. Recommended, non-blocking release-matrix evidence remains for Linux with a native Secret Service backend and Windows Credential Manager. Headless Linux fallback and the live CAP path are now exercised. This is platform coverage, not a known code defect or a blocker to opening/merging this PR.

## High-ROI follow-ups after GA

1. Add a permanent built-binary config and authentication smoke matrix to CI. This would exercise cross-package command paths as one process, especially migration, login ownership, keychain fallback, divergent `--config` accounts, Linux Secret Service, and Windows Credential Manager.

2. Document downgrade and re-login behavior for binaries that predate keychain sentinels.

3. Investigate bounded timeout behavior for macOS keychain access in stripped or sessionless environments.

4. Remove the duplicate layered-migration warning on write commands that intentionally load twice.

5. Define backward-compatible relative TLS file path behavior. Any change must preserve existing keychain binding digests or provide an explicit reauthentication path.

6. Add owner-aware OAuth refresh persistence to `gcx dev serve` if users need it. It currently fails closed with guidance because consuming a refresh token without safely persisting it would be unsafe.

7. Add a one-command resumable transaction across several legacy config files if migration volume justifies it. The current flow is safe and recoverable, but completing several layers can require more than one command.

8. Clean up historical orphaned keychain generations after exceptional failures and make the remaining reference-cleanup command missing-safe.

## Governance

1. `CONSTITUTION.md` remains unchanged.

2. ADR-022 is proposed pending explicit owner ratification. Existing accepted ADRs remain accepted and point to ADR-022 as a proposed amendment.

## Not included

1. Realm discovery during `gcx cloud login`, including automatic population of organizations and stacks, remains planned for PR 3.


dafydd-t:
added a published migration guide in docs/sources (`migrate-configuration.md`) with the manual legacy → v1 field mapping, and wired it into the migrator: the success notice, migration-failure errors, and the in-memory-migration warnings all point at it. The URL goes live when the docs sync after merge.

dafydd-t:
two more changes from the #890 discussion (https://github.com/grafana/gcx/issues/890#issuecomment-5029465359), one point at a time:

**stack/cloud entries are now atomic across config layers.** Same-named entries replace wholesale instead of merging field-by-field, so a credential and its destination always come from the same file - a repo-local `.gcx.yaml` can no longer graft its `server`/`api-url` onto an entry whose token lives in your user config. A hostile layer can only shadow an entry (break it), not combine with it. This mattered more after migration made entry names predictable (`grafana-com`). Contexts keep field-level merging: they carry refs and datasource defaults, no secrets or destinations. Token-refresh persistence now targets the layer that owns the effective stack entry for the same reason, and the diagnostics pre-read is legacy-aware so `telemetry: disabled` is honoured on the run that migrates.

**OAuth cloud tokens now live in `oauth-token`**, with `oauth-token-expires-at` populated when the token response reports a lifetime. Readers prefer `token` (CAP) and fall back to `oauth-token`; an expired OAuth token errors with "run `gcx cloud login`" instead of a raw 401. Setting one credential clears the other, so switching auth methods leaves nothing stale. This resolves the "fields nothing reads" point - both fields are now written and read.

dafydd-t:
third point from the #890 discussion: dropped `orgs`/`stacks` from cloud entries. They were login-time snapshots, not authoritative config, and the empty `/api/orgs` signal is ambiguous (stack realm vs missing scope vs transient failure) - auto-filling `stacks` on it could freeze a snapshot for an org-realm token. Config now holds user assertions only; PR 3 becomes runtime discovery with an explicit unknown state plus a guard on org-ambiguous mutations (#949 is the motivating case - a cached org list wouldn't have caught it since gcx can't see which org is the default). Plan updated with the redesign.

dafydd-t:
naming/identities point from the #890 discussion: `gcx cloud login` now suffixes the host-named entry when it would collide with a different credential (`grafana-com-<context>`), matching migration's convention - a stack-scoped CAP login can no longer quietly replace the org-wide entry other contexts share. Same-credential logins still dedup. On the product question: one credential per stack entry is now stated as the intended v1 model in the published configuration doc and the architecture doc (two identities against one stack = two stack entries; credential indirection can be added later without breaking v1 files).

dafydd-t:
path grammar change after discussion: `gcx config set` paths are now **literal** - no bare-path routing through the context's stack ref. The path you type is exactly the path `gcx config view` shows, and bare/legacy paths error with the absolute path spelled out from the current context (e.g. `grafana.server` → "use stacks.dev.grafana.server"). Removes the ownership mini-DSL and the silent shared-stack edit surprise.

awsome-o:
## GA stabilization update

The PR branch is now rebased onto current `main` at `9793d49d`. Dafydd's original format work is unchanged in scope; the two final commits keep the hardening and GA fixes separate for review.

### What improved

1. `--config` now selects the same account for resource reads, writes, deletes, provider adapters, and write back.

2. Keychain references are bound to their source, owner, field, and destination. Credential writes are transactional and unexpected backend errors remain fail closed.

3. Login and provider write back only change a file when ownership is clear. Shared Cloud credentials use copy on write during login.

4. Migration preflights every layer, verifies against an independent baseline, writes exact durable backups, and gives a raw `gcx config edit` recovery path.

5. Rejected, blank, or partial credentials stop before network use. The selected Grafana auth method is authoritative across direct consumers too.

### Verified

1. The full race suite passed after the rebase.

2. The repository test task passed, including all 44 linter rule tests.

3. Lint passed with zero issues and the production build passed.

4. Reference generation passed twice with identical output.

5. Focused regressions cover divergent `resources --config` accounts, layered migration recovery, ambiguous login ownership, moved multi context configs, keychain failure paths, shared Cloud entries, OAuth edge cases, rejected credentials, unsupported versions, and raw edit recovery.

The review environment does not have `mkdocs`, so PR CI remains the check for the complete site build.

### Still needed before GA

1. Let PR CI finish.

2. Run a short canary with the built binary, a real nonproduction stack, and a copy of a real legacy config.

3. Give `docs/reference/login.md` a final editorial read.

### Practical follow ups

1. Highest return: add a permanent built binary config and auth smoke matrix to CI.

2. Medium return: define backward compatible relative TLS path behavior.

3. Demand driven: add owner aware OAuth refresh persistence to `gcx dev serve` and a one command resumable multilayer migrator.

4. Conditional: tighten unchanged OAuth refresh token handling only if the proxy contract guarantees rotation.

5. Low urgency: historical keychain orphan cleanup and the remaining reference task idempotence cleanup.

The Constitution is unchanged. ADR 022 is proposed until it receives explicit owner ratification.

awsome-o:
## Final config GA verification

This is now the authoritative handoff and supersedes the earlier GA status snapshots.

The PR remains based on `main` at `817e7a83` and is pushed at `906a881a`.

### Final high-ROI fix

The remaining headless-Linux defect was narrow but worth fixing before GA: when no native keychain is available, a new credential can deliberately fall back to a mode-`0600` config file, but the existing structured warning was hidden at the default log level. The fallback itself worked; users were not being told clearly enough that encrypted storage was unavailable.

The fix sends one request-scoped warning to stderr at default verbosity, only after the config commit is durable. It does not echo the credential, does not duplicate under verbose logging, retains structured logging for non-command callers, and does not claim success on a failed commit. Regression tests cover both human and agent modes, and an independent review found no substantive GA issue.

### Verification

- Fresh CI is green on `906a881a`: Tests, Linters, Documentation/reference drift, preview build/deploy/link checks, Socket, TruffleHog, Semgrep, and zizmor all passed.
- `GCX_AGENT_MODE=false mise run gate` passed locally on the same head: full normal Go suite, all 44 linter-rule tests, lint, and production build. Reference generation produced no diff; the pinned MkDocs build passed separately.
- Focused warning regressions passed normally and with the race detector.
- A network-disabled Alpine canary with no Secret Service backend passed: warning exactly once on stderr, no secret in output, deliberate plaintext fallback preserved, file mode `0600`, default view redacted, and no keychain sentinel.
- A temporary stack-scoped access policy with only `stacks:read` passed a live Cloud read: exit zero, expected stack match, nonzero ID, byte-identical config, and no credential in output. The token and policy were revoked immediately afterward.
- The earlier real-stack/macOS-keychain, browser OAuth login plus two forced refresh/persistence cycles, released-version rollback/migration, 12-process concurrent migration, copied-sentinel rejection, destination-binding, environment-isolation, malformed/future-config recovery, symlink-authority, and local HTTP/TLS canaries still apply. The final commit only changes warning delivery and its tests.
- Temporary Cloud credentials and server-side OAuth access were removed; the original user config remained byte-identical.

### Review state

There are still no review submissions or inline review threads to resolve. The earlier design comments remain valid. Only their old progress snapshots were superseded, and the PR body now reflects the pushed head and completed evidence.

### Remaining before public GA

1. Owner ratification of ADR-022 remains the governance gate.
2. Linux with a native Secret Service backend and Windows Credential Manager are still recommended release-matrix smokes. This is platform coverage, not a known code defect or a blocker to opening/merging this PR. Headless Linux and the live CAP path are now exercised.
3. The highest-ROI follow-up remains a permanent built-binary config/auth smoke matrix in CI so this evidence is reproducible after future changes.

### Verdict

I found no remaining GA-blocking code defect. The config/auth/migration change is ready for human review; the PR intentionally remains draft for the owner to mark ready.

dafydd-t:
I manually tested these scenarios, and picked up a couple of presentation bugs which have now been committed. Only tested on macOS with keychain

1. Real-config migration — Converts your existing config correctly, loses nothing, no plaintext secrets on disk, keeps a backup, doesn't re-run. Keychain entries are copied, never moved.

2. Rollback — One file-copy restores a fully working setup on the old binary; the backup never gets overwritten. Old binary + new file fails loudly, not quietly.

3a. Read-only config — Unwritable file (CI): gcx converts in memory each run, never touches the file, warns.

3b. Locked keychain — Can't read your secrets → migration waits and retries later, rather than committing one it couldn't carry credentials into.

3c. Dangling refs — A context pointing at a nonexistent stack is named precisely by config check.

3d. Missing cloud binding — No cloud credentials on a context → clear how-to-fix error at use time, not a raw 401.

3e. Expired OAuth token — Says "run gcx cloud login" instead of a cryptic auth failure.

3f. Layering security — A repo's .gcx.yaml cannot pair its server/URL with your personal token; proved the token never left toward the repo's address. Hostile repos can only break things visibly.

3g. Layered migration — Multi-file configs are never half-migrated silently; you get one explicit command per file, backup each. Repo-local credentials get their own entry name.

3h. Garbage input — Broken configs give clear parse errors; nothing is migrated, written, or backed up.

4. Login flows — Fresh login builds the new structure; re-login preserves settings; a different cloud token gets its own entry instead of replacing the shared one; same token still shares; OAuth tokens land in the OAuth field.

5. Paths + env vars — config set paths are literal; old paths fail with the exact replacement to copy-paste; env-var credentials work but are never written to the file.

annanay25 (approved):
lgtm! manually tested and covers most use cases. nice work!

PR #1013, opened 2026-07-20
feat(agento11y)!: naming convergence batch (#387)

awsome-o:
This is the agento11y batch of the #387 naming workstream, one of the first two provider-sized batches (the datasources one is going up alongside it). Ten renames, all clean breaks per the pre-v1 policy: no aliases, no forwarders, old paths fail as unknown commands. Ready for review; merge #1009 first because it supplies the CONSTITUTION allowance for these list-* compounds.

Every record came from the classification worksheet, and each one was re-verified against the code before implementation (result cardinality, whose ID the positional carries, side effects), then the finished diff went through an adversarial three-lens review. Behavior, flags, positionals, output schemas and exit codes are unchanged on every command; only names moved.

Migration mapping, with the observed behavior behind each rename:

| old | new | why |
|---|---|---|
| `gcx agento11y agents versions <name>` | `gcx agento11y agents list-versions <name>` | returns []AgentVersion for one agent; the positional is the agent's name and versions have no ID of their own (the only read-one is `agents get --version`) |
| `gcx agento11y collections conversations list <id>` | `gcx agento11y collections list-conversations <id>` | lists the members of one collection; the positional is the collection's ID |
| `gcx agento11y collections conversations add <id> <saved-id>...` | `gcx agento11y collections add-conversations <id> <saved-id>...` | membership add (POST AddMembers); variadic saved-ids preserved |
| `gcx agento11y collections conversations remove <id> <saved-id>` | `gcx agento11y collections remove-conversation <id> <saved-id>` | membership unlink of exactly one saved conversation; the saved conversation itself survives |
| `gcx agento11y evaluators create` | `gcx agento11y evaluators upsert` | the POST is create-or-update (the client doc says so, and the adapter's UpdateFn calls the same Create) — the same shape as the alert templates decision, D6 |
| `gcx agento11y experiments scores <run-id>` | `gcx agento11y experiments list-scores <run-id>` | lists the scores of one run; this is the ADR §8 worked example, confirmed in code |
| `gcx agento11y judge models --provider <id>` | `gcx agento11y judge list-models --provider <id>` | flag-scoped catalog, no per-model operations anywhere |
| `gcx agento11y judge providers` | `gcx agento11y judge list-providers` | capability catalog, no read-one anywhere |
| `gcx agento11y scores list <generation-id>` | `gcx agento11y generations list-scores <generation-id>` | the positional is a generation's ID, so the command belongs under generations (which has its own get); the one-command scores group dissolves |
| `gcx agento11y templates versions <id>` | `gcx agento11y templates list-versions <id>` | same shape as agents versions |

Two groups dissolve as a consequence: `collections conversations` (its three children become compounds directly under collections) and the top-level `scores` group. Both old group paths now fail as unknown commands.

Kept, with explicit dispositions — these were the two human-review records in the slice:

- `saved-conversations collections` stays. A three-judge panel split 2-1 (the dissent: collections are independently addressable via their own group, which cuts against a list-collections compound), and the rule for this workstream is that a non-unanimous panel means keep. A keep is reversible; a wrong rename pre-GA is not.
- `saved-conversations save` stays. Unanimous: save is honest for the bookmark-from-a-live-conversation workflow, and renaming it to create would promise absence semantics we can't show the API honors (the client accepts both 200 and 201).

There are no alias records for agento11y.

Two things worth a close look in review:

- One sanctioned help-text fix rides along, flagged in the worksheet: the evaluators upsert example invoked `gcx agento11y templates show`, which doesn't exist; it now says `templates get`.
- The bundled skills reference these commands, so claude-plugin/skills/agento11y and agento11y-prod-setup are updated in the same commit (the skillsdrift test enforces that). The skill frontmatter still says "manage evaluators (create, test, delete)" — that's describing user intent rather than a command path, so I left it; happy to reword if you'd rather.

Verified locally: built the binary and ran every new path (--help renders with the right positionals and flags), every old path fails as unknown command, and spot-checked unrelated siblings (`agento11y conversations list` is a different command — live conversations — and is untouched). gofmt clean, lint clean, go test -race -count=1 ./... green, reference docs regenerated with GCX_AGENT_MODE=false and drift-clean. mkdocs isn't installed locally so the docs build runs in CI.

Post-census note (2026-07-21): #1029 merged two new conversation commands into main after this batch's census — `conversations list-annotations <conversation-id>` and `conversations annotate <conversation-id>`. Both are keeps: `list-annotations` is exactly the parent-scoped compound this workstream prescribes, and `annotate` is an honest domain action (recording it for the eventual operation registry). No rename or migration row needed; the rebase preserves their command annotations.


awsome-o:
External review caught two things the batch review missed, both fixed now:

- 5d1c6a25: the prod-setup skill's step-1 tip still said "agents get / versions" — the phrase was split across a line break, which is why the token sweep missed it. Now says agents list-versions. While in the file I also named the evaluators write op explicitly in the write warning (upsert evaluators, create/update rules and guards).
- f8800a4e: evaluators upsert printed "Evaluator X created" even when the POST replaced an existing evaluator — the exact misrepresentation the rename is removing. It now says upserted. I kept this out of the rename commit on purpose (the migration safety gates say behavior-text changes ride separately): it touches only the stderr status line, stdout payload, exit codes and flags are unchanged, and no test asserted the old string. If you'd rather it ride its own PR, say so and I'll split it.

Gates re-ran after both: lint 0 issues, full go test -race -count=1 ./... green (141 packages), binary rebuilt and upsert re-checked.

Looked at and left alone: the skill frontmatter still says users may want to "manage evaluators (create, test, delete)" — that's describing user intent, not a command path.

awsome-o:
Correction to my earlier note: a second-pass review (fresh readers going through the skill files end-to-end instead of grepping) changed my mind on the frontmatter. The description renders at runtime in the gcx agento11y --help related-skills trailer, and "(create, test, delete)" there reads as an operations list, not intent phrasing — an agent planning from the description alone would type evaluators create. Fixed in 39ef85d3 along with the two command-groups table rows that pattern-matched the same way ("view versions", "inspect scores" -> the real list-versions / list-scores compounds). The "create evaluator" trigger phrases on the next line stay — those genuinely are user intent. Root/skills/provider tests re-ran green.

awsome-o:
Ran a live end-to-end comparison against a real Grafana Cloud stack (old command on an origin/main binary vs new command on this branch's binary, same args, sandboxed config): agents list-versions returned the real version history byte-identical to agents versions; generations list-scores round-tripped the scores endpoint for a real generation identically to scores list; judge list-providers/list-models, experiments list-scores and collections list-conversations matched on their live responses and error identities; the three write commands (upsert, add-conversations, remove-conversation) matched on their pre-API argument-validation paths (not run against the tenant). 10/10 pairs: identical stdout, identical exit codes, stderr identical modulo the renamed token.

awsome-o:
Third-pass catch, fixed in a73a7cda: the Step 5 hard write gate in the prod-setup skill ("never run them before the developer's explicit yes") still enumerated the write commands as create/update, omitting upsert — the command that now writes evaluators. It reads upsert/create/update now. I swept every other write-op enumeration in the file this time (lines 61, 211-212, 287) and they were already correct. Root and skills test packages re-ran green.

dafydd-t (approved):

PR #1014, opened 2026-07-20
feat(datasources)!: naming convergence batch (#387)

awsome-o:
This is the datasources batch of the #387 naming workstream, going up alongside the agento11y batch. Four renames, clean breaks per the pre-v1 policy: no aliases, no forwarders, old paths fail as unknown commands. Ready for review; merge after #1009, which supplies the CONSTITUTION allowance for the list-* compounds.

Each worksheet record was re-verified against the code before implementation, and the finished diff went through an adversarial three-lens review, which came back clean. Behavior, flags, positionals, output schemas, error identity and exit codes are unchanged everywhere; only names moved.

| old | new | why |
|---|---|---|
| `gcx datasources influxdb field-keys` | `gcx datasources influxdb list-field-keys` | datasource-scoped discovery facet; field keys aren't in the closed signal-shorthand set |
| `gcx datasources influxdb measurements` | `gcx datasources influxdb list-measurements` | same — discovery facet with no ID of its own |
| `gcx datasources influxdb tag-keys` | `gcx datasources influxdb list-tag-keys` | same |
| `gcx datasources influxdb tag-values --key <k>` | `gcx datasources influxdb list-tag-values --key <k>` | values of one tag key (flag-supplied identity), not independently addressable |

athena describe-table keeps its name (see dispositions below) but still carries the worksheet-flagged help honesty fix: the old Short/Long claimed "column details including name and type" while the command only ever returned names; the help now says what it does. clickhouse describe-table is deliberately untouched — it returns a genuine schema (name, type, default, comment) and defensibly keeps describe.

Kept, with explicit dispositions:

- `gcx datasources athena describe-table` stays. The first revision of this PR renamed it to `list-columns` (the worksheet's suggestion — the /columns endpoint returns column names only, nothing schema-like); Dafydd's review kept `describe-table`. The rename was reverted; only the help-text honesty fix remains. Disposition recorded as keep, approved by dafydd-t.
- `gcx datasources health` stays. The first revision of this PR renamed it to `status` (the worksheet's suggestion); Dafydd's review kept `health` — the command executes Grafana's product-defined datasource health check (the documented operation is "Check data source health"), so health is the honest name in the product's own vocabulary. The rename was reverted commit-for-commit (health.go and its test restored byte-identical to main); disposition recorded as keep, approved by dafydd-t. Whether `health` gets a global registry entry or a scoped exception is decided in the registry PR at the end of the rollout.
- `pyroscope exemplars profile` and `exemplars span` (the two human-review records in the slice) stay. Unanimous panel: several defensible shapes exist (kind flag vs compounds, list vs query given the required selector), and the exemplars builder is mounted at both `gcx datasources pyroscope` and `gcx profiles`, so any rename has to converge with the profiles surface. Keeping is the explicit disposition for this batch.
- `pyroscope profile-types` → `list-profile-types` is not in this PR: that rename is decided (D4c, respelled in the #1009 reroll after review) and carried by #1009 at both mounts of its shared builder.
- The two tempo aliases keep-synonym: `labels` ⇢ `tags` and `query` ⇢ `search` are the declared permanent synonyms from the cross-signal ADR, and the operation-semantics ADR cites them as the worked example of an approved alias disposition.

Verified locally: built the binary, ran every new path (--help renders with the same positionals and flags), every old path fails as unknown command, and spot-checked siblings (clickhouse describe-table, pyroscope exemplars and the profile-types area untouched (its rename rides #1009), datasources health and athena describe-table restored and working — `list-columns` now fails as an unknown command). gofmt clean, lint clean, go test -race -count=1 ./... green, reference docs regenerated with GCX_AGENT_MODE=false and drift-clean. Also ran a live end-to-end comparison against a real Grafana Cloud stack (old command on an origin/main binary vs new command on this branch's binary, same args) — see the comments; with the health and athena reverts the live comparison stands at 4/4 renames identical. mkdocs isn't installed locally so the docs build runs in CI.

### DIFF
--- a/docs/reference/cli/gcx_datasources_status.md
+++ b/docs/reference/cli/gcx_datasources_status.md
@@ -1,4 +1,4 @@
-## gcx datasources health
+## gcx datasources status
 
 Check the health of one or more datasources
 
@@ -15,7 +15,7 @@ Exit codes distinguish resource failure from command failure:
   1/2/3 - the check could not run (operational, usage, or auth failure)
 
 ```
-gcx datasources health [UID] [flags]
+gcx datasources status [UID] [flags]
 ```
 
 ### Examples
@@ -23,13 +23,13 @@ gcx datasources health [UID] [flags]
 ```
 
 	# Check a single datasource
-	gcx datasources health my-ds-uid
+	gcx datasources status my-ds-uid
 
 	# Check all datasources
-	gcx datasources health
+	gcx datasources status
 
 	# Check all datasources of a given type
-	gcx datasources health --type grafana-sentry-datasource
+	gcx datasources status --type grafana-sentry-datasource
 ```
 
 ### Options
@@ -37,7 +37,7 @@ gcx datasources health [UID] [flags]
 ```
       --config string    Path to the configuration file to use
       --context string   Name of the context to use
-  -h, --help             help for health
+  -h, --help             help for status
       --jq string        jq expression to apply to JSON output. Mutually exclusive with --json.
       --json string      Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
   -o, --output string    Output format. One of: agents, json, table, yaml (default "table")
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,7 +131,7 @@ internal/
 ├── datasources/ Datasource HTTP client, DatasourceProvider interface + registry
 │   ├── clickhouse/  ClickHouse datasource commands (query, list-tables, describe-table, explore)
 │   ├── cloudwatch/  CloudWatch CLI commands (query, list-namespaces, list-metrics, list-dimensions, list-regions, list-accounts)
-│   ├── influxdb/  InfluxDB datasource command layer (query, field-keys, measurements)
+│   ├── influxdb/  InfluxDB datasource command layer (query, list-field-keys, list-measurements)
 │   └── query/   Shared query CLI utils (time parsing, codecs, opts, resolve helpers — used by signal providers and GenericCmd)
 ├── query/       Datasource query clients
 │   ├── dataframe/   Shared Grafana data frame wire types for unified datasource query API responses
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,7 +131,7 @@ internal/
 ├── datasources/ Datasource HTTP client, DatasourceProvider interface + registry
 │   ├── clickhouse/  ClickHouse datasource commands (query, list-tables, describe-table, explore)
 │   ├── cloudwatch/  CloudWatch CLI commands (query, list-namespaces, list-metrics, list-dimensions, list-regions, list-accounts)
-│   ├── influxdb/  InfluxDB datasource command layer (query, list-field-keys, list-measurements)
+│   ├── influxdb/  InfluxDB datasource command layer (query, list-field-keys, list-measurements, list-tag-keys, list-tag-values)
 │   └── query/   Shared query CLI utils (time parsing, codecs, opts, resolve helpers — used by signal providers and GenericCmd)
 ├── query/       Datasource query clients
 │   ├── dataframe/   Shared Grafana data frame wire types for unified datasource query API responses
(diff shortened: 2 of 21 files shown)


awsome-o:
External review caught one straggler: the canonical gcx skill's datasource-lifecycle list still taught health [UID] — a bare token without the gcx datasources prefix, so both the skillsdrift test and the phrase greps missed it. Fixed in 97edb901 to status [UID]. I kept the relative-token format of the surrounding list (create -f FILE, delete UID..., schemas get) rather than switching one entry to a full path; the drift test doesn't parse relative entries either way, so consistency won. Lint and the full race suite re-ran green.

awsome-o:
One more polish from the second-pass review, in 7c58dbfb: the influxdb package-map line in AGENTS.md carried over the old two-command sampling; it now names all four list-* subcommands. The same pass confirmed no surviving stale references to any of the six renamed paths and clean runtime output on all of them. It also turned up one pre-existing defect unrelated to this batch (claude-plugin/agents/grafana-debugger.md teaches gcx resources list, which has never existed) — leaving that out of here, flagged separately.

awsome-o:
Ran a live end-to-end comparison against a real Grafana Cloud stack (old command on an origin/main binary vs new command on this branch's binary, same args, sandboxed config): datasources status did a real health sweep and hit the exit-4 partial-failure path for real (the stack has two unhealthy datasources) — identical rows and exit code to datasources health; the four influxdb list-* commands and athena list-columns matched their old spellings exactly on resolution-error identity and exit codes (no influx/athena datasources on this stack, which is itself the §3a-relevant error path). 6/6 pairs identical. Also confirmed the sandboxed config picked up no hidden datasource-UID write.

docs/reference/cli/gcx_datasources_status.md
```diff
@@ -1,4 +1,4 @@
-## gcx datasources health
+## gcx datasources status
```
dafydd-t: we need to keep `health` here, as it maps directly to the API: https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/api-legacy/data_source/#check-data-source-health

dafydd-t (changes requested):
all LGTM except the `health` => `status` rename

awsome-o:
Addressed the review: gcx datasources health is restored, in 5ae1ae38. The revert covers everything the rename touched — health.go and its test in crud_test.go restored byte-identical to main, the gcx skill line reverted, and the CLI reference regenerated (gcx_datasources_status.md removed, gcx_datasources_health.md back). The branch is also rebased onto current main per the PR checklist. Gates re-ran green: lint 0 issues, full race suite 141 packages ok, reference-drift clean; verified with the built binary that health works again (including a live check against a real stack), status fails as unknown command, and the five surviving renames are intact.

This supersedes my earlier comments where they mention the status rename: the migration table is now five renames, and the live old-vs-new E2E comparison stands at 5/5 (health no longer needs one — it's byte-identical to main, smoke-checked live). PR body updated to match; health is recorded in the kept section as an owner-approved keep.

awsome-o:
Re-ran the live E2E comparison on the exact revision now on this PR (5ae1ae38, rebased) against a binary built from current main: 7/7 identical. health ran the full live sweep on both binaries and hit the exit-4 partial-failure path for real with byte-identical rows — the restore behaves exactly like main. status fails as unknown command identically on both (never existed on main, removed here). The five surviving renames matched on stdout, exit codes and error identity, and the sandboxed config confirmed no hidden datasource-UID write.

awsome-o:
Rebased the four datasource commits directly onto current main. The accidental duplicate #1029 patch is gone, so this is back to the intended 14-file datasource diff. The health keep from @dafydd-t’s review remains intact, and the stale commit message now says list-profile-types. The full local code gate and reference generation pass; CI is rerunning. @dafydd-t, re-requesting review on the corrected head.

dafydd-t:
i put `describe-table` back because its familiar from DB commands

dafydd-t (approved):

soniaAguilarPeiron (approved):

PR #1016, opened 2026-07-21
feat(assistant)!: naming convergence batch (#387)

awsome-o:
This is the assistant batch of the #387 naming workstream, wave 2 — and it grew a second, bigger change in review, so read this section first: on top of the renames, a follow-up commit removes six investigation commands whose backends don't exist for v2 investigations. That makes this PR part rename batch, part surface reduction, and the reduction needs an explicit nod from the assistant owner before this merges.

Commit 1 — the naming batch (what the workstream prescribed). Renames verified against the code with adversarial passes; behavior, flags, positionals, output schemas and exit codes unchanged; the list-* compounds build on the CONSTITUTION amendment that landed with #1009.

Commit 2 — the v2 surface reduction (owner-decision commit, separate on purpose): `list-todos`, `timeline`, `report`, `get-document`, and `list-approvals` all read legacy v1 multi-agent endpoints that return empty stubs or placeholder data for v2 investigations, and `regenerate-report` targets a backend route that does not exist. Rather than rename dead commands, they're removed. This goes beyond the naming workstream's rename-only rule, which is why it's isolated in its own commit and called out here. Note it also overrides two earlier panel dispositions (`regenerate-report` and the `todos` rename were keeps/renames on paper) — the panel had itself flagged the v2-consolidation removal option to the assistant owner, and this executes it.

Full migration mapping, both commits combined:

| old | outcome |
|---|---|
| `investigations tools <id>` | renamed → `investigations list-tool-calls <id>` (lists tool CALLS, not tools) |
| `investigations chat <id>` | renamed → `investigations list-messages <id>` (pure GET of the thread; `chat` implied an interaction that never happens) |
| `investigations approvals <id>` | removed (v1 stub on v2; transiently `list-approvals` within this branch's history) |
| `investigations todos <id>` | removed (v1 stub; transiently `list-todos`) |
| `investigations document <inv-id> <doc-id>` | removed (v1 placeholder; transiently `get-document`) |
| `investigations report <id>` | removed (v1 stub) |
| `investigations timeline <id>` | removed (v1 stub) |
| `investigations regenerate-report <id>` | removed (backend route absent) |

Remaining surface: `list`, `get`, `create`, `cancel`, `pause`, `resume`, `mode`, `share`, `list-messages`, `list-tool-calls`, `narrative`.

Kept, with dispositions — the human-review records that survive: `mode` (panel 2-1, minority target update-mode) and `narrative` (2-1, get-narrative) stay per the not-unanimous rule; `pause`, `resume`, `share`, and `assistant prompt` were unanimous keeps. The `mcp-server` alias on `mcp-servers` stays (singular variant). Still flagged for a separate decision, untouched here: `assistant dashboard` (panel-unanimous duplicate of `prompt --agent-id grafana_dashboarding`).

Disclosed help-truth edits in the rename commit: list-messages Short/Long "Show/Stream the chat thread" → list wording; one stale Go doc comment updated. No runtime-honesty commit needed.

Verified: gofmt/lint clean, full `-race` suite green (GCX_AGENT_MODE=false) on the rebased branch over post-merge main, reference docs regenerated drift-clean (removed commands' pages deleted). Live e2e against a real stack, re-run 2026-07-22 on the current base: the two surviving renames compare identical to their old spellings on a main binary (stdout + exit codes, live BAD_REQUEST identity; stderr differs only in server traceIds), and every removed or renamed old path fails as unknown command on the new binary. The earlier 6/6 pair run predates the removal commit and covered all five original renames.


claude-plugin/skills/gcx-demo/SKILL.md
```diff
@@ -115,9 +115,8 @@ gcx assistant prompt "Which service owns checkout-latency?" --continue
 gcx assistant prompt "Summarize CPU on prod" --json
 
 gcx assistant investigations list
-gcx assistant investigations todos <id>
-gcx assistant investigations timeline <id>
-gcx assistant investigations report <id>
+gcx assistant investigations get <id>
+gcx assistant investigations narrative <id>
```
dafydd-t: I think we need a verb on here? `get-narrative` maybe?

alexsniffin: Renamed to `get-narrative` in 2fb68bf0 — command, agent annotations, both skills, and the CLI reference page.

dafydd-t (approved):

PR #1017, opened 2026-07-21
feat(k6)!: naming convergence batch (#387)

awsome-o:
This is the k6 batch of the #387 naming workstream, wave 2. Five renames and one alias removal, all clean breaks: no forwarders, old paths fail as unknown commands. The compound spellings build on the CONSTITUTION amendment that landed with #1009.

Two records were pre-adjudicated in the list-* verdict table (docs/plans/list-subject-verdicts.md, committed in the kg batch PR); the rest were verified fresh with adversarial passes. Behavior, flags, positionals, output schemas and exit codes unchanged; only names (and two command homes) moved.

Migration mapping:

| old | new | why |
|---|---|---|
| `gcx k6 load-zones allowed-load-zones list <project-id>` | `gcx k6 projects list-allowed-load-zones <project-id>` | the positional is the PROJECT's id — the membership set has no id of its own, so the compound lives under the identity owner |
| `gcx k6 load-zones allowed-load-zones update <project-id>` | `gcx k6 projects update-allowed-load-zones <project-id>` | same re-homing, write side |
| `gcx k6 load-zones allowed-projects list <load-zone-id>` | `gcx k6 load-zones list-allowed-projects <load-zone-id>` | positional is the load zone's id; nested noun group dissolves into the compound |
| `gcx k6 load-zones allowed-projects update <load-zone-id>` | `gcx k6 load-zones update-allowed-projects <load-zone-id>` | same |
| `gcx k6 schedules delete <load-test-id>` | `gcx k6 load-tests delete-schedule <load-test-id>` | the landmine of the batch: the positional is the LOAD TEST's id (DELETE /loadtests/{id}/schedule) while sibling `schedules get/update` take the schedule's own id — a user who pattern-matches and passes a schedule id deletes a different schedule, and the client swallows 404 as success. The rename moves the command to where its identity actually points; args, flags and behavior are byte-identical |
| alias `gcx k6 run …` (on `runs`) | removed | `run` reads as an execution verb (upstream k6 spells it `k6 cloud run`) but gcx k6 cannot start runs — the group only lists history. Zero active-text references; the only spelling lost is `gcx k6 run list` |

Both `allowed-load-zones` and `allowed-projects` noun groups dissolve (they had no aliases). The `schedules` group keeps its other four children and its `schedule` alias.

Flagged for a decision, not executed: `gcx k6 test-run runs list` — panel unanimous REMOVE (verbatim duplicate of `gcx k6 runs list`, same client call and codec, mounted under a group whose declared purpose is TestRun CRD manifests). Removal wasn't a worksheet proposal, so it stays for now; happy to do the follow-up if you agree.

Kept, with dispositions: `k6 auth token` (2-1, minority get-token), `k6 runs list` (2-1, minority wanted load-tests list-runs), `k6 test-run status` (2-1), `k6 test-run emit` (unanimous keep). Aliases all resolve keep-synonym with a row each: `env`/`envvar`/`envvars` on env-vars, `test` (2-1 panel — one judge argued it half-promises generic tests) and `tests` on load-tests, `load-zone`/`lz`, `proj`/`project`, `schedule`, `testrun` (concatenated spelling matching the k6-operator CRD kind) — none mis-states an operation.

Disclosed help-truth fix riding the rename commit: both update-allowed-* commands now say "Replace the set of …" in Short and "complete replacement set" in the --filename help — verified against the client first (full-array PUT, no merge path). The old help read as incremental update, which it never was. Also updated: the k6 rows of docs/reference/migration-gap-analysis.md (that file's "Status in gcx" cells track the current surface and have been updated on every past rename), and a stale section comment that described the removed `run` alias. No runtime-honesty commit needed — all Success strings were already honest.

Verified locally: gofmt/lint clean, full `-race` suite green (GCX_AGENT_MODE=false), reference drift clean, git status clean. Live e2e with two binaries: 6/6 pairs identical stdout/exit (live comparison for the list re-home returned identical plugin-unavailable error identity on the eval stack; write commands compared on pre-API argument validation only — nothing was created or deleted), plus old-path-gone and run-alias-gone checks on the new binary. mkdocs runs in CI.


fornfrey (approved):
LGTM

dafydd-t (approved):

PR #1019, opened 2026-07-21
feat(cli)!: naming convergence tail sweep (#387)

awsome-o:
This is the tail sweep of the #387 naming workstream, wave 2 — the last provider batch before the registry work. It bundles the small slices: appo11y, cloud, dashboards, dev, fleet, frontend, logs, metrics, resources, synthetic-monitoring, traces, plus the config and slo alias records. Thirteen renames, all clean breaks: no aliases, no forwarders. #1009 has merged and this branch is rebased over it (the shared Constitution paragraph is already reconciled — see disclosures). Stays draft until the CONSTITUTION amendment in this PR gets explicit maintainer approval.

Six were pre-adjudicated in the list-* verdict table (docs/plans/list-subject-verdicts.md, committed in the kg batch PR); the rest were verified fresh with adversarial passes, and the three adaptive records went through the coupled subject+verb panel described below. Behavior, flags, positionals, output schemas and exit codes unchanged; only names moved.

Migration mapping:

| old | new | why |
|---|---|---|
| `gcx appo11y services labels <service>` | `gcx appo11y services list-labels <service>` | parent-scoped label facet, positional is the service's id (§8 case b; sibling of the existing list-operations) |
| `gcx cloud stacks regions` | `gcx cloud stacks list-regions` | plain value enumeration, no region read-one (§8 case a, the cloudwatch list-regions analog) |
| `gcx dashboards versions list <name>` | `gcx dashboards list-versions <name>` | versions have no own ID (only the parent's name + generation); the versions group keeps `restore` |
| `gcx dev lint rules` | `gcx dev lint list-rules` | rule names are only filter values, no read-one |
| `gcx frontend apps show-sourcemaps <app>` | `gcx frontend apps list-sourcemaps <app>` | returns the bundle collection for one app; show is not canonical |
| `gcx frontend apps remove-sourcemap <app> <bundle>…` | `gcx frontend apps delete-sourcemap <app> <bundle>…` | the operation permanently destroys the bundles; "remove" reads as detachment |
| `gcx fleet tenant limits` | `gcx fleet tenant get-limits` | fetches the one limits object for the tenant |
| `gcx resources examples [selector]` | `gcx resources list-examples [selector]` | example manifests are a catalog facet keyed by type, selector is a filter |
| `gcx resources schemas [selector]` | `gcx resources list-types [selector]` | its own Short says "List available Grafana API resource types"; schemas are opt-out enrichment (--no-schema) absent from default text output — list-types is the only name truthful in every output mode; it coexists cleanly with the profiles pilot, which now spells its command `list-profile-types` (#1009 reroll) |
| `gcx synthetic-monitoring probes token-reset <id>` | `gcx synthetic-monitoring probes reset-token <id>` | operation-first compound; noun-first was backwards |
| `gcx logs adaptive patterns show` | `gcx logs adaptive patterns list` | see the adaptive decision below |
| `gcx metrics adaptive recommendations show` | `gcx metrics adaptive recommendations list` | same |
| `gcx traces adaptive recommendations show` | `gcx traces adaptive recommendations list` | same |

The adaptive trio was one coupled decision (subject and verb together, so these commands rename exactly once). A three-lens panel — product docs, code behavior, cross-signal consistency — was unanimous: logs keeps the `patterns` subject and all three take `list`. The apparent subject mismatch with metrics/traces is faithful product terminology, not a gcx inconsistency: the Adaptive Logs UI is the Patterns page and a "recommendation" there is the drop rate attached to a pattern, while Adaptive Metrics/Traces genuinely call their items recommendations. The Go type `[]LogRecommendation` and the API path are transport naming, which the ADR forbids as a basis for meaning. Verb-wise all three are pure collection fetches with no single-item mode. Nobody gets a list-* compound: metrics recommendations are addressable per metric name (diff/apply take one), traces recommendations have IDs (apply/dismiss <id>), and logs patterns is a catalog child with no parent positional — all honest noun groups with plain `list`.

Kept, with dispositions: `appo11y services map`, `metrics cardinality label-names`/`label-values`, `providers list` — unanimous keeps. `synthetic-monitoring probes deploy` — 2-1 (minority: generate-manifests; it emits YAML and contacts no cluster). Two three-way rename splits stay per the not-unanimous-on-target rule, both worth owner follow-ups: `dashboards versions restore` (2× `dashboards restore`, 1× `restore-version` — the mutated subject is the dashboard itself) and `frontend apps apply-sourcemap` (votes split create-/upload-sourcemap; every invocation creates a new bundle, so `apply` is the wrong verb whichever wins).

Aliases — all 15 records resolve keep-synonym with a row each: dashboards `ls`/`rm` (shell conventions naming the exact operations), fleet `collector`/`pipeline`, frontend `app`, sm `sm`/`check`/`probe`, traces `tags`/`search` (declared permanent synonyms, cross-signal ADR), config `use`, slo `def`/`defs`/`report`. One needs your eye: `synth` on synthetic-monitoring went 2-1 (remove/remove/keep). The remove votes cite #631 verbatim — it was kept "to not break any existing configurations workflows", i.e. a compatibility alias, which the clean-start policy removes. The keep vote counters that `synth` is the provider's own registry/config identifier (`providers.synth` in config files). Not unanimous = keep, but it's a genuine judgment call you may want to overrule.

Disclosures, the big one first: this PR now carries a real CONSTITUTION amendment, not just a token edit, and it needs explicit maintainer approval. The adaptive renames create three bare `list` commands on provider-only resources (`logs adaptive patterns list`, `metrics/traces adaptive recommendations list`), and a strict reading of two existing clauses — "provider-only resources must not mimic adapter verbs" and "provider CRUD commands must use TypedCRUD" — would forbid that, along with most existing provider trees. The amendment now makes the intended scope explicit without leaving the old opposite rule in place: dual-path JSON/YAML parity and TypedCRUD apply to adapter-backed resources; provider-only commands use behavior-based operations; a genuine read-one is `get` and a genuine enumeration is `list` regardless of adapter registration; nonstandard behavior uses an honest query, view, domain operation, or operation-subject compound; and `create`/`update`/`upsert` join the core-operation enumeration in the extensions clause. #1009 edited the old provider-only paragraph, so rebasing over its merge produced the expected one-paragraph Constitution conflict — resolved (2026-07-21) toward this behavior-based rule while keeping #1009's concrete compound examples (`list-profile-types`, `list-tables`) in the text. Smaller CONSTITUTION touch that was already here: line 44's example reference to the `schemas` command now reads `resources list-types`. Help-truth fixes riding the rename commit: fleet tenant Shorts ("Show tenant limits."→"Get tenant limits.", "Manage …"→"Fleet Management tenant settings." — the group is read-only), frontend delete-sourcemap Short/examples ("Remove"→"Delete"), metrics adaptive error suggestions that cited `recommendations show` now cite `list`, and the appo11y codec-misuse error strings follow the family name. The separate runtime commit changes one string: faro's success line "Removed %d sourcemap(s) from app %s" → "Deleted …". The "Status in gcx" cells of docs/reference/migration-gap-analysis.md are updated to the new spellings (that file tracks the current surface, per its history and the #1009/k6/irm precedent), including its dashboards versions row which wrongly claimed the capability was missing. docs/bento_marketing_slide.html had its four renamed command tokens hand-edited.

Verified locally: gofmt/lint clean, full `-race` suite green (GCX_AGENT_MODE=false), reference drift clean, git status clean. The read-through review earned its keep here: it caught a skill file whose instructions became actively inverted by the adaptive rename ("the recommendations group uses `show`/`diff`, not `list`") and the gcx SKILL.md resource-operations list — all fixed before this PR opened. Live e2e with two binaries against a real stack: 13/13 pairs identical stdout and exit codes (regions/fleet/adaptive compared on identical live error identity, stderr differing only in server requestIds; `dev lint rules` output ordering is nondeterministic on main too — rule sets are identical; `resources list-types` was byte-identical on a fresh run after an initial diff traced to a server-side discovery change between calls), plus old-path-gone checks. mkdocs runs in CI.


awsome-o:
Follow-up to the Constitution disclosure: the old opposite provider-only rule is now removed, dual-path output parity and TypedCRUD are scoped to adapter-backed resources, and provider-only commands use behavior-based operations. This makes the amendment internally coherent. #1009 should merge first; its shared Constitution paragraph will cause one expected rebase conflict, which should resolve toward this rule while retaining the list-profile-types changes.

dafydd-t (approved):

PR #1020, opened 2026-07-21
Update installation.md

chrisfinazzo:
Remove instructions for using a custom Tap.

gcx can now be installed directly from Homebrew.


dafydd-t (approved):
Thanks a lot @chrisfinazzo !

PR #978, opened 2026-07-15
feat(telemetry): wire usage event lifecycle with log mode

dafydd-t:
Part of https://github.com/grafana/gcx-internal/issues/2

follow on from https://github.com/grafana/gcx/pull/944

## Summary

- wires the internal telemetry library from #944 into the CLI lifecycle
- `GCX_TELEMETRY=log` prints the event JSON to stderr. nothing is exported yet

behaviours:
  - suppression: `completion`,`version` and`telemetry` commands do not emit events.
  - `--help`, `gcx help <path>`, `gcx`, and e.g. `gcx resources` record `outcome=help` with the resolved target path
  - parse failures (e.g. `gcx invalid command`) emit nothing, this is a TODO in #578 
  - user cancellation exits fast with no event

## Examples

```
// valid
GCX_TELEMETRY=log ./bin/gcx commands >/dev/null
{"service":"gcx","version":"(devel)","os":"darwin","arch":"arm64","device_id":"36d9...","device_id_persisted":true,"command":"commands","flags":"","provider":"commands","outcome":"ok","exit_code":0,"error_kind":"","duration_ms":24,"is_tty":false,"is_ci":false,"ci_provider":"","is_agent":false,"agent":"","target_kind":"","output_format":"json"}

// valid, multiple args
GCX_TELEMETRY=log ./bin/gcx dashboards list >/dev/null
{"service":"gcx","version":"(devel)","os":"darwin","arch":"arm64","device_id":"36d9...","device_id_persisted":true,"command":"dashboards list","flags":"","provider":"dashboards","outcome":"ok","exit_code":0,"error_kind":"","duration_ms":1059,"is_tty":false,"is_ci":false,"ci_provider":"","is_agent":false,"agent":"","target_kind":"","output_format":"table"}

// help
GCX_TELEMETRY=log ./bin/gcx help dashboards list >/dev/null
{"service":"gcx","version":"(devel)","os":"darwin","arch":"arm64","device_id":"36d9...","device_id_persisted":true,"command":"dashboards list","flags":"","provider":"dashboards","outcome":"help","exit_code":0,"error_kind":"","duration_ms":18,"is_tty":false,"is_ci":false,"ci_provider":"","is_agent":false,"agent":"","target_kind":"","output_format":""}

// help, invalid command => command is "help"
GCX_TELEMETRY=log ./bin/gcx help foo bar >/dev/null
[...]

Issue #578, opened 2026-04-27
Telemetry: capture guessed/nonexistent commands and flags


wardbekker:
## Summary

When we add anonymous telemetry, also instrument the **failed** parse paths — commands and flags that don't exist but are guessed by users or agents (e.g. Claude Code, Cursor, etc.). Today we only see what *worked*; we're blind to what people *tried*.

## Motivation

Agents in particular pattern-match from training data and CLI conventions, so their guesses are a strong signal:

- Where our command grammar diverges from user/agent expectations.
- Which commands or flags we should consider adding.
- Where in-CLI guidance (stderr hints, `--help`, agent metadata) is insufficient.
- Which areas of the surface need better discoverability via `gcx commands` / help tree.

This is effectively a "404 log" for the CLI.

## Proposed scope

Hook into Cobra's unknown-command and unknown-flag paths, plus arg-validation failures, and emit an event with:

- The attempted command path (e.g. `gcx dashboards search`).
- The unknown flag name(s) (without values).
- Optional: fuzzy distance to the nearest real command/flag — lets us distinguish near-misses (likely UX gaps) from random noise.
- Standard envelope: gcx version, OS, agent-mode detection (`CLAUDECODE`, `CLAUDE_CODE`, etc.).

## Privacy / safety

- **Never** capture flag values, positional args, or resource names — only the command path and flag name(s).
- Gated by the same opt-in/opt-out as the rest of telemetry.
- Document in the telemetry reference what is and isn't collected for failed parses.

## Open questions

- Should we also capture *typos on real commands* (e.g. `dashbaords`) separately from "command does not exist at all"? The fuzzy-distance idea above subsumes this.
- Do we report once-per-invocation or aggregate? (Probably once — a single user typo shouldn't hammer telemetry.)
- Should agent-mode invocations be tagged so we can split agent guesses from human guesses in analysis?

PR #1021, opened 2026-07-21
feat(telemetry): export usage events as flat JSON over HTTP in enabled mode

dafydd-t:
Part of https://github.com/grafana/gcx-internal/issues/2, follow on from #978.

This is an alternative to #1002, : instead of encoding the event as an OTLP span, it emits the JSON-encoded telemetry vent directly.

## Summary

- need `GCX_TELEMETRY=enabled` for now, as its default off until the receiver is ready. It posts to `https://stats.grafana.org/gcx-usage-report` or wherever `GCX_TELEMETRY_ENDPOINT` is set to.
- 1s timeout on export, all errors ignored. telemetry should not cause an error or noticeably slow the CLI. We can delegate the export to another process if it's obviously too slow.

Example payload:

```json
{"service":"gcx","version":"v0.4.4","os":"darwin","arch":"arm64","device_id":"36d988e3-971e-4566-b0ca-7416f85b0da2","device_id_persisted":true,"command":"providers list","flags":"","provider":"providers","outcome":"ok","exit_code":0,"error_kind":"","duration_ms":17,"is_tty":true,"is_ci":false,"ci_provider":"","is_agent":false,"agent":"","target_kind":"","output_format":"table"}
```

Not included: we still dont have `target_kind` (needs the keychain-free context loader) and parse-failure events (#578).

Run locally with e.g. (if you have the stats service running locally) 

```
GCX_TELEMETRY=enabled GCX_TELEMETRY_ENDPOINT=http://localhost:9051/gcx-usage-report ./bin/gcx datasources list
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -169,7 +169,7 @@ internal/
 ├── notifier/    Update notifications (skills + gcx version checks; XDG state, throttling, message rendering — wired into root PersistentPostRun)
 ├── skills/      Portable Agent Skills installer primitives (BundledSkillNames, Install, Update — extracted from cmd/gcx/skills)
 ├── strcase/     String case conversion (snake_case, kebab-case, PascalCase)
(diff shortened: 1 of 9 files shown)


gerboland (approved):
Looks good

PR #1023, opened 2026-07-21
docs(design): command naming conventions guide

dafydd-t:
Contributor-facing guide for command naming: which verb to use, and where an operation sits in the tree. It covers the CRUD verbs, the two create-or-update workflows, query/view/domain operations, and placement. Now at `docs/design/command-naming.md`.

This is meant to be the single prescriptive source for command naming. #994 is an alignment record and will not merge, so the vocabulary work it drafted is folded in here.

The PR also carries one CONSTITUTION change, which the guide leans on. v1.0.0 shipped on 2026-07-28 with some known naming deviations already in it, and nothing in tier 1 said the command surface was stable. That meant any grammar rule in CONSTITUTION could be read as grounds for renaming a command we already released. CLI Grammar now says command paths, aliases, flags and positional syntax are stable within a major version, that naming consistency does not authorize breaking changes, and that the whole v1.0.0 surface is a supported compatibility exception for all of v1.x. A conforming replacement can be added, but only while the existing invocation keeps working.

## What changed after review

`push` is now described by what it does and nothing more. It applies the manifests you give it and does not prune resources you left out. It no longer claims a diff or a separate validation phase, because no push implements either. `gcx resources push` emits a summary, and `ValidateFn` only runs under dry-run and returns `ErrDryRunUnverified` when an adapter does not define one. Idempotency and safety stay where they already live, in CONSTITUTION § Push/Pull Philosophy and `docs/design/safety.md`.

Placement is decided per operation rather than once per subject. That keeps `gcx k6 load-tests delete-schedule <load-test-id>` correct, since the deletion needs the load test id, without contradicting the `schedules` noun group whose siblings take a schedule id. It also lines up with CONSTITUTION § Sub-resources, which lets a sub-resource keep a standalone get-by-ID. Composite identities and parent-required `create` are left open on purpose, pointing at a Constitution amendment or a recorded waiver, rather than being settled quietly in a tier-3 doc.

Domain verbs no longer point at an allowed-operation registry, because there is not one. The term only appears in unmerged #994, whose own draft defers the registry to a follow-up. Until something lands, a new canonical domain verb needs explicit PR review and a written definition.

Query operations have a section of their own now: `query`, `search`, and the shorthands already in use, `labels`, `series`, `metrics` and `metadata`. Without it, a rule derived from this guide would flag `gcx logs query`, which CONSTITUTION cites as an example of correct grammar.

There is a scope preamble saying the rules bind new or explicitly renamed canonical commands, that a shipped command is not automatically a precedent, and that the guide being silent about something does not make an existing command non-compliant. A new verb or shape it does not cover still needs review.

All twelve `gcx` examples are complete runnable paths and were checked against a built binary. Nine were previously incomplete, and four of those were missing an area segment outright, so `cloudwatch list-regions` is now `gcx datasources cloudwatch list-regions`.

Entry points repointed at this file: AGENTS.md, CONTRIBUTING.md, DESIGN.md, docs/design/README.md, naming.md, provider-checklist.md, and both migrate-provider skill files. The superseded #994-era plan docs are marked historical rather than deleted.

## Known, and deliberately not in this PR

- #1107, `gcx slo definitions push` and `gcx slo reports push` always create when a manifest has no `uuid`, which contradicts the idempotency guarantee in CONSTITUTION § Push/Pull Philosophy. Needs a code fix with idempotency tests.
- The allowed-operation registry and its lexical CI check.
- Whether nouns should be plural or singular. Still unspecified, and the tree is inconsistent today.
- CONSTITUTION.md and DESIGN.md are not part of the MkDocs site, so this guide's relative links to them work on GitHub but not on the published site. The docs build warns and passes. I kept them relative because absolute `blob/main` links have no precedent in this repo and would show main's content when read inside a PR. Worth noting `docs/adrs/dashboards-provider/001-dashboards-provider-design.md` already links `../../../CONSTITUTION.md` the same way, so this is 5 more warnings on an existing pile of 25.
- A markdown invocation validator for `docs/design/`, reusing `extractInvocations` from `cmd/gcx/root/skillsdrift_test.go`.


docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,106 @@
...
+## Behavior determines the name
+
+A command is named for what the *user* experiences: what it acts on, whether
+you address one thing by ID, whether it returns one item or many, and whether
+it changes anything. It is never named for the HTTP method, the API path, or
+how it is wired internally.
+
+A read-one is `get` whether the backend uses GET, POST, or three calls under
+the hood. And when a provider migrates from `/api` to `/apis`, its command
+names must not change - users never agreed to depend on the transport.
```
dafydd-t: i dont think we need this

docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,106 @@
...
+| `get` | Retrieve exactly one thing, addressed by ID (or a singleton) |
+| `create` | Make a new thing - fails if it already exists |
+| `update` | Change an existing thing - fails if it doesn't exist |
+| `delete` | Remove an existing thing |
+
+Two create-or-update verbs exist, distinguished by workflow, not transport:
+
+- `upsert` - one invocation creates the thing if absent, updates it if
+  present. Never split a true upsert into fake `create`/`update` commands:
+  that falsely promises existence checks and invites read-then-write races.
```
dafydd-t:
```suggestion
- `upsert` - one invocation creates the thing if absent, updates it if
  present. Do not split a true upsert into fake `create`/`update` commands:
  that falsely promises existence checks and invites read-then-write races.
```

docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,106 @@
...
+- `push` / `pull` - the manifest (GitOps) workflow: `push` applies local
+  manifest files to the remote, `pull` writes remote resources to local files.
+  This is a different workflow from `upsert`, not a synonym.
+
+`patch` is reserved for APIs that take an explicit patch document. "The update
+is partial" is not enough - most resource updates are.
```
dafydd-t: do we need this?

docs/design/naming-conventions-guide.md
dafydd-t: removed

docs/design/naming-conventions-guide.md
dafydd-t: applied

docs/design/naming-conventions-guide.md
dafydd-t: removed - no command uses patch yet, can document it when one exists

docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,61 @@
...
+gcx k6 load-tests list
+gcx k6 load-tests get <id>
+```
+
+**`list-things` is reserved for two cases:**
+
+1. When a resource is listable but they have no ID of their own: `cloudwatch list-regions` -
+   there is no `region get`; a region is just a value.
+2. When you filter a list based on the *parent's* ID: `alert-groups list-alerts
+   <group-id>` - the ID you pass belongs to the group, not to an alert.
```
awsome-o: The rule here is a little too broad. A subject having no fetch one command does not always force `list-<subject>`, and a parent filter does not automatically make something a parent scoped collection.

`incidents severities list` is the catalog child case we already decided to keep. Severity rows do have IDs, but there is no independent severity fetch command and this list does not take a parent ID.

I suggest replacing the two cases with this:

> Use `<subject> list` for an independently addressable resource group. A catalog child with no parent identity may also keep this shape, for example `incidents severities list`.
>
> Use `list-<subject>` when the subject is a discovery or catalog facet that is not independently addressable, or when it is a parent scoped collection that is not independently addressable and the required positional identity belongs to the parent. `cloudwatch list-regions` and `alert-groups list-alerts <group-id>` are the two examples.
>
> The number of commands implemented for the subject today is not part of the rule.

awsome-o (commented):
I read this against #994, #1009, the current command tree, and the current heads of #1013 through #1019. The guide is going in the right direction. Every example maps to a live command, none of those paths changes in the naming batches, and this branch is pairwise merge clean with each batch.

There are three content points I would fix before this is ready.

First, the list placement rule is too broad. A parent filter does not automatically mean `list-<subject>`. For example, `k6 load-tests list --project-id` stays a noun group because load tests can be fetched independently. The compound form applies when the child collection is not independently addressable and the required positional belongs to the parent. Catalog children with no parent identity, such as `incidents severities list`, may also keep a noun group. I left exact wording inline.

Second, `get` should include singletons, not only things addressed by an ID. `appo11y settings get` and `appo11y overrides get` take no positional ID. I suggest: `Retrieve one addressable thing or singleton in its stable structured representation.`

Third, the `upsert` and `push` distinction should be explicit, and the open question behind it is now settled on our side (owner decision, recorded in #994 and #1015): `upsert` is the direct create-or-update workflow — the caller supplies one or more explicit subjects and never chooses create vs update; multi-subject input is processed independently and may be non-atomic. `push` is the manifest/GitOps workflow — file-oriented apply with validation/diff/summary, potentially covering several resources, and it does not delete remote resources omitted from the input. Cardinality alone doesn't decide the verb; the workflow does. The KG batch's entity and relationship `upsert -f` array input stays under that definition, with help text and focused tests now documenting the ordered, non-atomic processing.

There is also a documentation ownership question. The PR body says this waits for #994, but #994 is now an alignment record that is not expected to merge. If this guide is the contributor source, the PR body should say that clearly and the required Constitution wording should land elsewhere. Otherwise we risk keeping two different prescriptive naming documents.

The parent identity and `datasources health` notes are optional improvements, and I left concise comments for both. With the content fixes above, I support this.

docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,61 @@
...
+`describe`) only when the output differs from a straight read of the resource. For example: `kg entities inspect` returns an RCA timeline and related entities, which is a diagnostic analysis, not the entity's fields.
+
+## Domain verbs
+
+Only diverge from the CRUD verbs when they are not accurate for your domain. For example, you `close` an incident or
+`acknowledge` an alert - you don't "update" it closed. Domain verbs like
+`resolve`, `silence`, and `escalate` must each be registered in the
+allowed-operation registry with a written definition of what they mean.
+
+## Where does a list command live?
```
awsome-o: This placement rule applies to operations other than `list` too. When the required positional belongs to the parent and the child is not independently addressable, the shape is `<parent> <operation>-<child> <parent-id>`.

The current batches introduce `k6 load-tests delete-schedule <load-test-id>`, `collections add-conversations <collection-id> <id>...`, and `projects update-allowed-load-zones <project-id>` using that rule. The parent compound rule already exists in the Constitution; #1009 broadens compound use for provider only resources. One sentence here would save the next author from having to infer it again.

docs/design/naming-conventions-guide.md
```diff
@@ -0,0 +1,61 @@
...
+  manifest files to the remote, `pull` writes remote resources to local files.
+  This is a different workflow from `upsert`.
+
+There is one documented exception to these rules: `gcx resources get` takes kubectl-style selectors
+(`dashboards`, `dashboards/foo,bar`) and may return one item or many.
+
+## View verbs
+
+Default to using `get`. Use a view verb (`status`, `timeline`, `inspect`, `diff`, `stats`, `report`,
+`describe`) only when the output differs from a straight read of the resource. For example: `kg entities inspect` returns an RCA timeline and related entities, which is a diagnostic analysis, not the entity's fields.
```
awsome-o: `datasources health` is now an owner approved keep from #1014 because it runs the Grafana product health check. The registry work still has to decide whether `health` becomes a governed operation or a scoped exception, so I would not call it a scoped exception yet.

A short sentence saying that approved exceptions are recorded case by case would keep this guide consistent with the live surface without making `health` a general pattern to copy.

docs/design/naming-conventions-guide.md
dafydd-t: addressed in a23711d6 - replaced the two cases with your wording: independently addressable groups keep the noun group, catalog children like `incidents severities list` may too, and `list-things` is for non-addressable facets or parent-scoped collections. also added the line that command count today isn't part of the rule

docs/design/naming-conventions-guide.md
dafydd-t: good shout - added a closing paragraph extending the placement rule to other operations with the `<parent> <operation>-<child> <parent-id>` shape and the delete-schedule / add-conversations examples

docs/design/naming-conventions-guide.md
dafydd-t: added a sentence to the view verbs section: approved exceptions are recorded case by case, with `datasources health` called out as an owner-approved keep rather than a pattern to copy

dafydd-t:
addressed the two content points from the review body in a23711d6 as well: `get` now covers singletons (`appo11y settings get`), and the upsert/push bullets now spell out that the workflow decides the verb - upsert is direct create-or-update with explicit subjects (multi-subject processed independently, may be non-atomic), push is the manifest workflow and never deletes omitted remote resources

dafydd-t:
good shout on the ownership point - #994 won't merge (it's an alignment record), so I've updated the PR body: this guide is the contributor-facing source of truth, and any constitution-level wording lands separately. taking this out of draft

docs/design/naming-conventions-guide.md
awsome-o: Took most of this, but I changed one part of your wording, so flagging it rather than burying it.

Your line about the number of commands implemented not being part of the rule is in. `incidents severities list` stays a noun-group example too, and there is now a second arm that covers it directly: use a noun group when the operation either addresses the child by its own identity, or enumerates the group without needing a parent identity. So the catalog-child case falls out of the rule instead of being a special case you have to remember.

What I dropped is "not independently addressable" from the parent-scoped half. Two reasons. It made the guide contradict its own example, because `gcx k6 schedules` ships list, get by id, create and update, so a schedule really is independently addressable, and yet `load-tests delete-schedule <load-test-id>` is still the shape we want. It is also stricter than CONSTITUTION § Sub-resources, which says get-by-ID can keep a standalone adapter when the API supports direct lookup. Applied literally it flags commands we already shipped, like `gcx dashboards list-versions <name>` and `gcx agento11y experiments list-trials <run-id>`.

What replaced it is a per-operation test instead of a per-subject one. Ask whose identity this particular operation needs. If it only needs the parent, use the compound. If it takes the child's own id, use the noun group. `delete-schedule` passes because that deletion needs the load test id, whatever `schedules` happens to offer elsewhere.

If you think the per-operation framing loses something your wording had, say so and I will put it back.

docs/design/naming-conventions-guide.md
awsome-o: Added, and generalised slightly. The guide now says to apply the same test outside `list`, using `gcx k6 load-tests delete-schedule <load-test-id>` and `gcx agento11y collections add-conversations <collection-id> <saved-id>...` as the worked examples. I checked that `gcx k6 projects update-allowed-load-zones` resolves as well, but left it out to keep the example block short.

Your note also surfaced a gap the rule did not cover. `add-conversations` takes a collection id plus N saved-conversation ids, so multi-id operations needed a tie-breaker. The guide now says placement follows the identity of the primary object or relationship being changed, and the remaining ids are operands. That is a different thing from a composite key where the parts together identify the child, and the composite case is left open on purpose rather than decided here.

Same caveat as the sibling thread: I dropped the "child is not independently addressable" half of your condition, for the reasons I wrote up over there.

docs/design/naming-conventions-guide.md
awsome-o: Your main point is in. It now reads "an owner-approved keep from #1014, it runs the Grafana product health check rather than reading the datasource record, and is not a pattern to copy". I agree that calling it a scoped exception pre-empts a registry decision nobody has made yet.

I did not take the second half, and I am leaving this thread open for you instead of resolving it myself. The first revision said approved exceptions are recorded case by case, but there is no such record. Grepping for "owner-approved" or "approved exception" across CONSTITUTION.md and docs/design/ turns up only that one line. A pointer to a register that does not exist reads like real process to a contributor or an agent, and there is nothing for them to go and check.

What the guide does instead is send new unresolved shapes to either a Constitution amendment or an explicit waiver recorded in the approving PR or ADR, so the record sits wherever the approval happened. If you would rather have your sentence back, or you want a real exceptions table built in this PR, say so and I will add it.

awsome-o (approved):

K-minutti (approved):
lgtm

PR #1025, opened 2026-07-21
feat(kg): add entity quality reports (kg quality) and fold them into kg diagnose

FCalero:
## Summary

Adds support for the Knowledge Graph **entity quality reports** capability to gcx. Quality reports grade how well each entity (currently `Service`) is instrumented — a set of checks (request metrics, service-map data, deployment environment, span cardinality, logs, profiles, …) with an overall quality percentage.

This PR exposes them two ways:

1. A new **`gcx kg quality`** command group (`list` / `get`) for reading reports directly.
2. An aggregate **instrumentation-quality check** folded into **`gcx kg diagnose`**, so a healthy pipeline also surfaces instrumentation debt across services.

The reports are served by the Asserts app plugin and proxied under `/asserts/kg-quality/v1/entities/...`.

### `gcx kg quality list`
Lists/ranks entity quality reports by quality percent.
- Flags: `--type` (default `Service`), `--env`, `--namespace`, `--site`, `--entity`, `--failed-check` (repeatable), `--sort asc|desc`, `--page`, `--page-size`.
- Returns the report items (not the page envelope) so `--json`/`--jq` operate on report fields; pagination is surfaced as a stderr hint, mirroring `kg entities list`.

### `gcx kg quality get <entity-name>`
Full report for a single entity: per-check state/impact, remediation doc links, and resolved query templates. Requires `--env`. The default table prints a `Details:` footer for each non-`SUCCESS` check (description + `docUrl` + `reference`) so the remediation metadata is legible without dropping to `-o json`/`-o yaml`; resolved query templates remain JSON/YAML-only.

### `gcx kg diagnose` integration
`runDiagnose` now always runs an aggregate "Instrumentation quality" check (via the KG client, independent of Prometheus availability):
- **skip** when reports are unavailable (endpoint not provisioned / 404) or empty — never fails the run.
- **pass** when all services are at 100%.
- **warn** when any service is below 100%, with a detail line (service count, worst %, top recurring failed checks) and a recommendation pointing at `kg quality list` / `kg quality get` (scoped to the diagnose env/namespace/site). This warn is surfaced in the check table and summary but is **excluded from the `Orientation.PipelineHealth` rollup** — instrumentation debt is not a pipeline fault, so a lone service below 100% won't flip the stack verdict to degraded.
- Also exposes a structured `Quality` summary on the JSON `DiagnoseResult` (total services, count below perfect, worst %, top failed checks, sampled flag). Aggregation focuses on the worst-instrumented services: it fetches a single worst-first page of up to 50 reports (`sortDirection=ASC`) and marks the summary `sampled` when more degraded services may lie beyond that window. `totalServices` comes from the page envelope, so it reflects the true count across the scope regardless of the window size.

## Changes by area
- **`internal/providers/kg/types.go`** — DTOs for quality reports (`QualityReport`, `QualityReportData`, `QualityCheckResult`, `QualityCheckReference`, `CheckCondition`, `QualityReportListItem`, `QualityReportPage`, `QualityState`).
- **`internal/providers/kg/client.go`** — path constants + `GetEntityQualityReport` and `ListQualityReports` (with `QualityReportQuery`).
- **`internal/providers/kg/quality.go`** *(new)* — the `kg quality` command tree, opts, and table codecs.
- **`internal/providers/kg/diagnose.go`** — `checkQuality` aggregation, `QualityCheckSummary`/`FailedCheckCount` types, `DiagnoseResult.Quality`, goroutine wiring, and sort ordering.
- **`internal/providers/kg/provider.go`** — mounts `newQualityCommand`.
- **`internal/agent/command_annotations.go`** — token-cost + LLM hints for the two new leaf commands; updated `kg diagnose` hint.
- **Docs** — regenerated CLI reference (`gcx_kg_quality*.md`, updated `gcx_kg.md`); updated the `kg` package-map line in `AGENTS.md`.

## Testing
- **`quality_test.go`** *(new)* — client methods (URL/query params, decoding, 404), both table codecs, and `checkQuality` (warn/pass/skip/empty + top-gap ordering).
- **`diagnose_test.go`** — updated pipeline-test assertions and added integration tests exercising the quality check through `runDiagnose` (warn + populated `Quality`; skip + nil `Quality` when the endpoint is unavailable).
- `GCX_AGENT_MODE=false mise run all` passes locally (lint: 0 issues, all tests ok, build + docs succeed).
- Manually verified end-to-end against a live Asserts stack: `kg quality list/get` return real reports and `kg diagnose` surfaces the aggregate quality check.

Tested locally against a demo stack:

KG diagnose

image

KG quality list

image

KG quality GET

image


internal/providers/kg/diagnose.go
```diff
@@ -1099,6 +1141,151 @@ func checkTracePropagation(ctx context.Context, client *prometheus.Client, datas
...
+	sampled := false
+	for page := 0; ; page++ {
+		p, err := client.ListQualityReports(ctx, QualityReportQuery{
+			Type:      defaultQualityEntityType,
+			Env:       scope.env,
+			Namespace: scope.namespace,
+			Site:      scope.site,
+			Page:      page,
+			PageSize:  qualityDiagnosePageSize,
+		})
```
annanay25: we are trying to fetch sometimes >500 reports and there might still be some left over, should we just try to grab the top 50 degraded services? just looking for a way to get maximum value out of the aggregation.

internal/providers/kg/diagnose.go
```diff
@@ -1099,6 +1141,151 @@ func checkTracePropagation(ctx context.Context, client *prometheus.Client, datas
...
+		}
+		detail += "; top gaps: " + strings.Join(parts, ", ")
+	}
+	if sampled {
+		detail += fmt.Sprintf(" [sampled first %d]", qualityDiagnoseMaxReports)
+	}
+
+	rec := ""
+	if status == CheckWarn {
+		rec = "Run 'gcx kg quality list --sort asc' to rank the worst-instrumented services, and 'gcx kg quality get <service> --env <env>' for per-check remediation."
```
annanay25: should we add the scope to this hint as well (`gcx kg quality list --sort asc --env xyz`)

annanay25 (commented):
mostly sound, left 2 comments. thanks!

internal/providers/kg/diagnose.go
FCalero: Makes sense. Will apply the change

internal/providers/kg/diagnose.go
FCalero: 👍  Good call

internal/providers/kg/diagnose.go
```diff
@@ -1099,6 +1143,178 @@ func checkTracePropagation(ctx context.Context, client *prometheus.Client, datas
...
+		TotalServices:   total,
+		BelowPerfect:    below,
+		WorstQuality:    worst,
+		TopFailedChecks: topFailedChecks(counts, qualityDiagnoseTopN),
+		Sampled:         truncated,
+	}
+
+	status := CheckPass
+	if below > 0 {
+		status = CheckWarn
```
nandadev, line 1296: Any service below 100% → `CheckWarn` → `PipelineDegraded`. Since virtually every real stack has at least one imperfect service, `kg diagnose` will now report "degraded" nearly everywhere, which dilutes the signal for actual pipeline problems (especially for agents keying off `Orientation.PipelineHealth`).

Consider either excluding this check from the pipeline-health rollup (instrumentation debt isn't a pipeline fault), or gating the warn on severity — e.g. median of the sampled window below a threshold, or any `CRITICAL`-impact failure — so a lone straggler at 95% doesn't flip the verdict.

Related nit: the PR description mentions "median/worst %" and "paged up to 500 reports", but the code computes only `WorstQuality` over one page of 50 — worth reconciling.

internal/providers/kg/quality.go
```diff
@@ -0,0 +1,282 @@
...
+	return errors.New("table format does not support decoding")
+}
+
+// QualityReportTableCodec renders a single entity quality report as a summary
+// header plus a table of checks.
+type QualityReportTableCodec struct{}
+
+func (c *QualityReportTableCodec) Format() format.Format { return "table" }
+
+func (c *QualityReportTableCodec) Encode(w io.Writer, v any) error {
```
nandadev, line 254: The default table renders only `CHECK | STATE | IMPACT | TITLE`, dropping the metadata that makes the report actionable — `description`, `docUrl`, and `reference` are only reachable via `-o json`/`-o yaml`, despite the PR description promising remediation doc links.

Suggestion: print a short detail block (description + docUrl + reference) for each non-`SUCCESS` check after the table — a footer rather than a URL column, since columns truncate. Query templates can stay JSON-only.

FCalero: done! 

image

internal/providers/kg/diagnose.go
FCalero: Fixed

nandadev (approved):

annanay25 (approved):

PR #1026, opened 2026-07-21
feat(profiles): add time-range flags to labels and profile-types

aleks-p:
`labels` and `profile-types` query a fixed last-1h window: the time flags every other profiles subcommand has (`--since`/`--from`/`--to`) aren't registered on them, and an empty result says nothing about the window that was searched — `{"names":null}`, exit 0. A tenant with no data and a tenant whose data sits 61 minutes in the past are indistinguishable from the output, and there's no way to widen the search even after diagnosing the cause.

In an internal profiling-agent benchmark this was the largest single source of wasted agent steps: agents ran `labels`/`profile-types` against data anchored a few hours back, got silent empties, retried with `-v 3` and `--log-http-payload` (neither reveals the window), guessed `--from`/`--to` (unknown flag on these two subcommands), and eventually fell back to raw curl against the datasource proxy.

Changes:
- register the shared `TimeRangeOpts` flags (`--since`/`--from`/`--to`) on both commands and pass the range through — the request structs already had `Start`/`End` fields, they were just never populated
- on an empty result, emit a stderr hint naming the effective window:

```
$ gcx profiles labels -d <uid>
{"names":null}
hint: no labels found in the default window (last 1h: 2026-07-16T19:29:40Z to 2026-07-16T20:29:40Z) — pass --since (e.g. --since 24h) or --from/--to to search further back
```

Agent mode gets the typed `{"class":"hint",...}` JSONL form; stdout payloads are unchanged in both modes, as is the default window when no flags are passed.

Notes:
- verified end-to-end against a Pyroscope holding only 5h-old data: `--since 6h` returns the labels/values/types that are silently dropped today
- will rebase once #1009 (`profile-types` → `list-profile-types`) lands; these changes are orthogonal to the rename

### DIFF
--- a/docs/reference/cli/gcx_datasources_pyroscope_labels.md
+++ b/docs/reference/cli/gcx_datasources_pyroscope_labels.md
@@ -20,6 +20,9 @@ gcx datasources pyroscope labels [flags]
 	# Get values for a specific label
 	gcx datasources pyroscope labels -d UID --label service_name
 
+	# Search a wider window than the default last hour
+	gcx datasources pyroscope labels -d UID --since 24h
+
 	# Output as JSON
 	gcx datasources pyroscope labels -d UID -o json
 ```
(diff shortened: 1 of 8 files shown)


awsome-o:
Heads-up: #1009 renames this exact command — profile-types becomes list-profile-types at both mounts (part of the #387 naming work, agreed with @simonswine and Dafydd). It's a GA item slated to merge today, so it'll likely land first and this PR will need a rebase: the command file moves to list_profile_types.go, the builder is now ListProfileTypesCmd, and the reference docs pages get regenerated under the new name. Your changes here are additive (flags + hints) so the rebase should be mechanical — happy to help with it or push the rebase to this branch if you'd like, just say the word. Sorry for the churn, and thanks for fixing the fixed-window blindspot, it's a real gap.

aleks-p:
Thanks @awsome-o, I've rebased it now.

dafydd-t (approved):

PR #985, opened 2026-07-15
docs: add anonymous usage statistics page

dafydd-t:
Public docs page for gcx anonymous usage stats, requested in grafana/gcx-internal#2: every collected field, the anonymity properties (per-install random device ID, delete-to-reset), what is never collected, the server-side region/whois enrichment, and opt-out via `GCX_TELEMETRY`, `DO_NOT_TRACK`, or the `diagnostics.telemetry` config field.

Modelled on [Tempo's equivalent page](https://grafana.com/docs/tempo/latest/configuration/anonymous-usage-reporting/), formatted like the sibling pages in `docs/sources/`.

**Blocked on #978**: the page documents the lifecycle wiring (suppression list, `log` mode, the config field having effect), which only exists on that branch - do not merge before it. Reporting itself remains off by default either way. This page is also an input to the privacy/legal review.

PR #1027, opened 2026-07-21
feat(telemetry): first-run notice and default-on usage stats

dafydd-t:
## Summary

- Do a one-time  notice to stderr telling interactive users thattelemetry is on and how to opt out.
- flip the telemetry default from `disabled` to `enabled`
- record the default in the config and env-var reference docs; the public docs page (#985) already documents enabled-by-default

## details

Renders as:

```
gcx collects anonymous usage statistics: command paths, flag names, and outcomes - never argument values, resource names, or hosts.
Opt out: set GCX_TELEMETRY=disabled, or diagnostics.telemetry: disabled in the gcx config file.
Details: https://grafana.com/docs/grafana/latest/as-code/observability-as-code/grafana-cli/gcx/anonymous-usage-statistics.md
```

If the state dir is unwritable the notice is skipped entirely.

The docs URL is added to the `internal/docs` registry (`.md` per the registry invariant); the page merged in #985 but hasn't synced to the website yet, so the link 404s until the next docs sync.

we don't show the message in non-interactive environments, as it just adds noise

### DIFF
--- a/cmd/gcx/telemetry.go
+++ b/cmd/gcx/telemetry.go
@@ -40,7 +40,14 @@ func emitUsageEvent(cmd *cobra.Command, start time.Time, exitCode int) {
 		return
 	}
 
-	switch telemetry.ResolveMode(diagnosticsTelemetryValue) {
+	mode := telemetry.ResolveMode(diagnosticsTelemetryValue)
+
+	// One-time opt-out notice for interactive users; the command's own output
+	// has already been written by this point.
+	_, isCI := telemetry.DetectCI()
+	telemetry.MaybeShowFirstRunNotice(os.Stderr, mode, terminal.StdoutIsTerminal(), isCI, agent.IsAgentMode())
+
+	switch mode {
 	case telemetry.ModeLog:
 		if data, err := json.Marshal(buildUsageEvent(info, start, exitCode)); err == nil {
 			fmt.Fprintln(os.Stderr, string(data))
--- a/cmd/gcx/telemetry.go
+++ b/cmd/gcx/telemetry.go
@@ -43,9 +43,11 @@ func emitUsageEvent(cmd *cobra.Command, start time.Time, exitCode int) {
 	mode := telemetry.ResolveMode(diagnosticsTelemetryValue)
 
 	// One-time opt-out notice for interactive users; the command's own output
-	// has already been written by this point.
+	// has already been written by this point. Gated on stderr's TTY state
+	// because that is where the notice goes: piped stdout must not hide it,
+	// and discarded stderr must not consume the one-shot flag.
 	_, isCI := telemetry.DetectCI()
-	telemetry.MaybeShowFirstRunNotice(os.Stderr, mode, terminal.StdoutIsTerminal(), isCI, agent.IsAgentMode())
+	telemetry.MaybeShowFirstRunNotice(os.Stderr, mode, terminal.StderrIsTerminal(), isCI, agent.IsAgentMode())
 
 	switch mode {
 	case telemetry.ModeLog:
--- a/docs/reference/configuration/index.md
+++ b/docs/reference/configuration/index.md
@@ -193,7 +193,7 @@ diagnostics:
   log-dir: string
   # Telemetry controls anonymous usage telemetry: "enabled", "disabled",
   # or "log" (prints to stderr). Enabled by default. Overridden by the
-  # GCX_TELEMETRY and DO_NOT_TRACK environment variables.
+  # GCX_TELEMETRY environment variable.
   telemetry: string
 
 ```
--- a/docs/reference/configuration/index.md
+++ b/docs/reference/configuration/index.md
@@ -192,8 +192,8 @@ diagnostics:
   # Default: $XDG_STATE_HOME/gcx/ (platform-specific).
   log-dir: string
   # Telemetry controls anonymous usage telemetry: "enabled", "disabled",
-  # or "log" (prints to stderr). Overridden by the GCX_TELEMETRY and
-  # DO_NOT_TRACK environment variables.
+  # or "log" (prints to stderr). Enabled by default. Overridden by the
+  # GCX_TELEMETRY and DO_NOT_TRACK environment variables.
   telemetry: string
 
 ```
--- a/docs/reference/environment-variables/index.md
+++ b/docs/reference/environment-variables/index.md
@@ -21,8 +21,8 @@ disables the notifier (NO_COLOR convention).
 
 Telemetry controls anonymous usage telemetry for this invocation:
 "enabled", "disabled", or "log" (print the event to stderr and send
-nothing). Takes precedence over DO_NOT_TRACK and the
-`diagnostics.telemetry` config field.
+nothing). Telemetry is enabled by default. Takes precedence over
+DO_NOT_TRACK and the `diagnostics.telemetry` config field.
 
 ## `GCX_TELEMETRY_ENDPOINT`
 
--- a/docs/reference/environment-variables/index.md
+++ b/docs/reference/environment-variables/index.md
@@ -21,8 +21,9 @@ disables the notifier (NO_COLOR convention).
 
 Telemetry controls anonymous usage telemetry for this invocation:
 "enabled", "disabled", or "log" (print the event to stderr and send
-nothing). Telemetry is enabled by default. Takes precedence over
-the `diagnostics.telemetry` config field.
+nothing). Telemetry is enabled by default. Any other non-empty value
+disables telemetry. Takes precedence over the `diagnostics.telemetry`
+config field.
 
 ## `GCX_TELEMETRY_ENDPOINT`
 
--- a/docs/reference/environment-variables/index.md
+++ b/docs/reference/environment-variables/index.md
@@ -22,7 +22,7 @@ disables the notifier (NO_COLOR convention).
 Telemetry controls anonymous usage telemetry for this invocation:
 "enabled", "disabled", or "log" (print the event to stderr and send
 nothing). Telemetry is enabled by default. Takes precedence over
-DO_NOT_TRACK and the `diagnostics.telemetry` config field.
+the `diagnostics.telemetry` config field.
 
 ## `GCX_TELEMETRY_ENDPOINT`
 
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -134,7 +134,7 @@ type DiagnosticsConfig struct {
 
 	// Telemetry controls anonymous usage telemetry: "enabled", "disabled",
 	// or "log" (prints to stderr). Enabled by default. Overridden by the
-	// GCX_TELEMETRY and DO_NOT_TRACK environment variables.
+	// GCX_TELEMETRY environment variable.
 	Telemetry string `json:"telemetry,omitempty" yaml:"telemetry,omitempty"`
 }
 
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -133,8 +133,8 @@ type DiagnosticsConfig struct {
 	LogDir string `json:"log-dir,omitempty" yaml:"log-dir,omitempty"`
 
 	// Telemetry controls anonymous usage telemetry: "enabled", "disabled",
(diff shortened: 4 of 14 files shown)


docs/reference/configuration/index.md
```diff
@@ -151,8 +151,8 @@ diagnostics:
   # Default: $XDG_STATE_HOME/gcx/ (platform-specific).
   log-dir: string
   # Telemetry controls anonymous usage telemetry: "enabled", "disabled",
-  # or "log" (prints to stderr). Overridden by the GCX_TELEMETRY and
-  # DO_NOT_TRACK environment variables.
+  # or "log" (prints to stderr). Enabled by default. Overridden by the
+  # GCX_TELEMETRY and DO_NOT_TRACK environment variables.
```
dafydd-t:
```suggestion
  # or "log" (prints to stderr). Enabled by default. Overridden by the
  # GCX_TELEMETRY environment variable.
```

docs/reference/environment-variables/index.md
```diff
@@ -21,8 +21,8 @@ disables the notifier (NO_COLOR convention).
 
 Telemetry controls anonymous usage telemetry for this invocation:
 "enabled", "disabled", or "log" (print the event to stderr and send
-nothing). Takes precedence over DO_NOT_TRACK and the
-`diagnostics.telemetry` config field.
+nothing). Telemetry is enabled by default. Takes precedence over
+DO_NOT_TRACK and the `diagnostics.telemetry` config field.
```
dafydd-t:
```suggestion
nothing). Telemetry is enabled by default. Takes precedence over
the `diagnostics.telemetry` config field.
```

internal/config/types.go
```diff
@@ -69,8 +69,8 @@ type DiagnosticsConfig struct {
 	LogDir string `json:"log-dir,omitempty" yaml:"log-dir,omitempty"`
 
 	// Telemetry controls anonymous usage telemetry: "enabled", "disabled",
-	// or "log" (prints to stderr). Overridden by the GCX_TELEMETRY and
-	// DO_NOT_TRACK environment variables.
+	// or "log" (prints to stderr). Enabled by default. Overridden by the
+	// GCX_TELEMETRY and DO_NOT_TRACK environment variables.
```
dafydd-t:
```suggestion
	// or "log" (prints to stderr). Enabled by default. Overridden by the
	// GCX_TELEMETRY environment variable.
```

internal/docs/links.go
```diff
@@ -75,6 +75,11 @@ const (
 
 	// IRMInvoice documents how Grafana IRM is billed per monthly active user.
 	IRMInvoice = "https://grafana.com/docs/grafana-cloud/cost-management-and-billing/manage-invoices/understand-your-invoice/irm-invoice.md"
+
+	// AnonymousUsageStats documents gcx's anonymous usage statistics: what is
+	// collected and how to opt out. Referenced by the first-run telemetry
+	// notice.
+	AnonymousUsageStats = "https://grafana.com/docs/grafana/latest/as-code/observability-as-code/grafana-cli/gcx/anonymous-usage-statistics.md"
```
dafydd-t, line 97: We need to wait for this to be live before merging

internal/config/types.go
dafydd-t: applied in e5d0ef01

docs/reference/configuration/index.md
dafydd-t: applied - this file is generated from the types.go comment, so the fix landed there and this was regenerated (e5d0ef01)

docs/reference/environment-variables/index.md
dafydd-t: applied - generated from the telemetry.go Env comment, fixed at source and regenerated (e5d0ef01). DO_NOT_TRACK precedence stays documented under its own entry

internal/docs/links.go
dafydd-t: agreed - PR stays draft until the docs page is live and the sign-offs land; both called out in the PR body

internal/telemetry/telemetry.go
```diff
@@ -25,18 +25,19 @@ const (
 )
 
 // defaultMode is the resolved mode when no env var or config setting applies.
-// It stays disabled until privacy/legal and usage-stats owner sign-off clear;
-// flipping to ModeEnabled is deliberately a one-line change gated on those.
-const defaultMode = ModeDisabled
+// Telemetry is opt-out: enabled by default, disabled via GCX_TELEMETRY,
+// DO_NOT_TRACK, or the diagnostics.telemetry config value. Interactive users
+// are told about this once via the first-run notice (firstrun.go).
+const defaultMode = ModeEnabled
```
awsome-o, line 31: With enabled as the default, the synchronous export now runs before os.Exit on every non-suppressed command, and the shared client's retry transport retries connection errors with a 500ms minimum backoff inside the 1s timeout, so an unreachable endpoint burns the whole budget. I reproduced it: gcx providers takes 0.03s with telemetry disabled and a consistent 1.03s with it enabled, same number for connection-refused and blackholed endpoints. As opt-in that was a fair trade, as the default it's a CLI-wide slowdown for anyone offline or firewalled, and agent loops pay it on every invocation. Could the telemetry POST use a single-attempt transport, so an unavailable endpoint costs one fast failure instead of the full timeout?

cmd/gcx/telemetry.go
```diff
@@ -40,7 +40,14 @@ func emitUsageEvent(cmd *cobra.Command, start time.Time, exitCode int) {
 		return
 	}
 
-	switch telemetry.ResolveMode(diagnosticsTelemetryValue) {
+	mode := telemetry.ResolveMode(diagnosticsTelemetryValue)
+
+	// One-time opt-out notice for interactive users; the command's own output
+	// has already been written by this point.
+	_, isCI := telemetry.DetectCI()
+	telemetry.MaybeShowFirstRunNotice(os.Stderr, mode, terminal.StdoutIsTerminal(), isCI, agent.IsAgentMode())
```
awsome-o: The notice is written to stderr but gated on stdout's TTY state. Someone whose gcx use is always piped, like gcx ... --output json | jq, never sees the notice even though stderr is a visible terminal, while events are exported on every run. The inverse also bites: stdout on a TTY with stderr sent to /dev/null consumes the one-shot flag while the notice goes nowhere. Since the notice targets stderr, could the gate check stderr instead?

internal/docs/links.go
```diff
@@ -75,6 +75,11 @@ const (
 
 	// IRMInvoice documents how Grafana IRM is billed per monthly active user.
 	IRMInvoice = "https://grafana.com/docs/grafana-cloud/cost-management-and-billing/manage-invoices/understand-your-invoice/irm-invoice.md"
+
+	// AnonymousUsageStats documents gcx's anonymous usage statistics: what is
+	// collected and how to opt out. Referenced by the first-run telemetry
+	// notice.
```
awsome-o, line 96: The page is live now, but two things on it don't match the code. It documents a gcx telemetry command ("the command that controls reporting doesn't report on itself") that isn't mounted, so the first thing a curious user tries gives an unknown-command error. And its error_kind examples, auth and validation, don't match what we emit: usage_error, auth_failure, partial_failure, version_incompatible, error. Since this page is the disclosure the notice links to, could those be corrected before this merges? Also, the notice prints the .md URL, which serves raw markdown. Right for the agent-facing registry, but the human notice would be friendlier with the rendered trailing-slash URL.

internal/telemetry/firstrun.go
```diff
@@ -0,0 +1,52 @@
...
+
+// firstRunNotice is the one-time message telling interactive users that
+// anonymous usage stats are on and how to opt out.
+const firstRunNotice = `gcx collects anonymous usage statistics so we can make gcx better. We do not collect potentially identifiable or sensitive information like argument or flag values or resource names.
+You can opt out by setting GCX_TELEMETRY=disabled, or putting diagnostics.telemetry: disabled in your gcx config file.
+Find out more at ` + docs.AnonymousUsageStats + "\n"
+
+// FirstRunNoticePath returns the flag file that records the notice was shown.
+func FirstRunNoticePath() string {
+	return filepath.Join(xdg.StateHome(), "gcx", firstRunNoticeFileName)
```
awsome-o: With HOME and XDG_STATE_HOME both unset, xdg.StateHome() returns an empty string, so the device-id and notice paths go relative and MkdirAll creates them in the current directory. I reproduced it: a single run in an env -i environment created ./gcx/device-id in the cwd. With the default flipped on, every HOME-less setup (minimal containers, service users) starts doing this, and a stray gcx/ directory in a repo checkout is the kind of thing that gets committed. Could an empty state home be treated as non-persistable, skipping the device id and notice bookkeeping?

internal/telemetry/telemetry.go
```diff
@@ -25,18 +25,19 @@ const (
...
 // tags are read by scripts/env-vars-reference (docs generation); resolution
 // itself happens in ResolveMode.
 type Env struct {
 	// Telemetry controls anonymous usage telemetry for this invocation:
 	// "enabled", "disabled", or "log" (print the event to stderr and send
-	// nothing). Takes precedence over DO_NOT_TRACK and the
-	// `diagnostics.telemetry` config field.
+	// nothing). Telemetry is enabled by default. Takes precedence over
+	// the `diagnostics.telemetry` config field.
 	Telemetry string `env:"GCX_TELEMETRY"`
```
awsome-o, line 42: Small one, feel free to push back. parseMode accepts only enabled, disabled, and log, and DO_NOT_TRACK only 1 and true, with everything else falling through to the default, which is now enabled. So GCX_TELEMETRY=off or =false, which used to harmlessly stay off, now silently opts in. Those spellings were never documented, so maybe this is fine, but for an opt-out control a typo that keeps telemetry on is the failure mode I'd least want to ship. Would a stderr warning on unrecognized values be worth a line? Related wording nit: the notice says to put diagnostics.telemetry: disabled in the config file, and pasted literally that's an unknown key under strict parsing, config-loading commands then fail loudly so it's recoverable, but showing the actual two-line YAML would remove the guesswork.

awsome-o (changes requested):
The mechanism itself looks solid, the show-once bookkeeping and the unwritable state dir handling are well tested. But I reproduced a couple of things that I think need fixing before the default flips on, details inline.

internal/telemetry/telemetry.go
dafydd-t: Yeah i think this is good enough reason to move away form the common transport. Good shout

dafydd-t: applied in 28fff3ff - plain http.Client with the same 1s timeout, User-Agent set on the request directly. added a test pinning exactly one attempt on a 503

cmd/gcx/telemetry.go
dafydd-t: applied in 330b9c92 - added terminal.StderrIsTerminal and the notice gate now checks that. the IsTTY event field stays on stdout, that one is an output-shape signal not a notice gate

internal/telemetry/firstrun.go
dafydd-t: applied in 5d3c0258 - DeviceIDPath and FirstRunNoticePath return "" when no state home is known, DeviceID goes ephemeral (persisted=false) and the notice is skipped, matching the unwritable-dir behaviour. guarded in telemetry rather than xdg so other StateHome consumers keep their current behaviour

internal/telemetry/telemetry.go
dafydd-t: went a slightly different way in 1a47d8db: unrecognised non-empty values now resolve to disabled instead of warning - a warning would repeat on every invocation since env vars persist, and for an opt-out control a typo should fail toward privacy. DO_NOT_TRACK keeps the cross-tool convention (only 1/true disable). the notice now shows the actual two-line YAML too

internal/docs/links.go
dafydd-t: Addressed in https://github.com/grafana/gcx/pull/1038

awsome-o (approved):

PR #1028, opened 2026-07-21
feat(telemetry): capture guessed commands and flags as parse_error events

### LINKED ISSUE #578: Telemetry: capture guessed/nonexistent commands and flags

dafydd-t:
closes #578

## Summary

- Adds the `parse_error` outcome to usage events: unknown commands, unknown flags, and argument-validation failures are now recorded instead of suppressed. This is the "404 log" for the CLI - what people and agents *try* that doesn't exist.
- `FallbackTelemetryInfo` classifies the failed parse by walking the command tree the way cobra's `Find` does: `parse_error_kind`, `parse_error_parent` (deepest valid command), `parse_error_token` (first unknown token), `parse_error_nearest`/`parse_error_distance` (cobra's suggestion rule, `-1` when nothing is near).
- Unknown flags are captured via the root `SetFlagErrorFunc` hook (propagates to children); only flag **names** are recorded, and flag errors that embed values (e.g. `invalid argument "x" for --limit`) record nothing from the message.
- Privacy is enforced by a command-shape filter on the token: lowercase kebab, max 20 chars, at most one hyphen, no digits, an entropy bound, and stdlib URL/IP/UUID checks. Failing tokens become `<redacted>` while the distance is kept, so novel guesses still count. `attempted_command` is truncated at the unknown token, and flag values can never be mistaken for the token (unknown flags conservatively consume the next arg, matching cobra's `stripFlags`).
- Parse errors under suppressed commands (`completion`, `version`, `telemetry`) stay unrecorded; typos *of* those commands (e.g. `gcx versoin`) still count.
- Golden PII test pins down exactly what leaves the process for every failure shape: typo, novel guess, truncation, flag-value skipping, and secret-looking/UUID/URL tokens.

No wire or server changes - the `parse_error_*` fields already exist in the event schema and usage-stats columns.

## Example events (`GCX_TELEMETRY=log`)

**We need to be very cautious of accidentally emitting sensitive or identifiable data in these error events. Are we being defensive enough?** 

Typo of a real command (`gcx dashboards serch`):

```json
{"outcome":"parse_error","exit_code":2,"error_kind":"usage_error","parse_error_kind":"unknown_command","parse_error_parent":"dashboards","parse_error_token":"serch","attempted_command":"dashboards serch","parse_error_nearest":"search","parse_error_distance":1}
```

Novel top-level guess (`gcx deploy`) - no near match, still counted:

```json
{"outcome":"parse_error","exit_code":2,"error_kind":"usage_error","parse_error_kind":"unknown_command","parse_error_token":"deploy","attempted_command":"deploy","parse_error_distance":-1}
```

Unknown flag (`gcx resources get --frmat json`) - name only, never the value:

```json
{"outcome":"parse_error","exit_code":1,"parse_error_kind":"unknown_flag","parse_error_parent":"resources get","attempted_command":"resources get","parse_error_flags":"frmat","parse_error_distance":-1}
```

Token failing the shape filter (`gcx some-uuid-looking-token-4f3a2b1c`) - redacted, distance preserved:

```json
{"outcome":"parse_error","exit_code":2,"parse_error_kind":"unknown_command","parse_error_token":"<redacted>","attempted_command":"<redacted>","parse_error_distance":-1}
```

(envelope fields trimmed for readability; full events carry the usual service/version/os/device_id/agent fields)

## Notes

- Entropy threshold is 3.7 bits/char: real command words sit around 3.2, and exceeding 3.7 needs 13+ distinct characters, which reads as a random identifier. The charset/length/hyphen rules do most of the redaction work; entropy is the backstop the design asks for.
- Cobra only assigns `SuggestionsMinimumDistance` lazily while rendering an unknown-command error, so the classifier applies the same default (2) itself and reuses `SuggestionsFor` for command matching; flags get the identical rule.

### DIFF
--- a/cmd/gcx/root/command.go
+++ b/cmd/gcx/root/command.go
@@ -226,7 +226,9 @@ func newCommand(version string, pp []providers.Provider) *cobra.Command {
 		},
 	}
 
-	rootCmd.SetFlagErrorFunc(func(_ *cobra.Command, err error) error {
+	rootCmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
+		// Record the failure for the parse_error usage event (#578).
+		recordFlagFailure(cmd, err)
 		if strings.Contains(err.Error(), "--log-http-payload") && strings.Contains(err.Error(), "flag has been renamed") {
 			return errors.New("--log-http-payload has been renamed; use --insecure-log-http-payload instead")
 		}
--- a/cmd/gcx/root/command.go
+++ b/cmd/gcx/root/command.go
@@ -227,12 +227,12 @@ func newCommand(version string, pp []providers.Provider) *cobra.Command {
 	}
 
 	rootCmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
-		// Record the failure for the parse_error usage event (#578).
-		recordFlagFailure(cmd, err)
 		if strings.Contains(err.Error(), "--log-http-payload") && strings.Contains(err.Error(), "flag has been renamed") {
-			return errors.New("--log-http-payload has been renamed; use --insecure-log-http-payload instead")
+			err = errors.New("--log-http-payload has been renamed; use --insecure-log-http-payload instead")
 		}
-		return err
+		// Carry the failing command out with the error so the exit path can
+		// classify the parse_error usage event.
+		return &flagParseError{cmd: cmd, err: err}
 	})
 
 	defaultHelp := rootCmd.HelpFunc()
--- a/cmd/gcx/root/parsefailure.go
+++ b/cmd/gcx/root/parsefailure.go
@@ -1,23 +1,21 @@
 package root
 
-// Parse-failure capture (#578): classify invocations cobra rejected before any
-// hook ran (unknown command, unknown flag, argument validation) into an
-// anonymous usage event. Privacy invariant: only command-shaped tokens and
-// flag names are ever recorded, never argument or flag values.
-
 import (
+	"context"
+	"errors"
 	"math"
 	"net"
 	"net/url"
 	"regexp"
+	"slices"
 	"strings"
-	"sync/atomic"
 
+	"github.com/grafana/gcx/internal/config"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
-// ParseFailure kinds.
+// Categories of parse failure.
 const (
 	parseErrorUnknownCommand = "unknown_command"
 	parseErrorUnknownFlag    = "unknown_flag"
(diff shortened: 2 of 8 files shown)


cmd/gcx/root/command.go
```diff
@@ -226,7 +226,9 @@ func newCommand(version string, pp []providers.Provider) *cobra.Command {
 		},
 	}
 
-	rootCmd.SetFlagErrorFunc(func(_ *cobra.Command, err error) error {
+	rootCmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
+		// Record the failure for the parse_error usage event (#578).
```
dafydd-t:
```suggestion
		// Record the failure so we can create a usage event from it.
```

cmd/gcx/root/parsefailure.go
```diff
@@ -0,0 +1,321 @@
+package root
+
+// Parse-failure capture (#578): classify invocations cobra rejected before any
+// hook ran (unknown command, unknown flag, argument validation) into an
+// anonymous usage event. Privacy invariant: only command-shaped tokens and
+// flag names are ever recorded, never argument or flag values.
```
dafydd-t: Put this on recordFlagFailure function docstring

cmd/gcx/root/parsefailure.go
```diff
@@ -0,0 +1,321 @@
...
+
+// redactedToken replaces tokens that fail the command-shape filter.
+const redactedToken = "<redacted>"
+
+// suggestionsMinimumDistance mirrors cobra's default SuggestionsMinimumDistance;
+// cobra only assigns it lazily while rendering an unknown-command error.
+const suggestionsMinimumDistance = 2
+
+// ParseFailure describes a failed command-line parse for the usage event
+// (#578). Token and Flags carry shape-filtered names only, never values.
```
dafydd-t:
```suggestion
// ParseFailure describes a failed command-line parse for the usage event.
// Token and Flags carry shape-filtered names only, never values.
```

cmd/gcx/root/parsefailure.go
```diff
@@ -0,0 +1,321 @@
...
+type ParseFailure struct {
+	Kind      string // unknown_command, unknown_flag or invalid_args
+	Parent    string // deepest valid command path reached; "" is the root
+	Token     string // first unknown token; redactedToken if it fails the shape filter
+	Attempted string // Parent plus Token, truncated at the unknown token
+	Flags     string // unknown flag names only, never values
+	Nearest   string // nearest real command or flag; "" if no near match
+	Distance  int    // Levenshtein distance to Nearest; -1 if no near match
```
dafydd-t:
```suggestion
	Kind      string 
	Parent    string // deepest valid command path reached; "" is the root
	Token     string // first unknown token; redactedToken if it fails the shape filter
	Attempted string // Parent plus Token, truncated at the unknown token
	Flags     string // unknown flag names only, never values
	Nearest   string // nearest real command or flag; "" if no near match
	Distance  int    // Levenshtein distance to Nearest; -1 if no near match
```

Issue #1031, opened 2026-07-21
Output workstream: list truncation contract and shared --limit semantics (#387)


awsome-o:
Part of #387. Filed per the #994 review discussion: the output-format and `--limit`/truncation work is tracked separately from the naming batches, as a sub-issue of #387 with its own PRs.

The starting point was Dafydd's ask on #387: wire in a global `--limit` if we can, and add a hint when we've truncated the response due to the limit. PR #988 made a first attempt at the truncation half for two commands and review found real problems with it (silent hard-cap truncation, broken `--json` field selection and discovery, hint wording). Before writing more code we ran a proper investigation, recorded in `docs/research/2026-07-17-global-limit-investigation.md` (lands with the groundwork PR).

What the investigation found:

There are ~64 non-test local `--limit` flag registrations today, with three systemic defects: many list commands truncate client-side with no signal at all, `0` variously means "all", "server default", "invalid", or "all up to a hidden cap", and there is no machine-readable way for an agent reading `-o json` to tell a truncated page from the complete set.

A root-level persistent `--limit` was investigated and rejected. Cobra merges persistent flags into each command's flag set and silently skips names that already exist, so every one of the ~64 local flags would shadow the root flag — the "global" flag would only reach commands that never declared one, including dozens of non-list commands where it's meaningless. A single global default also can't vary per command, and at least six existing `--limit` families aren't "number of list rows" at all (Loki's flag caps log lines, the SQL datasources rewrite a `LIMIT` clause, dashboards list uses it as a K8s page size paired with `--continue`, incidents list rejects `0`, incidents activity coerces `0` to a server default, resources get uses it as a per-type page size). Those must never be captured by a uniform contract.

What we're doing instead, in two stages:

[...]

Issue #1030, opened 2026-07-21
Output format consistency: safe corrections vs breaking default-format changes (#387 output track)


awsome-o:
This issue started as a decision menu from the output audit (docs/research/2026-07-17-output-format-consistency-audit.md on the #1032 branch). It is now two decisions with one recommendation each; the original menu body is preserved in this issue's edit history. Reply "approved" to take both, or name the row you want changed. #1032 is closed unmerged; its findings get re-evaluated and reused in the implementation PR that follows this decision, so nothing lands on provisional policy.

Framing corrections from re-verification, so the scope is honest:

Two items the audit listed as open decisions are not open. CONSTITUTION.md already says "STDOUT is the result, STDERR is the diagnostic — failure details and progress feedback go to stderr", so the 109-stdout status-message call sites are debt against a settled rule (output.md §1.4 says the opposite of the Constitution and gets corrected either way). And it already says "CRUD mutation commands output structured operation summaries", so mutation output needs an ADR about the envelope shape, not a decision about direction. Both get follow-up issues. For the same reason, mutation echo defaults are NOT normalized here — converting today's object echoes to one format would be an intermediate breaking change ahead of the envelope work. Known consequence: cloud stacks create/update keep their table echo until that ADR, so the create → inspect shape change stays for now.

On agents: these default changes are invisible to detected agent-mode invocations, because agent mode overrides every per-command default with the agents codec (internal/output/format.go). Undetected agents, --agent=false callers, and plain scripts consume the human defaults, so this is not agent-neutral — it is a default-visibility change for anything not running in detected agent mode. Machine consumers should pass an explicit format; -o json / -o yaml / agents output shapes are untouched by everything here.

Census on main (from the generated CLI reference): 64 get commands split 37 yaml / 14 table / 7 json / 6 text; 80 list commands split 75 table / 5 text.

## Decision 1 — human read defaults

Collections default to a table; an ordinary single-resource get defaults to yaml. The rule is semantic: a list is a summary you scan, a get is the complete object. Yaml-for-get ratifies the shipped majority (37 of 64), so this is convergence on the dominant convention, not a new invention.

Exactly 27 defaults change:

[...]

PR #1032, opened 2026-07-21
fix(output): honest output handling for pull/edit and typed hints (#387 output track)

awsome-o:
This is the safe-corrections extraction from the #387 output-format audit, tracked in #1030. The full audit record rides along as docs/research/2026-07-17-output-format-consistency-audit.md. The bar for inclusion was: makes existing behavior honest without changing any non-agent default format or any successful non-agent stdout payload.

What's in:

- resources pull and edit no longer let agent mode flip their default to the agents display codec. Their output format doubles as the on-disk file extension, the encoder, and (for edit) the round-trip decode format, so agent invocations wrote `<name>.agents` files — and resources over the 100KiB spill threshold got a spill-summary envelope written into the resource file instead of the resource. Both now pin json (their existing non-agent default, unchanged) via the new Options.PinDefaultFormat, reject an explicit -o agents at validation time, and hide agents from their format menus via the new Options.HideFormat, so help never advertises a format the command refuses.
- edit now actually calls its own Validate() (it never did), and both pull and edit reject --json/--jq upfront. pull advertised both flags but silently ignored them; a field-selected or jq-transformed document can't round-trip through the editor or as a resource file that push reads back. The rejections are typed usage errors (exit 2 per the DESIGN.md taxonomy).
- The truncation and pagination hints in kg entities list, resources get, and the agents codec's own spill notice now go through output.EmitHint — a "hint: ..." line on a TTY, a {"class":"hint"} JSONL record in agent mode. kg also stops writing to raw os.Stderr past the command's redirected stream, and resources get now emits its truncation hint on the --json field-selection path too, which previously dropped it for exactly the field-selecting consumers the hint exists for.
- IRM oncall list options: a stale comment and dead defaultFmt reassignments removed. No behavior change.
- docs/design/output.md §14 rewritten: it documented a --format flag on pull (yaml default) that never existed. It now describes the real -o flag and json default; the yaml design intent is recorded as an open decision, not silently adopted.

Deliberate behavior changes a reviewer should sign off on explicitly — this list is complete, no need to read #1030 to review this PR:

1. The agent-mode default for pull and edit changes from agents to json. Non-agent defaults are untouched, but this one is a real default change — it's in the safe bucket because the old default was writing display envelopes into resource files, not because nothing moved.
2. Previously-tolerated invocations now fail: pull -o agents used to exit 0 and write .agents files; pull --json/--jq were advertised but silently ignored; edit --json/--jq were inert because edit never validated. All six combinations now exit 2 with a clear error before any fetch or editor launch. Any automation that passed these flags by accident will start failing loudly.
3. The resources get truncation notice changed wording and stream shape: "Showing first N items per resource type." became a typed hint ("hint: showing first N items per resource type; ..." on a TTY, JSONL in agent mode). Anything grepping the old literal stops matching. That's the point of the FR-104 conversion, but it's not byte-neutral on stderr.

One decision surfaces here that lives outside the PR: CONSTITUTION.md's Push/Pull Philosophy says pull writes YAML by default. We traced that line — it was added 2026-03-25 in PR #17 as codified design intent, and on that same day the code already defaulted pull to json. The sentence has never matched shipped behavior at any point in the repo's history (same shape as the output.md §14 `--format` flag this PR fixes: documented, never implemented). This PR documents the shipped json behavior in §14 with the conflict called out inline, and changes nothing either way. The call needed from a maintainer: ratify json — follow-up is a one-line Constitution amendment PR we'll put up on a nod — or schedule a yaml migration as its own breaking change (it would break every workflow reading `.json` files from a bare pull). Either answer is fine for this PR; it doesn't block merge.

Follow-ups after merge, for the record: one rebase here once the naming train lands (kg/irm overlap, ours to do); the Constitution amendment or yaml-migration PR per the decision above; and PR #1033's rebase absorbs the output.md overlap (its workstream). The rest of the audit's breaking default-format questions are queued in #1030 and none of them touch this PR.

What's deliberately out, so the boundary is visible in review: every breaking default-format change the audit surfaced — get-default convergence, the adaptive family's table/json/yaml split, cloud stacks get vs its siblings, the pull json-vs-yaml on-disk default, the text/table canonical spelling (a resolution-time alias was tried on the feasibility branch and reverted after it leaked help-tree's prose codec into -o table), and the output.md §1.3 policy-vs-reality call. Those are presented as a decision table in #1030 and none of them are implemented here. Also out: the list-truncation contract (Track C), which is its own workstream in #1031 / PR #1033 — nothing here depends on it, and this PR should merge before #1033 (both touch output.md and format.go; #1033's rebase absorbs the overlap).

Verification: gofmt clean, golangci-lint 0 issues, go test -race -count=1 ./... green, reference docs regenerated with GCX_AGENT_MODE=false and drift-clean (the two pull/edit CLI pages pick up the menu changes).

The diff also went through an adversarial multi-agent review at xhigh effort before any E2E. Four confirmed findings landed as the second commit: pull/edit now hide the always-rejected --json/--jq from help (same honesty rule as hiding the agents format — the Validate rejections still fire on explicit use), the agent-mode field-selection nudge is suppressed for pinned file-writing commands (edit's own stderr was recommending exactly the flags edit rejects), get's --json path emits the truncation hint only after a successful encode like the other paths, and a test-helper comment that overstated writeGetOutput's contract was corrected. A second review handoff produced the third commit: the six rejections were plain errors and exited 1; they're now fail.UsageError and exit 2, with tests asserting the converted exit code. Left as recorded findings rather than fixed here: the shared Options.Validate errors (unknown -o format, --json/-o conflicts) and pull's --path check exit 1 everywhere today — reclassifying those is a repo-wide existing-behavior change; and writeFieldSelect's agent-mode partial-failure envelope declares exit_code 4 in the payload but the process exits 0 (pre-existing on main). The remaining review suggestions (a combined DisallowFormat API, deduplicating the twin pull/edit Validate blocks, unifying the stderr plumbing in writeGetOutput) are real but would be redesign, which this PR intentionally isn't.

E2E, against a live stack with a main-built binary vs this branch, identical args, scratch config, read-only throughout: get and pull JSON stdout plus pulled file trees are byte-identical; every hint is stderr-only, and the agent-mode stderr stream is now pure JSONL where main interleaved raw prose; agent-mode pull writes .json where main wrote .agents; pull -o agents, pull/edit --json/--jq all fail fast with exit 2 where main exited 0 (main demonstrably wrote a `<name>.agents` display-envelope file and silently ignored --json). The kg hint path couldn't be exercised live (Asserts isn't enabled on the test stack; both binaries fail byte-identically there) — it's pinned by unit tests, and its TTY text was already hint:-prefixed, so TTY output is byte-identical by construction.

Sequencing: this should merge after the #387 naming train (#1009 is already in; the batch PRs #1013-#1019 remain) to avoid churn on shared provider files — kg/commands.go and irm/oncall_commands.go are touched by both tracks, so one rebase here is expected — and before #1033 as noted above.

### DIFF
--- a/cmd/gcx/resources/edit.go
+++ b/cmd/gcx/resources/edit.go
@@ -39,6 +39,12 @@ func (opts *editOpts) setup(flags *pflag.FlagSet) {
 	// Bind all the flags
 	opts.IO.BindFlags(flags)
 	opts.flags = flags
+
+	// Validate rejects every use of --json/--jq (below), so hide both from
+	// help — advertising flags the command always refuses is the same
+	// dishonesty as advertising the rejected agents format.
+	_ = flags.MarkHidden("json")
+	_ = flags.MarkHidden("jq")
 }
 
 func (opts *editOpts) Validate() error {
--- a/cmd/gcx/resources/edit.go
+++ b/cmd/gcx/resources/edit.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 
 	cmdconfig "github.com/grafana/gcx/cmd/gcx/config"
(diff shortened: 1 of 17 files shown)


xujiaxj (dismissed):

awsome-o:
Closing unmerged, per the policy-first pivot: #1030 is now a two-decision proposal and nothing should land ahead of that sign-off. The audit record and the findings here stay valid — the implementation PR that follows #1030 re-evaluates and reuses what applies rather than re-landing this wholesale. One known defect to fix on the way back in: the typed rejections added here exit 2 only when the shared Options.Validate accepts the invocation first, so mixed-invalid combinations like `pull -o yaml --json x` still exit 1 through the shared path while `pull --json x` exits 2.

PR #1033, opened 2026-07-21
feat(output): list truncation contract with honest caps and agent-legible metadata

awsome-o:
This is the groundwork PR for #1031, the output workstream of #387 (split out of the naming train per the #994 review discussion). It supersedes the approach in #988 — same goals, and it fixes the four defects that PR's review found. The investigation behind it, including why a root-level persistent `--limit` was rejected, rides along as `docs/research/2026-07-17-global-limit-investigation.md`; the contract itself is documented as `docs/design/output.md` §15, marked PROPOSED until #387 settles it repo-wide.

The contract in short: a truncated list page carries a reserved `list_meta` key in its envelope — `{truncated, returned, total?, cap?, continue?}` — and absence means the output is the complete result set, so an agent reading `-o json` can't mistake a page for the whole set. `total` appears only when actually observed, never guessed. `cap` appears only when a client-side safety cap, not the user's `--limit`, was the bound. `continue` is a runnable command derived from the real invocation argv, so the user's filter flags survive — never a hardcoded string. The stderr hint uses the sentence shape from the #988 review ("showing first 5 of 219. See all results with: gcx datasources list --limit 0"); unknown totals get a doubled limit instead of a `--limit 0` promise, and cap-bounded pages say "safety cap" and suggest refining filters, because no `--limit` can beat the cap. `--json` field selection and discovery are now tolerant of the reserved key: selection re-attaches `list_meta` (this was the `{"uid": null}` break), discovery samples item fields and never lists `list_meta.*`. Any other second envelope key keeps today's behavior, pinned by test.

Two commands migrate as the reference implementations:

- `datasources list` — cheaply complete source. The silent client-side slice is gone; default `--limit` goes 50 → 0 (the full set is fetched from the API regardless, so a default cap only hid data — and the reviewer on #988 didn't object to 0 there); truncation now reports `list_meta` plus the hint, and negative `--limit` is rejected at validation.
- `irm oncall alert-groups list` — paginated source with a 1000-item runaway cap. `--limit 0` bounded by the cap now reports `{cap: 1000}` and the refine-filters hint instead of returning 1000 items indistinguishable from a complete set (the exact case flagged in #988 review). The pagination loop also counts a final page that overshoots the bound with no next cursor as truncation evidence — previously that truncated silently — and attaches the observed total only when the suggested continuation can honestly retrieve it. Negative `--limit` is rejected, and the non-rich fallback path over-fetches by one so truncation is detectable there too.

Deliberately out of scope, so review doesn't go looking for it:

- The shared `--limit` rollout across the remaining ~40 list commands stays out — that's the stretch item on #1031. Several sites need per-command decisions first (incidents list rejects `--limit 0` today, incidents activity coerces 0 to 50 in the client, dashboards list has cursor pagination to converge with).
- `alert rules list` is not migrated. Its JSON output is a bare array with no envelope to carry `list_meta`, and its `--limit` counts different units per format (flattened rules in the table, groups in JSON). The adjudicated review of the feasibility work said to keep it out of the first production truncation PR; the envelope and unit decisions are recorded in the investigation doc.

Verified locally: gofmt clean, lint clean, `go test -race -count=1 ./...` green (with `GCX_AGENT_MODE=false`), reference docs regenerated with `GCX_AGENT_MODE=false` and drift-clean. The diff also went through an adversarial multi-lens review workflow before going up; the confirmed findings were fixed (the capped command's validation error no longer borrows the "0 means all" wording, the research doc no longer describes prototype-only changes as landed, a dead second derivation path in the hint helper was removed so `AttachListMeta` is the single source of the continuation, and the os.Args/agent-mode test helpers moved to `internal/testutils` instead of being copy-pasted per package).

Live end-to-end comparison against a real Grafana Cloud stack, origin/main binary vs this branch with identical args (details in a comment below): the `datasources list` truncation behavior is exactly as promised — honest counts, `list_meta` in the payload, the hint sentence on stderr, filters surviving in the continuation, discovery and selection clean — and untouched commands (including `alert rules list`) are byte-identical modulo server-side evaluation timestamps that also differ between two runs of the same main binary. The IRM cap and overshoot paths could not be forced live without creating tenant alert groups, which I didn't do; they're covered by httptest-driven tests that drive the real pagination loop end to end, and the live runs verified the empty-complete-set parity and the negative-limit rejection. mkdocs isn't installed locally so the docs build runs in CI.

Sequencing: this merges after the #387 naming train (#1009 and the batches), one rebase expected. It also crosses `internal/output/format.go` with the output-safe-corrections branch from the same workstream — whichever lands second rebases trivially.

### DIFF
--- a/cmd/gcx/datasources/list.go
+++ b/cmd/gcx/datasources/list.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	cmdconfig "github.com/grafana/gcx/cmd/gcx/config"
@@ -30,7 +31,9 @@ func (opts *listOpts) setup(flags *pflag.FlagSet) {
 
 	flags.StringVarP(&opts.Type, "type", "t", "", "Filter by datasource type (e.g., prometheus, loki)")
 	flags.StringVar(&opts.Name, "name", "", "Filter by datasource name (case-insensitive substring match)")
-	flags.IntVar(&opts.Limit, "limit", 50, "Maximum number of datasources to return")
+	// Cheaply complete source (no server-side limit exists), so the default
+	// is the full set; a default cap would only hide data.
+	opts.IO.BindListLimit(flags, &opts.Limit, "datasources", 0)
 }
 
 func (opts *listOpts) Validate() error {
@@ -109,13 +112,20 @@ func listCmd() *cobra.Command {
 				})
 			}
 
-			if opts.Limit > 0 && len(infos) > opts.Limit {
-				infos = infos[:opts.Limit]
-			}
+			// The full set is already in hand (the /api/datasources transport
+			// has no server-side limit), so the limit is purely a display trim
+			// and the observed total is exact. Truncation is machine-legible
+			// (list_meta in the envelope) and human-legible (stderr hint).
+			infos, meta := cmdio.TruncateCompleteList(infos, opts.Limit)
+			meta = cmdio.AttachListMeta(meta, os.Args)
 
 			// Pattern 13: single shape for all formats. The table codec extracts
 			// .Datasources to render rows; JSON/YAML serialize the envelope.
-			return opts.IO.Encode(cmd.OutOrStdout(), &datasourceListResult{Datasources: infos})
+			if err := opts.IO.Encode(cmd.OutOrStdout(), &datasourceListResult{Datasources: infos, ListMeta: meta}); err != nil {
+				return err
+			}
+			cmdio.EmitListTruncationHint(cmd.ErrOrStderr(), meta)
+			return nil
 		},
 	}
 
@@ -139,6 +149,10 @@ type datasourceInfo struct {
 // extracts .Datasources to render rows (Pattern 13: format-agnostic data).
 type datasourceListResult struct {
 	Datasources []*datasourceInfo `json:"datasources" yaml:"datasources"`
+	// ListMeta is attached only when the output is a truncated page, so
+	// agents cannot mistake a page for the complete set. Reserved key —
+	// see docs/design/output.md § List Truncation Contract.
+	ListMeta *cmdio.ListMeta `json:"list_meta,omitempty" yaml:"list_meta,omitempty"`
 }
 
 type datasourceTableCodec struct{}
--- a/cmd/gcx/datasources/list_test.go
+++ b/cmd/gcx/datasources/list_test.go
@@ -9,12 +9,24 @@ import (
 	"testing"
 
 	"github.com/grafana/gcx/cmd/gcx/datasources"
+	"github.com/grafana/gcx/internal/testutils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func executeDatasourceCommand(t *testing.T, args []string) (string, error) {
 	t.Helper()
+	stdout, _, err := executeDatasourceCommandStreams(t, args)
+	return stdout, err
+}
+
+func executeDatasourceCommandStreams(t *testing.T, args []string) (string, string, error) {
+	t.Helper()
+
+	// Pin agent-mode detection off so TTY hint shapes and table-default
+	// output are asserted deterministically even when the test itself runs
+	// inside an agent harness (e.g. CLAUDECODE=1).
+	testutils.SetAgentMode(t, false)
 
 	root := helperRoot(datasources.Command())
 	var stdout, stderr bytes.Buffer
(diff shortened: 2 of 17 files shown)


awsome-o:
Live E2E details, as promised in the body. Setup: built two binaries — origin/main at 420046bd and this branch — and ran each command on both with identical args against a real Grafana Cloud stack (`https://<stack>.grafana.net`), config pointed at a scratch copy. No tenant data was created; everything below is read-only. Continuation commands below show `gcx ...` — the real output derives argv[0] from the invocation, same as the existing dashboards `--continue` hint.

datasources list (13 datasources on the stack):

- default invocation: byte-identical stdout on both binaries (13 items; old default 50 didn't truncate at this size, new default 0 fetches all).
- `--limit 5 -o json`: main silently returns 5 items. Branch returns the same 5 items plus `"list_meta": {"truncated": true, "returned": 5, "total": 13, "continue": "gcx datasources list -o json --limit 0"}` and stderr `hint: showing first 5 of 13. See all results with: gcx datasources list -o json --limit 0`. Piped stdout stays parseable JSON.
- `--limit 5 --json uid`: real per-item uids on both binaries (no `{"uid": null}`), and the branch re-attaches `list_meta` to the selection output.
- `--limit 5 --json list`: discovery output byte-identical to main — item fields only, no `list_meta.*` paths.
- `--name <substring> --limit 2 -o json`: the filter survives in both the payload's `continue` and the stderr hint (`gcx datasources list --name <substring> -o json --limit 0`), total honest under the filter (2 of 12).
- `--limit -1`: main silently treats it as unlimited; branch exits 1 with `Invalid --limit -1 / must be >= 0 (0 means all results are returned)`.
- table output piped: byte-identical.

irm oncall alert-groups list (stack has zero alert groups, so the cap/overshoot paths can't be forced live without creating alert groups, which was off-limits — those paths are locked by the httptest tests that drive the real pagination loop):

- default, `--all --limit 1`, `--all --limit 0`: stdout and stderr byte-identical on both binaries — `{"items": []}`, no `list_meta`, absence-means-complete holds on the live empty set.
- `--limit -1`: main accepts it; branch exits 1 with the capped-source wording `must be >= 0 (0 means as many results as the safety cap allows)`.

Untouched commands:

- `alert rules list` (`-o json`, `--limit 2 -o json`, table, `--limit -1`): byte-identical except `lastEvaluation`/`evaluationTime` fields in the full JSON dump, which differ identically between two back-to-back runs of the *same* main binary — server-side evaluation churn, not a binary difference. Its `--limit` help text is confirmed unchanged (`0 for unlimited, default 50`), i.e. the deliberate exclusion is real.
- `slo list -o json`: identical output and exit code on both binaries.

awsome-o:
Pushed b5a45316 addressing a hole found in a second review pass: the §15.5 selection/discovery guarantees only held for typed envelope structs. The direct `map[string]any` fast paths (in the field-selection type switch and in `marshalToSampleMap`) bypassed the envelope-aware logic, so a dynamic map carrying `list_meta` would get whole-object selection (`{"uid": null}`) and discovery would list `list_meta.*`. Neither migrated command emits map envelopes, so nothing shipped was broken — but the shared contract claimed more than the code delivered, and the migration wave will hit it.

The fix is gated on the reserved entry itself: a dynamic map only gets the envelope treatment when it carries an object-valued `list_meta` key. That scoping matters — `gcx api` passthrough responses are `map[string]any`, and treating every items-shaped map as an envelope would have changed field selection for raw API responses that happen to have that shape. Maps without the reserved key keep today's behavior exactly, pinned by tests. Discovery on an empty dynamic envelope (no element type to reflect on) degrades to the envelope's own keys, never `list_meta.*`.

`unstructured.UnstructuredList` deliberately stays out of the contract — no producer attaches truncation metadata to unstructured lists today; recorded in §15.5 and the research doc for the resources-pipeline migration where the producer will exist.

Gates rerun on the new head: gofmt clean, lint clean, full `go test -race -count=1 ./...` green, reference drift clean.

awsome-o:
Pushed 2af61e98: the previous commit's dynamic-map support only covered the JSON-decoded representation. A map assembled naturally in Go — `"list_meta": &output.ListMeta{...}` with a native `[]map[string]any` items slice — never passed the reserved-key gate (`isListMetaEntry` only recognizes `map[string]any`), so selection still returned `{"uid": null}`. And the hybrid shape (decoded metadata object, native items slice) was worse: it passed the gate but `toSliceOfMaps` only accepts `[]any`, so selection emitted an empty items array — silently dropping rows. That one was a regression the previous commit introduced for that shape.

Fix: both map paths now gate on presence of the reserved key (checked before normalization, so native metadata values opt in), JSON-normalize a copy of the map, validate the reserved shape on the copy, and run envelope handling there. Maps without the key keep whole-object behavior on the original value, so the gcx api passthrough pins are untouched. New regression tests exercise the native representations directly — native metadata plus native and typed item slices, the hybrid row-dropping shape, and native-value discovery. The `ListMetaKey` GoDoc now states exactly what the guarantees cover instead of implying universal handling.

Gates rerun on the new head: gofmt clean, lint clean, full `go test -race -count=1 ./...` green, reference drift clean.

dafydd-t (approved):

PR #1037, opened 2026-07-22
feature: Tier 0 agentic support for metrics labels match

### LINKED ISSUE #1034: Add match[] scoping to gcx metrics labels and label values
### Drilldown squad idea

`gcx metrics labels` lists all labels in a datasource, and `gcx metrics labels --label NAME` lists all values for a label. Neither can be scoped to a metric or series selector, so you cannot ask "what labels exist on http_requests_total" or "what job values exist on http_requests_total" from the CLI.

The Prometheus labels and label-values APIs both accept a `match[]` parameter for exactly this. gcx already sends `match[]` for the series endpoint (internal/query/prometheus/series.go builds the query string with it), but the labels client does not: `Client.Labels` (internal/query/prometheus/client.go:114) and `Client.LabelValues` (internal/query/prometheus/client.go:146) accept no match parameter, and `labelsOpts` (internal/datasources/prometheus/labels.go:18) has only `Datasource` and `Label`.

### Proposed change:

- Add a repeatable `--match` flag (series selector) and a `--metric` convenience flag (shorthand for `{__name__="NAME"}`) to `gcx metrics labels`
- Thread `match []string` through `Client.Labels` and `Client.LabelValues`, following the existing series.go pattern

Example usage:

```
# What labels exist on a metric
gcx metrics labels -d UID --metric http_requests_total

# What values does job take on a metric
gcx metrics labels -d UID --metric http_requests_total --label job

# Arbitrary selector scoping
gcx metrics labels -d UID --match '{job="api",env="prod"}'
```

### More context

This work addresses a gap found in the [Drilldown design doc research to support GCX](https://docs.google.com/document/d/1D29TOTXy7cs5Q2XlJ1nBRieGIsppeCmQGRsB3WJ1qXA/edit?tab=t.0#heading=h.5sybau7waq2q). 

### Future work

Drilldown has decided to start the implemtation in the metrics tool to get early testing done on the helpfulness of this feature. The drilldown apps plan to have a backend and port this functionality to that API. This work is the first step before we have the backend in place.

bohandley:
Fixes https://github.com/grafana/gcx/issues/1034

### What is this?

This is the first PR for drilldown apps to support GCX. The drilldown apps' expertise is both queryless experience (heuristics for building from primitives) and strategies for breaking down on dimensions.

This PR helps for breaking down dimensions by adding match[] param to the metrics and labels endpoint. This mimics the experience in Metrics Drilldown where you can breakdown label dimensions on a metric.

### What does it look like?

```
# What labels exist on a metric
gcx metrics labels -d UID --metric http_requests_total

# What values does job take on a metric
gcx metrics labels -d UID --metric http_requests_total --label job

# Arbitrary selector scoping
gcx metrics labels -d UID --match '{job="api",env="prod"}'
```

### Who does this help?

This helps agents explore metric signals in a more organized way. This provides [Tier 0 support in this doc](https://docs.google.com/document/d/1kzgCaz-hViwB1FeIXsEwY2_vTQAbmuq5rGIPRrrPqSU/edit?tab=t.0) by @cedricziel used to guide the [wg-o11y-agentic-ai](https://grafanalabs.enterprise.slack.com/archives/C088W73SEP8), so it helps o11y apps by supporting agentic workflows.

### More context 

[Read this issue](https://github.com/grafana/gcx/issues/1034) to learn more about how the drilldown apps are helping support agentic workflows first in gcx with the telemetry api and next with their own backends (still in the research phase).


awsome-o (commented):
Thanks for putting this together, match[] on the labels commands is something agents and the Drilldown style flows genuinely need, and the client plumbing plus the client_test coverage look good. I double checked the behavior against how Prometheus treats repeated match[] params and left notes inline. Short version: combining --metric with --match quietly returns more than the flags promise, and the bare metric shorthand breaks on dotted metric names. Both are small code changes but they change the answers users see, so I'd like them settled before merge.

One thing without an inline anchor: there's no labels_test.go, so nothing at the command layer proves --metric and --match actually make it into the request. client_test.go covers the client nicely; a small test that builds the command and asserts the selectors reach Labels/LabelValues would keep a future refactor from silently dropping the flags.

Happy to take another look as soon as you push, this is close :slightly_smiling_face:

internal/datasources/prometheus/labels.go
```diff
@@ -28,6 +30,18 @@ func (opts *labelsOpts) setup(flags *pflag.FlagSet) {
...
+	flags.StringVar(&opts.Metric, "metric", "", "Scope results to this metric (shorthand for --match METRIC)")
+	flags.StringArrayVar(&opts.Match, "match", nil, "Series selector(s) to scope results; repeatable")
+}
+
+// selectors combines --match selectors with the --metric shorthand. A bare
+// metric name is itself a valid series selector for the match[] parameter.
+func (opts *labelsOpts) selectors() []string {
+	match := append([]string{}, opts.Match...)
+	if opts.Metric != "" {
+		match = append(match, opts.Metric)
```
awsome-o: This is my main flag on the PR. Prometheus treats repeated match[] as a union: you get labels from series matching ANY selector, not all of them. So `--label job --metric http_requests_total --match '{cluster="prod"}'` returns job values from every cluster=prod series anywhere, plus every http_requests_total series in any cluster, when the person asking almost certainly wanted the intersection. Repeating --match twice does the same thing. The flag help ("Scope results to this metric", "Series selector(s) to scope results") reads like each flag narrows things further, so nobody gets warned. A user, or worse an agent, walks away believing a label exists on a metric when it doesn't. The spelling that actually intersects is a single selector, `--match 'http_requests_total{cluster="prod"}'`.

I see three ways out: reject the combination of --metric with --match, fold the metric into each provided selector, or keep union semantics and say so plainly in the help and examples. Any of them works for me, the current silent OR is the only version I'd block on.

internal/datasources/prometheus/labels.go
```diff
@@ -28,6 +30,18 @@ func (opts *labelsOpts) setup(flags *pflag.FlagSet) {
 
 	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.prometheus is configured)")
 	flags.StringVarP(&opts.Label, "label", "l", "", "Get values for this label (omit to list all labels)")
+	flags.StringVar(&opts.Metric, "metric", "", "Scope results to this metric (shorthand for --match METRIC)")
+	flags.StringArrayVar(&opts.Match, "match", nil, "Series selector(s) to scope results; repeatable")
+}
+
+// selectors combines --match selectors with the --metric shorthand. A bare
+// metric name is itself a valid series selector for the match[] parameter.
```
awsome-o: The bare name is valid for classic metric names but it's a parse error for the dotted UTF-8 style (OTel semconv names, supported since Prometheus 3.x and available in Mimir). `--metric http.server.duration` goes out as match[]=http.server.duration, the server answers 400, and the command fails with "failed to get labels" for exactly the metrics whose selector syntax people can't remember. Sending `{__name__="<name>"}` instead of the bare name works for every metric name, old and new, and then this comment can go.

internal/query/prometheus/client.go
```diff
@@ -110,15 +110,18 @@ func (c *Client) Query(ctx context.Context, datasourceUID string, req QueryReque
 	return convertGrafanaResponse(&grafanaResp, req.IsRange()), nil
 }
 
-// Labels returns all label names.
-func (c *Client) Labels(ctx context.Context, datasourceUID string) (*LabelsResponse, error) {
+// Labels returns all label names. Optional match selectors scope the result
+// to labels present on matching series.
+func (c *Client) Labels(ctx context.Context, datasourceUID string, match []string) (*LabelsResponse, error) {
```
awsome-o, line 115: The labels and label values endpoints accept start and end just like the series one does, and Series in series.go already threads them through (with --since, --from and --to on the command). Without them every lookup here scans the whole retention window, which on long retention Mimir tenants is slow, and it resurfaces label values from metrics that died months ago. The Drilldown and query editor flows this PR cites always pass a time window, so right now there's no way to reproduce what they do. Fine as a fast follow if you want to keep this PR tight, but worth landing before agents start depending on the command.

bohandley: Will add this as a fast follow!

internal/datasources/prometheus/labels.go
bohandley: Good catch! Handling this by passing metric always in brackets as {__name__="<metric>"}

internal/datasources/prometheus/labels.go
bohandley: Fixed! Updated to pass the metric with the labels selected. I also handled an additional scenario if the metric flag exists and the labels match, where the labels match has a contradictory/additonal __name__ selector. Tests added!

internal/query/prometheus/client.go
bohandley: https://github.com/grafana/gcx/issues/1064

awsome-o:
The updated commits look good to me. Folding --metric into the selectors with real parsing kills the union problem, the dotted name support plus tests is exactly what I hoped for, and the contradictory `__name__` guard is a nice touch. Nothing blocking on this PR anymore. One heads up: the stacked PR picked up a stale version of the datasource help text through the merge chain (3514077b had it, 89b8742b fixed it here, the merge over there kept the old line), details in my review on #1050.

Small take or leave items while you're in the file. The labels command still accepts and ignores positional args, so `gcx metrics labels http_requests_total` silently returns the full unscoped label list; `Args: cobra.NoArgs` like you did on list would close that. An explicitly empty --metric (think `--metric "$METRIC"` with the var unset) silently drops scoping instead of erroring. And --match only gets the friendly client side parse when --metric is also present, without it the same typo comes back as an opaque proxied 400.

nitpick: the brace plus strings.Join selector assembly is technically the string formatting that CLAUDE.md Pattern 14 warns about; since the parts are canonical matcher renderings and promql-builder can't express matcher only selectors, a one line comment saying why it's safe here would settle it. Time bounds we already agreed is a follow up.

awsome-o (approved):
Approving, this is in good shape and thanks again for the quick turnarounds. One ask on timing: our v1 tag goes out tomorrow morning and we want main quiet until it's cut, so please hold the merge until after the tag lands. Nothing about the change itself, it's additive and low risk, we just don't want anything moving under the release.

Same timing for #1050, and that one additionally needs the two items from my review over there fixed before it merges anywhere (the stale datasource help text and the list_meta truncation). Once this PR merges it retargets to main, so worth double checking those fixes are in first. Ping me when they are and I'll take a quick look, should be a fast one.

bohandley:
Thank you @awsome-o for the guidance and clarity!

- I won't merge this until tomorrow morning/after v1 is released to keep main clean
- I botched the [stacked PR](https://github.com/grafana/gcx/pull/1050) merge and will fix that all after this PR is merged 🙏🏻 
- The fixes on the subsequent PR will include
  - fix - labels command still accepts and ignores positional args 
  - fix - --match only gets the friendly client side parse when --metric is also present, without it the same typo comes back as an opaque proxied 400
  - one line comment for reasoning about using CLAUDE.md Pattern 14 warns about

PR #1040, opened 2026-07-22
feat(kg): add --dry-run to model-rules/prom-rules upsert

kblaschke512:
## Summary

Adds `--dry-run` to the KG `model-rules upsert` and `prom-rules upsert` commands.
With `--dry-run`, the command runs the backend's synchronous validator and, if valid,
prints a unified diff of what a real `upsert` would add or modify against the current
remote config — **without writing anything**. Reaching the diff implies validation passed;
validation failures surface the backend's field-level errors and exit non-zero.

This mirrors the validate-then-diff pattern already shipped for `kg suppressions upsert`
(#1015), and brings all three KG write commands into one consistent surface:

| Command | Verb | `--dry-run` |
|---|---|---|
| `kg suppressions upsert` | `upsert` | ✅ (pre-existing) |
| `kg model-rules upsert` | `upsert` | ✅ (this PR) |
| `kg prom-rules upsert` | `upsert` | ✅ (this PR) |

## Backend dependency

Depends on the two already-merged, already-deployed backend PRs (asserts-adi validate
endpoints + cell-gateway routes). E2E-validated live against `assertsdemo01`:

- New rule → `model rules are valid; would add …` + full add diff
- Modify existing rule → `would modify …` + minimal diff
- Empty name → rejected client-side, exit 1
- Malformed PromQL → backend **HTTP 422** with parser-precise location, exit 1
- prom-rules new rule → `prom-rules are valid; would add …` + diff
- Confirmed nothing persisted (all dry-run)

## Naming

Rebased onto `main` after #1015's `create` → `upsert` naming convergence. Resolution
kept this PR's dry-run feature and adopted the `upsert` verb, so model-rules/prom-rules
now match the convention suppressions already established (shared `newRulesUpsertCommand`
helper for the two single-config commands).

## ⚠️ Open question for the team: prom-rules `get` ↔ `upsert` round-trip (NOT fixed here)

Surfaced during E2E; **out of scope for this PR**, flagging for direction. `prom-rules`
cannot round-trip its own `get` output back into `upsert` (the natural pull → edit →
dry-run → apply loop):

| | `get -o yaml` emits | `upsert -f` consumes | round-trips? |
|---|---|---|---|
| `model-rules` | flat (`name`/`relations`) | flat `ModelRules` | ✅ |
| `prom-rules`  | K8s envelope (`apiVersion`/`kind`/`metadata`/`spec`) | flat `Rule` (`name`/`groups`) | ❌ (`Prom-rules file has no name`) |

Note this is a `get`↔`upsert` **shape** mismatch, not a validate-vs-upload one — validate
and upload receive the identical flat bytes. Both `Rule` and `ModelRules` are registered
resource-tier types with two-way converters, so this is a KG-provider-local choice.

**Options (follow-up, not blocking):**
- **A — fix consumer (non-breaking):** unwrap a K8s envelope via the existing
  `RuleFromResource` in `runPromRulesDryRun` + `UploadPromRules` before the flat fallback.
  Restores round-trip; `get` output unchanged.
- **B — fix producer (root cause, breaking):** pick one canonical `get` shape for both
  commands.

Recommendation: A as a follow-up issue; B as a separate consistency discussion. **Preference?**

## Testing

- `GCX_AGENT_MODE=false mise run gate` — lint 0 issues, full suite passes, build OK
- `GCX_AGENT_MODE=false mise run reference-drift` — clean
- 309 lines of new table tests in `internal/providers/kg/rules_dryrun_test.go`

### DIFF
--- a/docs/reference/cli/gcx_kg_model-rules_upsert.md
+++ b/docs/reference/cli/gcx_kg_model-rules_upsert.md
@@ -6,11 +6,24 @@ Upload model rules from a YAML file.
 gcx kg model-rules upsert [flags]
 ```
 
+### Examples
+
+```
+  gcx kg model-rules upsert -f model-rules.yaml
+
+  # Validate against the backend and preview the diff without uploading:
+  gcx kg model-rules upsert -f model-rules.yaml --dry-run
+```
+
 ### Options
 
 ```
-  -f, --file string   Input file (YAML)
-  -h, --help          help for upsert
+      --dry-run         Validate against the backend and show a diff without uploading.
+  -f, --file string     Input file (YAML), or '-' for stdin.
+  -h, --help            help for upsert
+      --jq string       jq expression to apply to JSON output. Mutually exclusive with --json.
+      --json string     Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
+  -o, --output string   Output format. One of: agents, json, text, yaml (default "text")
 ```
 
 ### Options inherited from parent commands
--- a/docs/reference/cli/gcx_kg_prom-rules_upsert.md
+++ b/docs/reference/cli/gcx_kg_prom-rules_upsert.md
@@ -6,11 +6,24 @@ Upload Knowledge Graph Custom Prometheus rules from a YAML file.
 gcx kg prom-rules upsert [flags]
 ```
 
+### Examples
+
+```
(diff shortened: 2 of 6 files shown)


xujiaxj (approved):

awsome-o (approved):
Looks safe and aligned with https://github.com/grafana/gcx/pull/1023

PR #1044, opened 2026-07-22
fix(cloud): migrate stack regions to /api/v1/stack-regions

dafydd-t:
- [ ] Needs the new `/api/v1/stack-regions` endpoint to have an oauth scope attached to it

## Summary

- point the GCOM client's `ListRegions` at `GET /api/v1/stack-regions` (was `/api/stack-regions`)
- drop `status`, `createdAt` and `updatedAt` from `cloud.Region` — the v1 endpoint no longer returns them (`sortOrder` was also removed upstream but gcx never read it)
- remove the STATUS column from the `gcx cloud stacks list-regions` table
- update the API-tiers reference doc

## Why

The grafana.com team is migrating region endpoints to v1 and wants clients off `/api/stack-regions` to ease its phase-out. The v1 endpoint is live in all environments with the same auth. Verified against the grafana-com OpenAPI spec and live responses: v1 keeps the `{"items": [...]}` envelope (minus pagination metadata gcx never used) and drops `createdAt`, `updatedAt`, `sortOrder` — and also `status`, which the Slack thread didn't mention but which gcx did surface.

## User-visible change

The STATUS column is gone from the table (it was `active` for every region — v1 only returns visible regions), and `status`/`createdAt`/`updatedAt` no longer appear in `-o json` output.

Before:

```
SLUG                 NAME               DESCRIPTION                   PROVIDER  STATUS
prod-us-west-0       AWS US West        AWS US West                   aws       active
us                   GCP US Central     United States                 gcp       active
...
```

After (verified live against grafana.com):

```
SLUG                 NAME               DESCRIPTION                   PROVIDER
prod-us-west-0       AWS US West        AWS US West                   aws
us                   GCP US Central     United States                 gcp
...
```

### DIFF
--- a/docs/reference/grafana-cloud-api-tiers.md
+++ b/docs/reference/grafana-cloud-api-tiers.md
@@ -45,7 +45,7 @@ above any individual stack.
 ```
 /api/orgs/{slug}                          org + members
 /api/stacks, /api/instances/{id}          stack lifecycle
-/api/stack-regions                        region catalog
+/api/v1/stack-regions                     region catalog
 /api/v1/accesspolicies, /api/v1/tokens    IAM for Cloud itself
 /api/plugins, /api/instances/{id}/plugins plugin catalog + install
 /api/stacks/{id}/integrations             prebuilt integrations installer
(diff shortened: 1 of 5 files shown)


dafydd-t:
cc @nachogiljaldo

Issue #796, opened 2026-05-30
Evaluate NDJSON output to survive stderr/stdout merging (2>&1) in agent harnesses


wardbekker:
### Idea / experiment

Run an experiment: when **stdout is not a TTY** (any pipe/redirect, which covers agent harnesses), default gcx's data output to **NDJSON** — newline-delimited JSON, one complete JSON object per line — instead of a single JSON document. Evaluate whether it removes the `2>&1` footgun and improves stream-processability for agentic harnesses.

### What happens today

In agent mode the streams are shaped differently:

- **stdout** — the `agents` codec writes the result as a *single* JSON object (`internal/output/agents.go`, contract in `docs/design/output.md` §1.1.1).
- **stderr** — hints/warnings/notes are written as **JSONL**, one object per line with a `class` field (`internal/output/format.go`, `EmitHint` et al.).

Agentic harnesses routinely merge streams with `2>&1` before parsing:

```bash
gcx datasources list -o json 2>&1 | jq '.[]'   # breaks
```

The stderr JSONL lines interleave into the single stdout JSON document, so the merged stream is no longer a parseable JSON document and `jq` fails. We already warn against this in the debug skill and recommend `2>/dev/null` or `--json <fields>` as workarounds — but harnesses don't reliably apply those, so the default path is fragile.

### What I'd like to change

Default to NDJSON whenever **stdout is not a TTY**, emitting the **data** itself as line-delimited JSON (one complete object per line). Then a `2>&1`-merged stream is *uniformly* NDJSON: every line is an independently parseable JSON value, and consumers demultiplex by a discriminator field (diagnostics already carry `class`; data lines can be distinguished by its absence or by a `kind`). NDJSON's line-orientation is inherently robust to stream interleaving — which is precisely the failure mode above — and it's a de-facto standard for agent-friendly CLI output.

We already emit NDJSON for streaming events (instrumentation `wait`, assistant), so this extends an existing in-house pattern rather than introducing a new one.

Pair this with the runtime I/O-shape discovery gcx already has, so agents learn schemas at runtime instead of guessing: field selection + introspection (`--json field1,field2`, `--json list`, `--json ?`) and command annotations / known-resource hints.

### Why it's better

- A merged `2>&1` stream stays machine-parseable; the most common harness invocation stops being a footgun.
[...]

Issue #264, opened 2026-03-30
Implement mutation summary table codec for push/pull/delete


radiohead:
## Problem

Current push/pull/delete output is unstructured prose that agents can't parse reliably and humans can't scan at scale (100s+ resources).

## Solution

Replace ad-hoc `cmdio.Success/Warning/Error` output in push/pull/delete commands with a structured summary table codec:

- **STDOUT:** kind-grouped summary table (kind, total, succeeded, skipped, failed)
- **STDERR:** individual failure details with error messages
- **JSON:** tiered structure with `summary[]`, `failures[]`, `skipped[]`

## Acceptance Criteria

- [ ] push, pull, delete commands use summary table codec
- [ ] text/wide/json/yaml all produce structured output
- [ ] Failures always enumerated individually
- [ ] Successes counted, not enumerated
- [ ] `cmdio.Success/Warning` remain for progress during execution only

PR #1046, opened 2026-07-22
feat(output)!: agent output contract — one JSON document per finite command (#387)

awsome-o:
This is the agent output unification for GA. The contract it implements and ratifies: when agent mode supplies the default (no explicit -o, --json or --jq), a finite command emits exactly one JSON value on stdout — its result or a fused error — and the process exit code agrees with that outcome. Streams emit typed, versioned JSONL with a terminal success/error event. Artifact commands (the pull family) emit one JSON receipt while -o stays the pinned file format. Interactive, server, shell, prose and raw commands are declared protocol classes. stderr stays advisory, so an agent never needs to parse both streams to know what happened. Explicit format flags always win.

The scope is all commands, per the settled direction. Every one of the 490 runnable leaves is classified in cmd/gcx/root/testdata/output_classes.json and a CI test fails on any new unclassified command or stale entry — it already caught the #1016 renames during this branch's rebase, which is exactly the failure mode it exists for. An exec-based conformance suite builds a fresh binary and pins the end-to-end wiring: one JSON value then EOF, in-band gcx.error documents whose embedded exitCode matches the process exit code, explicit -o overrides honored, stdin closed everywhere.

What lands here, roughly in dependency order: the re-landed #1032 safety corrections (pull/edit pinned to json so agent mode can never write .agents resource files or spill envelopes into manifests, with the exit-code ordering defect fixed); the protocol primitives (gcxerrors.EmittedError for atomic stdout ownership, discriminators on the spill receipt and error envelopes, the shared mutation result family, the artifact receipt emitter); the resources cohort including three proven-then-fixed partial-failure defects in resources get (one of them exited 0 while the envelope said 4); fourteen provider families migrated in parallel with per-family adversarial verification; assistant prompt/dashboard on typed JSONL streams with approval prompts explicitly declined in agent mode; instrumentation wait streams typed with guaranteed terminal events; and the CONSTITUTION amendment ratifying pull's json default (#1030 decision 2, maintainer-approved 2026-07-22) plus the contract itself as an invariant.

Human-mode default stdout is byte-identical across the migration — the text codecs reproduce the exact lines, pinned by tests — with a handful of disclosed exceptions where the old behavior was itself the bug: pull now counts file-write failures instead of reporting 0 errors with exit 0, resources get partial failures exit 4 instead of 1, diagnostics like dry-run notices moved from stdout to stderr with identical rendering, and logs/traces adaptive plus kg destructive verbs gained the same ConfirmDestructive guard their metrics siblings always had (scripts there now need --force or --yes). Mutations that had no -o at all now bind the standard output flags with text as the human default. Full inventory of the ~90 behavior changes is in the commit messages per family.

Verification story, since a diff this size needs one: an 18-family exhaustive inventory with adversarial verification and a completeness diff against the CLI reference seeded the classification (529 entries, zero missing, zero phantom); each migration family was independently verified; then a seven-lens adversarial review with refuter agents ran over the finished diff and produced 25 confirmed findings — all fixed, including two commands the migration itself had missed (datasources delete and health double-emitting, reproduced empirically against a fake server) and total-failure paths that wrongly wore the partial-failure exit code. Gates: golangci-lint 0 issues, full race suite green, reference docs drift-clean, skills-drift green, docs build not run locally (mkdocs unavailable in this environment — CI covers it). The suite also passes under CLAUDECODE=1 except for confirm-prompt tests in aio11y and dashboards that fail identically on main and are untouched here.

Known follow-ups, deliberately out of this PR: browser-OAuth login still blocks on its localhost callback in agent mode (documented as a known gap; token auth is the harness path); the assistant llm_hint annotations still steer agents to the legacy --json stream; the human-defaults normalization (#1030 decision 1, get→yaml/list→table) is its own PR by design; NDJSON evaluation stays #796.

Refs #387, #1030, #264, #1031.

---

## Review-fix round (2026-07-23)

A risk-focused review of this PR produced 12 findings; each is fixed in its own commit on top of the branch:

- `fix(fail)`: an error chain carrying both a `DetailedError` and an `EmittedError` could bypass the no-second-document guarantee — the EmittedError check now runs before DetailedError extraction
- `fix(resources)`: `get --jq` no longer drops the jq transformation on agent-mode partial failure (output keeps its shape; failure travels via stderr + exit 4)
- `fix(resources)`: push/pull/delete/validate partial failures restore the stderr failure-count diagnostic that `EmittedError` suppression removed
- `fix(output)!`: zero-success convention unified — `dashboards snapshot`, `dev generate` and `dev import` now return the raw error (exit 1) on total failure instead of a receipt + exit 4; `dev lint run` intentionally keeps exit 4 (violations are findings, not operation failures)
- `fix(instrumentation)`: both wait commands' timeout terminals now name a persistent poll error (bad token no longer reads as a bare timeout)
- `fix(instrumentation)`: `check` failures restore the stderr failure-count line
- `fix(assistant)`: canceled agent streams carry `ExitCancelled` (5), matching the PR's own convention; legacy `--json` modes keep exit 1
- `fix(assistant)`: a completed task's context ID persists even when the stream write failed (broken pipe no longer loses `--continue`)
- `fix(config)`: `check`'s Grafana<12 error path no longer double-emits in agent mode; `ErrorToDetailedError` derefs are nil-guarded
- `test(root)`: the conformance sweep now asserts in-band/process exit-code agreement for every swept leaf and covers stream-class leaves with a JSON-only stdout assertion
- `docs`: stale `handleError` reference fixed in errors.md

### Additional disclosed human-mode drifts

These stdout→stderr moves shipped in the migration but were missing from the disclosed list above (the moves are correct — diagnostics belong on stderr — and are disclosed here rather than reverted):

- `synth probes reset-token`: the "SM API does not return the new token" advisory moved stdout→stderr
- `k6 projects create` / `k6 tests create`: success one-liner moved stdout→stderr (the created object is the stdout document)
- `irm incidents create`: success line moved stdout→stderr
- `logs adaptive` drop-rules create/update: success line moved stdout→stderr
- `kg entities|relationships upsert` with a batch `-f` file: human output changed from N concatenated JSON documents to one array

### Correction

The classification fixture holds **484** runnable leaves (455 finite, 3 artifact, 7 raw, 7 interactive, 4 stream, 4 shell, 3 prose, 1 server), not 490 as stated above; 458 finite+artifact leaves are swept empirically (plus the 4 stream leaves as of the review-fix round).


awsome-o:
Live E2E evidence for this PR, run against a personal eval stack (main-vs-branch binary comparison, read-only commands plus server-side dry-runs only): 61 commands across resources, config, datasources, dashboards, alert, slo, synth, k6, kg, irm, signals, assistant and instrumentation, each run with both binaries in both agent and human mode.

Results: the branch was one-JSON-document compliant on every single agent-mode run including all error paths (strict decode plus EOF check), exit codes agreed everywhere, and agent-mode stderr was 100% JSONL. Human-mode stdout was byte-identical to main for every read command; the only human-visible diffs were live-data timestamp drift (confirmed by a main-vs-main control run), the commands/help-tree catalogs now annotating the new output flags, and the intentional removal of the redundant stderr error banner on emitted-result paths.

The comparison also caught main violating the contract live, which is exactly what this PR fixes: main emitted two JSON documents on the `datasources health` and `instrumentation check` failure paths, printed a human table for `config list-contexts` in agent mode and a bare non-JSON string for `current-context`, dropped every example manifest from `resources list-examples` in agent mode, and returned `null` instead of `[]` for an empty `irm incidents list`. The pull check on the branch produced clean `.json` manifests, zero `.agents` files, and a proper `gcx.artifact_receipt` on stdout.

One pre-existing, unrelated finding for the backlog: `kg entities list` enforces `--type` as required at runtime but its help does not mark it required (same on main).

awsome-o:
Follow-up to the E2E evidence above: the conformance suite now also sweeps every leaf command classified finite or artifact — 458 of the 490 (the rest are declared exempt classes: interactive, server, shell, prose, raw, stream). Each runs with no arguments in a fully isolated environment (no config, empty cwd, stdin closed, 20s timeout), and agent-mode stdout must hold exactly one JSON value whether the command succeeds offline or fails with a usage or missing-config error; a surviving prompt fails by timeout. All 458 pass, so the one-document contract is now checked per command by name in CI, not just per family.

dafydd-t:
some issues + reproduction steps

run from the root of the repo after running `mise run build`

---

### 1. `--jq` dropped on fused partial failure
**Where**: `cmd/gcx/resources/get.go:~317` · **Class**: bug

In agent mode, a partial-failure `gcx.partial_result` envelope emits raw item maps; the user's `--jq`
expression is never applied, so document shape silently differs between success and partial-failure runs.

```bash
GCX_AGENT_MODE=1 bin/gcx resources get dashboards/<real-uid> dashboards/nope --jq '.metadata.name'
# items are raw objects, jq ignored; drop the bad selector and jq works
```

### 2. Two zero-success conventions
**Where**: `internal/providers/dashboards/snapshot/snapshot.go:~318`, `cmd/gcx/dev/generate.go`,
`cmd/gcx/dev/import.go`, `cmd/gcx/linter/lint.go:189` · **Class**: consistency

Snapshot/generate/import/lint exit 4 ("partial") with a receipt when *everything* failed;
slo/synth/kg/metrics/traces return the raw error (exit 1) on zero success. Both sides are test-locked.

```bash
GCX_AGENT_MODE=0 bin/gcx dashboards snapshot does-not-exist; echo $?   # 4, receipt, 0 succeeded
GCX_AGENT_MODE=0 bin/gcx synth checks delete 999999999; echo $?        # 1, raw error
# offline variant:
bin/gcx dev lint run cmd/gcx/linter/testdata/missing-panel-title.json; echo $?  # 4 on total failure
```

### 3. apps wait retries persistent errors until timeout
**Where**: `cmd/gcx/instrumentation/clusters/apps/wait.go:153-162` · **Class**: behavior

Poll errors went fail-fast → retry-until-deadline in both modes: a bad token now burns the whole
`--timeout` and exits 1 as "timeout" (was immediate failure with the real exit code).
Disclosed in commit 096a2fbf only.

```bash
# with a revoked/bad token in the context:
time bin/gcx instrumentation clusters apps wait <cluster> <ns> --timeout 30s; echo $?
# branch: ~30s of stderr retries, exit 1 "timeout" — main: fails in <1s with the auth error
```

### 4. Conformance gate only verifies finite+artifact
**Where**: `cmd/gcx/root/agentconformance_test.go:223` · **Class**: gap

A command misclassified as `prose`/`raw`/`stream` is exempt from the sweep with zero protocol
verification; the sweep also discards exit codes (envelope/exit agreement asserted for exactly one command).

```bash
# flip any finite entry to "prose" in cmd/gcx/root/testdata/output_classes.json, then:
go test ./cmd/gcx/root/ -run 'TestOutputClass|TestAgentConformance' -count=1   # passes
```

### 5. Cancelled-stream exit 1 vs the PR's own convention of 5
**Where**: `internal/providers/assistant/stream_emitter.go:334` · **Class**: semantics

A server-canceled assistant task ends with `exitCode: 1`, while mcpservers delete decline and
instrumentation wait cancel (same PR) use 5.

```bash
bin/gcx assistant mcp-servers delete <id>   # answer "n"; echo $? → 5 (main: 0)
# server-side stream cancel isn't CLI-triggerable — see the "canceled" pin in
# stream_emitter_test.go showing exitCode 1
```

### 6. Broken pipe loses the `--continue` thread
**Where**: `internal/providers/assistant/stream_emitter.go:212-218` · **Class**: bug (minor)

A mid-stream stdout write failure skips `SaveLastContextID` even when the task completed;
previously the context ID always persisted.

```bash
GCX_AGENT_MODE=1 bin/gcx assistant prompt "count to 100" | head -n 2
bin/gcx assistant prompt --continue "where were we"   # new thread; main continued the old one
```

### 7. `config check` error path still double-emits
**Where**: `cmd/gcx/config/command.go:~495` · **Class**: semantics

The `hasIssues` path was fixed with `EmittedError`, but the retained `checkErr` path
(Grafana <12 `VersionIncompatibleError`) returns a raw error → agent mode appends a `gcx.error`
doc after the prose report. Same file ~L527 dereferences `ErrorToDetailedError` without a nil check
on a broadened nil-return set.

```bash
GCX_AGENT_MODE=1 bin/gcx config check   # needs a Grafana <12 instance to hit the path
```

### 8. Undisclosed stdout→stderr moves
**Where**: `internal/providers/synth/probes/commands.go:~775`, `internal/providers/k6/commands.go:406,749`,
`internal/providers/irm/incidents_commands_impl.go:~1002`, `internal/providers/logs/adaptive/commands.go:~186`
· **Class**: behavior

Success one-liners and the reset-token advisory moved to stderr; `| cat` pipelines lose lines they
always had. Not in the PR's disclosed-drift list.

```bash
bin/gcx k6 projects create --name repro 2>/dev/null                        # no "Created ..." line
(cd /tmp/gcx-main && bin/gcx k6 projects create --name repro) 2>/dev/null  # line present
```

### 9. `instrumentation check` failures: exit 1→4, stderr line gone
**Where**: `cmd/gcx/instrumentation/check/command.go:157-163` · **Class**: behavior

`ExitPartialFailure` for failed environment checks stretches the taxonomy, and `EmittedError`
suppresses the old `Error: N check(s) failed` stderr line.

```bash
bin/gcx instrumentation check; echo $?                        # 4, no stderr error line
(cd /tmp/gcx-main && bin/gcx instrumentation check); echo $?  # 1, line present
```

### 10. Partial-failure stderr loss for push/delete/pull/validate
**Where**: `cmd/gcx/resources/push.go`, `cmd/gcx/resources/delete.go` · **Class**: behavior

`EmittedError` suppresses the old `Error: N resource(s) failed to push` stderr line; get compensates
with an explicit warn, the others don't.

```bash
# dir with one valid + one broken manifest:
GCX_AGENT_MODE=0 bin/gcx resources push ./mixed-dir/; echo $?   # counts on stdout only, stderr silent
```

### 11. Safety-net comment overclaims
**Where**: `cmd/gcx/fail/convert.go:36-45` · **Class**: semantics

An error wrapping both a `DetailedError` and an `EmittedError` bypasses `convertEmittedError`
(DetailedError is extracted first), contradicting the "no caller can ever render a second document"
claim. No current call site constructs this — comment-vs-code gap only, demonstrable in a unit test,
not the CLI.

### 12. Doc/description drift
**Where**: `docs/design/errors.md` §4.4, PR body · **Class**: docs

Stale reference to nonexistent `handleError` (it's `reportError`, and the condition omits `--json`);
PR body claims 490 classified leaves, fixture has 484.

```bash
jq 'length' cmd/gcx/root/testdata/output_classes.json   # 484
grep -n 'handleError' docs/design/errors.md
```

---

Cleanup: `git worktree remove /tmp/gcx-main`.

dafydd-t:
☝️ adding fixes for these now..

dafydd-t:
@claude review

dafydd-t (approved):
My ✅ isn't effective because I added commits, but I'm comfortable enough with how this behaves to merge it.

petethepig (approved):

PR #1047, opened 2026-07-23
fix(aio11y): honor --config in direct agento11y CRUD commands

### LINKED ISSUE #951: aio11y: --config and config files are ignored — commands only work via GRAFANA_SERVER/GRAFANA_TOKEN env vars
## What happened

With a valid config file whose current context authenticates fine for every other provider:

```
gcx --config /path/to/config.yaml aio11y evaluators list
# → Invalid configuration found in '': grafana config is required

gcx --config /path/to/config.yaml --context eval aio11y evaluators list
# → invalid context "eval": context not found
```

The same `--config`/`--context` flags work for `slo`, `synthetic-monitoring`, `irm`, `api`, `resources` against the same stack. Only the env-var path works for aio11y:

```
GRAFANA_SERVER=https://<stack>.grafana.net GRAFANA_TOKEN=<sa-token> gcx aio11y evaluators list   # works
```

## Root cause

`internal/providers/aio11y/eval/evaluators/resource_adapter.go` — `NewTypedCRUD` constructs a fresh zero-value loader instead of the flag-bound one created in `provider.go` (`loader := &providers.ConfigLoader{}`, bound via `loader.BindFlags(...)`):

```go
func NewTypedCRUD(ctx context.Context) (*adapter.TypedCRUD[eval.EvaluatorDefinition], string, error) {
	var loader providers.ConfigLoader
	cfg, err := loader.LoadGrafanaConfig(ctx)
```

The zero-value loader has `configFile == ""`, which loads an empty config (the error's `found in ''`) rather than falling back to the default config path — so no config file is ever read, and `--context` then fails against the empty config. Env overrides are applied on top of the empty config, which is why the env-var path is the only one that works.

Same pattern is copy-pasted in the `rules`, `collections`, and `guards` resource adapters (`grep -rn 'var loader providers.ConfigLoader' internal/providers/aio11y/`).

## Expected

aio11y commands should honour `--config`/`--context` (and the default config path) like every other provider. The adapter factories presumably need the config path threaded through `context.Context` the same way the context *name* already is (`ConfigLoader.SetContextName` / `config.ContextNameFromCtx`), or the bound loader passed down instead of re-instantiated.

Worth noting for the bundled-skills evaluation: the `aio11y` skill instructs agents to use config-file auth, which currently cannot work — this blocks the skill eval until fixed or the skill documents the env-var requirement.

gcx: built from main. Found while dogfooding gcx for the bundled-skills evaluation work.

awsome-o:
Fixes #951.

The evaluators, rules, guards, and collections CRUD subcommands were building their clients from zero-value ConfigLoaders instead of the flag-bound loader the provider creates, so an explicit --config parsed fine and was then silently ignored. With a valid default config that means reads and mutations (delete, upsert) ran against the default context's stack while you thought you were targeting the --config one. #1012 fixed the resources tier by threading the selection through context, and it fixed default discovery and --context, but the direct agento11y command tree never consulted its bound loader.

The fix is the shape the SLO provider already uses: NewTypedCRUD takes the loader explicitly, commands pass the bound one, and the lazy adapter factories pass a local zero-value loader so the resources-tier path keeps inheriting the context threading unchanged. rules.Commands and guards.Commands gain the loader parameter they were missing, and collections update now builds its output envelope from the selected config too (before this its PATCH went to the right stack but the envelope namespace came from the default config, and the command failed outright when no default config existed).

Tests: cmd/gcx/root/aio11y_config_routing_test.go runs the real root command tree in-process against two recording servers. Config at the default location points at server A, --config points at server B, and every read and mutation for all four resource kinds must land on B with zero requests on A. Also covers --config beating GCX_CONFIG, --config with no default config present, the collections update envelope namespace (distinct stack-ids make the namespaces assertable), flag-before-subcommand position, and green smoke tests for the --context and env-override paths #1012 fixed. All the routing cases were written first and confirmed failing on unfixed main; the two smokes passed before and after.

Two adjacent findings that are deliberately not in this PR: the appo11y provider has the same defect (it binds --config to a loader it never passes to any subcommand, affecting services, overrides, and settings), and env-only auth (GRAFANA_SERVER/GRAFANA_TOKEN with no config file at all) currently fails for every ConfigLoader-based provider with a "$.stacks.default.grafana: node not found" error. Both reproduce on main without this change; issues to follow.

No CLI surface changes, so no reference drift. mise run gate is green locally; the docs step needs mkdocs which isn't installed here, CI covers it.

### DIFF
--- a/cmd/gcx/root/aio11y_config_routing_test.go
+++ b/cmd/gcx/root/aio11y_config_routing_test.go
@@ -7,10 +7,16 @@ package root_test
 // only the Agent Observability provider mounted, points two recording HTTP
 // servers A and B at distinct config files, and asserts every plugin-API
 // request lands on the server the user selected — and none on the other.
-// Distinct stack-ids give each config a deterministic namespace
-// (stacks-11111 vs stacks-22222) without any /bootdata discovery round-trip,
-// which also lets the collections-update case assert the output envelope's
-// namespace comes from the selected config.
+// Every fixed verb gets its own routing case because each subcommand
+// constructor captures the loader independently: a verb-local regression
+// cannot be caught by a neighboring verb's case.
+//
+// The two configs also differ in stack-id and token, so two same-shaped
+// mixed-path defect classes stay observable: distinct stack-ids give each
+// config a deterministic namespace (stacks-11111 vs stacks-22222, no
+// /bootdata round-trip) for the collections-update envelope assertion, and
+// distinct tokens let the recorder assert requests carry the selected
+// config's credentials, not the other config's or the environment's.
 
 import (
 	"bytes"
@@ -29,20 +35,30 @@ import (
 	"github.com/grafana/gcx/cmd/gcx/root"
 	"github.com/grafana/gcx/internal/providers"
 	"github.com/grafana/gcx/internal/providers/aio11y"
+	"github.com/grafana/gcx/internal/testutils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-// recorder captures every request that reaches a test server.
-type recorder struct {
-	mu   sync.Mutex
-	hits []string
+type hit struct {
+	method string
+	path   string
+	auth   string
 }
 
-func (r *recorder) record(method, path string) {
+func (h hit) String() string { return h.method + " " + h.path + " [" + h.auth + "]" }
+
+// recorder captures every request that reaches a test server, including the
+// Authorization header so credential routing is as observable as URL routing.
+type recorder struct {
+	mu   sync.Mutex
+	hits []hit
+}
+
+func (r *recorder) record(req *http.Request) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.hits = append(r.hits, method+" "+path)
+	r.hits = append(r.hits, hit{method: req.Method, path: req.URL.Path, auth: req.Header.Get("Authorization")})
 }
 
 func (r *recorder) count() int {
@@ -51,10 +67,10 @@ func (r *recorder) count() int {
 	return len(r.hits)
 }
 
-func (r *recorder) all() []string {
+func (r *recorder) all() []hit {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	return append([]string(nil), r.hits...)
+	return slices.Clone(r.hits)
 }
 
 func (r *recorder) reset() {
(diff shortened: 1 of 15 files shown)


awsome-o:
Addressed the xhigh review round in 259c50b1. Adjudication per finding:

Adopted (7): full verb routing matrix — 18 cases now, one per fixed verb, since every constructor captures the loader independently; I verified the matrix isn't decorative by temporarily reverting rules get to a zero-value loader and watching exactly that case go red with the wrong-target evidence in the failure message (GET on server A carrying token-a). Credential routing is now observable: the two configs carry distinct tokens, the recorder captures Authorization, and every case asserts the selected config's token — the env smoke also asserts the env token beats the config file's. Commands() in the four packages nil-guards its loader, which restores (and extends to all verbs) the Commands(nil) tolerance callers had on main; the collections tests are back to Commands(nil) and exercise the guard. The sandbox now isolates XDG_CONFIG_DIRS and the cwd .gcx.yaml layer and pins agent mode via testutils.SetAgentMode — the review was right that the t.Setenv pin was dead code, detection is cached at init. That sandbox moved to internal/testutils (SandboxConfigEnv) so the #1048 sibling reuses it instead of copying. The GCX_CONFIG test gained its positive control. And collections update now does one config load via a newCRUDAndClient helper, so the client and the envelope CRUD can't point at different stacks by construction.

Held (2), both previously adjudicated in the planning rounds: explicit loader passing stays (compile-enforced propagation was chosen over ctx-threading deliberately; ctx-threading remains an option for #1048 if we want to revisit at that scope), and the concrete *providers.ConfigLoader parameter stays for this bounded fix — the narrow-interface convention (slo/faro/kg/k6/synth) is real, but the other seven aio11y packages take the concrete type, and aligning all eleven is a follow-up sweep, not a release fix.

Partially adopted (1): harness extraction — the config/env sandbox is now shared in testutils; the recorder and plugin-server fakes stay local since they're aio11y-shaped. Consolidating configloader_test.go's isolateConfigSources onto the shared helper is left for the #1048 round.

One note on the review's collections claim: the PATCH itself was already routing through the bound loader before this round — the zero-loader half only fed ToUnstructured (envelope namespace) and made update fail when no default config existed. The single-load refactor removes the structural risk either way.

Gates: lint clean, full -race suite green, no reference drift.

awsome-o:
Both re-review findings closed in b663367f, and the second one was a fair hit — my claim that the reverted collections tests exercised the guard was wrong, they all return at flag validation before RunE touches the loader.

The guard moved to the choke point as suggested: resolvedContextName and resolvedConfigFile now treat a nil receiver like a zero-value loader, documented on the ConfigLoader type. That covers all eleven aio11y packages, both NewTypedCRUD paths, and any other provider handed a nil loader, in two one-line checks; the four pasted guards in the eval Commands() are removed, so the net diff is smaller than the round before. SaveDatasourceUID had an inline copy of the same context-name fallback and now calls resolvedContextName.

Teeth, empirically this time: two new tests, and I verified both go red by removing the nil checks and running them — the choke-point unit test (nil LoadGrafanaConfig equals zero-value, in configloader_test.go) and an end-to-end collections Commands(nil) list that reaches RunE and an HTTP round-trip through default discovery in SandboxConfigEnv. Both panicked without the guard, both green with it.

Gates: lint clean, full -race suite green.

awsome-o:
Closed the env-leak nit in the latest commit: the nil-loader test moved from isolateConfigSources onto testutils.SandboxConfigEnv, since its namespace assertion is hard-coded and an ambient GRAFANA_STACK_ID flipped it (reproduced with GRAFANA_STACK_ID=99999 before the fix, green after). Side benefit: the testutils import brings its env-clearing init into the providers test binary, so that package is now hermetic against agent-mode shells too. Lint and the full -race suite are green. That's the last open finding — from my side #1047 is merge-ready.

Live E2E addendum (head d247467b):

- Pointed `GCX_CONFIG` at a deliberately unusable decoy config, then passed the valid Grafana Cloud config with `--config`. Successful authenticated requests prove the flag-bound config wins and routes to the intended stack—the exact #951 failure mode.
- Exercised read/list routing for evaluators, rules, guards, and collections.
- Ran unique temporary lifecycles for an inert regex evaluator (upsert/get/list/delete), an empty collection (create/get/update/list/delete), and a disabled/warn impossible-match guard (create/get/update/list/delete).
- Exercised rule create/get/list/delete with a unique impossible-match rule.
- Verified cleanup with final filtered lists across all four resource families; no temporary resources remain.
- One separate pre-existing live incompatibility surfaced: rule creation returned `enabled: true` for an input of `false`, and `rules update` reached the selected stack but the backend rejected the existing payload with `400 unknown field "rule_id"`. That payload behavior is unchanged by this config-routing diff and does not affect the #951 fix.

Result: the #951 `--config` routing bug is fixed, and the change is ready for review.

edwardcqian (approved):
As you already called out, passing loader is a good patch fix. Revisiting might make sense in the future since this feels like it should just be implicit.

Edit: oof, don't think I'm listed as owner for this repo 😢

annanay25 (approved):

PR #1050, opened 2026-07-23
feature: Tier 0 agentic support for metrics list-names

### LINKED ISSUE #1035: Add gcx metrics list subcommand with name filtering
### Drilldown idea

There is no discoverable way to list metric names in gcx. The metrics provider (internal/providers/metrics/provider.go) registers query, labels, metadata, series, billing, cardinality, and adaptive, but no `list`. Listing metric names is technically reachable via `gcx metrics labels --label __name__` or a `group by (__name__)` query, but neither is discoverable as "list metrics" and neither supports prefix, suffix, or text filtering, which is what users and LLM agents actually need when hunting for a metric name.

### Proposed change: add `gcx metrics list`, a thin wrapper over the label-values endpoint for `__name__`, with:

- `--match` selector scoping (depends on the match[] plumbing from the labels enhancement)
- Client-side `--prefix`, `--suffix`, and `--contains` filters for name matching
- Standard output formats (table default for humans, json for agents)

Example usage:

```
# All metric names
gcx metrics list -d UID

# Find cart-related metrics
gcx metrics list -d UID --contains cart

# Counter discovery
gcx metrics list -d UID --suffix _total

# Metrics present on a job
gcx metrics list -d UID --match '{job="api"}'
```

### More context

This work addresses a gap found in the [Drilldown design doc research to support GCX](https://docs.google.com/document/d/1D29TOTXy7cs5Q2XlJ1nBRieGIsppeCmQGRsB3WJ1qXA/edit?tab=t.0#heading=h.5sybau7waq2q). 

### Future work

Drilldown has decided to start the implementation in the metrics tool to get early testing done on the helpfulness of this feature. The drilldown apps plan to have a backend and port this functionality to that API. This work is the first step before we have the backend in place.

bohandley:
Fixes #1035

### What is this?

This is the second PR for drilldown apps to support GCX. The drilldown apps' expertise is both queryless experience (heuristics for building from primitives) and strategies for breaking down on dimensions.

This PR adds `gcx metrics list-names` so agents can list metric names, similar to the Metrics Drilldown feature that shows all metrics. It builds on the match[] scoping added to the labels command in #1037 so agents can filter the list.

### What does it look like?

- `--match` selector scoping, validated client-side before any request (union semantics for repeated selectors, per the Prometheus match[] API)
- Client-side `--prefix`, `--suffix`, and `--contains` filters for name matching (case-sensitive)
- `--limit` capped at 100 names by default; truncated pages carry `list_meta` in the payload per the #1046 output contract, including under `--json` field selection
- Standard output formats (table default for humans, json for agents)

```
# Metric names (first 100 by default)
gcx metrics list-names -d UID

# Find cart-related metrics
gcx metrics list-names -d UID --contains cart

# Counter discovery
gcx metrics list-names -d UID --suffix _total

# Metrics present on a job
gcx metrics list-names -d UID --match '{job="api"}'
```

The command is named `list-names` (not `list`) per command-naming.md: metric names are a discovery facet with no independently addressable item, like `list-profile-types` and `list-regions`.

Also in this PR, review hardening on the labels command: positional args rejected, `--match` validated client-side, explicitly empty `--metric`, `--label`, and name filter values error instead of silently answering a different question, and table output routed through the shared codec.

Note: adding Args validation tightens three v1.0.0 commands (`gcx metrics labels`, `gcx metrics billing labels`, `gcx datasources prometheus labels`) that previously accepted and ignored positional args; flagged in review as the intended end state.

Fast follows from review: #1064 (time range on labels), #1065 (push filters and limit down to the server).

### Who does this help?

This helps agents explore metric signals in a more organized way. This provides [Tier 0 support in this doc](https://docs.google.com/document/d/1kzgCaz-hViwB1FeIXsEwY2_vTQAbmuq5rGIPRrrPqSU/edit?tab=t.0) by @cedricziel used to guide the [wg-o11y-agentic-ai](https://grafanalabs.enterprise.slack.com/archives/C088W73SEP8), so it helps o11y apps by supporting agentic workflows.

### More context

[Read this issue](https://github.com/grafana/gcx/issues/1034) to learn more about how the drilldown apps are helping support agentic workflows first in gcx with the telemetry api and next with their own backends (still in the research phase).


internal/providers/metrics/list.go
```diff
@@ -0,0 +1,156 @@
...
+		filtered = append(filtered, name)
+	}
+	return filtered
+}
+
+func listCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &listOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
```
awsome-o: cobra accepts positional args here and RunE drops them. Someone used to the series command will type `gcx metrics list '{job="api"}'` and get the entire unfiltered name list back, formatted exactly like a successful filtered answer, which is the worst kind of wrong because nothing hints the filter was ignored. `Args: cobra.NoArgs` turns that into an immediate error.

internal/providers/metrics/provider.go
```diff
@@ -77,6 +86,11 @@ func (p *Provider) descriptor() signals.Descriptor {
 				TokenCost: "medium",
 				LLMHint:   `gcx metrics series -d abc123 '{__name__="up"}' --since 1h -o json`,
 			},
+			{
+				Build:     listCmd,
+				TokenCost: "small",
```
awsome-o, line 91: small doesn't match what an unfiltered run returns. On a real cloud stack the __name__ list is easily tens of thousands of entries, so an agent that picks this command because it was advertised as cheap gets hundreds of KB of JSON, and a human at a terminal gets a table with 30k rows and no escape hatch. Other list commands in the tree cap output with --limit and print a truncation hint for exactly this reason. I'd add --limit here so small stays honest, or failing that call it medium or large.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,156 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			resp, err := client.LabelValues(ctx, datasourceUID, "__name__", opts.Match)
+			if err != nil {
+				return fmt.Errorf("failed to list metric names: %w", err)
+			}
+
+			resp.Data = filterMetricNames(resp.Data, opts.Prefix, opts.Suffix, opts.Contains)
```
awsome-o: Heads up that --prefix, --suffix and --contains all run client side, so `gcx metrics list -d UID --contains cart` still pulls the complete __name__ value set through the datasource proxy on every call just to keep a handful of names. Pushing the filter down as a match[] selector, `{__name__=~".*cart.*"}` style, keeps the transfer proportional to the answer and dodges any proxy body size limits. Not a blocker for a first cut if --limit lands, but big tenants will feel the difference.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,156 @@
...
+  # Output as JSON
+  gcx metrics list -d UID -o json`,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if err := opts.Validate(); err != nil {
+				return err
+			}
+
+			ctx := cmd.Context()
+
+			cfg, err := loader.LoadGrafanaConfig(ctx)
```
awsome-o: This block is an inline copy of dsquery.LoadContextAndConfig, which labels.go in this same stack already calls. Beyond tidiness there's a concrete reason to prefer the helper: the config loading path is where the recent --config bugs lived (#1047 and #1053 fix that family through the shared code), and a command carrying its own copy silently misses every future fix there. Swapping it in also drops the internalconfig, logging and slog imports.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return filtered
+}
+
+func listCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &listOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List metric names.",
```
awsome-o: nitpick: trailing period. The sibling short descriptions under gcx metrics don't end with one, so this entry renders with odd punctuation next to them in `gcx metrics --help` and in the generated command index.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,156 @@
...
+		Long: `List metric names from a Prometheus datasource via the label values endpoint for __name__.
```
awsome-o: nitpick: __name__ sits in plain prose here, and once the generated doc page renders as markdown the double underscores turn into bold, so the site shows 'the label values endpoint for name' with name bolded. Backticks around __name__ in this Long string fix the page on regeneration.

internal/query/prometheus/formatter.go
```diff
@@ -210,6 +210,15 @@ func FormatLabelsTable(w io.Writer, resp *LabelsResponse) error {
 	return t.Render(w)
 }
 
+// FormatMetricNamesTable formats a LabelsResponse holding metric names as a table.
+func FormatMetricNamesTable(w io.Writer, resp *LabelsResponse) error {
```
awsome-o: nitpick: this is FormatLabelsTable with a different header string. A small shared helper that takes the header would keep the two from drifting the first time one of them gets a styling or truncation fix.

awsome-o (commented):
Nice follow up to #1037, being able to grep the metric namespace from the CLI is genuinely handy. I checked out the branch and read through it, notes inline. Two things I'd fix before merge: the command tells agents it's a small token cost while an unfiltered run returns the tenant's complete metric name list with no --limit to cap it, and the config loading is a hand rolled copy of a shared helper that just had a round of bug fixes, so this command would quietly miss those. Whatever you settle on for the match[] union question over on #1037 applies to repeated --match here as well.

Two questions rather than findings. billing calls itself thin conveniences over the generic metrics subcommands and wraps query, labels and series; should it grow a `billing list` too, or is that deliberately out of scope? And on the name: the exploration siblings are nouns (labels, series, metadata) while list is a bare verb, something like `gcx metrics names` would say what comes back. We do have `gcx logs query` as verb precedent so either can be argued, I just want it to be a choice rather than an accident, since renames after release are breaking for agent skills and docs.

Last small one, TestListCmd_Flags proves the flags exist but nothing checks --match actually reaches LabelValues, so a refactor that drops opts.Match would still pass the suite.

internal/providers/metrics/list.go
bohandley: Fixed. Typing something after the command, like a selector, now returns an error instead of being silently ignored.

internal/providers/metrics/provider.go
bohandley: Fixed. Added --limit with a default cap of 100 names, and when the list is longer you get a note on stderr saying how many were held back and how to get them all with --limit 0.

internal/providers/metrics/list.go
bohandley: Agreed and deferred as you suggested: --limit landed in this PR, and I'll make a follow up issue for sending the filters to the server instead of filtering locally.

internal/providers/metrics/list.go
bohandley: Fixed. Swapped in the shared helper and dropped the three imports.

internal/providers/metrics/list.go
bohandley: Fixed, dropped the period to match the siblings, though heads up that the written standard in docs/design/help-text.md asks for the period 🤔 Happy to keep it consistent either way!

internal/providers/metrics/list.go
bohandley: Fixed, `__name__` is backticked in the Long text now and the reference docs were regenerated with it

internal/query/prometheus/formatter.go
bohandley: Fixed, both formatters now call one shared single-column helper that takes the header, so a future styling fix lands in both.

internal/providers/metrics/list.go
bohandley: https://github.com/grafana/gcx/issues/1065

awsome-o (commented):
Went through the new commits, and the round one items all landed well, NoArgs, the shared config loader, the command layer tests, the docs and formatter cleanups, thanks for the fast turnaround. Two new problems came in with this push though, both must fix before merge. Neither existed in the version I reviewed on Friday, so this isn't me moving goalposts, details inline.

Beyond those two the rest is non blocking. One decision worth making on purpose before v1 freezes the command surface: list vs names for the command name, and whether it should also mount under `gcx datasources prometheus` (it can't from its current package). Renames after v1 are breaking for agent skills and docs, so whatever you pick now is what we live with.

internal/datasources/prometheus/labels.go
```diff
@@ -32,7 +32,7 @@ func (opts *labelsOpts) setup(flags *pflag.FlagSet) {
 	opts.IO.DefaultFormat("table")
 	opts.IO.BindFlags(flags)
 
-	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.prometheus is configured)")
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless default-prometheus-datasource is configured)")
```
awsome-o: This line regressed through the merge chain. Fix commit 3514077b on the labels branch rewrote it to point at default-prometheus-datasource, commit 89b8742b later restored the right key over there, but merge fbc97909 into this branch resolved the line back to the stale version. So main and #1037 both say datasources.prometheus while this branch tells people to configure a key the config system removed. I built the branch to be sure: `gcx config set default-prometheus-datasource UID` fails hard with `Legacy path "default-prometheus-datasource" was removed, use contexts.<name>.datasources.prometheus`, so the help sends users to a dead end that contradicts the command's own error message. The docs regen also baked the dead key into three reference pages (gcx_metrics_labels, gcx_metrics_billing_labels, gcx_datasources_prometheus_labels). Re merging the current tip of the labels branch, taking its side for this line, and regenerating docs should make the whole thing disappear.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,157 @@
...
+
+			resp, err := client.LabelValues(ctx, datasourceUID, "__name__", opts.Match)
+			if err != nil {
+				return fmt.Errorf("failed to list metric names: %w", err)
+			}
+
+			resp.Data = filterMetricNames(resp.Data, opts.Prefix, opts.Suffix, opts.Contains)
+
+			if total := len(resp.Data); opts.Limit > 0 && total > opts.Limit {
+				resp.Data = resp.Data[:opts.Limit]
```
awsome-o: Thanks for adding the limit, but the way truncation is signaled will mislead the exact consumers the limit protects. Slicing resp.Data with the hint on stderr means a 5000 metric stack produces a JSON body that is status success, exactly 100 names, and nothing else. Under the output contract from #1046 a list payload without list_meta means the result set is complete, the rule is spelled out in [internal/output/listmeta.go](https://github.com/grafana/gcx/blob/main/internal/output/listmeta.go), so anyone piping to jq with stderr dropped reads the truncated page as the entire inventory. That's worse than the unbounded dump was, wrong but plausible instead of big but correct. The machinery already exists: BindListLimit for the flag, TruncateCompleteList and AttachListMeta for the payload, EmitListTruncationHint for the human message, and [cmd/gcx/datasources/list.go](https://github.com/grafana/gcx/blob/main/cmd/gcx/datasources/list.go) shows them wired together. I should have pointed at those helpers by name last round when I asked for the limit, sorry for the extra loop, that one's on me.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,157 @@
...
+			if err != nil {
+				return err
+			}
+
+			client, err := prometheus.NewClient(cfg)
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			resp, err := client.LabelValues(ctx, datasourceUID, "__name__", opts.Match)
```
awsome-o, line 152: Follow up material, not a blocker: the fetch still pulls the complete __name__ set and applies the name filters and the limit client side. The label values endpoint accepts a limit query parameter, and the name filters could push down as a __name__ regex match[]. On big Mimir tenants that's the difference between a response sized to the answer and one sized to the tenant.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,157 @@
...
+			return opts.IO.Encode(cmd.OutOrStdout(), resp)
+		},
+	}
+
+	opts.setup(cmd.Flags())
+
+	return cmd
+}
+
+type metricNamesTableCodec struct{}
```
awsome-o: nitpick: metricNamesTableCodec and labelsTableCodec are the same type with a different formatter call. Now that formatSingleColumnTable exists, one parameterized codec next to it would stop a third copy appearing with the next single column command.

bohandley:
Thank you @awsome-o, here's the work done for the PR

- Restored correct datasources.prometheus help text on labels, lost in a merge, plus three docs
- Rewired list truncation to the list_meta contract: envelope payload, shared limit flag, stderr hint
- Labels command now rejects positional args instead of silently returning the full unscoped list
- Invalid --match selectors fail with a clear client-side error even without --metric present
- Added Pattern 14 comment justifying brace-joining of canonical matcher renderings in selector assembly
- Explicitly empty --metric now errors, catching unset shell variables instead of dropping scoping
- Deduplicated labels and list table codecs into one shared parameterized single-column codec
- Drafted #1065 comment: push filters and limit down to the label values endpoint

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,161 @@
...
+	opts.setup(cmd.Flags())
+
+	return cmd
+}
+
+// metricNamesListResult is the single shape passed to every codec for
+// `gcx metrics list`. JSON/YAML serialize the envelope; the table codec
+// extracts .Data to render rows (Pattern 13: format-agnostic data).
+type metricNamesListResult struct {
+	Data []string `json:"data" yaml:"data"`
```
awsome-o, line 182: The envelope reads right for `-o json`, but `list_meta` drops out under `--json`, unless I'm misreading the selection code. I ran the encode path on the branch with five names and `--limit 2`:

```
-o json      -> {"data":["a_total","b_total"],"list_meta":{"truncated":true,"returned":2,"total":5,"continue":"gcx metrics list -d UID --limit 0"}}
--json data  -> {"data":["a_total","b_total"]}
```

[`singleKeyItems`](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/internal/output/field_select.go#L329-L346) only treats a payload as a list envelope when the items are objects and it gives up on a scalar array, so `envelopeFieldSelection` never runs and the reserved key gets dropped along with every other field nobody asked for. [output.md §15.5](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/docs/design/output.md#L430-L443) promises the opposite, that selection re attaches `list_meta` so the signal survives.

I don't read this as your mistake. It looks like the first `[]string` envelope to take up the #1046 contract, since `datasources list` and the oncall alert groups list both carry object items, so the gap is probably in `internal/output`. It bites here because agent mode prints a hint pointing agents at `--json field1,field2`, so the consumer the limit exists to protect is the one that loses the signal. Teaching `singleKeyItems` to pass a scalar array through and re attach the key would close it, or shaping the items as objects would too, with a test on a scalar envelope to keep it closed.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,161 @@
...
+		filtered = append(filtered, name)
+	}
+	return filtered
+}
+
+func listCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &listOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
```
awsome-o: Coming back to the list vs names thread now that #1023 is merged, because I think it settles it. The guide says a discovery or catalog facet with no independently addressable item takes an `<operation>-<subject>` compound, with `gcx datasources cloudwatch list-regions` as the example ([command-naming.md](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/docs/design/command-naming.md#L134-L139)). In the tree every `list` leaf in a signal family sits under a noun group, `gcx metrics adaptive rules list` and friends, and the leaves hanging straight off a signal are nouns or compounds: labels, series, metadata, query, [`gcx profiles list-profile-types`](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/internal/datasources/pyroscope/list_profile_types.go#L46). There's no bare `<signal> list` today, so this would be the first one.

CONSTITUTION also now freezes the v1 surface for all of v1.x and says a naming rule isn't grounds for renaming a released command ([CLI Grammar](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/CONSTITUTION.md#L75-L84)), so whatever ships here is what we keep. On that reading I'd lean `list-names`, it matches `list-profile-types` and reads clearly next to labels and series. If you read the shorthand rule as already covering a bare `list` I'm happy to hear that argument, I mostly don't want the name inherited by accident.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,161 @@
...
+			if err != nil {
+				return err
+			}
+
+			client, err := prometheus.NewClient(cfg)
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			resp, err := client.LabelValues(ctx, datasourceUID, "__name__", opts.Match)
```
awsome-o, line 152: `opts.Match` goes to the server exactly as typed here, while `selectors()` on labels now parses it client side, with a comment saying an opaque proxied 400 is the worse outcome. So `gcx metrics list --match '{job="api"'` comes back as a 400 wrapped in `failed to list metric names`, and the same typo on labels fails immediately with the selector quoted back. Reusing that parse loop would make the two behave the same, unless you're deliberately keeping this path thin and letting Prometheus be the judge.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (opts *listOpts) Validate() error {
+	return opts.IO.Validate()
+}
+
+// filterMetricNames returns the names passing every configured name filter.
+// Filters combine with AND; with no filters configured it returns names as-is.
+func filterMetricNames(names []string, prefix, suffix, contains string) []string {
+	if prefix == "" && suffix == "" && contains == "" {
```
awsome-o, line 61: Two small things about this short circuit.

An explicitly empty filter reads as no filter, so `--contains "$SVC"` with the variable unset returns the first 100 names in the tenant, formatted exactly like a filtered answer. That's the same trap the new `--metric` guard on labels catches, so it seemed worth raising for consistency.

Also, because the early return hands back the input slice while the filtered path allocates with `make`, `data` serialises as `null` with no filters and `[]` with any filter, so a `jq '.data[]'` pipeline lives or dies on which flags were passed. A `Changed()` check for the three filters plus returning an empty slice here would cover both, though if you'd rather not error on the empty case that's fair enough.

internal/providers/metrics/list.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			return result.Data, true
+		},
+	})
+	opts.IO.DefaultFormat("table")
+	opts.IO.BindFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.prometheus is configured)")
+	flags.StringArrayVar(&opts.Match, "match", nil, "Series selector(s) to scope results; repeatable (repeated selectors combine as a union, per the Prometheus match[] API)")
+	flags.StringVar(&opts.Prefix, "prefix", "", "Only include names starting with this string")
```
awsome-o: nit: the three filters match case sensitively and the help doesn't mention it. [`gcx datasources list --name`](https://github.com/grafana/gcx/blob/0a9281aa2427d62d04e23aed6fa285a2fc9f94c6/cmd/gcx/datasources/list.go#L33) goes the other way and says "case-insensitive substring match" while lowercasing both sides, so someone trying `--contains alerts` and missing `ALERTS` has nothing to go on. A few words in the flag help or in Long would do, no strong feeling on which, and no objection to case sensitive being the behaviour.

internal/providers/metrics/list_test.go
```diff
@@ -0,0 +1,212 @@
...
+	require.Equal(t, "list", cmd.Name())
+
+	for _, name := range []string{"datasource", "match", "prefix", "suffix", "contains", "limit", "output"} {
+		assert.NotNil(t, cmd.Flags().Lookup(name), "missing flag --%s", name)
+	}
+
+	// Name filters are provider-specific options and must not have shorthands
+	// (docs/design/naming.md 9.4).
+	for _, name := range []string{"match", "prefix", "suffix", "contains", "limit"} {
+		assert.Empty(t, cmd.Flags().Lookup(name).Shorthand, "--%s must not have a shorthand", name)
```
awsome-o, line 93: nit: this proves the three filters exist, but nothing drives them through `runListCmd`, so if `--prefix` were bound to `opts.Suffix` the suite would still be green. Same shape as the `--match` gap from round one, and the harness is already sitting right there, so it's a few lines.

internal/datasources/prometheus/labels.go
```diff
@@ -134,6 +154,13 @@ func LabelsCmdWithDefault(loader *providers.ConfigLoader, defaultDS string) *cob
 				return err
 			}
 
+			// An explicitly empty --metric (typically an unset shell variable,
+			// as in --metric "$METRIC") would silently drop scoping; error
+			// instead of returning unscoped results.
+			if cmd.Flags().Changed("metric") && opts.Metric == "" {
```
awsome-o: Good catch on the empty `--metric`, that's a real footgun. I think the same reasoning applies to `--label` down at line 191: an empty value falls into the `if opts.Label != ""` branch and returns the full label name list, in the same `{status, data}` shape a label values response uses, so `--label "$LBL"` with the variable unset quietly answers a different question than the one asked. If you'd rather keep the guard to just what review asked for and do `--label` separately, no argument from me, flagging it while we're in here.

internal/datasources/prometheus/labels.go
```diff
@@ -28,7 +26,16 @@ type labelsOpts struct {
 }
 
 func (opts *labelsOpts) setup(flags *pflag.FlagSet) {
-	opts.IO.RegisterCustomCodec("table", &labelsTableCodec{})
+	opts.IO.RegisterCustomCodec("table", &prometheus.SingleColumnTableCodec{
```
awsome-o, line 28: The shared codec is registered here, but RunE still returns `prometheus.FormatLabelsTable(...)` directly in both branches when the format is table, around lines 197 and 209, so on this command the registration mostly serves format validation and the codec's Encode never actually runs. Nothing renders wrong today, it just leaves two table paths alive, so a later header or truncation change inside the codec would silently skip `gcx metrics labels`. Dropping those two table branches looks like all it needs, unless they're doing something I'm not seeing.

internal/datasources/prometheus/labels.go
```diff
@@ -110,6 +129,7 @@ func LabelsCmdWithDefault(loader *providers.ConfigLoader, defaultDS string) *cob
 	cmd := &cobra.Command{
 		Use:   "labels",
 		Short: "List labels or label values",
+		Args:  cobra.NoArgs,
```
awsome-o, line 131: nit, or really a heads up: cobra's `legacyArgs` accepts and ignores positionals on a leaf command, so this tightens three commands that shipped in v1.0.0, `gcx metrics labels`, `gcx datasources prometheus labels` and `gcx metrics billing labels`. I asked for it and still think it's the right end state, I'd just call it out in the PR description now that the v1 surface is frozen, so the change is in the record.

awsome-o (commented):
Round three. The two blockers from Sunday look fixed to me. I checked out the branch and the labels flag help says `datasources.prometheus` again, and truncation now runs through `BindListLimit`, `TruncateCompleteList`, `AttachListMeta` and `EmitListTruncationHint`, which is the mechanism I was hoping for. Thanks for the turnaround, and for filing #1065.

One item I'd still fix before merge, and I don't think it's yours: `list_meta` disappears when the caller uses `--json`. Details on the envelope. The list vs names question also has an answer now that #1023 merged into main, and that one is worth settling before this ships rather than after. Everything else inline is small, and I've marked the nits as nits.

On the trailing period from last round, keep it the way you have it. The guide asks for one and the whole metrics family doesn't, so the guide is the piece that's out of step and it shouldn't sit on your PR. I'll pick that up on the docs side.

Not for this PR, just so it doesn't get lost: `create-dashboard/SKILL.md` and `manage-dashboards/SKILL.md` still tell agents to discover metric names with `labels -d <uid> --label __name__`, so until those point here the agents keep taking the uncapped path.

internal/providers/metrics/list.go
bohandley: Fixed in internal/output!

internal/providers/metrics/list.go
bohandley: Renamed to list-names. Read the guide after pulling main: metric names are a discovery facet with no addressable item, so rule 1 applies, and it reads right next to list-profile-types. Docs, output_classes, and tests updated.

internal/providers/metrics/list.go
bohandley: Fixed, same parse loop as labels: a typo now fails immediately with the selector quoted back, and valid selectors go to the server exactly as written.

internal/providers/metrics/list.go
bohandley: Both fixed: explicitly empty --prefix/--suffix/--contains now error like the --metric guard, and filterMetricNames always returns a non-nil slice so data is [] rather than null regardless of flags.

internal/providers/metrics/list.go
bohandley: Added "(case-sensitive)" to all three flag help strings and regenerated the doc. Behavior unchanged.

internal/providers/metrics/list_test.go
bohandley: Added: each filter plus their AND combination now runs through the command against the capture server, so a swapped binding fails the suite.

internal/datasources/prometheus/labels.go
bohandley: Took it in this PR since we were in the file: the empty-value guard now covers --label too, with the same error shape and a zero-requests test.

internal/datasources/prometheus/labels.go
bohandley: You read it right, they were legacy from before the codec existed. Both branches dropped, the now-orphaned FormatLabelsTable deleted, and a -o table test proves the codec path renders.

internal/datasources/prometheus/labels.go
bohandley: Called out in the PR description with the three affected commands, so it's in the record.

awsome-o (approved):
Rechecked the latest head (3a1b8339). The list_meta field-selection issue from the last round is fixed, I reproduced the old behavior on main and the fix here. Selector validation, the name filters, and the truncation signalling all look correct, both guards reject before any request goes out, and everything is green on my side: full suite, -race on the changed packages, isolated lint at 0 issues, build, and all the CI checks.

One thing I'm deliberately not asking for. These validators return exit 1 for invalid input where the taxonomy in DESIGN.md says 2. That isn't this branch's doing — it's how Options.Validate() has always behaved (unknown -o, negative --limit, bad --jq all exit 1), and it's the same across roughly 134 of the command files, so moving these six sites alone would just make list-names the one command in gcx that classifies usage errors correctly while every neighbor stays at 1. Worse for the wrappers the taxonomy exists for. Filing it separately as a centralized fix.
   - tracking here: https://github.com/grafana/gcx/issues/1117

Thanks for working through the rounds on this one. Approved.

PR #1051, opened 2026-07-23
fix(cloud): stop misreporting invalid stack slugs as "slug already taken"

### LINKED ISSUE #950: cloud stacks create: invalid slug (409 InvalidArgument) misreported as 'Stack slug already taken'
## What happened

```
gcx cloud stacks create --name "gcx eval" --slug my-gcx-eval --region us -o json
```

```json
{"error":{"summary":"Stack slug already taken","exitCode":1,"details":"failed to create stack: gcom client: unexpected status 409 Conflict: {\n  \"code\": \"InvalidArgument\",\n  \"message\": \"Invalid slug my-gcx-eval specified\",\n  ...}","suggestions":["Choose a different slug with --slug","List existing stacks: gcx cloud stacks list --org <org-slug>"]}}
```

The slug was not taken — it was *invalid*. GCOM rejects non-alphanumeric slugs (hyphens included; the [Cloud API docs](https://grafana.com/docs/grafana-cloud/developer-resources/api-reference/cloud-api/) note lowercase-only, and stack slugs become `<slug>.grafana.net` subdomains). gcx maps every 409 from this endpoint to "Stack slug already taken", so the summary and both suggestions point the user in the wrong direction. The real cause is only visible in the raw `details`.

## Expected

1. Distinguish the 409 body's `code`/`message` (`InvalidArgument` vs an actual duplicate-slug conflict) and surface GCOM's message as the summary.
2. Better: validate the slug client-side (lowercase alphanumeric) before the request, with a message like `stack slugs may only contain lowercase letters and digits (they become <slug>.grafana.net)`.
3. `--dry-run` currently passes an invalid slug without complaint — client-side validation would fix that too.

gcx: built from main. Found while dogfooding gcx for the bundled-skills evaluation work.

awsome-o:
Fixes #950.

GCOM's create-stack 409 turns out to be a generic invalid-arguments channel, not a duplicate-slug signal — the public OpenAPI describes the response as "Invalid arguments: wrong or missing parameters", and its ErrorConflict schema uses that same text as the example message for code Conflict. gcx was mapping every non-delete stacks 409 to "Stack slug already taken", so the invalid-slug 409 from the issue sent users off to pick a different slug when the slug was never taken at all. Update/list/get 409s got the same wrong label.

Three changes. The GCOM client now parses the error body's code and message best-effort into GCOMHTTPError — raw Body and the Error() string are byte-for-byte unchanged, so the login path and every existing consumer behave identically. convertStacksErrors then dispatches on operation plus parsed code: create with InvalidArgument becomes "Invalid stack request" with exit code 2 and GCOM's exact message as the first line of the details (slug advice only appears when the message actually mentions the slug, since the same code covers bad regions and the rest); every other stacks 409 becomes "Resource conflict". I retired "Stack slug already taken" entirely rather than guess — code Conflict is an error class, not a duplicate discriminator — but a duplicate-looking message still selects the choose-another-slug suggestions. And stacks create now validates the slug client-side (lowercase alphanumeric, the documented rule; hyphens were the issue's exact repro) in a proper Validate() that runs before the dry-run branch, so --dry-run rejects invalid slugs instead of previewing them.

One deliberate deviation from the issue text: it asked for GCOM's message as the error summary. Summaries stay static here (vocabulary stability, and the agent failure log truncates them at 200 chars); the message is surfaced verbatim as the first line of details in both the human rendering and the JSON envelope instead. Easy to flip if reviewers prefer the literal reading.

Tests: convertStacksErrors gets coverage for the first time (all the 409 shapes including unknown codes and update/list wrappers, plus the existing 403/401 branches), a body-parsing table on the GCOM client asserting Status, Body and Error() are preserved whether or not the parse succeeds, slug-validation tables on the create command including validation-precedes-dry-run, a docs.All() membership guard for the new CloudAPI link constant, and a process-level agent conformance case pinning exactly one gcx.error document with exit code 2 in both the process status and the in-band payload.

The new "Invalid stack request" summary is added to the errors.md vocabulary and the Resource conflict row now covers API-reported conflicts. Reference docs regenerated for the --slug help text change.

### DIFF
--- a/cmd/gcx/fail/convert.go
+++ b/cmd/gcx/fail/convert.go
@@ -1394,7 +1394,7 @@ func convertStacksErrors(err error) (*gcxerrors.DetailedError, bool) {
 		if strings.Contains(msg, "failed to delete stack") {
 			return &gcxerrors.DetailedError{
 				Summary: "Stack has delete protection enabled",
-				Details: msg,
+				Details: gcomErrorDetails(httpErr, msg),
 				Parent:  err,
 				Suggestions: []string{
 					"Disable delete protection first: gcx cloud stacks update <slug> --no-delete-protection",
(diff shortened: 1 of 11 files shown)


awsome-o:
## E2E verification

Tested the PR head (`fc1d3d1c`) using the compiled `bin/gcx` binary.

### CLI validation path

- Ran `gcx cloud stacks create` with the issue’s hyphenated slug under human output, explicit JSON output, and agent-mode defaults.
- Confirmed the command fails before rendering a dry-run preview or making a network request.
- Confirmed process exit code `2`, the `Invalid command usage` summary, actionable slug guidance, and the Cloud API documentation link.
- Confirmed agent mode emits exactly one `gcx.error` document whose in-band `exitCode` matches the process status.
- Ran a valid lowercase/digit slug with `--dry-run` and confirmed the expected structured POST preview.

### Full GCOM error path

Ran the compiled CLI against a temporary local GCOM-compatible HTTP endpoint so the complete request → HTTP response → `GCOMHTTPError` parsing → converter → process output path was exercised without provisioning infrastructure:

- `409 {"code":"InvalidArgument","message":"Invalid region narnia specified"}` → `Invalid stack request`, exit `2`, exact server message first in details, and general flag/dry-run remediation.
- `409 {"code":"Conflict","message":"that url has already been taken"}` → `Resource conflict`, exit `1`, exact server message first in details, and create-specific slug remediation.

A read-only request was also made to the real Grafana Cloud API using the existing local configuration. The stored token is expired, so a live 409 could not be exercised; however, the real `401 InvalidCredentials` response verified that the newly parsed server message leads the rendered details while authentication classification and exit code `3` remain intact. No credentials were changed and no Cloud resources were created or modified.

### Regression/safety checks

- Targeted changed-package tests passed.
- `go test ./...` passed.
- `go test -race -count=1 ./...` passed.
- Lint passed with zero issues.
- Binary build and CLI reference drift check passed.
- The branch is current with `main`.
- All required GitHub checks pass.

Result: the issue reproduction is fixed, the primary success paths are unchanged, and no blocking regression or safety concern was found. The only live-E2E limitation was the expired Cloud token preventing a real Grafana Cloud 409 response test.

annanay25 (approved):
lgtm!

PR #1052, opened 2026-07-23
feat(serve): enable OAuth for `gcx dev serve`

jkeljo:
`gcx dev serve` proxies browser requests to the real Grafana instance to render local dashboards. It previously hand-rolled authentication (basic auth or API token only) and its own TLS setup, so it never worked with OAuth login. This reworks the proxy to reuse the shared REST transport that the rest of gcx already uses, and fixes the URL-path handling that change exposes.

The change has the following parts:

* Transport reuse (core): the serve command now loads a NamespacedRESTConfig and builds the proxy's HTTP transport from it via rest.TransportFor, so proxied requests carry whichever auth mode the context is configured for (basic, service-account bearer, or OAuth with token refresh) plus TLS, retry, and logging. server.New now returns an error since building the transport can fail.

* Centralized proxy target: NamespacedRESTConfig.ProxyTarget() owns what the Grafana host/path means per auth mode -- empty path in direct mode (the Grafana subpath is already carried by the incoming request) or the assistant-backend proxy prefix in OAuth mode. Both proxy paths (the reverse proxy and the dashboard-HTML handler) now assemble outgoing URLs through it instead of each re-deriving the branch.

* Subpath doubling fix: routing the dashboard-HTML handler through ProxyTarget()/joinProxyPath stops it from doubling a Grafana subpath (e.g. https://host/grafana/grafana/d/... -> 404). The reverse proxy was already correct; the dashboard handler was not.

* OAuth redirect fix: a relative redirect from Grafana resolves against the proxy-prefixed URL and would drop the /api/cli/v1/proxy prefix. The dashboard handler's CheckRedirect now re-applies the prefix so followed redirects still route through the proxy.

* Mock /api/login/ping: the server now answers the frontend's login-state ping locally with a 200 {"message":"Logged in"}, alongside the other mocked endpoints.

* Tests: adds coverage for ProxyTarget (direct vs OAuth), the dashboard handler URL assembly (subpath and OAuth prefix), the OAuth-redirect re-prefixing, the login-redirect error page, and the login/ping mock -- the proxy/auth path previously had no tests.

Co-Authored-By: Claude Code

### DIFF
--- a/cmd/gcx/resources/serve.go
+++ b/cmd/gcx/resources/serve.go
@@ -102,7 +102,7 @@ support for file notifications.
 				return err
 			}
 
-			cfg, err := configOpts.LoadConfig(cmd.Context())
+			restCfg, currentContext, err := configOpts.LoadGrafanaConfigWithContext(cmd.Context())
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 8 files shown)


dafydd-t:
@claude review

dafydd-t:
@jkeljo 👋 sorry this took so long to get round to! Could you fix the linting errors?

jkeljo:
Fixed! Also rebased to the latest `main`

Issue #1049, opened 2026-07-23
config: env-only auth (GRAFANA_SERVER/GRAFANA_TOKEN, no config file) fails with "$.stacks.default.grafana: node not found"


awsome-o:
Found while writing the #951 regression tests (PR #1047). With no config file anywhere and only env vars set, every ConfigLoader-based provider command fails:

```
HOME=$(mktemp -d) GRAFANA_SERVER=https://example-region.grafana.net GRAFANA_TOKEN=<sa-token> \
  gcx agento11y evaluators list
# → failed to load REST config for Agent Observability evaluators:
#   failed to find path ( $.stacks.default.grafana ): node not found

# same for e.g.
gcx slo definitions list
# → Failed to load REST config for SLO: failed to find path ( $.stacks.default.grafana ): node not found
```

The stack-id discovery request does go out (a GET /bootdata reaches the server), so the failure is after env parsing — it looks like something in the load path (write-back or token-persistence wiring) tries to resolve the $.stacks.default.grafana YAML node for the env-synthesized context, which doesn't exist in the auto-created empty config file.

Env vars overriding an existing config file work fine — this only breaks the no-config-file case, which is the typical CI / ephemeral-environment setup, and it was the documented workaround path in #951. Likely a #1012 regression: the env-only path predates the stacks/contexts split and used to synthesize a working in-memory config.

There's a green regression test for the env-override-with-config-file case in cmd/gcx/root/aio11y_config_routing_test.go; the no-config-file case is called out there as out of scope for #951. Whoever picks this up can extend that harness with a no-config-file case as the red test.

PR #1053, opened 2026-07-24
fix(appo11y): honor --config across the direct appo11y command tree

### LINKED ISSUE #1048: appo11y: --config is ignored by direct services/overrides/settings commands (same defect as #951)
Same defect class as #951, found while fixing it (see PR #1047). The appo11y provider binds --config to a ConfigLoader it never passes to any subcommand:

- internal/providers/appo11y/provider.go creates the loader and calls loader.BindFlags on the appo11y persistent flags, then mounts overrides.Commands(), services.Commands(), settings.Commands() — none of which take it.
- The services command files (commands.go, get.go, map.go, labels.go, operations.go) each construct a zero-value ConfigLoader in their RunE, and the overrides/settings direct commands call a NewTypedCRUD(ctx) that does the same.

So `gcx appo11y ... --config /path/to/config.yaml` parses, is silently ignored, and the command runs against the default config's stack instead. With a valid default config that includes the settings/overrides mutations hitting the wrong stack. Default discovery and --context work (both come through context threading from the root command).

The fix shape is the one PR #1047 applied to aio11y, which copies the SLO provider: NewTypedCRUD takes the loader explicitly, subcommand constructors receive the bound loader, lazy adapter factories pass a local zero-value loader so the resources-tier path keeps its context threading. The routing regression harness from cmd/gcx/root/aio11y_config_routing_test.go can be reused nearly as-is.

Per the v0.6 discussion this is release-significant: it's the same silent wrong-stack mutation class we considered blocking for aio11y. Preferably fixed before the 0.6 test release, definitely before GA.

awsome-o:
Fixes #1048. 

Same defect class as #951: the appo11y provider bound --config to a ConfigLoader it never handed to any subcommand, so every direct verb (services list/get/map/list-operations/list-labels, overrides get/update, settings get/update) parsed the flag and then silently ran against the default config's stack. Reads hit the wrong stack, overrides/settings update could mutate the wrong stack, and datasource resolution plus the discovered-UID save-back followed the wrong config too.

The fix is the shape #1047 applied to aio11y (copied from SLO): NewTypedCRUD takes the loader explicitly, subcommand constructors receive the flag-bound loader from provider.go, and the lazy adapter factories keep passing a zero-value loader so the generic resources path keeps its ctx threading. The services handlers now call dsquery.LoadContextAndConfig instead of inlining its body, which is what the datasource commands already do. One drive-by production fix out of review: settings update was the only appo11y leaf without cobra.NoArgs, so it silently ignored stray positional args on a mutating command — it now rejects them.

Testing: cmd/gcx/root/appo11y_config_routing_test.go mirrors the #1047 harness (same package, reuses its recorder and sandbox helpers rather than duplicating them). Two recording servers with distinct URLs, tokens, stack ids and datasource UIDs; every verb must land on the --config server carrying that config's token while the default server receives zero requests. It also asserts the query body carries the selected config's datasource UID, the stacks-<id> namespace shows up in both the query path and the output envelope, --config works with no default config present (and doesn't silently create one), GCX_CONFIG precedence holds with a positive control, datasource auto-discovery runs against the selected server and the save-back writes the UID into the selected config file only (the default file stays byte-identical), and --context / GRAFANA_* env overrides still behave. All routing cases fail on the unfixed implementation — I ran the suite before applying the fix and every verb hit the default server with its token. Package-level tests prove the lazy factories honor a ctx-threaded config file, via a new shared testutils.AssertFactoryHonorsThreadedConfigFile helper that sibling providers can reuse for the same gap.

This PR is stacked on #1047 (base is that PR's branch) because it reuses its SandboxConfigEnv, the nil-safe ConfigLoader choke point, and the routing-harness helpers. The repo squash-merges, so once #1047 lands I'll rebase --onto main so only the two appo11y commits remain, retarget to main, and mark this ready for review.

Gates: full race suite green, reference regen produces no drift, build ok. mise run lint fails locally with two pre-existing staticcheck SA1019 hits in scripts/ (parser.ParseDir deprecation) that are unrelated to this diff and green in CI on main — looks like a local toolchain artifact. mkdocs isn't installed locally so the docs build is left to CI.

### DIFF
--- a/cmd/gcx/root/aio11y_config_routing_test.go
+++ b/cmd/gcx/root/aio11y_config_routing_test.go
@@ -7,10 +7,16 @@ package root_test
 // only the Agent Observability provider mounted, points two recording HTTP
 // servers A and B at distinct config files, and asserts every plugin-API
 // request lands on the server the user selected — and none on the other.
-// Distinct stack-ids give each config a deterministic namespace
-// (stacks-11111 vs stacks-22222) without any /bootdata discovery round-trip,
-// which also lets the collections-update case assert the output envelope's
-// namespace comes from the selected config.
+// Every fixed verb gets its own routing case because each subcommand
+// constructor captures the loader independently: a verb-local regression
+// cannot be caught by a neighboring verb's case.
+//
+// The two configs also differ in stack-id and token, so two same-shaped
+// mixed-path defect classes stay observable: distinct stack-ids give each
+// config a deterministic namespace (stacks-11111 vs stacks-22222, no
+// /bootdata round-trip) for the collections-update envelope assertion, and
+// distinct tokens let the recorder assert requests carry the selected
+// config's credentials, not the other config's or the environment's.
 
 import (
 	"bytes"
@@ -29,20 +35,30 @@ import (
 	"github.com/grafana/gcx/cmd/gcx/root"
 	"github.com/grafana/gcx/internal/providers"
 	"github.com/grafana/gcx/internal/providers/aio11y"
+	"github.com/grafana/gcx/internal/testutils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-// recorder captures every request that reaches a test server.
-type recorder struct {
-	mu   sync.Mutex
-	hits []string
+type hit struct {
+	method string
+	path   string
+	auth   string
 }
 
-func (r *recorder) record(method, path string) {
+func (h hit) String() string { return h.method + " " + h.path + " [" + h.auth + "]" }
+
+// recorder captures every request that reaches a test server, including the
+// Authorization header so credential routing is as observable as URL routing.
+type recorder struct {
+	mu   sync.Mutex
+	hits []hit
+}
+
+func (r *recorder) record(req *http.Request) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.hits = append(r.hits, method+" "+path)
+	r.hits = append(r.hits, hit{method: req.Method, path: req.URL.Path, auth: req.Header.Get("Authorization")})
 }
 
 func (r *recorder) count() int {
@@ -51,10 +67,10 @@ func (r *recorder) count() int {
 	return len(r.hits)
 }
 
-func (r *recorder) all() []string {
+func (r *recorder) all() []hit {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	return append([]string(nil), r.hits...)
+	return slices.Clone(r.hits)
 }
 
 func (r *recorder) reset() {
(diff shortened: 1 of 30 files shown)


awsome-o:
## Independent review and E2E validation

Validated at head `441414f9`.

The review scope was checked against #1048 and the related defect report #951. The reported #1048 failure is the silent wrong-stack class: the App O11y parent binds `--config`, but the direct services/overrides/settings commands previously discarded that flag-bound loader and loaded the default configuration instead. That made both reads and settings/overrides mutations capable of targeting the wrong stack. The separate env-only/no-config-file defect in #1049 is not part of this PR.

**Result: the pre-fix code reproduces #1048, while PR head passes the same routing cases. Code review and validation produced no blocking findings.**

Evidence:

- Isolated binaries were built for the stacked base and this PR head.
- With `GCX_CONFIG` pointed at an unusable decoy and a valid file passed through `--config`, the pre-fix binary selected the decoy. The PR binary selected and authenticated to the explicit configuration.
- The regression suite covers all nine affected direct leaves: services list/get/map/list-operations/list-labels, overrides get/update, and settings get/update.
- Against the pre-fix base, all 9/9 routing cases failed by reaching the default/decoy server. At this head they pass. The tests verify the exact method/path, selected credential, zero requests to the default server, namespace, datasource UID, discovery routing, and datasource save-back isolation.
- The App O11y routing suite passed 10 repeated race-enabled runs plus shuffled race runs.
- `go test -race -count=1 ./...`, lint, isolated builds, formatting/diff checks, reference generation, skill validation, and all 44 linter-rule tests passed. GitHub Tests, Linters, Documentation, Semgrep, Socket, and zizmor are green at this head.

Live Grafana Cloud read-only E2E used the same hostile-`GCX_CONFIG`/explicit-`--config` control:

- direct `settings get` succeeded with the selected stack namespace;
- generic `resources get settings/default` succeeded, exercising the lazy adapter path;
- combined `--config --context` succeeded;
- `services list` reached the selected Prometheus datasource successfully, and get/map/list-operations/list-labels reached the live datasource and returned the expected no-data result on this test stack;
- `overrides get` reached the selected stack, where App Observability is currently not activated.

Live settings/overrides mutations were not performed: these commands have no dry-run, the feature is not activated on the available stack, and replaying a partial settings payload could alter real configuration. Both update POST routes are exercised end-to-end through the real root command tree against two recording servers, including credentials and wrong-stack isolation.

The current head passes the listed code review, automated validation, and live read-only checks. The PR remains draft/stacked on #1047. After #1047 merges, rebase only `56e9272e` and `441414f9` onto fresh `main`, verify the intended 15-file diff, rerun the gates/live read-only E2E, then mark ready and request reviewers.

annanay25 (dismissed):
lgtm!

annanay25 (approved):

PR #1058, opened 2026-07-24
fix(fleet): preserve configType so OTel pipelines round-trip

### LINKED ISSUE #1057: Fleet pipeline serialization drops `configType`, making OTel pipelines unmanageable
## Summary

`internal/providers/fleet.Pipeline` has no `ConfigType` field, so the Fleet Management API's `configType` is silently dropped on every serialization path. Any `gcx` write to an OTel pipeline sends no `configType`, the API falls back to Alloy River validation, and the request is rejected:

```
fleet: update pipeline <id>: status 400: {"code":"invalid_argument",
"message":"invalid river contents provided: 1:10: illegal character U+003A ':' (and 218 more diagnostics)"}
```

The `U+003A ':'` is the first colon of the OTel YAML being parsed as River.

## Impact

OTel Fleet pipelines (`configType: CONFIG_TYPE_OTEL`) cannot be managed by `gcx` at all:

- `gcx fleet pipelines update -f <manifest>` fails, even when the manifest declares `spec.configType: CONFIG_TYPE_OTEL`.
- `gcx fleet pipelines create -f <manifest>` cannot create an OTel pipeline.
- `gcx fleet pipelines get`/`list` and `gcx resources pull` render manifests with no `configType`, so a pulled manifest is not round-trippable — pushing it back fails.
- `gcx resources push` via the pipeline `TypedCRUD` has the same defect.

Alloy pipelines are unaffected in practice because their `configType` is the API's default.

The field was always present in API *responses* — every pipeline in a live tenant returns one — so this is purely a client-side modelling gap.

## Root cause

`internal/providers/fleet/types.go` — the `Pipeline` struct:

```go
type Pipeline struct {
	ID       string         `json:"id,omitempty"`
	Name     string         `json:"name"`
	Enabled  *bool          `json:"enabled,omitempty"`
	Contents string         `json:"contents"`
	Matchers []string       `json:"matchers,omitempty"`
	Metadata map[string]any `json:"metadata,omitempty"`
}
```

All four serialization paths marshal/unmarshal this one struct, so a single missing field breaks all of them:

1. API response → `Pipeline` (`client.go` `GetPipeline`/`ListPipelines`) — `encoding/json` discards the unknown `configType` key.
2. `Pipeline` → K8s-style resource envelope (`provider.go` `PipelineToResource`) — marshals the struct into `spec`, so `spec.configType` is never emitted.
3. Manifest → `Pipeline` (`provider.go` `PipelineFromResource`, `readPipelineFromFile`) — a manifest that *does* declare `spec.configType` has it dropped on decode.
4. `Pipeline` → API request (`client.go` `CreatePipeline`/`UpdatePipeline`) — the outgoing `pipeline` object carries no `configType`.

## Reproduction

```bash
# against a tenant with an existing CONFIG_TYPE_OTEL pipeline
gcx fleet pipelines get <my-otel-pipeline> -o json > pipeline.json
# pipeline.json has no spec.configType (defect 2)

# even after hand-adding spec.configType: CONFIG_TYPE_OTEL to the manifest:
gcx fleet pipelines update <my-otel-pipeline> -f pipeline.json
# -> 400 invalid river contents (defects 3 + 4)
```

## Expected

`configType` survives all four paths. `omitempty` keeps Alloy pipelines and existing `configType`-less manifests byte-compatible — no empty `configType` is sent or rendered.

## Acceptance criteria

- [ ] `Pipeline` carries `ConfigType string \`json:"configType,omitempty"\``
- [ ] API response decode preserves `configType`
- [ ] `PipelineToResource` renders `spec.configType`
- [ ] `PipelineFromResource` / `readPipelineFromFile` decode `spec.configType`
- [ ] `CreatePipeline` / `UpdatePipeline` send `configType`
- [ ] Alloy backward compatibility: empty `configType` is omitted, not sent as `""`
- [ ] Pipeline JSON schema documents `configType` with its enum
- [ ] Regression tests covering each path

rknightion:
Fixes #1057.

## Problem

`internal/providers/fleet.Pipeline` had no `ConfigType` field, so the Fleet Management API's `configType` was silently dropped everywhere. Writing to an OTel pipeline sent no `configType`, the API fell back to Alloy River validation, and rejected the request:

```
status 400: {"code":"invalid_argument","message":"invalid river contents provided:
1:10: illegal character U+003A ':' (and 218 more diagnostics)"}
```

The `U+003A ':'` is the first colon of the OTel YAML being parsed as River. OTel Fleet pipelines were therefore unmanageable via `gcx` — `create`, `update`, and `resources push`/`pull` through the pipeline `TypedCRUD` all failed or produced non-round-trippable manifests.

## Fix

One field, because all four serialization paths marshal the same struct:

```go
ConfigType string `json:"configType,omitempty"`
```

| Path | Where | Now |
|---|---|---|
| API response → `Pipeline` | `client.go` `GetPipeline`/`ListPipelines` | `configType` decoded |
| `Pipeline` → resource envelope | `provider.go` `PipelineToResource` | `spec.configType` rendered |
| manifest → `Pipeline` | `provider.go` `PipelineFromResource`, `readPipelineFromFile` | `spec.configType` decoded |
| `Pipeline` → API request | `client.go` `CreatePipeline`/`UpdatePipeline` | `configType` sent |

`omitempty` preserves Alloy behaviour exactly: no empty `configType` is rendered or sent, so existing `configType`-less manifests stay byte-compatible.

Two adjacent changes, both driven by the same defect:

- `pipelines list -o wide` gains a `CONFIG TYPE` column. OTel and Alloy pipelines were previously indistinguishable in gcx output, which is partly why this went unnoticed. Default (non-wide) output is unchanged.
- The pipeline JSON schema documents `configType` with its `CONFIG_TYPE_ALLOY`/`CONFIG_TYPE_OTEL` enum, and the example declares it.

## Tests

TDD — tests written first, confirmed failing against the old struct (compile error: the missing field *is* the bug), then the fix.

New regression tests, one per path plus backward-compat guards:

- `TestClient_GetPipeline_PreservesConfigType` — API decode
- `TestClient_UpdatePipeline_SendsConfigType` — outgoing request
- `TestClient_UpdatePipeline_OmitsEmptyConfigType` — Alloy: no empty field on the wire
- `TestPipelineToResource_PreservesConfigType` — pull/render + round-trip
- `TestPipelineFromResource_ReadsConfigType` — manifest read/push
- `TestPipelineToResource_OmitsEmptyConfigType` — Alloy: no empty `spec.configType`
- `TestPipelineTableCodec_WideShowsConfigType` — wide column, `-` when empty

## Verification

```
go test -race -count=1 ./internal/providers/fleet/...   ok
golangci-lint run -c .golangci.yaml                     0 issues (whole repo)
go build -buildvcs=false -o bin/gcx ./cmd/gcx/          ok
go run ./scripts/validate-skills                        ok
dev lint test ./internal/linter/bundle/gcx/             all pass
CLI reference drift                                     none
```

Five failures in the full `go test ./...` run are pre-existing and environmental — confirmed reproducing on a clean `main` checkout with these changes stashed (four TTY/confirm-prompt tests, plus a login test that reads an ambient `GRAFANA_CLOUD_TOKEN` from the shell).

### Live tenant

Validated end-to-end against a real OTel pipeline, with a backup taken first:

1. **Pre-fix binary** (built from clean `main`), no-op update with the live manifest → reproduces the 400. Pipeline verified byte-identical afterwards, so the failure is a safe rejection.
2. **Fixed binary**, same manifest → succeeds, `configType` intact.
3. **Write-through proof**: a semantic edit (one scrape `collection_interval`) landed and read back correctly. A comment-only edit is *not* a valid probe here — Fleet canonicalizes OTel YAML server-side, sorting keys and stripping comments.
4. **Restored** to the exact original; `get -o json` byte-identical to the pre-validation backup. Collector confirmed still enabled.

Also confirmed live that the API returns `configType` for *every* pipeline, Alloy included — so this was purely a client-side modelling gap, not a recent API addition.


wardbekker:
I pushed one commit to this branch — `280812d3`, `test(fleet): close two gaps in the configType regression tests`. It touches tests only. Your fix in `a904ecf5` is unchanged. Sorry for moving the branch under you; here is what changed and why.

The fix itself is right, and I verified it end to end: `PipelineToResource`, `PipelineFromResource`, and the client methods all marshal the whole struct, so the single `ConfigType` field really does cover all four paths. Two things in the test suite needed work.

### 1. The wide-table assertion could not fail

`TestPipelineTableCodec_WideShowsConfigType` ended with:

```go
assert.Contains(t, lines[2], "-", "empty configType should render as -")
```

The Alloy row renders as `p-2  alloy-one  true  -  -`. The ID `p-2` and the name `alloy-one` already contain a hyphen, so the check passes whatever the codec puts in the config type column. I proved it by changing the fallback in `provider.go` from `"-"` to `"MUTANT"` — the test still passed.

The assertion now indexes the column, the fixtures carry no hyphen (`p1`/`otelone`, `p2`/`alloyone`), and both rows declare a matcher. A correct codec therefore emits exactly one `-`: the empty config type of the Alloy row. The same `"MUTANT"` mutation now fails the test.

Note that the older matchers assertion at the end of `TestPipelineTableCodec_Encode` has the same defect. It predates this PR, so I left it alone.

### 2. Two client paths had no direct test

The suite covered `GetPipeline` decode and `UpdatePipeline` encode. It did not cover:

- `TestClient_CreatePipeline_SendsConfigType` — `configType` on the create request, and on the create response decode.
- `TestClient_ListPipelines_PreservesConfigType` — `configType` on the list decode, which is the path that feeds your new wide column.

### Checks

```
gofmt -l internal/providers/fleet/                      clean
go test -race -count=1 ./internal/providers/fleet/...   ok
golangci-lint run -c .golangci.yaml ./internal/...      0 issues
```

### Two notes, no action needed

- The PR body says existing `configType`-less manifests stay byte-compatible. That is true for manifests gcx **reads**. It is not true for manifests gcx **writes**: since the API returns `configType` for every pipeline, a re-pull now adds `spec.configType: CONFIG_TYPE_ALLOY` to every stored Alloy manifest. GitOps users see a one-time diff. This is correct behaviour, but it is worth one line in the description.
- The schema `enum` pins the value to two options. If Fleet ever adds a third config type, the schema goes stale. The descriptor schema feeds `resources list-schemas` and field-path completion rather than push-time validation, so the risk is small.

Nice find on the root cause, and thanks for the live-tenant verification with the restore step.

wardbekker (approved):

awsome-o (approved):

PR #1059, opened 2026-07-24
feat(datasources): add support for MSSQL datasource

ktw4071:
## Summary

Adds a Microsoft SQL Server (MSSQL) DatasourceProvider with three subcommands and auto-detect integration, enabling agents and developers to query MSSQL datasources through Grafana's query API. 

- `gcx datasources mssql query` — execute arbitrary T-SQL with `TOP (n)` injection, `$__timeFilter`/`$__timeGroup` macro support, and time-range binding
- `gcx datasources mssql list-tables` — list tables and views via `INFORMATION_SCHEMA.TABLES` (filterable with `--schema`)
- `gcx datasources mssql describe-table` — inspect column schema via `INFORMATION_SCHEMA.COLUMNS` (`--schema` to disambiguate)
- Auto-detect via `gcx datasources query <uid> <sql>` routes MSSQL datasources automatically

## Architecture

- **HTTP client** (`internal/query/mssql/client.go`): reuses the shared `grafanaquery` transport (K8s query API `/apis/query.grafana.app/.../query` with `/api/ds/query` fallback) and `internal/query/sql` for response parsing — same pattern as ClickHouse/Athena. No AWS/DB auth in gcx; Grafana's datasource proxy handles the SQL Server connection.
- **Provider registration** (`internal/datasources/providers/mssql.go`): standard DatasourceProvider interface with `init()` self-registration. Plugin ID `mssql` equals the short kind, so no `NormalizeKind` mapping is required (added a test-table entry for documentation).
- **Schema discovery via SQL**, not resource endpoints: `list-tables`/`describe-table` query `INFORMATION_SCHEMA` (mirrors ClickHouse's `system.tables` approach). Identifier validation via regex allowlist + single-quote escaping guards the interpolated `--schema`/table values.
- **Output**: `query`, `list-tables`, and `describe-table` all return the shared `querysql.QueryResponse`, already rendered by the existing datasource codecs (text/wide/json/yaml) — no new codec needed.

### 1. `format: "table"` (string), not the integer used by sqlds plugins

ClickHouse/Athena (built on `grafana/sqlds`) send `format: 1`. The **core MSSQL plugin returns HTTP 500** on an integer format code — it requires the string `format: "table"`. Confirmed by probing the live query API. Pinned by `TestQuery_SendsStringTableFormat` and `TestGenericQueryRoutesMSSQL`.

### 2. `TOP (n)` injection instead of `LIMIT`

T-SQL has no `LIMIT` keyword. `--limit` (default 100, max 1000; `--limit 0` disables) injects `TOP (n)` after a leading `SELECT`/`SELECT DISTINCT`. It **bails safely** (leaves SQL unchanged) for statements where injecting `TOP` would be invalid or change semantics: non-SELECT statements, CTEs (`WITH`), queries that already use `TOP`, set operations (`UNION`/`INTERSECT`/`EXCEPT`), and paged queries (`OFFSET … FETCH`). Covered by `TestEnforceTop`.

## Sample output

image

$ gcx datasources mssql query -d <mssql-uid> 

'SELECT BASE_COUNTRY, GDP, CO2 FROM dbo.WORLD_DATA ORDER BY GDP DESC' --limit 3
BASE_COUNTRY                                    GDP      CO2
{"Name":"Guadeloupe","Capital":"Basse-Terre"}  4999932  934
{"Name":"American Samoa","Capital":"Fagatogo"}  4999932  857
{"Name":"Guam","Capital":"Aga"}                 4999812  228

## Test plan

- [x] Unit tests for `EnforceTop`, identifier validation/escaping, client response parsing, request-body shape (`format:"table"`), and the Explore-link builder
- [x] Verified live against a real MSSQL datasource on a dev instance: `query` (incl. `TOP` injection, `WHERE`, `GROUP BY`/`HAVING`, `JSON_VALUE`, `-o json`, `--share-link`), `list-tables`, `describe-table`, the auto-detect path, and typed error handling
- [x] `GCX_AGENT_MODE=false mise run all` passes (lint 0 issues + `-race` tests + build + docs); reference docs regenerated

Part of the datasource-coverage effort (https://github.com/grafana/plugins-private/issues/4127).


dafydd-t:

dafydd-t:
thanks for this. two minor things worth a look:

- `EnforceTop` leaves an existing `TOP 999999` untouched, while the shared `EnforceLimit` clamps an existing `LIMIT` above the max down to 1000 - so the sql datasources now behave differently on user-supplied caps. bailing is probably the right call here (rewriting `TOP n PERCENT`/`WITH TIES` is risky), just flagging the divergence in case you want a comment noting it
- `describe-table dbo.WORLD_DATA` fails with "invalid table: must contain only letters, numbers, and underscores" - schema-qualified names are the natural thing to type, so the error should probably hint at `--schema` (or split on the dot)

neither is blocking 👍

cmd/gcx/datasources/query.go
```diff
@@ -215,3 +101,155 @@ that do not have a dedicated subcommand.`,
...
+	cmd *cobra.Command,
+	shared *dsquery.SharedOpts,
+	cfg config.NamespacedRESTConfig,
+	dsType, datasourceUID, expr string,
+	start, end time.Time,
+	step time.Duration,
+	limit int,
+	profileType string,
+	maxNodes int64,
+) error {
```
dafydd-t: I think this will conflict with some of @alyssajoyner 's diffs, which also extract these due to complexity rules. Just something to be aware of.

ktw4071:
@dafydd-t 
Addressed your comments, requesting re-review after changes

cmd/gcx/datasources/query.go
```diff
@@ -215,3 +213,44 @@ that do not have a dedicated subcommand.`,
...
+	}
+
+	req := clickhouse.QueryRequest{
+		RawSQL: clickhouse.EnforceLimit(expr, 100, 1000),
+		Start:  start,
+		End:    end,
+	}
+	if step > 0 {
+		req.IntervalMs = step.Milliseconds()
+	}
```
dafydd-t, line 231: Do we want something like this in the MSSQL client too? we're currently ignoring `--step` in the MSSQL queries

internal/query/mssql/types_test.go
```diff
@@ -0,0 +1,83 @@
...
+		name       string
+		input      string
+		wantSchema string
+		wantTable  string
+		wantErr    bool
+	}{
+		{"bare table", "WORLD_DATA", "", "WORLD_DATA", false},
+		{"schema qualified", "dbo.WORLD_DATA", "dbo", "WORLD_DATA", false},
+		{"three parts errors", "db.dbo.WORLD_DATA", "", "", true},
+		{"trailing dot yields empty table part", "dbo.", "dbo", "", false},
```
dafydd-t: should we reject on this instead? your call. It seems like a waste to end up calling e.g. describe table on an empty table name, due to a typo (e.g. `describe-table dbo.`)

docs/architecture/project-structure.md
```diff
@@ -99,7 +100,8 @@ gcx/
 │   │   ├── influxdb/         # InfluxDB HTTP query client
 │   │   ├── infinity/         # Infinity HTTP query client
 │   │   ├── loki/             # Loki HTTP client (log + metric queries)
-│   │   └── clickhouse/       # ClickHouse HTTP client
+│   │   ├── clickhouse/       # ClickHouse HTTP client
+│   │   └── mssql/            # Microsoft SQL Server HTTP client (format:"table" string, TOP/EnforceTop, INFORMATION_SCHEMA helpers)
```
dafydd-t:
```suggestion
│   │   └── mssql/            # Microsoft SQL Server HTTP client
```

docs/reference/cli/gcx_datasources_mssql_describe-table.md
```diff
@@ -0,0 +1,57 @@
+## gcx datasources mssql describe-table
+
+List columns for an MSSQL table
```
dafydd-t: Should this description be more related to the command name?


```suggestion
## gcx datasources mssql describe-table

Describe a MSSQL table
```

dafydd-t: something similar for the long description below

docs/reference/cli/gcx_datasources_mssql_query.md
```diff
@@ -0,0 +1,76 @@
+## gcx datasources mssql query
+
+Execute a SQL query against an MSSQL datasource
+
+### Synopsis
+
+Execute a T-SQL query against a Microsoft SQL Server datasource.
```
dafydd-t: Is T-SQL important here? is it the default for mssql?

internal/query/mssql/types.go
```diff
@@ -0,0 +1,108 @@
...
+// it. Rewriting TOP is not a safe find-and-replace — it has `TOP (n) PERCENT`
+// and `TOP (n) WITH TIES` variants whose semantics a naive clamp would corrupt —
+// so a user-supplied TOP is always honoured as-is.
+func EnforceTop(sql string, limit, maxLimit int) string {
+	if limit == 0 {
+		return sql
+	}
+	if limit > maxLimit {
+		limit = maxLimit
+	}
```
dafydd-t, line 114: I'd like to make sure we're advertising to the user when we've capped their results.  We should show a warning when  they have more than 1000 results, but we are capping them.


One way to do this: fetch `maxLimit + 1` results, and if we get 1001 rows back, warn the user(e.g. "only showing the first 1000 results"). There may be other ways too

internal/query/mssql/types.go
```diff
@@ -0,0 +1,108 @@
...
+// Note the deliberate divergence from the shared EnforceLimit: when the SQL
+// already carries a cap, EnforceLimit rewrites an over-max `LIMIT n` down to
+// maxLimit, but EnforceTop leaves an existing TOP untouched rather than clamping
+// it. Rewriting TOP is not a safe find-and-replace — it has `TOP (n) PERCENT`
+// and `TOP (n) WITH TIES` variants whose semantics a naive clamp would corrupt —
+// so a user-supplied TOP is always honoured as-is.
+func EnforceTop(sql string, limit, maxLimit int) string {
+	if limit == 0 {
+		return sql
+	}
```
dafydd-t: should we also check in case limit is <0 by mistake?

cmd/gcx/datasources/query.go
```diff
@@ -215,3 +213,44 @@ that do not have a dedicated subcommand.`,
...
+
+	resp, err := client.Query(ctx, datasourceUID, req)
+	if err != nil {
+		return nil, fmt.Errorf("query failed: %w", err)
+	}
+	return resp, nil
+}
+
+// runMSSQLQuery executes the auto-detected mssql branch of QueryCmd.
+func runMSSQLQuery(ctx context.Context, cfg config.NamespacedRESTConfig, datasourceUID, expr string, start, end time.Time) (*querysql.QueryResponse, error) {
```
dafydd-t: for `gcx datasources query` with `--limit`, limit is ignored. could we respect that here? I think the same issue exists in clickhouse but maybe we can address that separately

dafydd-t: hm, maybe we can do this one separately actually, and solve it for all providers in one PR. Lets leave it as-is in this PR

internal/query/mssql/types.go
ktw4071: I've also used Claude to apply this over the cap use results. Let me know if it's done in a right way

internal/query/mssql/types.go
ktw4071: updated

docs/reference/cli/gcx_datasources_mssql_query.md
ktw4071: updated

cmd/gcx/datasources/query.go
ktw4071: I've used Claud to apply this change to ignore `--step`. Let me know if it's done in a right way!

ktw4071:
```
➜  gcx git:(ktw4071/mssql) ✗ GCX_AGENT_MODE=false bin/gcx datasources mssql query -d mssql-ds-m "SELECT 1" --step 1m

Error: Unknown flag
│
│ --step
│
└─

```

ktw4071: UPDATE: I've added `--step` flag to be usable in MSSQL.

```
gcx git:(ktw4071/mssql) ✗ TS='SELECT $__timeGroup(DATE_TIME, $__interval) AS t, COUNT(*) AS n FROM dbo.WORLD_DATA GROUP BY $__timeGroup(DATE_TIME, $__interval) ORDER BY t'
GCX_AGENT_MODE=false bin/gcx datasources mssql query -d mssql-ds-m "$TS" --since 24h --step 1h -o table

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ T                                                                                                       │ N                                                                                                       │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1785384000                                                                                              │ 81                                                                                                      │
│ 1785387600                                                                                              │ 720                                                                                                     │
│ 1785391200                                                                                              │ 719                                                                                                     │
│ 1785394800                                                                                              │ 720                                                                                                     │
│ 1785398400                                                                                              │ 720                                                                                                     │
│ 1785402000                                                                                              │ 720                                                                                                     │
│ 1785405600                                                                                              │ 720                                                                                                     │
│ 1785409200                                                                                              │ 720                                                                                                     │
│ 1785412800                                                                                              │ 720                                                                                                     │
│ 1785416400                                                                                              │ 720                                                                                                     │
│ 1785420000                                                                                              │ 720                                                                                                     │
│ 1785423600                                                                                              │ 720                                                                                                     │
│ 1785427200                                                                                              │ 720                                                                                                     │
│ 1785430800                                                                                              │ 720                                                                                                     │
│ 1785434400                                                                                              │ 720                                                                                                     │
│ 1785438000                                                                                              │ 720                                                                                                     │
│ 1785441600                                                                                              │ 720                                                                                                     │
│ 1785445200                                                                                              │ 720                                                                                                     │
│ 1785448800                                                                                              │ 720                                                                                                     │
│ 1785452400                                                                                              │ 718                                                                                                     │
│ 1785456000                                                                                              │ 720                                                                                                     │
│ 1785459600                                                                                              │ 720                                                                                                     │
│ 1785463200                                                                                              │ 143                                                                                                     │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

internal/query/mssql/types.go
ktw4071: Done — we now warn the user when we cap their results, for MSSQL, ClickHouse, and Athena.

I used your n+1 idea: we fetch one row past the limit, and if that extra row comes back, we know there were more rows than we're showing — so we drop it and print a warning. It keys off whatever --limit is in effect (default 100, max 1000), not a fixed 1000.

What I added:

A shared helper EnforceLimitSentinel in internal/query/sql that injects LIMIT eff+1 and reports whether we added the cap. ClickHouse and Athena call it directly; MSSQL has the TOP equivalent, EnforceTopSentinel.
A shared SurfaceRowLimits helper that drops the extra row, prints the warning when we hid results, and also passes through the datasource's own "results limited" notices. (ClickHouse/Athena now show those notices too — before, only MSSQL did.)
A few things it gets right:

It only warns when we did the capping. If the user wrote their own LIMIT/TOP, or the query can't be safely capped, we leave the results untouched and stay quiet.
It still works at the max (--limit 1000 fetches 1001 to detect more).
The Explore share link shows the real limit, not the +1.
Tested: unit tests for all the cases, plus checked live — a small --limit now returns N rows and prints showing the first N rows; more rows match — raise --limit …, while a query with fewer rows (or the user's own limit) gives no false warning.

Let me know if this is done right way :)

ktw4071:
Rebased onto current `main` to pick up the postgres/mysql/elasticsearch/cloudmonitoring datasource work merged since this branch was opened. Notable reconciliation: `internal/query/sql.EnforceLimit` (used by mysql/postgres, already reviewer-hardened this cycle) is left untouched; the new sentinel-based truncation detection (`EnforceLimitSentinel`, `Truncate`, `TruncationHint`) added here lives alongside it as an independent, additive mechanism rather than a shared low-level helper, so mysql/postgres's shipped behavior can't drift as a side effect of this PR.

Re-verified live against `mssql-ds-m` post-rebase: the `--limit`/sentinel warning still fires correctly (`--limit 3` → 3 rows + "showing the first 3 rows; more rows match" hint), and the generic `gcx datasources query` auto-detect path picks up mssql's dispatch route (including the `--step` → `intervalMs` threading) through the query-routes refactor that landed on main since.

Both previously-unresolved threads here (`--step` support, the n+1 truncation-warning sentinel) already have their fixes and live verification posted — carried through the rebase intact. Re-review requested to clear them.

ktw4071:
Went through the four sibling datasource PRs (postgres #969, mysql #970, elasticsearch #971, cloudmonitoring #974) and the new `add-datasource` raw-SQL guidance in #1227, checking which review findings from those apply here too. Found and fixed three:

- **negative `--limit` rejected** (`69fe41f6`) — mssql's `EnforceTop`/`EnforceTopSentinel` treat any `limit <= 0` as "disable enforcement", so `--limit -5` silently behaved like `--limit 0` instead of erroring. Same fix as postgres's `09d037dd`/mysql's equivalent.
- **`--schema ""` rejected on both `list-tables` and `describe-table`** (`eae20f01`) — same class as postgres's `93ddf1ae`: an explicit empty value read as omitted and silently widened scope from one schema to every schema (the classic unset-`$SCHEMA`-variable case). Fixed with `cmd.Flags().Changed("schema")` on both commands.

Checked and deliberately left alone:
- **CTE queries never get a `TOP` cap** — `EnforceTop`'s `leadingSelectRe` only matches statements starting with `SELECT`, so `WITH x AS (...) SELECT ...` bails entirely (no cap at all), unlike postgres/mysql which include `WITH` in their allow-list. This is already disclosed in the command's own `Long` text ("CTEs (WITH)... are left unchanged") and fails safe (no cap ≠ corrupted query), so I didn't treat it as a defect — flagging here in case you'd rather it be explicitly capped for consistency with the other dialects.
- **`list-tables` has no row cap at all** (unlike postgres/mysql's capped-with-warning `list-tables`) — since it's genuinely uncapped rather than silently truncated, there's no honesty gap to fix; adding a defensive cap would be a consistency choice, not a bug fix, so I left it as-is pending your call.

All three fixes live-verified against `mssql-ds-m`, full gate green.

ktw4071:
@awsome-o — pulled these two fixes directly from your own review comments on the sibling PRs, so flagging the connection explicitly:

**`69fe41f6` — negative `--limit` rejected.** Same fix postgres/mysql already got, from your comments on both:
- postgres #969: https://github.com/grafana/gcx/pull/969#discussion_r3724119159 — "A negative value isn't checked anywhere, so `--limit -5` emits `LIMIT -5`..."
- mysql #970: https://github.com/grafana/gcx/pull/970#discussion_r3724121626 — "Same limit contract as the Postgres sibling..."

mssql's `EnforceTop`/`EnforceTopSentinel` treated any `limit <= 0` as "disable enforcement," so `--limit -5` silently behaved identically to `--limit 0` — no error, no explanation, just a lower effective cap than intended.

**`eae20f01` — `--schema ""` rejected on both `list-tables` and `describe-table`.** Exact same bug class from:
- postgres #969: https://github.com/grafana/gcx/pull/969#discussion_r3724119164 — "`--schema ""` is indistinguishable from omission here, so it silently returns every schema..."
- mysql #970: https://github.com/grafana/gcx/pull/970#discussion_r3724121642 — "`--database ""` reads as omitted and broadens scope to every database..."

Same fix here: `cmd.Flags().Changed("schema")` catches the explicit-empty case that `value != ""` can't.

Both live-verified against `mssql-ds-m` (before/after outputs in the commit messages), full gate green.

ktw4071:
Full live validation against `mssql-ds-m`, covering the rebase + all fixes from this round.

## Query surface sweep (9 commands, unaffected by the rebase/fixes)

**1. Default query (no `--limit`)**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT * FROM [grafanadb].dbo.WORLD_DATA" -o json
```
stderr: `Results have been limited to 100 because the SQL row limit was reached`
stdout: columns `[BASE_COUNTRY, BIRTH_RATE, CO2, GDP, DATE_TIME, TIMESTAMP_VALUE]`, **100 rows**, first row `["Sao Tome and Principe/São Tomé", 4, 962, 3724416, 1787443203000, 1787443203482]`

**2. Small `--limit 5` (gcx's own sentinel hint)**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT * FROM [grafanadb].dbo.WORLD_DATA" --limit 5 -o json
```
stderr: `showing the first 5 rows; more rows match — raise --limit (max 1000), pass --limit 0 to remove the cap, or add your own row limit to the query`
stdout (full):
```json
{
  "columns": [ {"name":"BASE_COUNTRY","type":"string"}, {"name":"BIRTH_RATE","type":"number"}, {"name":"CO2","type":"number"}, {"name":"GDP","type":"number"}, {"name":"DATE_TIME","type":"time"}, {"name":"TIMESTAMP_VALUE","type":"number"} ],
  "rows": [
    ["{\"Name\":\"Sao Tome and Principe\",\"Capital\":\"São Tomé\"}",4,962,3724416,1787443203000,1787443203482],
    ["{\"Name\":\"Croatia\",\"Capital\":\"Zagreb\"}",8,679,4047721,1787443208000,1787443208482],
    ["{\"Name\":\"Sao Tome and Principe\",\"Capital\":\"São Tomé\"}",8,934,1022906,1787443213000,1787443213483],
    ["{\"Name\":\"Australia\",\"Capital\":\"Canberra\"}",6,727,3002792,1787443218000,1787443218484],
    ["{\"Name\":\"Bahrain\",\"Capital\":\"al-Manama\"}",3,869,1397140,1787443223000,1787443223484]
  ]
}
```

**3. User's own `TOP(7)` — must not trigger a false warning**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT TOP(7) * FROM [grafanadb].dbo.WORLD_DATA" -o json
```
stderr: *(no truncation warning — only the generic `--json`/`--jq` usage hint)*
stdout: **7 rows**

**4. `--limit 0` (gcx cap disabled)**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT * FROM [grafanadb].dbo.WORLD_DATA" --limit 0 -o json
```
stderr: `Results have been limited to 100 because the SQL row limit was reached` *(datasource's own notice — unaffected by gcx's cap being off)*
stdout: **100 rows**

**5. `--step 1h` time-series bucketing**
```
$ TS='SELECT $__timeGroup(DATE_TIME, $__interval) AS t, COUNT(*) AS n FROM [grafanadb].dbo.WORLD_DATA GROUP BY $__timeGroup(DATE_TIME, $__interval) ORDER BY t'
$ gcx datasources mssql query -d mssql-ds-m "$TS" --since 24h --step 1h -o table
```
stdout (42 buckets, first/last shown):
```
T           N
1787443200  720
1787446800  720
...
1787590800  720
1787594400  632
```

**6. `list-tables`**
```
$ gcx datasources mssql list-tables -d mssql-ds-m -o table
```
```
TABLE_SCHEMA  TABLE_NAME               TABLE_TYPE
dbo           WORLD_DATA               BASE TABLE
sys           database_firewall_rules  VIEW
```

**7. `describe-table` — bare name vs schema-qualified (identical output, both forms work)**
```
$ gcx datasources mssql describe-table WORLD_DATA -d mssql-ds-m -o table
$ gcx datasources mssql describe-table dbo.WORLD_DATA -d mssql-ds-m -o table
```
```
COLUMN_NAME      DATA_TYPE  IS_NULLABLE  CHARACTER_MAXIMUM_LENGTH  COLUMN_DEFAULT
BASE_COUNTRY     nvarchar   YES          -1                        -
BIRTH_RATE       int        YES          -                         -
CO2              int        YES          -                         -
GDP              int        YES          -                         -
DATE_TIME        datetime2  YES          -                         -
TIMESTAMP_VALUE  bigint     YES          -                         -
```

**8. Generic auto-detect path (`c0251824` fix)**
```
$ gcx datasources query mssql-ds-m "SELECT * FROM [grafanadb].dbo.WORLD_DATA" -o json
```
stderr: `Results have been limited to 100 because the SQL row limit was reached`
stdout: **100 rows** — now matches #1's behavior; before `c0251824` this printed no warning at all.

**9. `--share-link`, matching a real Explore URL exactly**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT TOP(50) * FROM [grafanadb].dbo.WORLD_DATA" --since 1h --share-link
```
stdout: **50 rows**, first row `["Sao Tome and Principe/São Tomé", 4, 962, 3724416, 1787443203000, 1787443203482]`
Explore link (decoded `rawSql`): `SELECT TOP(50) * FROM [grafanadb].dbo.WORLD_DATA` — identical to a live Explore session against the same table.

## Before/after for the two new fixes

**`69fe41f6` — negative `--limit` rejected**
```
$ gcx datasources mssql query -d mssql-ds-m "SELECT * FROM [grafanadb].dbo.WORLD_DATA" --limit -5 -o json
before: 100 rows returned, no error (identical to --limit 0)
after:  {"type":"gcx.error","schema_version":"1","error":{"summary":"--limit must be >= 0, got -5","exitCode":1}}
```

**`eae20f01` — `--schema ""` rejected on `list-tables` and `describe-table`**
```
$ gcx datasources mssql list-tables -d mssql-ds-m --schema "" -o table
before: TABLE_SCHEMA  TABLE_NAME               TABLE_TYPE
        dbo           WORLD_DATA               BASE TABLE
        sys           database_firewall_rules  VIEW      (identical to --schema omitted — silently widened)
after:  {"type":"gcx.error","schema_version":"1","error":{"summary":"--schema must not be empty","exitCode":1}}

$ gcx datasources mssql describe-table WORLD_DATA -d mssql-ds-m --schema "" -o table
after:  {"type":"gcx.error","schema_version":"1","error":{"summary":"--schema must not be empty","exitCode":1}}

# sanity: a real --schema value is unaffected on both commands
$ gcx datasources mssql list-tables -d mssql-ds-m --schema dbo -o table
        TABLE_SCHEMA  TABLE_NAME  TABLE_TYPE
        dbo           WORLD_DATA  BASE TABLE
$ gcx datasources mssql describe-table WORLD_DATA -d mssql-ds-m --schema dbo -o table
        COLUMN_NAME      DATA_TYPE  IS_NULLABLE  CHARACTER_MAXIMUM_LENGTH  COLUMN_DEFAULT
        BASE_COUNTRY     nvarchar   YES          -1                        -
        ...
```

Full gate (lint + tests + build + docs) green on every commit in this round.

ktw4071:
Dug into the `BuildRawQueryBody` finding from the earlier review pass, since I wasn't sure it was really a bug rather than a stylistic inconsistency. It was a real one.

`internal/query/mssql/client.go` previously omitted `intervalMs` entirely when `--step` was unset, on the theory that the plugin would apply a sensible default for the `$__interval` macro. Tested that theory live against `mssql-ds-m` rather than trusting it — it's wrong. An absent `intervalMs` makes the plugin resolve `$__interval` to a near-zero width:

```
$ TS='SELECT $__timeGroup(DATE_TIME, $__interval) AS t, COUNT(*) AS n FROM [grafanadb].dbo.WORLD_DATA GROUP BY $__timeGroup(DATE_TIME, $__interval) ORDER BY t'
$ gcx datasources mssql query -d mssql-ds-m "$TS" --since 24h -o table

before (intervalMs omitted):
T           N
1787443203  1
1787443208  1
1787443213  1
...                          # 1 row per bucket — no real aggregation

after (mssql now uses querysql.BuildRawQueryBody, defaults intervalMs to 60000 like mysql/postgres):
T           N
1787443200  12
1787443260  12
1787443320  12
...                          # 60s buckets, 12 rows/bucket — matches mysql/postgres's own default exactly
```

Fixed in `5a4c8535` by switching mssql onto the shared `querysql.BuildRawQueryBody` (mysql/postgres already use it) instead of hand-rolling the request body — closes the duplication and the bad default at the same time.

Regression-checked: explicit `--step 1h` (still 720 rows/bucket, unaffected), the `--limit` sentinel warning, negative `--limit` rejection, and the generic auto-detect dispatch path all still behave correctly. Full gate green.

internal/query/mssql/types.go
```diff
@@ -0,0 +1,139 @@
...
+// returning the rewritten SQL and whether the injection happened. It bails
+// (returns the SQL unchanged, false) for statements where injecting TOP would be
+// invalid or change semantics: non-SELECT statements, CTEs (WITH ...), queries
+// that already use TOP, set operations, and paged (OFFSET ... FETCH) queries.
+func injectTop(sql string, n int) (string, bool) {
+	loc := leadingSelectRe.FindStringIndex(sql)
+	if loc == nil {
+		return sql, false // not a leading-SELECT statement (e.g. WITH, EXEC, INSERT)
+	}
+	if existingTopRe.MatchString(sql) || setOpRe.MatchString(sql) || offsetRe.MatchString(sql) {
```
awsome-o: `SELECT ... INTO` reaches this line without bailing, and it is a write.

`leadingSelectRe` matches it:

```sql
SELECT * INTO dbo.backup FROM dbo.orders
```

None of `existingTopRe`, `setOpRe` or `offsetRe` fires, so it comes back as:

```sql
SELECT TOP (101) * INTO dbo.backup FROM dbo.orders
```

That is valid T-SQL. The new table gets 101 rows instead of the whole result set, and gcx then prints `showing the first 100 rows; more rows match` as though it had capped a read.

MySQL already bails on exactly this class. `internal/query/mysql/types.go:55` carries `\bINTO\s+(OUTFILE|DUMPFILE)\b` and `\b(INSERT|UPDATE|DELETE)\b` in `limitBailRe`, with a comment explaining why. And the shared raw SQL contract in `.claude/skills/add-datasource/references/raw-sql.md` has "a write or a metadata statement" as the first row of its bail table.

An `INTO` bail alongside the other three would close it. Worth a test row that asserts the SQL comes back byte-identical, not just that it "was left unchanged", since the reference warns that unchanged can pass for the wrong reason.

One thing that would settle this the other way: if the Grafana MSSQL plugin refuses non-read statements outright, this cannot happen. Your live sweep only ran reads, so I could not tell from the PR.

internal/datasources/clickhouse/query.go
```diff
@@ -72,7 +72,11 @@ open it in your browser after the query succeeds.`,
 				return err
 			}
 
-			sql := clickhouse.EnforceLimit(expr, limit, maxLimit)
+			// Inject LIMIT (eff+1) so we can detect and warn when our own cap hid
+			// rows; displaySQL carries the user-facing LIMIT (eff) for the Explore
+			// link so it never leaks the +1 sentinel.
+			querySQL, eff, capped := clickhouse.EnforceLimitSentinel(expr, limit, maxLimit)
```
awsome-o, line 81: This changes behaviour for a negative `--limit`, and I don't think that was intended.

`EnforceLimit` disables enforcement only at `limit == 0` (`internal/query/sql/types.go:44`). `EffectiveLimit` disables at `limit <= 0` (`types.go:73`), and `EnforceLimitSentinel` returns early on that.

So before this PR:

```
gcx datasources clickhouse query 'SELECT * FROM big' --limit -5
```

sent `... LIMIT -5` and ClickHouse rejected it. After this PR the same command runs with no cap at all, against a table the user was explicitly trying to bound.

The line right below is the other half of it. `displaySQL` still goes through `EnforceLimit`, which does append `LIMIT -5`, so that invalid SQL is what ends up in the `--share-link` Explore URL. Execute path and display path now disagree about what ran.

postgres, mysql and your own new mssql command all reject this up front. `internal/datasources/postgres/query.go:35`:

```go
if opts.Limit < 0 {
    return fmt.Errorf("--limit must be >= 0, got %d", opts.Limit)
}
```

and `internal/datasources/mssql/query.go:65` does the same inline. The same guard here and in `athena/query.go` would close it.

I am flagging this harder than the rest because clickhouse and athena are already released and this PR's stated reason for touching them was adding a warning.

internal/datasources/athena/query.go
```diff
@@ -75,7 +75,11 @@ open it in your browser after the query succeeds.`,
 				return err
 			}
 
-			sql := athena.EnforceLimit(expr, limit, maxLimit)
+			// Inject LIMIT (eff+1) so we can detect and warn when our own cap hid
+			// rows; displaySQL carries the user-facing LIMIT (eff) for the Explore
+			// link so it never leaks the +1 sentinel.
+			querySQL, eff, capped := athena.EnforceLimitSentinel(expr, limit, maxLimit)
```
awsome-o, line 84: Same negative `--limit` problem as `internal/datasources/clickhouse/query.go:78`, see the detail there.

Short version: `EnforceLimit` disabled at `limit == 0`, `EffectiveLimit` disables at `limit <= 0`, so `--limit -5` used to send `LIMIT -5` and get rejected, and now runs unbounded. `displaySQL` on the next line still produces `LIMIT -5` for the Explore link.

A `--limit < 0` guard before the query, matching `internal/datasources/postgres/query.go:35`, fixes both.

internal/query/sql/types.go
```diff
@@ -59,3 +66,71 @@ func EnforceLimit(sql string, limit, maxLimit int, bail func(string) bool) (stri
...
+	if bail != nil && bail(sql) {
+		return sql, eff, false
+	}
+
+	trimmed := strings.TrimRight(sql, "; \t\n")
+	suffix := sql[len(trimmed):]
+
+	if m := limitClauseRe.FindStringSubmatchIndex(trimmed); m != nil {
+		existing, _ := strconv.Atoi(trimmed[m[2]:m[3]])
+		if existing > maxLimit {
```
awsome-o, line 121: This branch clamps the user's own `LIMIT` and returns `false`, so nothing tells them.

With `SELECT * FROM t LIMIT 5000` and `maxLimit` 1000, the SQL that runs becomes `LIMIT 1000`, `capped` comes back false, and `SurfaceRowLimits` stays quiet. The user asked for 5000 rows in their own SQL, got 1000, and heard nothing about it.

`EnforceLimit` returns `true` on the equivalent branch (line 58), which is why mysql and postgres do warn here.

I read the doc comment above explaining why the two bools deliberately mean different things, and the argument about not changing shipped behaviour for mysql and postgres is fair. But the net effect is that the one code path this PR adds disclosure to is also the path that keeps this case silent, and the raw SQL reference calls out silent truncation as the known athena/clickhouse defect. Some signal on this branch, so the caller can say "your `LIMIT 5000` was lowered to 1000", would finish the job the PR started.

MSSQL is not affected, since `EnforceTop` honours a user-supplied `TOP` as-is. This is only about the two LIMIT dialects the PR touches.

internal/datasources/mssql/list_tables.go
```diff
@@ -0,0 +1,102 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			resp, err := client.Query(ctx, datasourceUID, mssql.QueryRequest{RawSQL: sql})
+			if err != nil {
+				return fmt.Errorf("query failed: %w", err)
+			}
+
+			return opts.IO.Encode(cmd.OutOrStdout(), resp)
```
awsome-o, line 100: Worth one line before the `Encode`: this PR teaches `ParseResponse` to collect plugin notices and `query` surfaces them through `SurfaceRowLimits`, but these two schema commands drop them on the floor.

That matters here specifically. Your own live validation comment shows this datasource returning:

```
Results have been limited to 100 because the SQL row limit was reached
```

So `list-tables` against a database with more than 100 tables returns a truncated list and says nothing about it, which is the exact failure the rest of this PR is about.

`SurfaceRowLimits` already handles the notices-only case, since `capped` false skips the truncation branch:

```go
dsquery.SurfaceRowLimits(cmd.ErrOrStderr(), resp, false, 0, 0)
```

Same in `describe_table.go:115`.

Not a regression, to be fair. `Notices` is new in this PR, so no sibling command surfaces them either. But these two are the commands most likely to hit that server-side cap.

internal/datasources/mssql/describe_table.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			schema := opts.Schema
+			if schemaFromName != "" {
+				if schema != "" {
+					return errors.New("specify the schema in the table name (SCHEMA.TABLE) or via --schema, not both")
+				}
+				schema = schemaFromName
+			}
+
+			if err := mssql.ValidateIdentifier(table, "table"); err != nil {
```
awsome-o, line 87: Nit: `describe-table ""` still gets all the way to the wire.

`SplitSchemaQualifiedTable("")` has no dot, so it returns `("", "", nil)`, and `ValidateIdentifier` returns nil for the empty string by design (`internal/query/mssql/types.go:65`, which is its "schema was omitted" case). The command then builds `WHERE TABLE_NAME = ''` and comes back with zero rows.

This is the other half of the `dbo.` case dafydd raised. `SplitSchemaQualifiedTable` now rejects `dbo.` and `.WORLD_DATA`, but a bare empty argument goes straight through. mysql's `splitTableArg` has the same gap, so it isn't specific to this PR, but an explicit empty check here is cheap and you are already in the file.

internal/datasources/mssql/describe_table.go
```diff
@@ -0,0 +1,126 @@
...
+			// Accept a schema-qualified table (e.g. "dbo.WORLD_DATA"); the
+			// schema segment is equivalent to passing --schema.
+			schemaFromName, table, err := mssql.SplitSchemaQualifiedTable(args[0])
+			if err != nil {
+				return err
+			}
+			schema := opts.Schema
+			if schemaFromName != "" {
+				if schema != "" {
+					return errors.New("specify the schema in the table name (SCHEMA.TABLE) or via --schema, not both")
```
awsome-o, line 82: Nit: this rejects `describe-table dbo.WORLD_DATA --schema dbo`, even though the two values agree.

mysql only errors on a genuine conflict (`internal/datasources/mysql/describe_table.go:32`):

```go
if databaseFlag != "" && databaseFlag != parts[0] {
```

An agent that habitually passes `--schema` and also copies a qualified name out of `list-tables` output hits an error for no reason. `if schema != "" && schema != schemaFromName` would match mysql and drop the false positive.

internal/datasources/mssql/list_tables.go
```diff
@@ -0,0 +1,102 @@
...
+				return err
+			}
+
+			// Resolve datasource UID from -d flag, config, or Grafana auto-discovery.
+			datasourceUID, _, err := dsquery.ResolveValidateAndSaveDatasource(ctx, loader, opts.Datasource, cfgCtx, cfg, "mssql")
+			if err != nil {
+				return err
+			}
+
+			sql := "SELECT TABLE_SCHEMA, TABLE_NAME, TABLE_TYPE FROM INFORMATION_SCHEMA.TABLES"
```
awsome-o, line 79: Nit, docs only: `INFORMATION_SCHEMA` is per-database, so this command and `describe-table` can only ever see the datasource's configured database.

That is a reasonable limit, but it isn't stated anywhere, and the error you get when you try to work around it doesn't explain it either. `SplitSchemaQualifiedTable` rejects `db.schema.table` with `use SCHEMA.TABLE (e.g. dbo.WORLD_DATA) or pass --schema`, which reads like a syntax correction rather than "gcx cannot cross databases here".

Meanwhile `query` happily accepts `[grafanadb].dbo.WORLD_DATA`, so the three MSSQL commands disagree about what a valid table reference is.

A sentence in the `Long` text, something like "lists tables in the datasource's configured database", would save someone the round trip.

awsome-o (commented):
Went through the whole diff at `5a4c8535` and re-read the thread history. Overall this is solid, and the MSSQL feature itself is close. `TOP (n)` instead of `LIMIT` is the right call for T-SQL, and I agree with the reasoning in the comment on `EnforceTop` about not rewriting a user's existing TOP, since `TOP (n) PERCENT` and `WITH TIES` really would break under a naive clamp.

Two things I would fix before this merges. Both are small, and both are in the truncation-warning work rather than in MSSQL itself:

1. `injectTop` will inject a row cap into `SELECT ... INTO`, which is a write statement. Details inline on `internal/query/mssql/types.go`.
2. A negative `--limit` now runs unbounded on `datasources clickhouse query` and `datasources athena query`, where it used to error. Those are already-released commands and I don't think this PR meant to change them. Details inline.

Everything else I left inline is smaller, and I have called out which ones are nits.

**On the review threads.** I checked all nine and every point dafydd raised has a code answer. `--step` landed in `0a4a3797`, the truncation warning in `bd19609c` using his own n+1 suggestion, the generic dispatch path in `c0251824`, the `TOP 999999` case got the documented divergence comment, and schema-qualified names are now split. The `--limit` gap on the generic `datasources query` was explicitly deferred to a separate all-provider PR, and `dispatchMSSQL` hardcoding 100 matches what clickhouse, mysql and postgres already do. So the two threads still showing open look open for sign-off rather than for missing work, and dafydd should be the one to close them.

**One structural note, not blocking.** `EffectiveLimit` is exported but has exactly one caller outside tests, `EnforceLimitSentinel` in the same file. And three dialects now run two nearly identical injectors over the same SQL, one for what executes and one for what the Explore link shows. If the sentinel helpers returned both strings, `(querySQL, displaySQL, eff, capped)`, you could delete the separate `EnforceLimit` display calls in clickhouse, athena and mssql, unexport `EffectiveLimit`, and the execute path and display path could no longer disagree. That last part is exactly what point 2 above is.

**Small thing on the description.** Decision 1 is now stale. Since the client moved to `querysql.BuildRawQueryBody` (`internal/query/mssql/client.go:27`), and that helper hardcodes `"format": "table"` (`internal/query/sql/request.go:45`), every raw SQL dialect sends the same string. There is no MSSQL-specific divergence left to pin. `TestQuery_SendsStringTableFormat`'s comment still says it "guards the key MSSQL divergence", which now describes shared behaviour, and `mssql.QueryFormatTable` survives only for the Explore link in `internal/datasources/mssql/explore.go:22`.

ktw4071:
Went through the batch of 8 review comments from 2026-08-25. Grouped by severity:

## Correctness/disclosure bugs

**`types.go:133` (SELECT ... INTO not bailed)** — `23c447d0`. Added a bail check for `\bINTO\b` alongside the existing CTE/UNION/OFFSET/existing-TOP cases, plus a byte-identical-SQL test assertion per your note (not just "was left unchanged"):

```go
{"select into bails", "SELECT * INTO dbo.backup FROM dbo.orders", 100, "SELECT * INTO dbo.backup FROM dbo.orders"},
```

Mutation-tested (reverting the bail check turns this red, restored and green), and now also live-verified against a local MSSQL + Grafana instance instead of just unit tests — set this up specifically to settle this without touching the shared datasource. Full session:

```
$ gcx datasources list --context local
{"datasources":[{"uid":"mssql-e2e","name":"MSSQL","type":"mssql","url":"mssql:1433", ...}]}

# picked a source with enough rows to make truncation visible — WORLD_DATA only has 48,
# sys.all_columns (a real system catalog view) has plenty:
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.WORLD_DATA"
{"rows":[[48]]}
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM sys.all_columns"
{"rows":[[12364]]}

# --- BEFORE: pre-fix binary (built from commit 5a4c8535 via a git worktree) ---
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx_before datasources mssql query -d mssql-e2e "SELECT column_id INTO dbo.gcx_selectinto_test FROM sys.all_columns"
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.gcx_selectinto_test"
{"rows":[[101]]}   # SELECT INTO reached the wire as "SELECT TOP (101) column_id INTO ..." —
                    # 12263 rows silently dropped from the write

# --- AFTER: current (fixed) binary ---
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx datasources mssql query -d mssql-e2e "SELECT column_id INTO dbo.gcx_selectinto_test FROM sys.all_columns"
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.gcx_selectinto_test"
{"rows":[[12365]]}  # full copy — the +1 is sys.all_columns picking up gcx_selectinto_test's own
                      # new column, not truncation (confirmed by re-running the COUNT on
                      # sys.all_columns itself, which also read 12365 afterward)

# cleanup — back to a clean local database
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx datasources mssql list-tables -d mssql-e2e
{"rows":[["dbo","WORLD_DATA","BASE TABLE"]]}
```

This also settles your closing question: the Grafana MSSQL plugin does **not** refuse non-read statements outright — the write went through both before and after, confirming the bail is necessary rather than moot.

**`clickhouse/query.go:78`, `athena/query.go:81` (negative --limit regression)** — `5e65b4ed`. Same `--limit must be >= 0` guard postgres/mysql/mssql already have. Live before/after:

```
$ gcx datasources clickhouse query -d clickhouse-native-ds-m 'SELECT name FROM system.tables' --limit -5
# before: ran unbounded, returned all ~195 rows, no error
{"columns":[{"name":"name","type":"string"}],"rows":[["CHARACTER_SETS"],["COLLATIONS"], ... 195 rows total]}
# after:
{"error":{"summary":"--limit must be >= 0, got -5"}}

$ gcx datasources athena query -d athena-ds-m 'SELECT 1' --limit -5
# before:
{"columns":[{"name":"_col0","type":"number"}],"rows":[[1]]}
# after:
{"error":{"summary":"--limit must be >= 0, got -5"}}
```

**`sql/types.go:112` (silent clamp when user's own LIMIT exceeds maxLimit)** — `78fa61a4`. Instead of a static "we changed your literal" warning, this now sentinels (`maxLimit+1`, eff reported as maxLimit) so the existing Truncate+warn path *detects* whether the query actually had more rows than the ceiling allows. Live before/after on ClickHouse's `system.numbers`:

```
$ gcx datasources clickhouse query -d clickhouse-native-ds-m "SELECT number FROM system.numbers LIMIT 5000"
# before: 1000 rows, no warning at all
# after: same 1000 rows, plus:
{"class":"hint","summary":"showing the first 1000 rows; more rows match — raise --limit (max 1000), pass --limit 0 to remove the cap, or add your own row limit to the query"}
```

**`list_tables.go:91` (dropped plugin notices)** — `314c91f1`. Both `list-tables` and `describe-table` now call `SurfaceRowLimits` before `Encode`, matching `query`.

Reproduced directly on `list-tables` itself (created 110 scratch tables in `mssql-ds-m` to push `INFORMATION_SCHEMA.TABLES` past 100, dropped them after):

```
$ gcx datasources mssql list-tables -d mssql-ds-m   # 112 tables in the database
# before (pre-fix): 100 rows, no notice
# after (current):
{"class":"hint","summary":"Results have been limited to 100 because the SQL row limit was reached"}
100 rows
```

A genuine server-side cap on `mssql-ds-m`, independent of gcx — confirmed it now surfaces correctly.

## Nits

**`describe_table.go:77` (empty table name)** and **`describe_table.go:72` (false-positive schema conflict)** — `fedfb337`. Live before/after:

```
$ gcx datasources mssql describe-table "" -d mssql-ds-m
# before:
{"columns":null,"rows":null}
# after:
{"error":{"summary":"Table name is required"}}

$ gcx datasources mssql describe-table dbo.WORLD_DATA --schema dbo -d mssql-ds-m
# before:
{"error":{"summary":"Specify the schema in the table name (SCHEMA.TABLE) or via --schema, not both"}}
# after: succeeds
{"columns":[{"name":"COLUMN_NAME",...}],"rows":[["BASE_COUNTRY","nvarchar","YES",-1,null], ...]}
```

Genuinely conflicting schemas (`--schema sales` against `dbo.WORLD_DATA`) still correctly reject, unchanged.

**`list_tables.go:75` (docs-only db-scope note)** — folded into `314c91f1`/`fedfb337`. Added a sentence to both commands' Long text; `docs/reference/cli/gcx_datasources_mssql_{list-tables,describe-table}.md` regenerated accordingly.

All 5 commits: full suite green, lint 0 issues, `mise run reference`/`mise run docs` clean with no unexpected drift.

internal/query/mssql/types.go
ktw4071: Added a bail check for `\bINTO\b` alongside the existing CTE/UNION/OFFSET/existing-TOP cases, plus a byte-identical-SQL test assertion per your note (not just "was left unchanged"):

```go
{"select into bails", "SELECT * INTO dbo.backup FROM dbo.orders", 100, "SELECT * INTO dbo.backup FROM dbo.orders"},
```

Mutation-tested (reverting the bail check turns this red, restored and green), and now also live-verified against a local MSSQL + Grafana instance instead of just unit tests — set this up specifically to settle this without touching the shared datasource. Full session:

```
$ gcx datasources list --context local
{"datasources":[{"uid":"mssql-e2e","name":"MSSQL","type":"mssql","url":"mssql:1433", ...}]}

# picked a source with enough rows to make truncation visible — WORLD_DATA only has 48,
# sys.all_columns (a real system catalog view) has plenty:
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.WORLD_DATA"
{"rows":[[48]]}
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM sys.all_columns"
{"rows":[[12364]]}

# --- BEFORE: pre-fix binary (built from commit 5a4c8535 via a git worktree) ---
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx_before datasources mssql query -d mssql-e2e "SELECT column_id INTO dbo.gcx_selectinto_test FROM sys.all_columns"
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.gcx_selectinto_test"
{"rows":[[101]]}   # SELECT INTO reached the wire as "SELECT TOP (101) column_id INTO ..." —
                    # 12263 rows silently dropped from the write

# --- AFTER: current (fixed) binary ---
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx datasources mssql query -d mssql-e2e "SELECT column_id INTO dbo.gcx_selectinto_test FROM sys.all_columns"
$ gcx datasources mssql query -d mssql-e2e "SELECT COUNT(*) AS n FROM dbo.gcx_selectinto_test"
{"rows":[[12365]]}  # full copy — the +1 is sys.all_columns picking up gcx_selectinto_test's own
                      # new column, not truncation (confirmed by re-running the COUNT on
                      # sys.all_columns itself, which also read 12365 afterward)

# cleanup — back to a clean local database
$ gcx datasources mssql query -d mssql-e2e "DROP TABLE IF EXISTS dbo.gcx_selectinto_test"
$ gcx datasources mssql list-tables -d mssql-e2e
{"rows":[["dbo","WORLD_DATA","BASE TABLE"]]}
```

This also settles your closing question: the Grafana MSSQL plugin does **not** refuse non-read statements outright — the write went through both before and after, confirming the bail is necessary rather than moot.

Fixed in `23c447d0`.

internal/datasources/clickhouse/query.go
ktw4071: Same `--limit must be >= 0` guard postgres/mysql/mssql already have. Live before/after:

```
$ gcx datasources clickhouse query -d clickhouse-native-ds-m 'SELECT name FROM system.tables' --limit -5
# before: ran unbounded, returned all ~195 rows, no error
{"columns":[{"name":"name","type":"string"}],"rows":[["CHARACTER_SETS"],["COLLATIONS"], ... 195 rows total]}
# after:
{"error":{"summary":"--limit must be >= 0, got -5"}}
```

Fixed in `5e65b4ed` (same commit fixes athena/query.go:81 below).

internal/datasources/athena/query.go
ktw4071: Same fix as `clickhouse/query.go:78` above, same commit:

```
$ gcx datasources athena query -d athena-ds-m 'SELECT 1' --limit -5
# before:
{"columns":[{"name":"_col0","type":"number"}],"rows":[[1]]}
# after:
{"error":{"summary":"--limit must be >= 0, got -5"}}
```

Fixed in `5e65b4ed`.

internal/query/sql/types.go
ktw4071: Instead of a static "we changed your literal" warning, this now sentinels (`maxLimit+1`, eff reported as maxLimit) so the existing Truncate+warn path *detects* whether the query actually had more rows than the ceiling allows. Live before/after on ClickHouse's `system.numbers`:

```
$ gcx datasources clickhouse query -d clickhouse-native-ds-m "SELECT number FROM system.numbers LIMIT 5000"
# before: 1000 rows, no warning at all
# after: same 1000 rows, plus:
{"class":"hint","summary":"showing the first 1000 rows; more rows match — raise --limit (max 1000), pass --limit 0 to remove the cap, or add your own row limit to the query"}
```

Fixed in `78fa61a4`. `EnforceLimit` (mysql/postgres's simpler, already-shipped contract) is untouched.

internal/datasources/mssql/list_tables.go
ktw4071: Both `list-tables` and `describe-table` now call `SurfaceRowLimits` before `Encode`, matching `query`.

Reproduced directly on `list-tables` itself (created 110 scratch tables in `mssql-ds-m` to push `INFORMATION_SCHEMA.TABLES` past 100, dropped them after):

```
$ gcx datasources mssql list-tables -d mssql-ds-m   # 112 tables in the database
# before (pre-fix): 100 rows, no notice
# after (current):
{"class":"hint","summary":"Results have been limited to 100 because the SQL row limit was reached"}
100 rows
```

A genuine server-side cap on `mssql-ds-m`, independent of gcx — confirmed it now surfaces correctly. Fixed in `314c91f1`.

internal/datasources/mssql/describe_table.go
ktw4071:
```
$ gcx datasources mssql describe-table "" -d mssql-ds-m
# before:
{"columns":null,"rows":null}
# after:
{"error":{"summary":"Table name is required"}}
```

Fixed in `fedfb337`.

internal/datasources/mssql/describe_table.go
ktw4071:
```
$ gcx datasources mssql describe-table dbo.WORLD_DATA --schema dbo -d mssql-ds-m
# before:
{"error":{"summary":"Specify the schema in the table name (SCHEMA.TABLE) or via --schema, not both"}}
# after: succeeds
{"columns":[{"name":"COLUMN_NAME",...}],"rows":[["BASE_COUNTRY","nvarchar","YES",-1,null], ...]}
```

Genuinely conflicting schemas (`--schema sales` against `dbo.WORLD_DATA`) still correctly reject, unchanged. Fixed in `fedfb337`.

internal/datasources/mssql/list_tables.go
ktw4071: Added a sentence to both `list-tables` and `describe-table`'s Long text; `docs/reference/cli/gcx_datasources_mssql_{list-tables,describe-table}.md` regenerated accordingly. Folded into `314c91f1`/`fedfb337`.

awsome-o (dismissed):
LGTM

internal/query/sql/types.go
```diff
@@ -59,3 +66,83 @@ func EnforceLimit(sql string, limit, maxLimit int, bail func(string) bool) (stri
...
+		return false
+	}
+	r.Rows = r.Rows[:eff]
+	return true
+}
+
+// TruncationHint is the stderr message shown when gcx's row cap dropped rows.
+// It is dialect-agnostic (applies to both LIMIT and TOP datasources).
+func TruncationHint(shown, maxLimit int) string {
+	return fmt.Sprintf("showing the first %d rows; more rows match — raise --limit (max %d), pass --limit 0 to remove the cap, or add your own row limit to the query", shown, maxLimit)
```
dafydd-t:
```suggestion
	return fmt.Sprintf("showing the first %d rows. See more rows by setting --limit (max %d).
```

Just a nit about the truncation hint reading

dafydd-t (commented):
Overall I defer to @awsome-o 's review (so LGTM too), but please can we make these two truncation hints read consistently?


e.g. we see both of these in this PR:

```
Results have been limited to 100 because the SQL row limit was reached
```

```
showing the first %d rows; more rows match...
```

ktw4071:
@dafydd-t re: the truncation-hint wording — `TruncationHint` (`internal/query/sql/types.go`) now reads as two complete, period-terminated sentences instead of one dash-joined clause, closer to the plugin notice's style:

```
$ gcx datasources clickhouse query -d clickhouse-native-ds-m "SELECT number FROM system.numbers LIMIT 5000"
{"class":"hint","summary":"showing the first 1000 rows; more rows match. Raise --limit (max 1000), pass --limit 0 to remove the cap, or add your own row limit to the query."}
```

I can't change the plugin's own wording ("Results have been limited to 100 because the SQL row limit was reached") since that's server-generated and surfaced verbatim, but gcx's own hint should no longer clash stylistically with it.

Also caught and fixed a syntax break in the in-progress edit (the string literal was left unterminated) — full suite green, lint 0 issues, in `0b119355`.

Separately: rebased onto main to pick up the BigQuery merge and resolve the resulting conflicts in `query_routes.go` / `project-structure.md` / the two golden-list test files — pushed as a force-update since the rebase rewrote history.

dafydd-t:
@claude review

dafydd-t (approved):

PR #1066, opened 2026-07-26
fix(config): let env credentials override keychain failures

### LINKED ISSUE #1060: config: GRAFANA_TOKEN can't rescue an unreadable keychain credential — validation runs before the rejection-clearing pass
Found while setting up a Viewer-token smoke test on a machine where the login keychain is unreadable for the invoking process tree (headless/locked macOS security session — the error-message side of this is #1061, and the keychain kill-switch follow-up is #1062).

With a config whose grafana token is keychain-backed (a `keychain:gcx:v2:...` sentinel) and the keychain unreadable, setting GRAFANA_TOKEN doesn't rescue anything:

```
GRAFANA_TOKEN=<sa-token> gcx slo definitions list
# → configured credential "stack:<stack>" field "grafana-token" was rejected
#   before network use: the OS keychain could not be read; review the file and
#   re-authenticate with --config "<path>"
```

Same for the K8s tier (gcx resources / datasources / api). On a headless machine this bricks gcx entirely — there's no way past the stored sentinel short of hand-editing the config file.

The frustrating part is the rescue mechanism already exists, it just runs too late. The sequence in Load/LoadLayered:

1. sentinel resolution runs first and, on an unreadable keychain, blanks the field and records a rejection on the stack (resolveSentinelsForOwner, internal/config/keychain.go:959-980)
2. the overrides loop runs; ParseEnvIntoContext detaches the stack view, sets the env token, and marks runtimeSecretOverrides (internal/config/envparse.go:42)
3. enforceRuntimeCredentialBindings runs after all overrides and would clear the stale rejection for env-overridden fields (internal/config/loader.go:629 and 1146; the clear is keychain.go:1063)

But both major callers pass Context.Validate as an override, so it runs between 2 and 3: internal/providers/configloader.go:273-280 (all provider commands) and cmd/gcx/config/command.go:127 plus 133-144 (K8s tier). Validate consults grafanaCredentialRejectionForMode (internal/config/types.go:497-503), sees the stale rejection, and LoadLayered returns the CredentialRejectedError before the clearing pass ever runs. selectGrafanaAuth itself already honors the override (grafana_auth.go:96-98 forces token mode), so the env token gets selected and then immediately rejected.

Fix sketch: have ParseEnvIntoContext clear the rejection for each env-overridden field right there, on the detached owners (StackEntry for grafana-token/password/sm-token, CloudEntry for cloud-token). It operates on the detached clone, so persisted rejection evidence stays visible to sibling contexts, and the local-layer rule (keychain.go:1050-1054, env creds + auto-discovered repository destination) still re-rejects afterwards since enforce runs after all overrides. Blank env values never mark runtimeSecretOverrides, so the clear-inherited-CI-state semantics are untouched.

Why tests miss it: the failing-store tests in internal/config/keychain_bound_internal_test.go never set GRAFANA_TOKEN, and the GRAFANA_TOKEN tests all use a working store. Nothing exercises the callers' validate-as-override chain with a failing store. The red test is exactly that combination: bound sentinel + a store whose Get returns credentials.ErrUnavailable + GRAFANA_TOKEN set + overrides [env override, Validate-as-override] → should load with the env token winning.

Adjacent to #1049 but a different root cause: #1049 is the env-only/no-config-file path, which never touches sentinel resolution at all; this one needs a config file that carries a sentinel while the keychain is unreadable. Env override with a config file and a working keychain is fine today.

Worth a release-notes mention when fixed — the failure mode is "gcx unusable on headless machines with no workaround".

sidsri14:
## Summary

- clear stale keychain-read rejection evidence when a nonblank environment credential supplies the effective runtime value
- limit the clearing to detached runtime owners so persisted stack and cloud entries retain their rejection state
- add regression coverage for an unavailable keychain entry rescued by `GRAFANA_TOKEN`

## Root cause

Keychain loading recorded an unreadable credential as rejected, but validation ran before the later binding pass that could clear that rejection for an explicit environment override. `ParseEnvIntoContext` now clears the matching rejection immediately after parsing nonblank credential overrides, on the already-detached runtime view.

## Testing

- `go test ./...`
- `go vet ./...`
- `golangci-lint run -c .golangci.yaml` (v2.12.2)
- `go run ./cmd/gcx/ dev lint test ./internal/linter/bundle/gcx/`
- `go run ./scripts/validate-skills`
- `go build -buildvcs=false -o bin/gcx ./cmd/gcx/`
- regenerated all reference documentation with no drift
- `mkdocs build -f mkdocs.yml`

Fixes #1060

### DIFF
--- a/internal/config/keychain_bound_internal_test.go
+++ b/internal/config/keychain_bound_internal_test.go
@@ -541,6 +541,33 @@ func TestBoundKeychainUnavailableReferenceIsPreservedUntilExplicitUnset(t *testi
 	assert.Contains(t, store.deletes, account)
 }
 
+func TestBoundKeychainUnavailableGrafanaTokenCanBeOverriddenByEnvironment(t *testing.T) {
+	store := newBoundTestStore()
+	useBoundTestStore(t, store)
+	path := filepath.Join(t.TempDir(), "config.yaml")
+	cfg := boundStackTestConfig("https://example.invalid", "stored-token")
+	cfg.Stacks["default"].Grafana.StackID = 12345
+	require.NoError(t, Write(t.Context(), ExplicitConfigFile(path), cfg))
+
+	store.getErr = credentials.ErrUnavailable
+	store.gets = nil
+	t.Setenv("GRAFANA_TOKEN", "env-token")
+
+	loaded, err := Load(t.Context(), ExplicitConfigFile(path),
+		func(cfg *Config) error {
+			return ParseEnvIntoContext(cfg.Contexts["default"])
+		},
+		func(cfg *Config) error {
+			return cfg.GetCurrentContext().Validate(t.Context())
+		},
+	)
+	require.NoError(t, err)
+	assert.Equal(t, "env-token", loaded.Contexts["default"].Grafana.APIToken)
+	assert.Empty(t, loaded.Stacks["default"].Grafana.APIToken,
+		"the environment override must not mutate the persisted stack view")
+	require.Error(t, loaded.Stacks["default"].credentialRejection(credentials.FieldGrafanaToken))
+}
+
 func TestBoundKeychainWholeOwnerDeletionRemovesExactAccount(t *testing.T) {
 	store := newBoundTestStore()
 	useBoundTestStore(t, store)
--- a/internal/config/envparse.go
+++ b/internal/config/envparse.go
@@ -58,6 +58,7 @@ func ParseEnvIntoContext(ctx *Context) error {
 		return err
 	}
 	CleanupAfterEnvParse(ctx)
+	clearRuntimeCredentialRejections(ctx)
 	if ctx.StackEntry != nil {
 		// PrepareForEnvParse may have created Grafana for a named stack that had
 		// no persisted Grafana block. Keep binding checks on the detached stack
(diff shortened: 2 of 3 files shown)


dafydd-t:
@sidsri14 when you are ready, take this out of draft and I'll take a proper look. Overall, it looks good! Thanks for making a PR

internal/config/keychain_bound_internal_test.go
```diff
@@ -541,6 +541,33 @@ func TestBoundKeychainUnavailableReferenceIsPreservedUntilExplicitUnset(t *testi
...
+	store := newBoundTestStore()
+	useBoundTestStore(t, store)
+	path := filepath.Join(t.TempDir(), "config.yaml")
+	cfg := boundStackTestConfig("https://example.invalid", "stored-token")
+	cfg.Stacks["default"].Grafana.StackID = 12345
+	require.NoError(t, Write(t.Context(), ExplicitConfigFile(path), cfg))
+
+	store.getErr = credentials.ErrUnavailable
+	store.gets = nil
+	t.Setenv("GRAFANA_TOKEN", "env-token")
```
dafydd-t, line 554: If you wanted to, you could expand this enhancement in general to also cover when `GRAFANA_PASSWORD` is set.  We only cover GRAFANA_TOKEN here: https://github.com/grafana/gcx/blob/main/internal/config/grafana_auth.go#L96-L98 - adding a similar clause for the existence of credentials.FieldGrafanaPassword in the context overrides would mean we could clear keychain warnings for that use case too.

Up to you if you want to add that here.

dafydd-t (approved):
LGTM, we can make another PR for the GRAFANA_PASSWORD clear if you want it

dafydd-t:
@sidsri14 once you push with these commits signed, I can re-approve and merge. Thanks a lot for taking the time to make this PR

dafydd-t:
I amended the commit to get it signed 👍  lets get this merged

PR #1069, opened 2026-07-27
fix(cloud): drop status, createdAt, updatedAt from stack regions output

dafydd-t:
Prep for #1044, which we can't merge yet (waiting on an oauth scope for the new `/api/v1/stack-regions` endpoint). This lands the output-shape half now so the command surface and result shapes don't change when #1044 applies: drop `status`, `createdAt` and `updatedAt` from `cloud.Region` (the v1 endpoint won't return them; the old endpoint's extra fields are ignored on unmarshal) and remove the STATUS column from `gcx cloud stacks list-regions`. The endpoint itself is unchanged.

Once this merges, #1044 rebases down to just the endpoint swap and a doc line - zero user-visible change at that point.

Before:

```
SLUG                 NAME               DESCRIPTION                   PROVIDER  STATUS
prod-us-west-0       AWS US West        AWS US West                   aws       active
us                   GCP US Central     United States                 gcp       active
```

After:

```
SLUG                 NAME               DESCRIPTION                   PROVIDER
prod-us-west-0       AWS US West        AWS US West                   aws
us                   GCP US Central     United States                 gcp
```

### DIFF
--- a/internal/cloud/gcom.go
+++ b/internal/cloud/gcom.go
@@ -77,13 +77,10 @@ type StackInfo struct {
 // Region describes a Grafana Cloud stack region as returned by the GCOM API.
 type Region struct {
 	ID          int    `json:"id"`
-	Status      string `json:"status"`
 	Slug        string `json:"slug"`
 	Name        string `json:"name"`
 	Description string `json:"description"`
 	Provider    string `json:"provider"`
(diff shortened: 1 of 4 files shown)


annanay25 (approved):
seems harmless 👍

PR #1071, opened 2026-07-27
enhancement(telemetry): populate target_kind in usage events

dafydd-t:
Part of https://github.com/grafana/gcx-internal/issues/2

## Summary

- populate `target_kind` in anonymous usage events, to show whether the config associated with the current gcx invocation was interacting with a self-hosted or cloud grafana instance.
- it uses all layers of config used in the invocation to capture the target kind

### DIFF
--- a/cmd/gcx/telemetry.go
+++ b/cmd/gcx/telemetry.go
@@ -75,9 +75,8 @@ func buildUsageEvent(info *root.TelemetryInfo, start time.Time, exitCode int) te
 		IsTTY:   terminal.StdoutIsTerminal(),
 		IsAgent: agent.IsAgentMode(),
 		Agent:   agent.Name(),
-		// TargetKind needs the resolved config context; resolving it requires
-		// a keychain-free context loader, which is still a follow-up. Empty
-		// until then.
+		// Keychain-free classification: only ever "cloud", "self-managed", or "".
+		TargetKind: internalconfig.LoadTargetKind(context.Background()),
 	}
 
 	// The provider is the top-level command, the first segment of the path.
--- a/cmd/gcx/telemetry.go
+++ b/cmd/gcx/telemetry.go
@@ -72,11 +72,10 @@ func buildUsageEvent(info *root.TelemetryInfo, start time.Time, exitCode int) te
 		DurationMS:   time.Since(start).Milliseconds(),
 		OutputFormat: info.OutputFormat,
 
-		IsTTY:   terminal.StdoutIsTerminal(),
-		IsAgent: agent.IsAgentMode(),
-		Agent:   agent.Name(),
-		// Keychain-free classification: only ever "cloud", "self-managed", or "".
-		TargetKind: internalconfig.LoadTargetKind(context.Background()),
+		IsTTY:      terminal.StdoutIsTerminal(),
+		IsAgent:    agent.IsAgentMode(),
+		Agent:      agent.Name(),
+		TargetKind: internalconfig.CapturedTargetKind(),
 	}
 
 	// The provider is the top-level command, the first segment of the path.
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -381,6 +381,10 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 	if flags.AllowServerOverride {
 		opts.AllowOverride = true
 	}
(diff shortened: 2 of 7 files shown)


cmd/gcx/telemetry.go
```diff
@@ -75,9 +75,8 @@ func buildUsageEvent(info *root.TelemetryInfo, start time.Time, exitCode int) te
 		IsTTY:   terminal.StdoutIsTerminal(),
 		IsAgent: agent.IsAgentMode(),
 		Agent:   agent.Name(),
-		// TargetKind needs the resolved config context; resolving it requires
-		// a keychain-free context loader, which is still a follow-up. Empty
-		// until then.
+		// Keychain-free classification: only ever "cloud", "self-managed", or "".
```
dafydd-t: dont need this comment here

awsome-o:
I pushed 7 commits on top of 17debaee. Fast forward, nothing rewritten.

## What was Noticted

1. The docs page said the value is `cloud` or `self-managed`. The code emits `self-hosted`, so the page named a value that never appears.
2. `target_kind` was only recorded when the config load succeeded. Commands run context validation as an override, so a config that resolved fine but failed validation (a self-hosted stack with no `org-id`, say) reported nothing and looked the same as having no config at all.
3. Recording on every error path then went too far. `--context does-not-exist` fails selection without changing the current context, so the event reported whichever context was previously current. A wrong label is worse than a blank one.
4. `gcx login` reported the current context rather than the target it was aiming at. Four cases: a new context built from `--server`, an existing context re-pointed with `--allow-server-override`, `--cloud` against a custom domain, and a new context with no server yet.
5. Every login rejection gate runs before the point where the target was recorded, so any refusal reported the old target. The earliest gate, refusing fresh credentials for an auto-discovered config, runs before any config is read.
6. The recapture after `login.Run` had no test. Deleting the call left every test in the package passing.

## Measured with the built binary, `GCX_TELEMETRY=log`

| invocation | before | after |
|---|---|---|
| self-hosted stack, no `org-id` | *(blank)* | `self-hosted` |
| `config view --context does-not-exist` | `self-hosted` | *(blank)* |
| `login --server <stack>.grafana.net` (no matching context) | `self-hosted` | `cloud` |
| `login dev --server <stack>.grafana.net --allow-server-override` | `self-hosted` | `cloud` |
| `login --cloud --server https://grafana.mycorp.example` | `self-hosted` | `cloud` |
| `login new-context --yes` (no server) | `self-hosted` | *(blank)* |
| `login dev --server <stack>.grafana.net` (stored token rejected) | `self-hosted` | `cloud` |
| `login default --cloud --cloud-token X` (auto-local refusal) | *(blank)* | `cloud` |

Unchanged: cloud config `cloud`, self-hosted config `self-hosted`, `login --yes` refreshing an existing self-hosted context `self-hosted`.

One thing to know before reading the data: blank means "unknown" and nothing narrower. It covers no config, a config that failed to load, a context selection that failed, a context holding only a cloud token, and the config write paths. It cannot separate those. What changed here is that blank is now correct rather than a wrong label; it is still not informative. Splitting those cases apart would need a separate reason field.

## How it works now

Requested intent (`--cloud`, `--server`) is recorded as the first thing `runLogin` does, so every rejection gate reports it. After the config load it is refined from the resolved context, which covers logins that pass neither flag. After `login.Run` it is overwritten from `opts.Target`, which is the only thing that can see a Cloud stack behind a custom domain. On the config side, classification survives validation and credential errors but not a failed context selection, keyed on `ErrContextNotFound` rather than on override ordering.

## Testing

Full suite green with race detection, lint clean, no reference drift. Each fix was checked by reverting it on its own and running the whole package, no `-run` filter. That matters here: the first version of the validation test passed against the broken code in a full run, because an earlier test left `self-hosted` in the process wide value. Tests now seed the opposite value first. The `login.Run` recapture is covered by driving `runLoginLoop` with an injected `DetectFn`, which asserts detection actually ran.

## Left alone, worth a second opinion

A context holding only a cloud token stays unclassified, since it names no Grafana instance. `config set --file` stays unclassified, because those paths load one named layer rather than the merged view and could report a kind the command never targeted. The auto-local gate stays unclassified when neither `--cloud` nor `--server` is given, because the target is knowable only by reading config and that gate deliberately runs first.

jalevin (approved):

edwardcqian (approved):

awsome-o (approved):

Issue #241, opened 2026-03-30
Add confirmation prompts and --yes/auto-approve flag for destructive operations


radiohead:
## Problem

Destructive operations like `delete` execute immediately without confirmation, creating risk for accidental data loss. There is no safety mechanism for interactive users, nor a bypass mechanism for automation.

## Solution

Add safety confirmation prompts for destructive operations with bypass mechanisms for automation. Depends on agent mode (Phase 3).

## Acceptance Criteria

- [ ] `delete` command prompts for confirmation before proceeding
- [ ] `--yes` / `-y` flag bypasses confirmation on individual commands
- [ ] `GRAFANACTL_AUTO_APPROVE=1` env var bypasses all confirmations globally
- [ ] Agent mode automatically bypasses all confirmations
- [ ] Confirmation prompt shows resource count and types being affected

PR #1073, opened 2026-07-27
update README and AGENTS

dafydd-t:
AGENTS.md
- releasing steps put into a skill
- clean up AGENTS.md fluff to rely on references more
- remove Beads references

README.md
- update command examples after the CLI surface refactor

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,7 +538,8 @@ jobs:
           gcx resources push -p ./resources --on-error abort
 ```
 
-- `push`, `pull`, and `validate` are non-interactive; add `--yes` to `delete` to skip its confirmation prompt
+- `gcx resources` commands are designed for non-interactive use, apart from `edit`, which opens an external editor
+- `resources delete` has no confirmation prompt. Named selectors proceed without `--force`; a type-only selector such as `dashboards` requires `--force` (`--yes` also enables it) and can delete every matching resource in the selected context and namespace
 - `--dry-run` on `push` and `delete` to preview changes
 - `--on-error abort|fail|ignore` to control error behavior
 - `-o json` or `-o yaml` for machine-parseable output
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ The agentic workflow above is one example. gcx supports a wide range of workflow
 - **Resource GitOps** — Pull resources to local files, let your agent edit them, push back to Grafana (`gcx resources pull` / `gcx resources push`)
 - **Explore your data** — Discover datasources, metrics, labels, and log streams before writing queries (`gcx datasources list`, `gcx metrics labels`)
 - **SLO management** — Create, monitor, and investigate SLOs from your terminal (`gcx slo definitions list`, `gcx slo reports list`)
-- **Onboarding & setup** — Instrument a Kubernetes cluster and configure Grafana Cloud products (`gcx setup instrumentation`)
+- **Onboarding & setup** — Instrument a Kubernetes cluster and configure Grafana Cloud products (`gcx instrumentation setup`)
 - **Observability as Code** — Scaffold a project, import existing dashboards as Go code, lint, and deploy (`gcx dev scaffold`, `gcx dev import`)
 
 ## Compatibility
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -538,7 +538,7 @@ jobs:
           gcx resources push -p ./resources --on-error abort
 ```
 
-- All commands except `edit` are non-interactive — safe for pipelines
+- `push`, `pull`, and `validate` are non-interactive; add `--yes` to `delete` to skip its confirmation prompt
```
awsome-o: I think this line may be describing planned behaviour rather than current behaviour. I traced `gcx resources delete` and couldn't find a confirmation call or a stdin read anywhere in its execution path, including the cobra hooks, the fetch, the adapters and the remote deleter. What `--yes` does instead is enable `--force` (delete.go line 115, with a notice on stderr at 118), and `--force` is what gets past the named-selector guard at line 132. So with a type-only selector such as `dashboards` it isn't skipping a prompt, it's authorising a delete of every matching resource in the selected context and namespace. Named selectors, including several at once, don't need it.

Looks like issue #241 is still open and asks for exactly this prompt, so my read is that `--yes` shipped ahead of the confirmation it was meant to bypass.

Worth saying the README isn't the only place with this. The command's own long help says it (delete.go line 70, "Use --yes to skip confirmation prompts"), and since the CLI reference is generated from that, docs/reference/cli/gcx_resources_delete.md line 7 publishes it too. DESIGN.md line 137 also says delete prompts, although the `GCX_AUTO_APPROVE` half of that line is accurate.

So it's really two options. Either #241 lands before GA and this sentence becomes true, or we describe current behaviour and fix the help text in the same pass so the generated reference stops repeating it. Given the timing I'd guess the second, but that's a call for whoever owns #241. If it's the second, maybe something like:

    - `gcx resources` commands are designed for non-interactive use, apart from `edit`, which opens an external editor
    - `resources delete` has no confirmation prompt. Named selectors proceed without `--force`; a type-only selector such as `dashboards` requires `--force` (`--yes` also enables it) and can delete every matching resource in the selected context and namespace. Try `--dry-run` first

awsome-o:
nit, not blocking, just noting while the branch is open.

The `internal/` pointer sends people to docs/architecture/project-structure.md and I don't think that file has caught up with everything removed here. Grepping it I get no hits for `irm`, `aio11y`, `queryerror` or `assistanthttp`, so the providers tree is there but a few packages aren't. Delegating is the right instinct, it'd just be good to fill those in, or keep the map until they are.

Related: step 4 now names project-structure.md, but docs/reference/doc-maintenance.md still points at the CLAUDE.md package map in a few places and stage-checklist.md line 25 does too, so those will drift against each other. Beads is the same shape, stage-checklist.md lines 31 and 32 and the docs/_templates/ files still mention it.

All fine as a follow up after GA.

awsome-o:
Unrelated question while you're in the README, not a request.

Anonymous usage telemetry is on by default and I can't find it mentioned in the README anywhere. It's documented under docs/reference/environment-variables (`GCX_TELEMETRY`, and `DO_NOT_TRACK`), but for a public GA I'd half expect a line in the README itself, since that's where people tend to look. Worth adding, or is it deliberately kept to the reference docs?

dafydd-t:
> Anonymous usage telemetry is on by default and I can't find it mentioned in the README anywhere. 


Good shout, I prefer to advertise it too. Added a note and a link at the bottom of the README

dafydd-t:
> All fine as a follow up after GA.


Yeah these do look a bit off, but we can follow up after 👍

README.md
dafydd-t: Thanks for raising this, its definitely one we want to document accurately. I think i've addressed these in https://github.com/grafana/gcx/pull/1073/changes/accc9a0f8e5de85a44fbd2aae69c3283f57863b5

awsome-o: That covers it, thanks for turning it round so quickly.

Checked all four on accc9a0f: the README bullets, delete.go's long help, the generated reference (reference-drift runs clean, so it matches what the generator produces), and DESIGN.md now pointing at #241 instead of claiming a prompt. `grep` for "skip confirmation prompts" across the repo comes back empty now.

You also fixed something I'd missed. The old DESIGN.md line said agent mode auto-approves, but `CheckDestructiveBypass` returns `ErrAgentModeRequiresForce` when agent mode is on without `--force`, so your wording is more correct than what I suggested.

Usage statistics section looks good too, the link resolves and the page does cover the opt-out env vars.

awsome-o (approved):
Approving. The delete wording is now accurate in all four places it was wrong, and reference-drift, `go test ./cmd/...` and lint on the changed package all pass locally on accc9a0f.

The package map nits are unaddressed and that's fine by me, happy for those to be a follow up after GA.

dafydd-t:
https://github.com/grafana/gcx-internal/issues/36

Issue #959, opened 2026-07-10
gcx kg: --dry-run for model-rules & prom-rules create (validate + diff)


xujiaxj:
Add `--dry-run` to `gcx kg model-rules create` and `gcx kg prom-rules create`, following the validate-then-diff contract from the suppressions work (#577): validate the payload server-side, show a diff (remote → local), never persist.

## Behavior
- Validate first → on failure, render field errors, exit non-zero.
- Valid → fetch current (upsert-by-name) and render a canonical-YAML diff, or "no changes".
- `--dry-run` issues zero writes.

## Backend changes needed (asserts-adi)
Both commands need a **synchronous** validate-only endpoint (dry-run must return an immediate answer):

- **model-rules** — no validate endpoint today. Add one reusing the existing `CustomModelRulesValidator` (the PUT's validation block minus persist).
- **prom-rules** — an async submit/poll validate endpoint exists but is **unused by asserts-app-plugin** (the UI saves via direct PUT). Add a new synchronous endpoint running the PUT's `basicValidator` — the check the write actually gates on — instead of relying on the orphaned async path.

Both are the same small pattern already used by suppressions' `disabled-alerts-validate`.

> **Note:** each new backend endpoint also requires a corresponding route change in the **api-gateway** repo: backend-enterprise.

## Notes
- #940 (prom-rules schema) becomes authoring-only, not on the dry-run critical path.
- Diff semantics: both PUTs are upsert-by-name, so the diff is scoped to the named config — no spurious deletions.

## Acceptance criteria
- [ ] `--dry-run` on both create commands: validate + diff, no writes (assert via fake client)
- [ ] Invalid payload → non-zero exit + field errors
- [ ] Diff golden tests: create (all-added), replace-with-changes, no-op
- [ ] Manual E2E: author a rule → `create --dry-run` → real create → edit → `create --dry-run` → delete

PR #1077, opened 2026-07-27
feat(kg): add notifications CRUD over alert notification configs (#1011)

### LINKED ISSUE #1011: gcx kg notifications — CRUD over alert notification configs (mirror suppressions)
## Summary
Add `gcx kg notifications {create, list, get, delete}` over the alert notification configs at `/v1/config/alert(s)` (`AlertConfigController`). Distinct from `suppressions` (which manages *disabled*-alert configs) — this manages the notification-rule lifecycle.

## Approach: mirror the `suppressions` provider
The backend controllers are structural twins, so this is ~90% a clone of the existing suppressions implementation (`internal/providers/kg/`) — client methods, command wiring, table codec, and the `-f`/stdin YAML pattern — swapping the DTO type and paths. Build from that template and apply the deltas below.

## Backend endpoint map (verified in asserts-adi)

| gcx verb | Endpoint | Notes |
|---|---|---|
| `list` | `GET /v1/config/alerts` → `AlertConfigsDto` | + category variants `GET /v1/config/alerts/{request\|resource\|health\|slo}` |
| `get <name>` | — **no by-name GET** | client-side: fetch list + filter by `.name` |
| `create` (upsert) | `POST /v1/config/alert` (single, name in body); batch `POST /v1/config/alerts` | **POST**, not PUT |
| `delete <name>` | `DELETE /v1/config/alert/{name}` | |

## Deltas vs suppressions
1. **DTO is richer** — `AlertConfigDto` = `{name, matchLabels, alertLabels, annotations, duration, silenced, managedBy}` (vs suppressions' `{name, matchLabels, managedBy}`).
2. **Add `get`** — suppressions never shipped it. No by-name GET exists, so implement as list-then-filter. (Consider backfilling `get` to suppressions later for symmetry.)
3. **`list --category`** — supports an extra `slo` category beyond suppressions' request/resource/health.
4. **No validate endpoint** — unlike suppressions (`disabled-alerts-validate`), there is **no** alert-config validate endpoint. So:
   - Do **not** promise a server-validated `--dry-run` here.
   - If `--dry-run` is wanted, it's **diff-only**, or requires a new backend validate endpoint (same situation as model-rules in #959) — out of scope for this ticket.

## Command shape (align verbs to suppressions)
```bash
gcx kg notifications list [--category request|resource|health|slo]
gcx kg notifications get <name>
gcx kg notifications create -f notifications.yaml   # or stdin; one or more
gcx kg notifications delete <name>
```

## Acceptance criteria
- [ ] `list` (with `--category`), `get <name>` (list-filter), `create` (`-f`/stdin, single + batch), `delete <name>`
- [ ] Reuses the suppressions table codec / output conventions; `-o json|yaml` supported
- [ ] `get` returns a clean not-found error when no config matches the name
- [ ] Client methods mirror `GetSuppressions`/`UploadSuppressions`/`DeleteSuppression` (no `Validate*`)
- [ ] Unit tests: list, category filter, get-by-name filter (hit + miss), create payload mapping, delete
- [ ] Reference docs regenerated (`GCX_AGENT_MODE=false mise run reference`)

## Other follow-ups while we are here
- `--dry-run` (no backend validate endpoint; so could use a new backend validate endpoint)
- Any change to `suppressions` (the `get`-backfill suggestion is a separate follow-up)

kblaschke512:
## Summary

Adds `gcx kg notifications {list, get, upsert, delete}` over the Knowledge Graph alert **notification** configs at `/v1/config/alert(s)` (`AlertConfigController`). Closes #1011.

These govern how a matched alert notifies — the labels it must match, extra alert labels and annotations to attach, the `for` duration, and the `silenced` flag. It is the structural twin of `gcx kg suppressions` (which manages *disabled*-alert configs), built by mirroring that provider and swapping in the richer DTO.

## Backend endpoint map (no backend change — endpoints already exist)

| gcx verb | Endpoint | Notes |
|---|---|---|
| `list` | `GET /v1/config/alerts` → `AlertConfigsDto` | optional category suffix `/v1/config/alerts/{request\|resource\|health\|slo}` |
| `get <name>` | — (no by-name GET) | client-side list-then-filter by `.name`; clean not-found on miss |
| `upsert` | `POST /v1/config/alert` (single) | create-or-update by name; never deletes remote entries absent from the file |
| `delete <name>` | `DELETE /v1/config/alert/{name}` | routed through the shared guarded-delete helper |

**Wire shape verified live before coding:** the batch wrapper is `{"alertConfigs":[...]}`, and the duration field on the wire is `for`, not `duration` as the issue draft assumed.

**No `--dry-run`:** there is no server-side validate endpoint for alert configs (unlike suppressions' `disabled-alerts-validate`), so — per the issue — a server-validated dry-run is deliberately **out of scope** here. It would be a separate follow-up (new backend validate endpoint + gateway routes + CLI flag, same shape as model-rules/prom-rules in #959/#1040).

## Usage

```bash
gcx kg notifications list [--category request|resource|health|slo]
gcx kg notifications get <name>
gcx kg notifications upsert -f notifications.yaml   # or stdin; single or batch
gcx kg notifications delete <name>
```

## Implementation

Mirrors the `suppressions` provider. Net diff vs `main`: 13 files, +786/−6.

- `internal/providers/kg/client.go` — `alertConfig*` paths; `NotificationCategory` enum; `AlertConfig`/`AlertConfigs` types; `ListNotifications` (category-aware), `GetNotification` (list-filter + not-found), `UpsertNotification`, `DeleteNotification`
- `internal/providers/kg/commands.go` — `newNotificationsCommand`, `NotificationTableCodec`; `delete` uses the `guardedDeleteOpts`/`runGuardedDelete` pattern (kind `AlertConfig`)
- `internal/providers/kg/provider.go` — mount under the `kg` tree
- `internal/agent/command_annotations.go` — token_cost + agent hints
- `cmd/gcx/root/testdata/output_classes.json` — classify the four commands as `finite`
- `internal/providers/kg/notifications_test.go` + `export_test.go` — unit tests
- `docs/reference/cli/gcx_kg_notifications*.md`, `gcx_kg.md` — regenerated reference

## Testing

- **Unit (11 tests, all pass):** client + command level — list, category filter, invalid category, get hit/miss, upsert single + empty-file, delete.
- **Full gate:** `GCX_AGENT_MODE=false` — `mise run lint` 0 issues; full `go test ./...` green (incl. the `output_classes.json` gate); CLI reference regenerated, no drift.
- **Manual end-to-end (dev stack, all four verbs):** `list` (table + `--category slo` + json), `get` (hit + clean not-found), `upsert` (stdin), `delete` → follow-up `get` returns not-found. Exercised with a throwaway config and cleaned up — nothing left on the stack.

## Notes / follow-ups

- **Verb naming:** shipped as `upsert` (not `create`) to align with suppressions/model-rules/prom-rules after the #387 naming convergence; the backend call is a POST create-or-update.
- **`get` backfill to suppressions** (which never shipped `get`) is a separate symmetry follow-up suggested in the issue.


xujiaxj (dismissed):

awsome-o:
Took a careful pass over this one since it lands on the alert path. Overall it's in good shape: the endpoint map in your description matches what the client actually does, the `guardedDeleteOpts` / `runGuardedDelete` reuse for `delete` is exactly right, and on naming it lines up with the guide in #1023. `upsert` is the blessed create-or-update verb there, and alert configs are addressable by name, so `notifications get <name>` correctly earns its own noun group instead of a `list-notifications` shape.

The one part I'd hold the merge on is `upsert`. `list`, `get` and `delete` all go through the migrated plumbing (`cmdio.Options` on the first two, `runGuardedDelete` on the third), but `upsert` was written by hand and ended up skipping it. The suppressions `upsert` about a hundred lines above it in the same file is the shape it wants.

## 1. `upsert` has no output plumbing, so it can't honour the `finite` class this PR declares for it

[`notificationsCreateOpts`](https://github.com/grafana/gcx/blob/92a7429b7958904abdf112c2d7adfc42af1efd2a/internal/providers/kg/commands.go#L1162-L1168) carries only `File`, no `IO cmdio.Options`, and the success path writes a prose line straight to stdout:

```go
cmdio.Success(cmd.OutOrStdout(), "%d notification config(s) upserted", total)
```

Two things follow from that. In agent mode stdout gets `✔ 2 notification config(s) upserted` rather than a JSON document, which contradicts the `"gcx kg notifications upsert": "finite"` entry this PR adds to `output_classes.json` (finite means stdout parses as exactly one JSON value, per `cmd/gcx/root/agentconformance_test.go`). And since nothing calls `IO.BindFlags`, `gcx kg notifications upsert -f n.yaml -o json` fails with `unknown shorthand flag: 'o' in -o`, while that same flag works fine on `gcx kg suppressions upsert`. So an agent that reads the class table and parses stdout as JSON gets a parse error, and any script reaching for `-o json` breaks outright.

Worth knowing why CI stays green: `TestAgentConformance_EveryFiniteLeafEmitsOneJSONValue` invokes the leaf with no args, that path fails before reaching the success line, and the root error document is what lands on stdout. So the passing case is never exercised.

Should be a small fix. Add `IO cmdio.Options` to the opts, `BindFlags` in `setup`, `IO.Validate()` at the top of RunE, and encode a mutation document in place of the prose line.

## 2. Partial upsert failure loses the record of what was applied

[The loop](https://github.com/grafana/gcx/blob/92a7429b7958904abdf112c2d7adfc42af1efd2a/internal/providers/kg/commands.go#L1109-L1114) returns a wrapped error on the first failure and nothing else. So if entry 4 of 10 fails, the first three are already written to the tenant, but the exit code is 1, which is the same code you get when nothing was applied at all, and stdout is empty. There's no machine-readable record of which entries landed, so automation branching on the exit code can't tell a total failure from a partial one and may re-drive or roll back the wrong set.

The suppressions `upsert` handles exactly this case: it builds `cmdio.NewBatchMutation("upserted")`, fills in succeeded / failed / skipped, encodes the document, then returns `gcxerrors.NewEmittedError(gcxerrors.ExitPartialFailure, failure)` for exit 4. It also special-cases `i == 0`, since when nothing succeeded no result document is owed and the plain error path is correct. Lifting that block over should be close to mechanical.

## 3. `silenced: false` disappears in both directions

[`Silenced bool`](https://github.com/grafana/gcx/blob/92a7429b7958904abdf112c2d7adfc42af1efd2a/internal/providers/kg/client.go#L487) is tagged `omitempty`, and `omitempty` on a bool drops `false`. Since `UpsertNotification` posts the struct directly and `get` encodes it directly, `silenced` is the one field in the new DTO that can't survive a round trip.

The read side is certain. `gcx kg notifications get foo` on a config that isn't silenced prints no `silenced` key at all, so the flag people most want to eyeball is invisible in precisely the state where its absence is ambiguous, and a caller can't tell "not silenced" from "field not returned".

On the write side I'm less sure, so flagging rather than asserting: based on the note right above `UpsertSuppression` ("uses the single-item endpoint so the backend performs a read-modify-write upsert"), it looks like a POST that omits `silenced` would leave the stored value untouched. If the alert controller behaves the same way, un-silencing a config through gcx wouldn't be possible: you'd write `silenced: false`, gcx would strip it, and the config would stay silenced while printing a success line. Quick check on a dev stack would settle it. The read side holds either way.

Small note that makes this more likely to bite: the example in the command's own help text is `silenced: false`, so the documented first case is the one that gets stripped.

Dropping `,omitempty` from that one field (both tags) fixes both directions. `*bool` if you'd rather still distinguish absent from false on input.

## 4. `get` and `list` output can't be fed back into `upsert -f`

`get` encodes the bare `*AlertConfig` and `list` encodes the bare slice, but `upsert` unmarshals into `AlertConfigs`, so it needs the `alertConfigs:` wrapper. That means `gcx kg notifications get foo -o yaml > foo.yaml` followed by `gcx kg notifications upsert -f foo.yaml` fails with `no notification configs found in file`, which reads like the file is empty rather than like it needs a wrapper key.

Suppressions has the same asymmetry, but it never shipped a `get`, so nobody runs into it there. This PR is the first place in kg where `get` and `upsert` sit next to each other, which makes read, edit, apply the obvious thing to try. Cheapest fix is to accept both shapes on input (try the wrapper, fall back to a single config or a bare list), or at minimum say "expected top-level alertConfigs:" in that error.

## 5. The test gap that let 1 and 2 through

`notifications upsert` and `delete` are absent from both `TestKgSingleMutations_OutputContract` and the suppressions upsert contract test in `output_contract_test.go`, which is where every other kg mutation gets its human line, its agent-mode single document, its explicit `-o json`/`yaml`, and its exit 4 path pinned. Adding notifications to those tables is what stops this drifting back later. As it stands `notifications_test.go` asserts the prose string, so it actively pins the behaviour items 1 and 2 want to change.

## Smaller things, all take or leave

These are minor and most of them match what kg already does, so no objection if you'd rather leave them:

1. Invalid `--category` returns a bare `fmt.Errorf`, so it exits 1 where `exit-codes.md` reserves 2 for bad flag values. Practical effect is that a wrapper retrying on exit 1 will keep retrying a typo. That said, nothing in kg uses a typed usage error today, so this is really a package-wide cleanup rather than yours.
2. `list` on a tenant with no configs encodes the nil slice as `null` rather than `[]`, so `--jq '.[].name'` errors instead of returning nothing. `suppressions list` does the same, so same story.
3. No `wide` codec on `list`, so `alertLabels`, `annotations` and `managedBy` aren't reachable in table output. `output.md` §11 marks wide as required for list and get, though only one kg list registers one today.
4. `DeleteNotification` hand-rolls request / do / status / close where `doJSON` would do it. All five existing kg deletes hand-roll it the same way, so it's consistent with its neighbours and purely a consistency thing.

## On the missing `--dry-run`

Your reasoning in the description holds up: no validate endpoint means no server-validated preview, and that's a separate piece of work. One thing that is still cheap without any backend help, though. `yaml.Unmarshal` here is non-strict, so `matchlabels:` with a lowercase l, or `duration: 5m` instead of `for:`, gets silently dropped and you POST a config with no matchers at all. Paired with `silenced: true` that's a match-everything rule, and with no dry-run there's no step where a person would notice. `yaml.UnmarshalStrict` (or a decoder with known-fields on) plus a check that `name` is non-empty would catch that class locally, and it's much cheaper to add now than after someone hits it.

## Where I land

Three of the four leaves look ready to go. Items 1, 2 and 3 are what I'd want fixed before this merges, plus item 5 so they stay fixed, and 1 and 2 are mostly "copy the suppressions upsert block" so hopefully quick. Item 4 is your call on whether it belongs here or in a follow-up. Happy to be wrong on the write side of item 3 if the alert controller replaces rather than merges.

kblaschke512:
Thanks for the careful pass — this was a genuinely useful review. Addressed the four hold-for-merge items; two of the smaller notes are deferred to follow-ups (called out below). Pushed in `39ca5675`.

### 1. `upsert` output plumbing — fixed
`notificationsCreateOpts` now carries `IO cmdio.Options`, `setup` calls `BindFlags`, and RunE starts with `IO.Validate()`. The success path encodes a mutation document instead of the prose line (via a new `NotificationsUpsertTextCodec` that renders the `cmdio.BatchMutation` as the one-line human summary and lets `-o json/yaml` emit the structured doc). `-o json` works now, and agent mode emits exactly one JSON value — so the `finite` class it declares is honoured.

### 2. Partial-failure record — fixed
Lifted the suppressions block: the loop builds `cmdio.NewBatchMutation("upserted")`, and on the first failure past entry 0 it records `succeeded/failed/skipped` + the failed target, encodes the result document, and returns `gcxerrors.NewEmittedError(gcxerrors.ExitPartialFailure, …)` for exit 4. Entry 0 still takes the plain error path (no result document owed).

### 3. `silenced: false` dropped by `omitempty` — fixed, and the write side does bite
Dropped `,omitempty` from both tags on `Silenced`. You were right to flag the write side rather than assert it — I confirmed it against a dev Asserts stack: with the old tag, a `true → false` upsert left the config silenced (the `false` was stripped before the POST, and the controller kept the stored value). With the fix, un-silencing round-trips correctly, and `get` now always shows the key. Verified create → toggle → un-silence → delete end to end.

### 5. Test gap — fixed
Added `notifications delete` to `TestKgSingleMutations_OutputContract` and a full `TestKgNotificationsUpsert_OutputContract` mirroring the suppressions contract: byte-identical human summary, one batch-mutation document in agent mode and under explicit `-o json`, partial-failure exit 4 with an enumerated failure, and no result document on total failure.

### 4. `get`/`list` → `upsert -f` round-trip — deferred
Real gap, agreed. Deferring to a follow-up so this PR stays scoped to the review-blocking items; I'll make `upsert` accept the bare-config and bare-list shapes in addition to the `alertConfigs:` wrapper (and improve that error in the meantime).

### Strict unmarshal — deferred (tracking)
Good catch that this is cheap even without a validate endpoint. Also folding into the follow-up: `yaml.UnmarshalStrict` (or known-fields) + a non-empty `name` check, so a mistyped `matchlabels:`/`duration:` fails locally instead of silently POSTing a match-everything rule.

### Smaller notes
Acknowledged — the `--category` exit-2, `[]` vs `null`, and `wide`-on-`list` items are package-wide consistency cleanups that match what the rest of kg does today, so leaving them out of this PR.

kblaschke512:
Follow-up on the review discussion: the non-empty `name` check (item 2) landed here in `1857001e` — it has clear precedent in the `entities`/`relationships` upsert paths, so no divergence. The strict-YAML-parse half stays in #1098: nothing in the codebase currently uses `KnownFields`/`UnmarshalStrict` (every kg parse path is non-strict), so it's worth deciding notifications-only vs repo-wide rather than adding a one-off here. The `get`/`list` → `upsert -f` round-trip (item 3) is tracked there too.

dafydd-t (approved):

kblaschke512:
Feature writeup: https://docs.google.com/document/d/1PiGXbgdmjCWAlfKxQSNV498JeEW_-V0njaEPloXvTx8/edit

PR #1086, opened 2026-07-28
feat(profiles): add series discovery

simonswine:
## Summary
- add `gcx profiles series` for Pyroscope profile label-set discovery without a profile type
- provide table, wide, JSON, YAML, and agent output plus selector error guidance
- return complete label sets by default; promote `--label-name` projection to reduce response size and speed up high-cardinality discovery

Replaces #1075


aleks-p (commented):
What do you think about using the existing `gcx profiles labels`, but adding multiple label support that flows through the Series API (and works as today for a single label / no label)? 

Something like `gcx profiles labels <EXPR> -l service_name -l namespace`. I made a draft PR with this and other small improvements in https://github.com/grafana/gcx/pull/1090.

I think either way works, expanding the existing command might be slightly more intuitive. The dedicated command would support dumping all series but not sure if this is useful or desired.

simonswine:
> What do you think about using the existing `gcx profiles labels`, but adding multiple label support that flows through the Series API (and works as today for a single label / no label)?
> 
> Something like `gcx profiles labels <EXPR> -l service_name -l namespace`. I made a draft PR with this and other small improvements in #1090.
> 
> I think either way works, expanding the existing command might be slightly more intuitive. The dedicated command would support dumping all series but not sure if this is useful or desired.

I do prefer it slightly more in this PR's way. I also have based this mostly on the existing `gcx metrics series` call.

What is a bit arbitrary with `profiles labels` is that it lists only label values that go together in a series, that is not intuitive in my view. I would rather expect it to be more of "batch" API for listing label values independent of series.

aleks-p (dismissed):
Makes sense, I also prefer this shape after thinking about it some more!

wardbekker:
`gcx profiles series` is the only Pyroscope leaf command that the `gcx datasources pyroscope` group does not mount. `internal/datasources/providers/pyroscope.go` registers `QueryCmd`, `LabelsCmd`, `ListProfileTypesCmd`, `MetricsCmd`, and `ExemplarsCmd`. It does not register `SeriesCmd`. The built binary confirms this. `gcx datasources pyroscope --help` lists five subcommands and omits `series`.

As a result, 16 lines in `internal/datasources/pyroscope/series_discovery.go` never run:

- lines 64-75, the `Example` block
- lines 119-122, the `cmd.Annotations` map

`internal/signals/command.go:89` replaces `cmd.Example`, and lines 98 and 101 replace both annotations, when the provider descriptor sets them. The `profiles` descriptor sets all three at `internal/providers/profiles/provider.go:80`. The `profiles` group is the only group that mounts this command, so the descriptor values always apply. `gcx profiles series --help` prints the two descriptor examples, not the three examples in the file.

There are two ways to correct this:

1. Add `pyroscope.SeriesCmd` to `internal/datasources/providers/pyroscope.go`. This gives `series` the same two mounts as its five sibling commands. The in-file `Example` and annotations then become active. Change the in-file hint to the `gcx datasources pyroscope series ...` form that every sibling uses.
2. Delete the unreachable `Example` block and the `cmd.Annotations` map, and keep the command on the `profiles` mount alone.

Two smaller notes:

- The `Example` block mixes tab indentation and space indentation between the three examples. Option 1 makes this visible in the help output and in the generated reference doc.
- `gcx profiles series` is absent from `internal/agent/command_annotations.go`. Lines 500-504 list every other `profiles` leaf command. This changes no behaviour, because `ApplyAnnotations` at line 657 fills gaps only. It is a consistency item.

Everything else is correct. All CI checks pass. The tests and the build also pass on a local checkout of this branch.

dafydd-t (changes requested):
I think if we can also add the command to `gcx datasources pyroscope series`, we're good to go. 


Thinking about ways to guard against this kind of thing happening for the MTLP datasources - see https://github.com/grafana/gcx/issues/1218

dafydd-t:
I would have expected this to work? am i misunderstanding something?

```
./bin/gcx datasources pyroscope series -d <UID>
Error: Invalid Pyroscope profile series query
│
│ Authentication to data source failed
│
├─ Suggestions:
│
│ • Run 'gcx profiles series --help' for usage and examples
│
├─ Learn more:
│
│ https://grafana.com/docs/pyroscope/latest/view-and-analyze-profile-data.md
│
└─
```

This was on an oauthed stack

simonswine:
@dafydd-t hm this works for me. That said it times out in big instances without a matcher but with an error:


```
./bin/gcx datasources pyroscope series
Error: Pyroscope profile series query failed (HTTP 502)
│
│ Pyroscope returned HTTP 502
│
├─ Suggestions:
│
│ • Run 'gcx profiles series --help' for usage and examples
│
├─ Learn more:
│
│ https://grafana.com/docs/pyroscope/latest/view-and-analyze-profile-data.md
│
└─
```


```
 ./bin/gcx datasources pyroscope series --match '{container="prometheus"}'
┌────────┬────────────────────
│ SERIES │ LABELS                                                                                                                                                                                                                                                                                                                                                                                                                        │
├────────┼────────────────────
```

> I would have expected this to work? am i misunderstanding something?
> 
> ```
> ./bin/gcx datasources pyroscope series -d <UID>
> 
> This was on an oauthed stack

dafydd-t (approved):

PR #1094, opened 2026-07-29
feat(telemetry): break down api command usage by route, method, and datasource type

tolzhabayev:
Usage events for `gcx api` only record `command: api`, so we can't tell what the passthrough gets used for. We want that breakdown to decide which first-class commands and datasource integrations to build next.

This adds three fields to the usage event, set only for the api command:

- `api_method`: the HTTP verb, from the fixed method list.
- `api_route`: the path matched against a built-in table of known Grafana API route templates. Variable segments become placeholders (`/api/dashboards/uid/{uid}`), the query string is dropped before matching, and unmatched paths are sent as `other`.
- `api_datasource_types`: for datasource query routes only, the `queries[].datasource.type` values. A type is sent only when it's on a fixed list of Grafana-authored plugin IDs (core datasource types plus the Grafana-published catalog datasources). Everything else, including third-party and private plugins, becomes `other`.

Every recorded value comes from a finite vocabulary baked into the binary. There are no pattern rules: a value either matches a listed entry or is recorded as `other`. gcx never stores the raw path or body, parses the body only to read the datasource type field, and never sends the datasource UID or name, the query text, or the time range. `GCX_TELEMETRY=log` prints exactly what an invocation would report, and the usage statistics doc has a new section that spells out the contract.

The tables are curated, not exhaustive. When the share of `other` grows we extend a table, we don't loosen the filtering.

Receiver side: the usage-stats flattener and schema gain the three columns in grafana/usage-stats#154. Merge order doesn't matter (the receiver keeps the raw JSON either way), the columns fill once both are deployed.


cmd/gcx/api/command.go
```diff
@@ -135,39 +143,46 @@ func Command() *cobra.Command {
...
 		b, err := os.ReadFile(data[1:])
 		if err != nil {
 			return nil, fmt.Errorf("failed to read file: %w", err)
 		}
-		return bytes.NewReader(b), nil
+		return b, nil
 	}
-	return strings.NewReader(data), nil
+	return []byte(data), nil
 }
```
dafydd-t, line 165: the diff on `resolveBody` seems unrelated to this PR?

dafydd-t: same with `doRequest` i think, unless i am missing something

internal/telemetry/api.go
```diff
@@ -0,0 +1,447 @@
...
+
+// grafanaDatasourceTypes lists the Grafana-published datasource plugin IDs
+// from the public catalog. Generated 2026-07-29 from
+// https://grafana.com/api/plugins (typeCode datasource, orgSlug grafana).
+// A finite list rather than a grafana- prefix rule so no user-typed value
+// can ever pass: a plugin missing here is undercounted as "other" until the
+// list is regenerated, which is the safe direction to fail.
+//
+//nolint:gochecknoglobals
+var grafanaDatasourceTypes = map[string]bool{
```
dafydd-t: We're adding a bit of maintenance burden here aren't we.. not sure about an alternative. If we keep it, this definition list should be somewhere more visible (in its own file, or at least at the top of this file?

internal/telemetry/api.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+	return otherValue
+}
+
+// maxQueryBodyBytes caps the body size datasourceTypes will parse. Query
+// request bodies are small; anything larger is skipped rather than parsed.
+const maxQueryBodyBytes = 1 << 20
+
+// maxDatasourceTypes caps how many distinct types one event can carry.
+const maxDatasourceTypes = 8
```
dafydd-t: how did you choose 8?

dafydd-t:
claude:

 reviewed this alongside a separate security pass. the privacy design holds up well: I traced every value that can reach the event and they're all provably drawn from compile-time constants, then fuzzed `normalizeAPIRoute` (3.2M execs) and `datasourceTypes` (4.3M) against the finite expected-output set with zero escapes. uids, names, query text, query strings and the `secureJsonData` body case all stay out. replacing the `grafana-` prefix rule with the finite catalog list in 6b94d9c4 was the right call, and `TestRecordAPIRequest` asserting the whole struct against a body containing `SELECT email, full_name, address FROM customers` is a good way to pin it. **no security findings.**

two things I think need addressing before merge, both signal quality rather than privacy:

## 1. `{rest}` never matches the collection root, so the highest-value buckets record `other`

`matchRouteTemplate` requires at least one segment *after* the prefix (`internal/telemetry/api.go:177`):

```go
if t == "{rest}" && i == len(tmpl)-1 {
    return len(path) > i
}
```

there are no bare-root entries for the `{rest}` routes, so every list/create endpoint falls through to `other`. verified against a build:

| path | recorded |
|---|---|
| `/api/playlists` | `other` |
| `/api/snapshots` | `other` |
| `/api/library-elements` | `other` |
| `POST /api/teams` | `other` |
| `/api/serviceaccounts` | `other` |
| `/api/users`, `/api/orgs` | `other` |

those are the list and create endpoints, which are the most common thing anyone does through a raw passthrough and exactly the signal this PR is meant to capture. `/api/annotations` and `/api/datasources` are fine because they have explicit bare roots, which is what makes the others look like an oversight rather than a decision.

either add the bare roots ahead of the `{rest}` lines, or change `{rest}` to match zero-or-more (`return true`). the second is a one-liner but it also makes `/api/teams` and `/api/teams/foo/bar` indistinguishable, so probably the former.

three smaller ones in the same table, worth fixing in the same pass since it's all first-match ordering:

- `POST /api/annotations/graphite` records `/api/annotations/{id}` (`api.go:106`) because `{id}` swallows a real literal sibling. inflates the per-id bucket
- the legacy numeric datasource proxy isn't covered: only `/api/datasources/proxy/uid/{uid}/{rest}` is listed (`api.go:101`), so `/api/datasources/proxy/1/api/v1/query` records `other`. still widely used, and a strong "build a datasource query command" signal
- `/api/v1/provisioning/{rest}` (`api.go:110`) collapses the whole provisioning surface into one bucket, so `alert-rules` and `contact-points` are indistinguishable. both second segments are fixed API literals with no user data, so they're safe to keep, and "should we build first-class contact-point commands?" is exactly the question the data can't answer as it stands. same fix shape as `/api/teams/search` sitting ahead of its catch-all

## 2. confirm the receiver-side flattener knows about the three new fields

`event_test.go` says the field inventory "must be reflected in the usage-stats flattener and the docs". docs are updated here, the flattener isn't in this repo. `Export` is deliberately fire-and-forget and doesn't check the response status, which is the right call for telemetry, but it does mean that if the receiver drops or rejects unknown keys, all three fields land nowhere and nothing surfaces it. the `Export` doc comment says the wire contract is pinned by the receiver's own tests, so I assume there's a corresponding change somewhere - could you link it in the PR body, or confirm no receiver change is needed? otherwise the first anyone notices is when the column comes back empty for every row.

## smaller stuff, take or leave

- the privacy sweep test only covers the `apiRouteTemplates` branch, not `k8sAPIRoute`, which is the branch that actually appends caller-supplied segments. it's safe today (I fuzzed both), but a `FuzzNormalizeAPIRoute` asserting the output never contains an input segment absent from a constant would stop a future change quietly breaking it
- nothing asserts the api command calls `RecordAPIRequest`, and `buildUsageEvent` has no test at all, so deleting the wiring keeps the suite green. I checked it by hand with `GCX_TELEMETRY=log`, which is the bit that should be a test
- empty `api_datasource_types` means four different things (not a query route, unparseable body, oversized body, no `queries` key) and `omitempty` makes them identical on the wire. emitting `other` when the route *is* a query route but extraction failed would let you measure whether extraction works in the wild
- the `internal/telemetry` line in `docs/architecture/project-structure.md` doesn't mention api-request sanitization, which is now the most privacy-sensitive thing in the package
- verb list is duplicated between `apiOpts.Validate` and `knownHTTPMethod`. the comment frames the second as a last line of defence which is fair, but if a verb is added to `Validate` only, telemetry silently drops the field for it

nice work on the allowlist approach generally, the undercounting-is-the-safe-direction reasoning in the comments made this much easier to review.

internal/telemetry/api.go
```diff
@@ -0,0 +1,447 @@
...
+	"dashboards": true, "folders": true, "search": true,
+	"query": true, "queryconvert": true,
+	"datasources": true, "connections": true,
+	"playlists": true, "shorturls": true, "preferences": true, "stars": true,
+	"alertrules": true, "recordingrules": true,
+	"receivers": true, "routingtrees": true, "templategroups": true, "timeintervals": true,
+	"users": true, "teams": true, "serviceaccounts": true, "ssosettings": true,
+	"securevalues": true, "keepers": true,
+	"repositories": true, "jobs": true,
+}
```
dafydd-t: can we do one per line, alphabetically, please?

dafydd-t: same request for other maps please

cmd/gcx/api/command.go
tolzhabayev: It's needed by the telemetry call. `RecordAPIRequest` inspects the body for `queries[].datasource.type`, and with an `io.Reader` that would consume the stream before the real request (stdin isn't re-readable at all). So `resolveBody` now materializes bytes and `doRequest` wraps them at send time. The reorder above `LoadGrafanaConfig` is so failed invocations still record their route.

internal/telemetry/api.go
tolzhabayev: Agreed on the burden, I don't see an alternative that doesn't reintroduce pattern rules. Moved all the vocabulary tables into their own file, `internal/telemetry/api_vocabulary.go`, with a header comment saying these tables are the privacy boundary and how to grow them.

internal/telemetry/api.go
tolzhabayev: Arbitrary. Most requests name one type and mixed panels would probably have a handful, the cap is to prevent it growing too much on edge cases

internal/telemetry/api.go
tolzhabayev: yep, done

tolzhabayev:
> claude:
> 
> reviewed this alongside a separate security pass. the privacy design holds up well: I traced every value that can reach the event and they're all provably drawn from compile-time constants, then fuzzed `normalizeAPIRoute` (3.2M execs) and `datasourceTypes` (4.3M) against the finite expected-output set with zero escapes. uids, names, query text, query strings and the `secureJsonData` body case all stay out. replacing the `grafana-` prefix rule with the finite catalog list in [6b94d9c](https://github.com/grafana/gcx/commit/6b94d9c4a42b7cfa791aaf80107d2ed851ddd88a) was the right call, and `TestRecordAPIRequest` asserting the whole struct against a body containing `SELECT email, full_name, address FROM customers` is a good way to pin it. **no security findings.**
> 
> two things I think need addressing before merge, both signal quality rather than privacy:
> 
> ## 1. `{rest}` never matches the collection root, so the highest-value buckets record `other`
> `matchRouteTemplate` requires at least one segment _after_ the prefix (`internal/telemetry/api.go:177`):
> 
> ```go
> if t == "{rest}" && i == len(tmpl)-1 {
>     return len(path) > i
> }
> ```
> 
> there are no bare-root entries for the `{rest}` routes, so every list/create endpoint falls through to `other`. verified against a build:
> 
> path	recorded
> `/api/playlists`	`other`
> `/api/snapshots`	`other`
> `/api/library-elements`	`other`
> `POST /api/teams`	`other`
> `/api/serviceaccounts`	`other`
> `/api/users`, `/api/orgs`	`other`
> those are the list and create endpoints, which are the most common thing anyone does through a raw passthrough and exactly the signal this PR is meant to capture. `/api/annotations` and `/api/datasources` are fine because they have explicit bare roots, which is what makes the others look like an oversight rather than a decision.
> 
> either add the bare roots ahead of the `{rest}` lines, or change `{rest}` to match zero-or-more (`return true`). the second is a one-liner but it also makes `/api/teams` and `/api/teams/foo/bar` indistinguishable, so probably the former.
> 
> three smaller ones in the same table, worth fixing in the same pass since it's all first-match ordering:
> 
> * `POST /api/annotations/graphite` records `/api/annotations/{id}` (`api.go:106`) because `{id}` swallows a real literal sibling. inflates the per-id bucket
> * the legacy numeric datasource proxy isn't covered: only `/api/datasources/proxy/uid/{uid}/{rest}` is listed (`api.go:101`), so `/api/datasources/proxy/1/api/v1/query` records `other`. still widely used, and a strong "build a datasource query command" signal
> * `/api/v1/provisioning/{rest}` (`api.go:110`) collapses the whole provisioning surface into one bucket, so `alert-rules` and `contact-points` are indistinguishable. both second segments are fixed API literals with no user data, so they're safe to keep, and "should we build first-class contact-point commands?" is exactly the question the data can't answer as it stands. same fix shape as `/api/teams/search` sitting ahead of its catch-all

On the route table: added the bare collection roots (`users`, `orgs`, `teams`, `serviceaccounts`, `library-elements`, `playlists`, `snapshots`) rather than making `{rest}` zero-or-more, for the reason you gave. Also added `/api/annotations/graphite` ahead of `{id},` the legacy numeric proxy route, and split provisioning into `alert-rules`, `contact-points`, `policies`, `mute-timings` and `templates` with the old catch-all kept last. All pinned in tests, and I replayed your table against the new binary with `GCX_TELEMETRY=log`, every row now records its proper template.

> ## 2. confirm the receiver-side flattener knows about the three new fields
> `event_test.go` says the field inventory "must be reflected in the usage-stats flattener and the docs". docs are updated here, the flattener isn't in this repo. `Export` is deliberately fire-and-forget and doesn't check the response status, which is the right call for telemetry, but it does mean that if the receiver drops or rejects unknown keys, all three fields land nowhere and nothing surfaces it. the `Export` doc comment says the wire contract is pinned by the receiver's own tests, so I assume there's a corresponding change somewhere - could you link it in the PR body, or confirm no receiver change is needed? otherwise the first anyone notices is when the column comes back empty for every row.

On the receiver: the flattener change exists, it's [grafana/usage-stats#154](https://github.com/grafana/usage-stats/issues/154) (schema, flattener, tests). Linked in the description now. Merge order doesn't matter since the receiver keeps the raw JSON either way, the columns fill once both are deployed.

> ## smaller stuff, take or leave
> * the privacy sweep test only covers the `apiRouteTemplates` branch, not `k8sAPIRoute`, which is the branch that actually appends caller-supplied segments. it's safe today (I fuzzed both), but a `FuzzNormalizeAPIRoute` asserting the output never contains an input segment absent from a constant would stop a future change quietly breaking it
> * nothing asserts the api command calls `RecordAPIRequest`, and `buildUsageEvent` has no test at all, so deleting the wiring keeps the suite green. I checked it by hand with `GCX_TELEMETRY=log`, which is the bit that should be a test
> * empty `api_datasource_types` means four different things (not a query route, unparseable body, oversized body, no `queries` key) and `omitempty` makes them identical on the wire. emitting `other` when the route _is_ a query route but extraction failed would let you measure whether extraction works in the wild
> * the `internal/telemetry` line in `docs/architecture/project-structure.md` doesn't mention api-request sanitization, which is now the most privacy-sensitive thing in the package
> * verb list is duplicated between `apiOpts.Validate` and `knownHTTPMethod`. the comment frames the second as a last line of defence which is fair, but if a verb is added to `Validate` only, telemetry silently drops the field for it
> 
> nice work on the allowlist approach generally, the undercounting-is-the-safe-direction reasoning in the comments made this much easier to review.

Added `FuzzNormalizeAPIRoute` covering the `k8sAPIRoute` branch (every segment of a non-other result must be a placeholder or known constant, ran 6.7M execs clean), wiring tests so deleting `RecordAPIRequest` or the event mapping now fails the suite, `api_datasource_types` records other when the route is a query route but extraction fails (empty now always means "not a query route"), the project-structure line mentions the sanitization, and the verb list is shared between flag validation and telemetry so they can't drift.

docs/sources/anonymous-usage-statistics.md
```diff
@@ -61,6 +61,18 @@ When the invocation fails to parse, these additional fields are set. They captur
 | `parse_error_nearest` | The nearest real command or flag name, if one is close. | `search` |
 | `parse_error_distance` | The edit distance to the nearest real name, or `-1` if there is no near match. | `2` |
 
+### The api command
+
+`gcx api` is a raw passthrough: the interesting usage signal (which endpoint, which method) lives in its argument values, which are never sent as-is. Instead, three additional fields carry derived values, each filtered through a fixed vocabulary built into the binary before anything is recorded. We use them to learn which endpoints and datasources people reach for through the raw passthrough, so we know which first-class `gcx` commands to build next.
```
dafydd-t:
```suggestion
The `gcx api` command is unique because it allows users to send an arbitrary request to the Grafana API. To get useful usage information (which endpoints, and methods are used) we need to inspect the argument values. This is an exception to our claim of not sending argument values as part of usage telemetry. We mitigate the risk of sending sensitive or identifying information by filtering argument values through a fixed vocabulary built into the binary, before anything is recorded. The filtered argument values are only used to learn which endpoints and queries our users use the `api` command for, so we can spot where we can add value with new gcx commands in the future.
```

(will need formatting I guess)

dafydd-t (changes requested):
just an edit to make the documentation more human friendly, then LGTM

tolzhabayev:
@dafydd-t hey, please take a look if there is more to change

dafydd-t (approved):

PR #1097, opened 2026-07-29
add data-sources codeowners for internal/datasources

dafydd-t:
Adds `@grafana/data-sources` as code owners for the per-datasource directories under `internal/datasources/` that were only covered by the `*` fallback: `athena`, `clickhouse`, `cloudwatch`, `infinity`, `influxdb`. Dirs with existing team owners (loki, prometheus, pyroscope, tempo) are untouched, and the `providers`/`query` plumbing packages stay with gcx.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -33,9 +33,7 @@
 /internal/datasources/influxdb/ @grafana/grafana-gcx @grafana/data-sources
 /internal/datasources/loki/ @grafana/grafana-gcx @grafana/loki-team
 /internal/datasources/prometheus/ @grafana/grafana-gcx @grafana/cortex-team
-/internal/datasources/providers/ @grafana/grafana-gcx @grafana/data-sources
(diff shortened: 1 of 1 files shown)


wardbekker (approved):
Approved. The paths match the five `internal/datasources/` directories that only had the `*` fallback, the existing team-owned directories stay untouched, and the `providers`/`query` plumbing packages correctly stay with gcx.

One follow-up: GitHub's CODEOWNERS validation currently reports `@grafana/data-sources` as an unknown owner on all five lines, because the team does not have write access to this repository:

```
Unknown owner on line 29: make sure the team @grafana/data-sources exists,
is publicly visible, and has write access to the repository
```

The lines are harmless as-is, because `@grafana/grafana-gcx` stays a valid owner on each one. But the data-sources team will not get review requests until a repository admin grants the team write access. Please ask an admin to add the team, then the entries take effect.

PR #1099, opened 2026-07-29
docs(skills): pick eval-rule selector by agent shape, not by habit

soniaAguilarPeiron:
## What

An online eval rule with `selector: user_visible_turn` scores **nothing** on a multi-agent DAG / programmatic pipeline. Those generations carry no user-visible-turn flag (`is_user_visible` is null), so the selector matches zero generations — the rule looks live on the stack but silently never fires, and the conversation shows no eval scores.

Verified live: a DAG agent's rule with `user_visible_turn` produced `eval_summary.total_scores = 0`; switching the same rule to `all_assistant_generations` produced scores immediately (2 per planner generation).

## Why the skill missed it

The skill already mentioned `all_assistant_generations`, but only for the narrow case "agent has generations but **no conversation**." A DAG that *does* group into a conversation fell through that carve-out, so the skill defaulted to `user_visible_turn` and produced a dead rule.

## Fix (skill guidance only)

Generalize: pick the selector from **how the agent produces generations**, not by habit.
- **DAG / pipeline nodes, single-shot agents** → `all_assistant_generations`, scoped with `match.agent_name`.
- **Genuine chat/assistant agents** (a turn is what the user sees) → `user_visible_turn`.
- **Verify:** after creating the rule, run the agent once and confirm `eval_summary.total_scores > 0`. Zero scores on a matching agent almost always means the selector is wrong for the agent shape.

Also annotates the recommendation table's quality-drift row with the same selector choice.

### DIFF
--- a/claude-plugin/skills/agento11y-prod-setup/SKILL.md
+++ b/claude-plugin/skills/agento11y-prod-setup/SKILL.md
@@ -300,6 +300,15 @@ type** — getting this wrong yields a guard that runs but silently does nothing
   (Postflight `redact` is not useless in general — some runtimes explicitly consume
   `transformed_input.output` to rewrite **tool-call payloads/arguments** postflight — but that is a
   narrow tool-arg case, not final-response scrubbing.)
+  - **Disambiguation — decide by where the sensitive value ENTERS, not by where it's most visible.**
+    An agent can both *receive* a secret in its input AND *echo/generate* one in its output; don't
+    let the more eye-catching output occurrence pull you to postflight. If the secret enters through
+    the input (e.g. a key pasted into the prompt / incident text / a config blob), the guard that
+    actually protects it is **`redact` preflight on the input** — that scrubs it before the model,
+    the logs, and every downstream node see it. A postflight guard cannot undo a secret that already
+    entered upstream. Add a postflight **detector** on top only if the model *also* independently
+    produces secrets in its final text; that is a second, separate concern, not a replacement for the
+    preflight redact.
 - **`tool_filter` → `postflight`.** It inspects the tool calls the model wants to make.
 - **`evaluator_ids` (detector) → phase follows what you evaluate.** `preflight` to judge the
   **input** (block a bad request before spending tokens), `postflight` to judge the **output**
--- a/claude-plugin/skills/agento11y-prod-setup/SKILL.md
+++ b/claude-plugin/skills/agento11y-prod-setup/SKILL.md
@@ -255,18 +255,13 @@ enabled at a low `sample_rate`.
 **Guards** — the shape the `agento11y` skill omits (`gcx agento11y guards create -f guard.yaml`; the
 resource `Kind` is `HookRule`).
 
-> **Hard gate: do not draft a guard until you have captured the exact accepted create-file shape
-> from the current `gcx` version.** Run `gcx agento11y guards create --help` and inspect a real
-> definition (`gcx agento11y guards get <id> -o yaml`, or `guards list -o yaml`); if none exists, get
-> the schema from `--help` / the resource definition. Only then write a guard file.
-
-This skill (not the `agento11y` skill) carries the guard shape, and field names/nesting can drift by
-`gcx` version and by stack — so the shape you capture from a real stored guard
-(`gcx agento11y guards get <existing-id> -o yaml`) is the source of truth, not the snippet below.
-The snippet is **illustrative only**. A guard drives its decision from one of three (mutually
+This skill (not the `agento11y` skill) carries the guard shape — draft the guard file directly from
+it, no schema-discovery step needed. A guard drives its decision from one of three (mutually
 exclusive) shapes — `evaluator_ids`, `redact`, or `tool_filter` — plus `action_on_fail`, `phase`,
 `priority`, `selector`. Use the `redact` schema exactly as the callout above shows (`{id, regex}`,
-no `replacement`). On create the server fills defaults you don't set — notably `selector: all`,
+no `replacement`). If the server ever rejects a field, its 400 names the correct one (e.g.
+`redact.patterns`) — fix and retry then, rather than probing up front. On create the server fills
+defaults you don't set — notably `selector: all`,
 `phase: preflight`, and `short_circuit: false` — so a `guards get -o yaml` right after create
 shows more fields than you sent; that's expected, not drift. A guard is drafted in **warn** (and
 `enabled: false` if the server honors it — it may not; see Step 5):
--- a/claude-plugin/skills/agento11y-prod-setup/SKILL.md
+++ b/claude-plugin/skills/agento11y-prod-setup/SKILL.md
@@ -128,18 +128,26 @@ from the traffic; read it from the code.
    (add `status = "error"`, time windows, `tool.name`, `eval.passed = false`) and
    `gcx agento11y generations get <id>` for detail. Look for long tool loops, over-refusals, PII
    echoed back, off-topic drift, malformed outputs, error clusters.
-   - **Pick the selector from how the agent produces generations, not by habit.**
-     `user_visible_turn` only matches generations flagged as a user-facing turn; a
-     **multi-agent DAG / programmatic pipeline** (fan-out/fan-in, internal nodes, one conversation
-     per run) produces generations with **no user-visible-turn flag** (`is_user_visible` is null),
-     so a `user_visible_turn` rule matches **nothing and scores zero** — the rule looks live but
-     silently never fires. For these agents (and for single-shot agents that have generations but no
-     conversation at all), use **`selector: all_assistant_generations`** and scope with
-     `match.agent_name` to the node you care about. Reserve `user_visible_turn` for genuine
-     chat/assistant agents where a turn is what the user sees. Verify before trusting it: after
-     creating the rule, run the agent once and confirm scores appear (`conversations search` →
-     `eval_summary.total_scores > 0`); zero scores on a matching agent almost always means the
-     selector is wrong for this agent shape.
+   - **Some agents have generations but no conversation** (e.g. single-shot agents whose spans
+     aren't grouped into a conversation). If `conversations search` is empty but `agents get` shows a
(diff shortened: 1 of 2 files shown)


claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -128,11 +128,18 @@ from the traffic; read it from the code.
...
    echoed back, off-topic drift, malformed outputs, error clusters.
-   - **Some agents have generations but no conversations** (e.g. single-shot agents whose spans
-     aren't grouped into a conversation). If `conversations search` returns empty but `agents get`
-     shows a non-zero `generation_count`, don't stop — sample generations directly
-     (`gcx agento11y generations get <id>`) and use `selector: all_assistant_generations` for rules
-     rather than a conversation-scoped one.
+   - **Pick the selector from how the agent produces generations, not by habit.**
+     `user_visible_turn` only matches generations flagged as a user-facing turn; a
+     **multi-agent DAG / programmatic pipeline** (fan-out/fan-in, internal nodes, one conversation
+     per run) produces generations with **no user-visible-turn flag** (`is_user_visible` is null),
```
awsome-o: nit: I grepped for `is_user_visible` and it doesn't seem to appear anywhere else in the repo, only `user_visible_turn` as a [selector value](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/internal/providers/aio11y/eval/types.go#L86), so as far as I can tell there's no `gcx agento11y` command that lets the agent actually look at it. Reads fine as background, but an agent might go hunting for a way to check it. Could be worth leading with the symptom (rule matches traffic, scores stay at zero) and keeping the field name as the explanation behind it.

claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -128,11 +128,18 @@ from the traffic; read it from the code.
...
+     so a `user_visible_turn` rule matches **nothing and scores zero** — the rule looks live but
+     silently never fires. For these agents (and for single-shot agents that have generations but no
```
awsome-o: Both shapes get the same selector advice here, which I think is right. What I'd flag is that the verification just below points at `conversations search`, and for an agent with no conversation at all there's nothing there to search.

The old bullet's "don't stop, sample generations directly" clause also went away in the rewrite, and that part felt more like evidence gathering than selector advice. Without it I'd worry the evidence bar at 144 to 150 makes the skill conclude "essentially no traffic, wrong skill" for an agent that actually has plenty of generations. Maybe keep one clause of it?

claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -128,11 +128,18 @@ from the traffic; read it from the code.
...
+     conversation at all), use **`selector: all_assistant_generations`** and scope with
+     `match.agent_name` to the node you care about. Reserve `user_visible_turn` for genuine
+     chat/assistant agents where a turn is what the user sees. Verify before trusting it: after
+     creating the rule, run the agent once and confirm scores appear (`conversations search` →
```
awsome-o: Based on line 79 the skill tells the agent to start rules around `sample_rate: 0.1`, so a single run would score nothing most of the time, and this check could then read a perfectly good selector as broken. Line 79 also warns that 1.0 costs real money, so verifying at 1.0 isn't quite free either. Maybe phrase it as confirming over enough sampled traffic, or as a deliberate temporary bump that gets dropped again afterwards.

Separately, [`EvalSummary` is a pointer with `omitempty`](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/internal/providers/aio11y/conversations/types.go#L35) on the conversation, so when nothing scored the field tends to be missing from the output rather than showing up as `0`. "absent or 0" might be safer wording.

claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -168,7 +175,7 @@ watch or to stop — and remember a guard is dead config until the agent is wire
 
 | If, in code or traffic, the agent… | Surface | Shape (prefer a predefined template) |
 | --- | --- | --- |
-| gives answers whose quality can drift | online **rule** | fork `template.helpfulness` / `template.relevance` (`llm_judge`) over `user_visible_turn` |
+| gives answers whose quality can drift | online **rule** | fork `template.helpfulness` / `template.relevance` (`llm_judge`); selector `user_visible_turn` for a chat agent, `all_assistant_generations` for a DAG/pipeline node (see Step 1) |
```
awsome-o: nit: good that this row spells it out. The rows below for groundedness, `json_valid` and public facing text are still quiet about the selector, though the refusal and drift row already says `all_assistant_generations`. One sentence above the table, something like pick the selector per Step 1 for every online rule row, might cover all of them, and it would keep this cell from being the longest thing in the table.

claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -248,18 +255,13 @@ enabled at a low `sample_rate`.
...
-> **Hard gate: do not draft a guard until you have captured the exact accepted create-file shape
-> from the current `gcx` version.** Run `gcx agento11y guards create --help` and inspect a real
-> definition (`gcx agento11y guards get <id> -o yaml`, or `guards list -o yaml`); if none exists, get
-> the schema from `--help` / the resource definition. Only then write a guard file.
-
-This skill (not the `agento11y` skill) carries the guard shape, and field names/nesting can drift by
-`gcx` version and by stack — so the shape you capture from a real stored guard
-(`gcx agento11y guards get <existing-id> -o yaml`) is the source of truth, not the snippet below.
-The snippet is **illustrative only**. A guard drives its decision from one of three (mutually
+This skill (not the `agento11y` skill) carries the guard shape — draft the guard file directly from
```
awsome-o, line 270: Agree with dropping the gate given this skill carries the shape itself. One leftover though: the YAML block at [line 270](https://github.com/grafana/gcx/blob/f2cb93813744ba0dc2673cc9d85265e82e76a150/claude-plugin/skills/agento11y-prod-setup/SKILL.md#L270) still opens with `# ILLUSTRATIVE, confirm fields with gcx agento11y guards create --help before use`, which says the opposite of this sentence. Since that comment travels inside the snippet the agent copies, my guess is it quietly wins over the prose up here, so it probably wants to go as well.

claude-plugin/skills/agento11y-prod-setup/SKILL.md
```diff
@@ -248,18 +255,13 @@ enabled at a low `sample_rate`.
...
+it, no schema-discovery step needed. A guard drives its decision from one of three (mutually
 exclusive) shapes — `evaluator_ids`, `redact`, or `tool_filter` — plus `action_on_fail`, `phase`,
 `priority`, `selector`. Use the `redact` schema exactly as the callout above shows (`{id, regex}`,
-no `replacement`). On create the server fills defaults you don't set — notably `selector: all`,
+no `replacement`). If the server ever rejects a field, its 400 names the correct one (e.g.
+`redact.patterns`) — fix and retry then, rather than probing up front. On create the server fills
+defaults you don't set — notably `selector: all`,
 `phase: preflight`, and `short_circuit: false` — so a `guards get -o yaml` right after create
```
awsome-o: Based on the [`redact` callout at line 195](https://github.com/grafana/gcx/blob/f2cb93813744ba0dc2673cc9d85265e82e76a150/claude-plugin/skills/agento11y-prod-setup/SKILL.md#L195), the 400 actually names the wrong field (`400 unknown field "replacement"`) rather than the correct one, so an agent could sit waiting for a corrected name that never arrives.

The suggestion keeps the reassurance without promising more than the server gives, and picks up the stray wrap after "notably `selector: all`," while it's there.

```suggestion
no `replacement`). If the server ever rejects a field, the 400 names the offending field, so a bad
shape surfaces at create time rather than needing a probe up front. On create the server fills
defaults you don't set — notably `selector: all`, `phase: preflight`, and `short_circuit: false` —
so a `guards get -o yaml` right after create
```

awsome-o (commented):
Read this a couple of times and I think the call is right, so nothing here is blocking. Based on [`eval/types.go:86`](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/internal/providers/aio11y/eval/types.go#L86) the selector really is one of five values, so choosing it from the agent's shape instead of by habit makes sense to me, and it lines up with the evidence in the description.

One thing that doesn't fit on a line because it lives in another file: three of the five templates in [`agento11y/references/rule-templates.md`](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/claude-plugin/skills/agento11y/references/rule-templates.md) hardcode `selector: user_visible_turn` ([L10](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/claude-plugin/skills/agento11y/references/rule-templates.md#L10), [L22](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/claude-plugin/skills/agento11y/references/rule-templates.md#L22), [L65](https://github.com/grafana/gcx/blob/a2ff8c5bf84a5d102693bb44387dc6621015f662/claude-plugin/skills/agento11y/references/rule-templates.md#L65)), and this skill at [line 210](https://github.com/grafana/gcx/blob/f2cb93813744ba0dc2673cc9d85265e82e76a150/claude-plugin/skills/agento11y-prod-setup/SKILL.md#L210) hands rule mechanics over to the `agento11y` skill. So an agent that forks a template as told could quietly bring back the same dead rule this PR is fixing. A note on those templates, or a line here where the skill defers, would close that. Same thought for the rule follow through at [line 445](https://github.com/grafana/gcx/blob/f2cb93813744ba0dc2673cc9d85265e82e76a150/claude-plugin/skills/agento11y-prod-setup/SKILL.md#L445), which still only mentions raising the sample rate and could point back at this selector reasoning.

awsome-o (approved):

PR #1106, opened 2026-07-29
feat(datasources): add BigQuery datasource support

ktw4071:
## Summary

Adds a **BigQuery** datasource type to gcx, part of the datasource-coverage expansion initiative. Users can now generate/execute BigQuery SQL and browse schema directly from the CLI.

Commands (under `gcx datasources bigquery`):
- `query [EXPR]` — execute GoogleSQL, with `--limit` enforcement and `--share-link`/`--open` Explore links
- `list-datasets` — datasets in a project (`--project` optional; defaults to the datasource's default project)
- `list-tables --dataset D` — tables in a dataset
- `describe-table TABLE --dataset D` — column name/type/nullability

Also reachable via the auto-detecting `gcx datasources query <uid> <sql>`.

## Design

- **Follows the ClickHouse pattern**: query goes through the unified datasource query API (`internal/query/grafanaquery`, `/apis/query.grafana.app` with `/api/ds/query` fallback); responses parse and format via the shared `internal/query/sql` helpers.
- **Metadata via `INFORMATION_SCHEMA`** (SCHEMATA/TABLES/COLUMNS), not the plugin's resource proxy — those routes (`/datasets`, `/tables`, …) return HTTP 500 through the current API transport, whereas `INFORMATION_SCHEMA` queries work over the proven query path.
- **No epoch-ms timestamp heuristics**: genuine `TIMESTAMP` columns arrive as dataframe wire-type `time` and already render as RFC3339 through the shared formatter; `DATETIME`/`DATE` come back as readable strings. A plain `INT64` holding epoch-ms stays an integer — guessing would corrupt legitimate integer columns.
- **Identifier safety**: project IDs may contain hyphens (validated separately from dataset/table names); qualified paths are backtick-quoted after validation.

## Sample Output
```
gcx datasources bigquery list-datasets
gcx datasources bigquery list-tables --dataset grafana_example_dataset
gcx datasources bigquery describe-table WORLD_DATA --dataset grafana_example_dataset
gcx datasources bigquery query "SELECT country_name, ROUND(AVG(co2),1) AS avg_co2, SUM(gdp) AS total_gdp FROM \`partner-datasources.grafana_example_dataset.WORLD_DATA\` GROUP BY country_name ORDER BY total_gdp DESC LIMIT 5"
```
image
image

## Testing

- [x] Unit tests for the query client (request construction, response parsing, typed errors, `/api/ds/query` fallback), types (validation, `INFORMATION_SCHEMA` prefix, limit enforcement, row parsers), formatters, and the Explore URL builder. `NormalizeKind` regression case added.
- [x] `GCX_AGENT_MODE=false mise run all` passes (validate-skills, lint 0 issues, tests, build, docs).
- [x] **Verified live** against a JWT-auth BigQuery datasource on dev instance:

Part of the datasource-coverage effort (https://github.com/grafana/plugins-private/issues/4127).


dafydd-t:
reviewed at `d64a7dbc`. structure closely mirrors the ClickHouse/Athena precedent, wiring is complete, tests are good. build, full test suite, `golangci-lint` over the five touched packages and `reference-drift` all clean in a worktree at head. approve with minor changes.

## the one I'd fix here

**`limitBailRe` silently disables limit enforcement on ordinary multi-line queries** - `internal/query/bigquery/types.go:103`

the `(?im)` flag makes `^\s*DESC(RIBE)?\b` match at *any* line start, not statement start. so this loses its row cap entirely:

```
gcx datasources bigquery query 'SELECT * FROM `p.d.events`
ORDER BY ts
DESC'
```

`--limit 100` is silently skipped, the full result set comes back, and nothing tells you the cap didn't apply. `DESC` on its own line is ordinary formatting, so this isn't exotic. verified empirically at head:

```
EnforceLimit("SELECT a\nFROM t\nORDER BY a\nDESC", 100, 1000)
  => "SELECT a\nFROM t\nORDER BY a\nDESC"   // unchanged
```

fix is `(?i)` plus `\A\s*` on the statement-leading keywords, keeping the intentionally-unanchored `LIMIT ... OFFSET` alternative separate. this is inherited from ClickHouse, which has the same bug - filed separately as #1114 so you don't have to fix theirs here, but worth getting right in new code.

## three bugs shared with #1059, now tracked separately

these aren't yours to fix in this PR, but flagging so it's visible how much is being copy-propagated across the parallel datasource PRs. #1059 (mssql) has the same first two, character for character:

- **#1113** - `describe-table my_dataset.` and `describe-table ""` accept an empty table name, build `WHERE table_name = ''`, and print `No data` rather than erroring (`describe_table.go:111`, `ValidateName` at `internal/query/bigquery/types.go:32-40`). same root cause in both PRs: the validator returns `nil` for empty because the dataset part is legitimately optional, and the splitter hands back an empty table segment without complaint
- **#1112** - `--step` is registered by `shared.Setup` and advertised in the generated reference, but discarded at `internal/datasources/bigquery/query.go:79`. now the fifth provider to do this; influxdb already works around it with `MarkHidden("step")`, which is the tell that the shared helper is the thing to fix
- **#1114** - the `limitBailRe` bug above, pre-existing in ClickHouse

## smaller stuff in this PR

- DML statements aren't in the bail list (`types.go:103` covers `CREATE|DROP|ALTER|EXPLAIN|DESC|SHOW` but not `INSERT|UPDATE|DELETE|MERGE|TRUNCATE|BEGIN|DECLARE|CALL`), so `DELETE FROM ... WHERE x = 1` gets ` LIMIT 100` appended and fails with a syntax error on a statement the user typed correctly. one-line addition
- hard-coded `LIMIT 1000` in all three metadata queries (`list_datasets.go:75`, `list_tables.go:86`, `describe_table.go:116`) truncates silently. a dataset with >1000 tables just returns a short list. at minimum worth a line in the `Long` text
- `format: 1` at `internal/query/bigquery/client.go:46` - Athena names this (`QueryFormatTable`), ClickHouse doesn't. mild preference for the named constant
- table-name validation is stricter than it needs to be (`describe_table.go:111`). unlike project/dataset, the table name goes into a *string literal* and already passes through `EscapeSQLString`, so `nameRe` blocks BigQuery flexible table names containing `-` for no safety benefit. worth keeping the strict check on project/dataset, which are interpolated as identifiers via `InfoSchemaPrefix`
- `docs/architecture/project-structure.md:91-102` doesn't list `bigquery`. athena is missing too so the map was already stale, but it's PR checklist step 4

## two questions

- no `--location` control - do `INFORMATION_SCHEMA` queries work against datasets outside the datasource's default processing location, or do they fail with "Dataset ... not found in location US"? Athena exposes `--region` for the analogous thing. haven't verified, just want to know if you've hit it
- the query body sends `from`/`to` (`client.go:49-50`) but the `Long` text doesn't mention `$__timeFilter`-style macros the way ClickHouse's does. deliberate, or an omission?

## what's good

- the injection safety is properly thought through: separate `nameRe`/`projectRe` (`types.go:24-27`), neither of which admits a backtick or a dot, plus backtick quoting in `InfoSchemaPrefix` (`types.go:97`) and an explicit doc comment saying callers must validate first. the qualified path can't be broken out of
- deliberately not adding epoch-ms timestamp heuristics is the right call, guessing would corrupt legitimate `INT64` columns
- tests are genuinely good: exact `/apis/query.grafana.app/...` path assertions, the `/api/ds/query` fallback with call counting (`client_test.go:107-131`), typed `queryerror.APIError` assertions, short-row skipping in all three parsers, and the `NormalizeKind` regression case at `resolve_test.go:287`
- wiring is complete - self-registering provider, `NormalizeKind` mapping, codecs in all three output paths with explicit graph-unsupported errors, `output_classes.json`, generated reference with no drift
- pulling `runClickHouseQuery` out of the auto-detect switch was a good call, keeps `QueryCmd` under the complexity limit and minimises conflicts with the other datasource PRs in flight

one architecture note for later, not this PR: `StringList`/`TableInfo`/`ColumnInfo` (`types.go:122-138`) are structurally identical to Athena's, and `internal/datasources/query/codecs.go` now carries the same three-case type switch in three codecs. with several datasource PRs in flight that grows linearly - worth extracting shared discovery types into `querysql` in a follow-up.

ktw4071:
Thanks for the thorough review. Went through it point by point against the current head (post-rebase onto main).

## "the one I'd fix here" — already fixed

`limitBailRe`'s `(?im)` → `(?i)` + `\A\s*` anchoring landed in `5c31e776` before I rebased this branch. Re-verified live: `EnforceLimit("SELECT a\nFROM t\nORDER BY a\nDESC", 100, 1000)` now returns `"SELECT a\nFROM t\nORDER BY a\nDESC LIMIT 100"`, not unchanged. `TestEnforceLimit/appends_LIMIT_to_multi-line_ORDER_BY_DESC` and the window-function variant both pass.

## Tracked separately (#1113, #1112, #1114) — untouched here, as you asked

Confirmed all three are out of scope for this PR and left alone.

## "smaller stuff in this PR" — already fixed, before I touched anything

- DML bail list (`INSERT|UPDATE|DELETE|MERGE|TRUNCATE|BEGIN|DECLARE|CALL`) — landed in `024c3fa9`.
- `format: 1` → named `QueryFormatTable` — landed in `62ab8863`.
- Table-name validation relaxed for hyphenated BigQuery flexible table names — same commit.
- `docs/architecture/project-structure.md` missing `bigquery`/`athena` — fixed by the rebase itself (`6b1fcd18`, cherry-picked from main).

The one item `62ab8863` documented but didn't actually implement — the 1000-row metadata cap had a `Long` text mention but no runtime signal when it actually triggers — I fixed for real in `6b173786`: fetch `MetadataRowLimit+1`, trim, warn on stderr. Same n+1-sentinel idea as postgres/mysql/mssql's `list-tables` fixes.

## New, found while rebasing and cross-checking against #969/#970/#1059

- `bigquery.EnforceLimit` was discarding the `capped` bool the shared `querysql.EnforceLimit` now returns — same anti-pattern the in-flight `add-datasource` raw-sql.md guidance calls out for athena/clickhouse. No negative-`--limit` check either. Both fixed in `a5bb33bc`, threaded through the typed command and the generic auto-detect dispatch.
- `list-datasets`/`list-tables`/`describe-table` all ran `ValidateProject`/`ValidateName` *after* `LoadContextAndConfig`/`ResolveValidateAndSaveDatasource` — same ordering bug postgres's `list-tables` had before `147e8f63`. Moved ahead of any I/O in `6b173786`.

## Your two questions

**`$__timeFilter`-style macros: omission, not deliberate.** Tested live — it works exactly like ClickHouse's, with one real BigQuery-specific catch: it substitutes a TIMESTAMP literal, and BigQuery's type system won't compare that against a DATETIME column:
```
$__timeFilter(date_time)                    -> googleapi 400: Could not cast literal "..." to type DATETIME
$__timeFilter(CAST(date_time AS TIMESTAMP)) -> real rows within --since 1h
```
Documented in `68bba30f` (Long text + a new Example).

**`--location`: reasoned answer, not live-verified against a real cross-region dataset (none exists in our test project).** Our three metadata commands use either the project-level unqualified form (`` `project`.INFORMATION_SCHEMA.SCHEMATA `` in `list-datasets`) or the dataset-qualified form (`` `project.dataset`.INFORMATION_SCHEMA.TABLES `` in `list-tables`/`describe-table`). Both are documented BigQuery conveniences specifically for *not* needing to know a dataset's region ahead of time — unlike the region-qualified form (`` `region-us`.INFORMATION_SCHEMA.TABLES ``), which we don't use. So I don't think `--location` is needed the way Athena's `--region` is, but I want to flag that this is inference from BigQuery's documented behavior, not something I watched fail or succeed against a dataset outside the default location. If you have one available to test against, that would settle it for real.

Full gate green on every commit; live-verified against a real BigQuery datasource throughout.

ktw4071:
Full live validation against a real BigQuery datasource, covering the rebase + all fixes from this round.

**1. Default query (no `--limit`)**
```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA`' -o json
```
stderr: *(no cap warning)* · stdout: **100 rows**

**2. `--limit 5000` (above max, capped and now warned — was silent before this round)**
```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA`' --limit 5000 -o json
```
stderr: `LIMIT in query exceeds the maximum of 1000 and was capped; use --limit 0 to disable enforcement` · stdout: **1000 rows**

**3. `--limit -5` (new: rejected instead of silently ignored)**
```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA`' --limit -5
{"error":{"summary":"--limit must be >= 0, got -5"}}
```

**4. User's own inline `LIMIT 50` (respected, no false warning)**
```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA` LIMIT 50' -o json
```
stderr: *(no warning)* · stdout: **50 rows**

**5. `list-datasets`**
```
$ gcx datasources bigquery list-datasets -d bigqueryjwt-ds-m -o table
DATASET
GCS
grafana_example_dataset
```

**6. `list-tables`**
```
$ gcx datasources bigquery list-tables -d bigqueryjwt-ds-m --dataset grafana_example_dataset -o table
NAME        TYPE
WORLD_DATA  BASE TABLE
```

**7. `list-tables` with `--dataset` omitted (required check now runs before any I/O)**
```
$ gcx datasources bigquery list-tables -d bigqueryjwt-ds-m
{"error":{"summary":"--dataset is required (run 'gcx datasources bigquery list-datasets' to discover datasets)"}}
```

**8. `describe-table`, schema-qualified form**
```
$ gcx datasources bigquery describe-table grafana_example_dataset.WORLD_DATA -d bigqueryjwt-ds-m -o table
NAME             TYPE      NULLABLE
base_country     JSON      YES
country_name     STRING    YES
capital_name     STRING    YES
birth_rate       INT64     YES
co2              INT64     YES
gdp              INT64     YES
date_time        DATETIME  YES
timestamp_value  INT64     YES
```

**9. Invalid `--dataset` identifier (new: rejected before any config/datasource I/O)**
```
$ gcx datasources bigquery describe-table WORLD_DATA --dataset "bad; DROP" -d bigqueryjwt-ds-m
{"error":{"summary":"Invalid dataset","details":"must contain only letters, numbers, and underscores"}}
```

**10/11. `$__timeFilter` — the type-mismatch caveat, live-reproduced**
```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA` WHERE $__timeFilter(date_time)' --since 1h
{"error":{"details":"googleapi: Error 400: Could not cast literal \"2026-08-24T20:15:23Z\" to type DATETIME"}}

$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA` WHERE $__timeFilter(CAST(date_time AS TIMESTAMP))' --since 1h -o json
-> 100 rows, no error
```

**12. Generic auto-detect path, inline `LIMIT 5000` in the SQL itself (dispatchBigQuery's warning wiring)**
```
$ gcx datasources query bigqueryjwt-ds-m 'SELECT * FROM `grafana_example_dataset.WORLD_DATA` LIMIT 5000' -o json
```
stderr: `LIMIT in query exceeds the maximum of 1000 and was capped` · stdout: **1000 rows**

Full gate (lint + tests + build + docs) green on every commit in this round.

internal/datasources/bigquery/metadata.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/grafana/gcx/internal/query/bigquery"
+	querysql "github.com/grafana/gcx/internal/query/sql"
+)
+
+// warnIfMetadataTruncated drops rows beyond bigquery.MetadataRowLimit (the
+// caller must request MetadataRowLimit+1 rows via its own LIMIT clause) and
+// warns on warnw that more rows matched, so a capped metadata listing never
+// reads as the complete inventory. Shared by list-datasets, list-tables, and
+// describe-table, which all cap the same INFORMATION_SCHEMA shape the same way.
+func warnIfMetadataTruncated(warnw io.Writer, resp *querysql.QueryResponse, noun string) {
```
awsome-o, line 16: nitpick: this helper is the fix that landed this round and it has no test, and the package's only test file is `explore_test.go`, which is `package bigquery_test`, so an unexported helper can't be reached from it. flip the `<=` to `<`, or drop the trim on the next line, and all three commands report 1001 rows with no warning on a green suite. same gap covers the other two fixes from this round: the negative `--limit` rejection in `query.go` and the validate-before-I/O reordering here and in `list_tables.go` / `describe_table.go` are all untested.

it's also the third copy of this helper: `postgres/list_tables.go:54` and `mysql/list_tables.go:54` have byte-equivalent bodies differing only in the message, and both of those have a `TestWarnIfTruncated` in a `*_internal_test.go`. hoisting it into `internal/query/sql` with the noun and narrowing hint as parameters fixes the duplication and inherits postgres's existing test in one move. follow-up is fine.

internal/datasources/bigquery/list_tables.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			datasourceUID, _, err := dsquery.ResolveValidateAndSaveDatasource(ctx, loader, opts.Datasource, cfgCtx, cfg, "bigquery")
+			if err != nil {
+				return err
+			}
+
+			sql := fmt.Sprintf(
+				"SELECT table_name, table_type FROM %s.TABLES ORDER BY table_name LIMIT %d",
+				bigquery.InfoSchemaPrefix(opts.Project, opts.Dataset),
```
awsome-o, line 89: question, and the one thing I'd check before merging: has the `--project` path ever actually run? all twelve live checks in the validation comment omit `--project`, so the prefix `InfoSchemaPrefix` builds here, which backtick-quotes the whole project.dataset path before `.INFORMATION_SCHEMA.TABLES`, has never been sent to BigQuery. the documented form quotes only the project and leaves the dataset outside the backticks. quoting the whole path prefix should parse the same way, but "should" is doing the work.

one `list-tables --project <id> --dataset <d>` run settles it. same prefix in `list_datasets.go:79` and `describe_table.go:120`. flagging it as unverified rather than as a defect, I couldn't establish it either way from the code.

internal/datasources/bigquery/explore.go
```diff
@@ -0,0 +1,26 @@
...
+		return ""
+	}
+
+	from, to := dsquery.ExploreRange(query.From, query.To, false)
+
+	q := map[string]any{
+		"refId":      "A",
+		"rawSql":     query.Expr,
+		"editorMode": "code",
+		"format":     1,
```
awsome-o: nitpick: this PR introduces `bigquery.QueryFormatTable` for exactly this value and uses it at `client.go:46`, so the constant currently has one caller and one untyped twin. use `bigquery.QueryFormatTable` here.

internal/query/bigquery/types.go
```diff
@@ -0,0 +1,202 @@
...
+// QueryRequest represents a BigQuery query request.
+type QueryRequest struct {
+	RawSQL string
+	Start  time.Time
+	End    time.Time
+}
+
+// StringList is a single-column discovery result (e.g. dataset names) with a
+// header for table rendering.
+type StringList struct {
```
awsome-o, line 140: nitpick, follow-up rather than this PR: `StringList` is field-for-field and tag-for-tag identical to `athena.StringList` (`internal/query/athena/types.go:28`), and `FormatStringList` in `formatter.go` is identical to `athena.FormatStringList`.

the cost isn't the ~30 lines, it's the codec arms downstream: each dialect that copies this pattern adds one `case` per codec, and this PR adds six in `internal/datasources/query/codecs.go` plus three in the graph codec. with several datasource PRs in flight that grows linearly and the JSON shape then gets frozen per dialect. hoisting `StringList` + `FormatStringList` into `internal/query/sql` and collapsing the athena and bigquery copies onto it replaces those six arms with one `querysql.StringList` arm in the table and wide codecs.

internal/datasources/query/codecs.go
```diff
@@ -188,6 +201,12 @@ func (c *queryGraphCodec) Encode(w io.Writer, data any) error {
 		return errors.New("graph output is not supported for ClickHouse describe-table; use -o table/json/yaml")
 	case athena.StringList:
 		return errors.New("graph output is not supported for Athena discovery; use -o table/json/yaml")
+	case bigquery.StringList:
```
awsome-o, line 204: nitpick: these three arms are unreachable. all three metadata commands call `RegisterCodecs(&opts.IO, false)`, so the graph codec is never registered for them and `-o graph` isn't an accepted value there, meaning the error string can't be printed. the clickhouse and athena arms above are dead for the same reason, so this copies an existing dead pattern rather than inventing one. deleting all three is a no-behaviour-change cleanup whenever the `StringList` hoist happens.

awsome-o (dismissed):
approve. reviewed at `68bba30f` against the ClickHouse/Athena precedent, read the four command files, the query package and the codec wiring, and checked the previous round's items at head rather than taking the reply on trust.

`limitBailRe` is genuinely fixed: `(?i)` with `\A\s*` on the statement-leading keywords, the `LIMIT ... OFFSET` branch left unanchored on purpose, and a comment saying why `(?m)` must not come back. `TestEnforceLimit` covers the multi-line `ORDER BY ... DESC` case, the window-function variant, and two "keyword appears mid-query, must not bail" cases. the rest of that round landed too (DML bail list, `QueryFormatTable`, relaxed table-name validation, `project-structure.md`, the `$__timeFilter` docs with the DATETIME cast caveat), and #1112 / #1113 / #1114 stay deferred as agreed. no open threads, all checks green, branch level with main.

the injection handling is the part I'd call out as done well rather than done: `nameRe` and `projectRe` are separate, neither admits a backtick or a dot, and the table name reaches SQL as a string literal through `EscapeSQLString` instead of being interpolated as an identifier.

everything below is a nitpick. none of it blocks the merge, and the duplication ones are fine as a follow-up rather than in this PR. the one I'd actually spend five minutes on before merging is the `--project` question on `list_tables.go`, since it is a verification gap rather than a code smell.

ktw4071:
Addressed the three action items from the latest review round, plus a status update on the two already-deferred items.

## Fixed, with new regression tests (mutation-tested)

**`metadata.go:16` (`warnIfMetadataTruncated` untested)** — added `metadata_internal_test.go` covering the exact boundary: exactly at `MetadataRowLimit` (no warning), one row over (trimmed + warned), well under, and that the noun is threaded into the message. Mutation-tested by flipping `<=` to `<` in the boundary check — confirmed the "exactly at cap" case goes red, then restored.

**`list_tables.go:89` (validate-before-I/O ordering untested)** — added `list_tables_test.go`, `list_datasets_test.go`, `describe_table_test.go`, `query_test.go`, each asserting on the specific validation error message with a zero-value `ConfigLoader` (so any path that reaches config/datasource I/O fails with an unrelated error instead). Mutation-tested by moving validation back after `LoadContextAndConfig`/`ResolveValidateAndSaveDatasource` — 2 of 3 `list_tables` subtests failed as expected, then restored (confirmed byte-identical via `diff` against the committed version).

**On the `--project` flag itself** (flagged as "unverified, not a defect"): ran all three metadata commands live against `bigqueryjwt-ds-m` with `--project partner-datasources`:

```
$ gcx datasources bigquery list-datasets -d bigqueryjwt-ds-m --project partner-datasources --context dev
{"items":["GCS","grafana_example_dataset"]}

$ gcx datasources bigquery list-tables -d bigqueryjwt-ds-m --project partner-datasources --dataset grafana_example_dataset --context dev
[{"name":"WORLD_DATA","type":"BASE TABLE"}]

$ gcx datasources bigquery describe-table WORLD_DATA --project partner-datasources --dataset grafana_example_dataset -d bigqueryjwt-ds-m --context dev
[{"name":"base_country","type":"JSON","nullable":"YES"},{"name":"country_name","type":"STRING","nullable":"YES"},{"name":"capital_name","type":"STRING","nullable":"YES"},{"name":"birth_rate","type":"INT64","nullable":"YES"},{"name":"co2","type":"INT64","nullable":"YES"},{"name":"gdp","type":"INT64","nullable":"YES"},{"name":"date_time","type":"DATETIME","nullable":"YES"},{"name":"timestamp_value","type":"INT64","nullable":"YES"}]
```

All three built `` `partner-datasources.grafana_example_dataset`.INFORMATION_SCHEMA `` under the hood (`InfoSchemaPrefix`, `internal/query/bigquery/types.go:105`) and returned correct results — confirming the whole-path backtick-quoting is valid BigQuery syntax. No code change needed.

**`explore.go:21` (hardcoded `"format": 1`)** — now uses `bigquery.QueryFormatTable`, matching `client.go`'s existing usage. Added an assertion in `explore_test.go` pinning it, and validated live:

```
$ gcx datasources bigquery query -d bigqueryjwt-ds-m 'SELECT 1 AS one' --context dev --share-link
{"columns":[{"name":"one","type":"number"}],"rows":[[1]]}
🛈 Explore link: https://datasourcedev.grafana-dev.net/explore?panes=...
```

Decoded `panes` param: `"format":1,"rawSql":"SELECT 1 AS one LIMIT 100","refId":"A"` — confirms the constant flows through correctly.

## Left as-is, per your own note

**`types.go:140` / `codecs.go:204`** (StringList/FormatStringList duplication with athena; 3 dead codec arms) — you'd already flagged these as fine to defer, so no change here.

Commits: `2eed9ae3` (format constant fix), `59294c9e` (regression tests).

Ready for another look.

awsome-o (approved):

PR #1110, opened 2026-07-30
feat(profiles): add data-range command exposing Pyroscope ingestion health

marcsanmi:
Adds `gcx profiles data-range` (and `gcx datasources pyroscope data-range`), exposing the `querier.v1.QuerierService/GetProfileStats` RPC.

Part of grafana/pyroscope-squad#744 (§6).

**Why**: empty query results are ambiguous — wrong selector, wrong window, or no data at all? `data-range` answers that in one cheap call: `dataIngested: false` means fix ingestion; otherwise the oldest/newest bounds show the currently queryable window. The range is tenant-wide (everything behind the datasource), not per-service. Also gives agent skills a deterministic "is profiling set up?" probe. profilecli does not expose this RPC (canary probe only).

**Usage**:

```console
$ gcx profiles data-range
DATA_INGESTED  OLDEST_PROFILE        NEWEST_PROFILE
true           2026-06-28T23:47:31Z  2026-07-30T08:03:05Z

$ gcx profiles data-range -o json
{"dataIngested": true, "oldestProfileTime": 1782690451103, "newestProfileTime": 1785398572873}
```

**Semantics**: `dataIngested` is a lifetime flag — it stays true even after all data ages out of retention (31d default, tenant-configurable), in which case the bounds render as `-` (0 in JSON). Unknown bounds are normalized to zero in the client (v1 returns MaxInt64/MinInt64 sentinels when nothing was ingested; v2 metastore shard indexes may report a zero min-time).

Verified against Grafana Cloud and a local Grafana → Pyroscope v2 setup.


internal/datasources/pyroscope/stats.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (opts *statsOpts) Validate() error {
+	return opts.IO.Validate()
+}
+
+func StatsCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &statsOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "stats",
```
simonswine: Not sure we should call it like that, it is a very specific command for showing data pyroscope holds for a tenant.

Maybe

* `data-window`
* `data-range`
* `available-data-rage`

internal/datasources/pyroscope/stats.go
```diff
@@ -0,0 +1,111 @@
...
+		Short: "Show ingestion stats (data ingested, oldest/newest profile times)",
+		Long: "Show ingestion stats for a Pyroscope datasource: whether any " +
+			"profiling data was ever ingested, and the oldest and newest profile times.\n\n" +
+			"Use this to disambiguate empty query results: if no data was ever ingested, " +
+			"fix ingestion before adjusting selectors; otherwise the oldest/newest bounds " +
+			"show the actual queryable time window.\n\n" +
+			"If gcx auto-discovers the datasource from your Grafana Cloud stack, " +
+			"the discovered datasource UID may be saved to your gcx configuration " +
+			"for future commands.",
```
simonswine: We should probably explain why data ingested can be true and still oldest/newest profile times empty. (retention)

Also worth double checking what pyroscope returns as timestamps and how that renders if there is no data available.

internal/datasources/pyroscope/data_range.go
```diff
@@ -0,0 +1,111 @@
...
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			resp, err := client.GetProfileStats(ctx, datasourceUID)
+			if err != nil {
+				return fmt.Errorf("failed to get profile stats: %w", err)
+			}
+
+			return opts.IO.Encode(cmd.OutOrStdout(), resp)
+		},
```
simonswine, line 90: Adding `Args: cobra.NoArg` will make this more robust

simonswine (commented):
Yes that's a good comand to have

simonswine (dismissed):

internal/query/pyroscope/types.go
```diff
@@ -60,6 +63,51 @@ type ProfileTypesRequest struct {
 	End   time.Time
 }
 
+// ProfileStatsResponse mirrors types.v1.GetProfileStatsResponse. Times are
+// milliseconds since epoch, zero when unknown. The wire encodes int64 as
+// proto-JSON strings; UnmarshalJSON accepts both string and number forms so
+// that gcx's own JSON output re-encodes them as plain numbers, consistent
+// with the other profiles commands.
```
wardbekker: This comment gives a wrong reason. It states that the plain number output is "consistent with the other profiles commands". The sibling `Flamegraph` type marshals `total` and `maxSelf` with `json:",string"`, so those commands emit strings.

The design choice is good, and I want to keep it. Plain numbers are the right output for an agent to parse. Please correct the reason in the comment. For example: the wire encodes int64 as proto-JSON strings, and gcx re-encodes them as plain numbers so that consumers do not parse a string.

internal/query/pyroscope/types.go
```diff
@@ -60,6 +63,51 @@ type ProfileTypesRequest struct {
...
+}
+
+// parseProtoInt64 parses a proto-JSON int64, which may be encoded as a JSON
+// string or number; absent or null values parse as zero.
+func parseProtoInt64(raw json.RawMessage) (int64, error) {
+	s := strings.TrimSpace(string(raw))
+	if s == "" || s == "null" {
+		return 0, nil
+	}
+	return strconv.ParseInt(strings.Trim(s, `"`), 10, 64)
```
wardbekker: ``strings.Trim(s, `"`)`` turns the empty proto-JSON string `""` into an empty string. `strconv.ParseInt` then returns an error, and the whole decode fails.

The stated intent of the function is to parse an absent value as zero. An empty string is the same case. Please trim the quotation marks first, then test for the empty string:

```go
s = strings.Trim(strings.TrimSpace(string(raw)), `"`)
if s == "" || s == "null" {
	return 0, nil
}
return strconv.ParseInt(s, 10, 64)
```

This case is unlikely on the real wire. The correction costs one line, so I prefer to close it.

wardbekker (commented):
Thank you for this command. It fills a real gap: no other gcx command reports lifetime ingestion or the true data bounds, and `datasources health` reports connectivity only. Continuous integration passes. The local quality gate passes here too. The three earlier comments from @simonswine are all answered in the code.

I have two small items. Both are safe, and both are in `internal/query/pyroscope/types.go`.

I also want to record one thing that I checked and that is correct, so that a later reviewer does not repeat the work. The custom `UnmarshalJSON` must stay. A `json:",string"` tag cannot replace it, for two reasons:

1. The tag accepts the string form only. It fails on the number form.
2. The tag also encodes as a string. The plain number output in `-o json` is deliberate, and this pull request body documents it.

One more note, for the record: `data-range` takes no selector and no time range, so it is a metadata command like `labels` and `list-profile-types`. The Explore link rule in CLAUDE.md covers query-class leaf commands, so it does not apply here. This pull request adds no new gap.

wardbekker (approved):
Thank you for the two corrections. Both of my comments are closed.

1. The comment on `ProfileStatsResponse` now gives the correct reason. It also records why a `json:",string"` tag cannot replace `UnmarshalJSON`, which saves the next reader the same check.
2. `parseProtoInt64` trims the quotation marks before it tests for the empty string. The new table case in `client_test.go` covers the empty-string time fields.

Approved.

PR #1115, opened 2026-07-30
feat(instrumentation): add fix plan to check that uses Assistant

maryliag:
This PR adds `--fix-plan` to `gcx instrumentation check` as a string-valued flag with two explicit modes with no silent fallback between them, so the user always knows which path ran and whether it was billable.

Modes:

- `--fix-plan=local`: deterministic aggregation of each finding's explanation doc's "How to fix" section. Offline, no billing, works on OSS / Enterprise. No config or Assistant calls.
- `--fix-plan=assistant`: Grafana Assistant synthesizes one prioritized plan covering every finding. Billable

`--fix-plan` alone (bareword) and `--fix-plan=` (explicit empty) are both rejected with an error naming the two modes. Anything other than `local` or `assistant` is also rejected.

Assistant integration:

To share auth + streaming with `gcx assistant prompt` without duplicating logic, the assistant provider exports two small helpers from `internal/providers/assistant/`:
- `ResolveClientOptions` — same auth resolution (OAuth vs. service-account, keychain, cloud-only precondition) used by `gcx assistant prompt`.
- `RequireGrafanaCloud` — cloud-only precondition, called before building a client so the fix-plan path fails fast with a clear message.

Agent catalog:

`gcx instrumentation check`, `explain`, and `list-explanations` are now correctly marked as available on self-hosted (OSS/Enterprise) in `gcx commands`, matching that they all run locally. The Instrumentation Hub subtrees (`setup`, `status`, `clusters`, `services`) remain Cloud-only.

Example of successful results:
Screenshot 2026-07-30 at 9 25 39 AM

Example of when something goes wrong and Assistant fails:

Screenshot 2026-07-30 at 9 17 23 AM


maryliag:
@claude review

maryliag:
@claude review

maryliag:
intentionally skipped item b, d and e from latest claude review

maryliag:
@claude review

cmd/gcx/instrumentation/check/command.go
```diff
@@ -28,13 +30,20 @@ type checkOpts struct {
...
+	PrintPrompt    bool
+	TimeoutSeconds int
 }
 
 func (o *checkOpts) setup(flags *pflag.FlagSet) {
 	o.IO.DefaultFormat("table")
 	o.IO.RegisterCustomCodec("table", &CheckTableCodec{})
 	o.IO.RegisterCustomCodec("wide", &CheckTableCodec{Wide: true})
-	o.IO.SetJSONFieldValidator(cmdio.MakeFieldValidator(otelutils.Results{}))
+	o.IO.SetJSONFieldValidator(cmdio.MakeFieldValidator(ResultsWithFixPlan{}))
```
dafydd-t, line 44: --json <field> no longer accepts any real field name

MakeFieldValidator calls reflectFields (internal/output/format.go:395), which iterates t.Fields(). That iterator yields the struct's own fields. It does not yield promoted fields from an embedded struct. So the embedded otelutils.Results appears as one field named Results.

cmd/gcx/instrumentation/check/codec.go
```diff
@@ -49,14 +87,74 @@ func (c *CheckTableCodec) Encode(w io.Writer, v any) error {
...
+	if _, err := fmt.Fprintln(w); err != nil {
+		return err
+	}
+	switch {
+	case plan.Preview:
+		fmt.Fprintln(w, "Prompt preview — Grafana Assistant was NOT called; no billing.")
+		fmt.Fprintln(w)
+	case plan.Source == "local":
+		if plan.Fallback && plan.Reason != "" {
+			fmt.Fprintf(w, "Grafana Assistant not available (%s). Showing combined explanation docs instead — no AI reasoning applied.\n\n", plan.Reason)
```
dafydd-t: Do you want to write this kind of thing to stderr instead? Otherwise some redirect of the plan to a file will contain this stuff too.

cmd/gcx/instrumentation/check/fixplan/fixplan.go
```diff
@@ -0,0 +1,170 @@
...
+		resp, err := runner(ctx, opts.Loader, assistantprov.PromptRequest{
+			Message:        prompt,
+			AgentID:        opts.AgentID,
+			TimeoutSeconds: opts.TimeoutSeconds,
+			// Opt in to auto-approve: the fix-plan prompt asks for markdown
+			// synthesis, not tool calls, and there is no interactive stdin
+			// here. RunPrompt's fail-closed default would deny any approval
+			// request, which manifests server-side as "HTTP 500: Permission
+			// check failed" on Grafana Cloud.
+			ApprovalHandler: assistantprov.AlwaysApprove{},
```
dafydd-t: I am a bit unsure about this, I think we need to be careful. If we dont make tool calls, will we ever get approval requests? I think something in this paragraph contradicts

cmd/gcx/instrumentation/check/fixplan/fixplan.go
```diff
@@ -0,0 +1,170 @@
...
+			// produces a usable plan. Surface the reason on Plan.Reason
+			// rather than treating it as fatal.
+			return Plan{
+				Source:   SourceLocal,
+				Content:  buildLocalPlan(findings, docs),
+				DocsUsed: docIDs,
+				Fallback: true,
+				Reason:   fmt.Sprintf("Assistant request failed: %s", err),
+			}, nil
+		}
```
dafydd-t: maybe we should check for context cancelled errors before calling the fallback? Otherwise a ctrl+c during an assistant call will trigger a fallback plan

cmd/gcx/instrumentation/check/codec.go
```diff
@@ -49,14 +87,74 @@ func (c *CheckTableCodec) Encode(w io.Writer, v any) error {
...
+// examples, and URLs in the plan stay on one logical line for click-through
+// and copy/paste.
+func renderFixPlan(w io.Writer, plan *FixPlanEnvelope) error {
+	if _, err := fmt.Fprintln(w); err != nil {
+		return err
+	}
+	switch {
+	case plan.Preview:
+		fmt.Fprintln(w, "Prompt preview — Grafana Assistant was NOT called; no billing.")
+		fmt.Fprintln(w)
```
dafydd-t: I think this will still go through glamour's rendering, so we should skip glamour here, otherwise you're not getting the raw prompt as the agent would get it.

internal/agent/command_annotations.go
```diff
@@ -140,7 +140,7 @@ var commandAnnotations = map[string]annotation{
 	// top-level single commands
 	"gcx instrumentation setup":             {Cost: "medium", Hint: "<cluster> --use-defaults -o json | Docs: " + docs.KubernetesMonitoring},
 	"gcx instrumentation status":            {Cost: "medium", Hint: "-o json | Docs: " + docs.KubernetesMonitoring},
-	"gcx instrumentation check":             {Cost: "small", Hint: "validates the LOCAL workstation's OTel setup (env vars, SDK deps, collector/Beyla/Alloy config, Grafana Cloud env creds) — does not query any Grafana stack. [components] --language <lang> -o json"},
+	"gcx instrumentation check":             {Cost: "small", Hint: "validates the LOCAL workstation's OTel setup (env vars, SDK deps, collector/Beyla/Alloy config, Grafana Cloud env creds) — does not query any Grafana stack. [components] --language <lang> -o json. Add --fix-plan to synthesize one fix plan for every finding: uses Grafana Assistant when the current context is Grafana Cloud (billable, see " + docs.AssistantPricing + ", cost medium), otherwise falls back to a local aggregation of the explain docs. Use --print-prompt to preview the Assistant prompt without billing."},
```
dafydd-t: can you reword this, so the --fix-plan clarification isnt just bolted on? the original part states the grafana stack is never queried, which isnt true with --fix-plan, for example

cmd/gcx/instrumentation/check/command.go
```diff
@@ -50,6 +59,17 @@ func (o *checkOpts) setup(flags *pflag.FlagSet) {
 		"Path to the OpenTelemetry Collector config file.")
 	flags.BoolVar(&o.Debug, "debug", false,
 		"Print additional diagnostic output from the checker.")
+
+	flags.BoolVar(&o.FixPlan, "fix-plan", false,
+		"After running the checks, synthesize a single fix plan for every finding. "+
+			"Uses Grafana Assistant when the current context is a Grafana Cloud stack (billable); "+
+			"falls back to a local aggregation of the explanation docs otherwise.")
+	flags.StringVar(&o.AgentID, "agent-id", "",
+		"With --fix-plan: target a specific Grafana Assistant agent (defaults to the CLI agent).")
```
dafydd-t: is there any use for choosing another agent?

cmd/gcx/instrumentation/check/command.go
```diff
@@ -50,6 +59,17 @@ func (o *checkOpts) setup(flags *pflag.FlagSet) {
...
+	flags.BoolVar(&o.PrintPrompt, "print-prompt", false,
+		"With --fix-plan: build and print the Assistant prompt to stdout, then exit. Assistant is NOT called; no billing.")
+	flags.IntVar(&o.TimeoutSeconds, "assistant-timeout", 0,
+		"With --fix-plan: Grafana Assistant response timeout in seconds. 0 uses the library default (300s).")
```
dafydd-t: do we need this? whats wrong with the library default?

cmd/gcx/instrumentation/check/codec.go
dafydd-t: or, can we remove the --print-prompt flag? just poking at ways to reduce the diff & scope

internal/providers/assistant/embed.go
dafydd-t, line 1: I think we can make this smaller - maybe we don't need the full abstraction? For example, reduce to two exported aliases: `ResolveClientOptions` and `RequireGrafanaCloud` and build the client and call ChatWithApproval in fixplan.

cmd/gcx/instrumentation/check/fixplan/fixplan.go
```diff
@@ -0,0 +1,170 @@
...
+// findings that need a fix plan; callers should skip rendering in that case.
+type Plan struct {
+	Source   Source
+	Content  string   // markdown (or the raw prompt when Preview == true)
+	DocsUsed []string // explain IDs the plan consulted
+	Empty    bool     // true when there are no error/warning findings
+	Fallback bool     // true when Assistant was requested but local was used
+	Reason   string   // when Fallback: one-line human explanation
+	Preview  bool     // true when Content holds the built Assistant prompt (--print-prompt), not a real plan
+}
```
dafydd-t: Can we combine this and FixPlanEnvelope somehow? they have the same fields

cmd/gcx/instrumentation/check/codec.go
```diff
@@ -49,14 +87,74 @@ func (c *CheckTableCodec) Encode(w io.Writer, v any) error {
...
+// renderFixPlan prints a separator, a source notice, and the plan body.
+// When the destination is a terminal AND styled output is enabled (respecting
+// --no-color / NO_COLOR), the plan (markdown) is styled via glamour;
+// otherwise the raw markdown is written so piping, no-color mode, and tests
+// stay clean.
+//
+// Word wrap is disabled (WithWordWrap(0)) so shell commands, env-var
+// examples, and URLs in the plan stay on one logical line for click-through
+// and copy/paste.
+func renderFixPlan(w io.Writer, plan *FixPlanEnvelope) error {
```
dafydd-t: there is something very similar in cmd/gcx/instrumentation/explain/command.go:110-155. Same for `isTerminalWriter`. can we combine them?

cmd/gcx/instrumentation/check/fixplan/local.go
```diff
@@ -0,0 +1,84 @@
...
+var nextH2 = regexp.MustCompile(`(?m)^##\s+.*$`)
+
+// extractHowToFix returns the body of the doc's "## How to fix" section, or
+// the whole body when no such section exists. Leading and trailing
+// whitespace on the returned block is trimmed.
+func extractHowToFix(body string) string {
+	loc := howToFixHeading.FindStringIndex(body)
+	if loc == nil {
+		return strings.TrimSpace(body)
+	}
```
dafydd-t: how important is it to check for his, when all checks in  otel-checker@v0.3.1 contain a "how to fix" header?

cmd/gcx/instrumentation/check/fixplan/prompt_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	assert.Empty(t, collectFindings(results))
+}
+
+func TestResolveDocs_DedupsAndSorts(t *testing.T) {
+	// Grab two real IDs from the embedded registry so we test against
+	// actual data. Emitting the same ID twice should collapse.
+	ids := otelexplain.All()
+	require.GreaterOrEqual(t, len(ids), 2, "registry must have at least 2 docs to run this test")
+	pick := []string{ids[3], ids[0], ids[3]} // out of order + duplicate
```
dafydd-t: it checks ids[3], but only checks that there are > 2 ids

cmd/gcx/instrumentation/check/fixplan/fixplan_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// A single real explain ID we use across tests. Picking from the registry
+// (instead of a hardcoded string) keeps the tests robust across doc renames.
+func firstRealExplainID(t *testing.T) string {
+	t.Helper()
+	res := otelutils.Results{
+		Errors: []otelutils.ComponentResult{{Component: "Grafana Cloud", Message: "no headers", ExplainID: "grafana-cloud.headers.missing-auth"}},
```
dafydd-t: the function comment says it doesn't use hard-coded strings, but there's one here?

cmd/gcx/instrumentation/check/command.go
```diff
@@ -61,6 +81,17 @@ func (o *checkOpts) Validate() error {
...
+	if !o.FixPlan {
+		// Reject fix-plan sub-flags when the parent flag is off so users get
+		// a clear error rather than silently ignored flags.
+		if o.AgentID != "" {
+			return errors.New("--agent-id requires --fix-plan")
+		}
+		if o.PrintPrompt {
+			return errors.New("--print-prompt requires --fix-plan")
+		}
+	}
```
dafydd-t: what about assistant timeout here (if we keep it)?

cmd/gcx/instrumentation/check/codec.go
```diff
@@ -49,14 +87,74 @@ func (c *CheckTableCodec) Encode(w io.Writer, v any) error {
...
+	if _, err := fmt.Fprintln(w); err != nil {
+		return err
+	}
```
dafydd-t, line 97: is it reasonable to ignore the err of the later calls to Fprintln?

CHANGELOG.md
```diff
@@ -127,7 +128,6 @@
 - Refine output formatting (format and jq handling)
 - Remove obsolete publish-technical-documentation workflows
 
-
```
dafydd-t, line 216: We dont need to remove this

dafydd-t (dismissed):
--json change is a breaking one, and then some more comments about scope. Thanks a lot for this!

cmd/gcx/instrumentation/check/codec.go
maryliag: removed the flag

cmd/gcx/instrumentation/check/command.go
maryliag: assistant timeout got removed

cmd/gcx/instrumentation/check/command.go
maryliag: I was just trying to align with the assistant option, but let's keep this simple. Removed

internal/providers/assistant/embed.go
maryliag: done

cmd/gcx/instrumentation/check/command.go
maryliag: fixed

cmd/gcx/instrumentation/check/fixplan/fixplan.go
maryliag: fixed

cmd/gcx/instrumentation/check/codec.go
maryliag: updated

cmd/gcx/instrumentation/check/fixplan/fixplan.go
maryliag: updated

cmd/gcx/instrumentation/check/codec.go
maryliag: updated, now it has a shared helper for render markdown

cmd/gcx/instrumentation/check/codec.go
maryliag: this no longer exists after the refactor

cmd/gcx/instrumentation/check/fixplan/fixplan.go
maryliag: The fail-closed default in the Assistant SDK auto-denies every approval request, which surfaces as HTTP 500: Permission check failed rather than a clean stream error, even though the fix-plan prompt only asks for markdown synthesis and never triggers a real tool call. Passing alwaysApprove{} short-circuits that 500. I added the rationale is in the runAssistant doc comment

cmd/gcx/instrumentation/check/fixplan/local.go
maryliag: removed this to simplify

cmd/gcx/instrumentation/check/fixplan/prompt_test.go
maryliag: nice catch, fixed

cmd/gcx/instrumentation/check/fixplan/fixplan_test.go
maryliag: urgh, forgot to remove this part. Updated

CHANGELOG.md
maryliag: added back

cmd/gcx/instrumentation/check/command.go
maryliag: removed

internal/agent/command_annotations.go
maryliag: updated

maryliag:
@dafydd-t any other changes required?

wardbekker:
Please make the fix-plan mode explicit.

At present, the same `--fix-plan` option can produce two very different results:

- A local mode copies instructions from the explanation documents. It does not use AI and does not incur billing costs.
- An Assistant mode creates and prioritizes a plan. It uses Grafana Assistant and incurs billing costs.

Please consider this interface:

```text
--fix-plan=local       Create a deterministic plan from the explanation documents. No billing.
--fix-plan=assistant   Create a prioritized plan with Grafana Assistant. This request is billable.
```

If the user selects `assistant` and Assistant is not available, return a clear error. Do not silently return the different local result. If a fallback mode is necessary, add an explicit `auto` value.

The command help and the agent hint already contain billing information. Please keep the billing warning next to the `assistant` mode. The command annotation has `Cost: "small"` because the annotation applies to the complete command. Therefore, the hint must clearly say that `--fix-plan=assistant` makes a billable request. It must also include the pricing documentation link.

Please also update this branch with the latest `main` branch and run the checks again.

maryliag:
feedback addressed

cmd/gcx/instrumentation/check/command.go
```diff
@@ -61,6 +74,13 @@ func (o *checkOpts) Validate() error {
 		return err
 	}
 
+	switch o.FixPlan {
```
wardbekker, line 77: **required** `--fix-plan=` sets this field to the same empty value as an omitted flag, so this switch accepts it and the command silently runs without a fix plan. I confirmed this through the built command. Please use `cmd.Flags().Changed("fix-plan")` to reject an explicitly empty value and add a command-level test. Is an empty assignment intended to mean omission despite the changelog statement that users must specify a mode?

ARCHITECTURE.md
```diff
@@ -123,12 +123,12 @@ Action-verb command tree for Grafana Cloud's Instrumentation Hub. Backed by flee
 
 - **`instrumentation setup <cluster>`** — End-to-end onboarding wizard; calls `SetupK8sDiscovery`, applies declared K8s monitoring config, prints a parameterized helm command
 - **`instrumentation status`** — Cross-cutting observed view across cluster → namespace → service hierarchy
-- **`instrumentation check [components]`** — Local OTel-instrumentation validation via the [otel-checker](https://github.com/grafana/otel-checker) library; runs entirely against the workstation's env vars, package manifests, and collector/Beyla/Alloy config. No Grafana stack calls
+- **`instrumentation check [components]`** — Local OTel-instrumentation validation via the [otel-checker](https://github.com/grafana/otel-checker) library; runs entirely against the workstation's env vars, package manifests, and collector/Beyla/Alloy config. No Grafana stack calls unless `--fix-plan` is set — that flag optionally hands the findings and their explanation docs to Grafana Assistant to synthesize one fix plan (billable), with a local aggregation fallback on OSS/Enterprise or when Assistant is unreachable
```
wardbekker: **recommended** This still describes the removed automatic fallback. Line 131 also names `assistant.RunPrompt`, which no longer exists, and describes an Assistant/local-fallback decision. Please update these lines, the matching `project-structure.md` entries, the `fixplan` package comment, and the PR description to describe the two disjoint modes.

internal/agent/command_annotations.go
```diff
@@ -141,7 +141,7 @@ var commandAnnotations = map[string]annotation{
...
 	"gcx instrumentation setup":             {Cost: "medium", Hint: "<cluster> --use-defaults -o json | Docs: " + docs.KubernetesMonitoring},
 	"gcx instrumentation status":            {Cost: "medium", Hint: "-o json | Docs: " + docs.KubernetesMonitoring},
-	"gcx instrumentation check":             {Cost: "small", Hint: "validates the LOCAL workstation's OTel setup (env vars, SDK deps, collector/Beyla/Alloy config, Grafana Cloud env creds) — does not query any Grafana stack. [components] --language <lang> -o json"},
+	"gcx instrumentation check":             {Cost: "small", Hint: "validates the LOCAL workstation's OTel setup (env vars, SDK deps, collector/Beyla/Alloy config, Grafana Cloud env creds). Without --fix-plan, no Grafana stack is queried. [components] --language <lang> -o json. Two fix-plan modes: --fix-plan=local aggregates the explain docs deterministically (offline, no billing). --fix-plan=assistant synthesizes a prioritized plan with Grafana Assistant — BILLABLE (see " + docs.AssistantPricing + ", cost medium), requires a Grafana Cloud context. Cost annotation stays 'small' because it reflects the base command; --fix-plan=assistant is the billable path."},
```
wardbekker, line 144: **recommended** The hint says that `--fix-plan=local` works offline, but `gcx commands --flat -o json` still marks this command as `grafana-cloud-only` through the `gcx instrumentation` path rule. Agents can therefore avoid this local mode on OSS or Enterprise. Please make the catalog availability match the new mode contract, or remove the offline and OSS/Enterprise availability claim.

wardbekker (commented):
The explicit `local` and `assistant` split is sound. Local mode avoids config and Assistant calls. Assistant mode now fails without falling back. The billing warning and pricing link are also present.

One required input fix remains: `--fix-plan=` is accepted as if the flag was omitted. I also found stale architecture text and an agent-availability mismatch.

Verdict: fix the explicit-empty mode before merge. The other two comments should also be addressed in this PR because they describe the new mode contract.

cmd/gcx/instrumentation/check/command.go
maryliag: fixed

ARCHITECTURE.md
maryliag: updated

internal/agent/command_annotations.go
maryliag: catalog updated

wardbekker (approved):

Issue #1062, opened 2026-07-24
config: env var to disable the OS keychain backend (escape hatch for headless environments)


awsome-o:
Follow-up to #1060 / #1061. There's currently no user-facing way to tell gcx "don't touch the OS keychain". The only disable mechanisms are internal: defaultKeychainStore returns a noop store under testing.Testing() (internal/config/loader.go:80-92, every op returns ErrUnavailable so everything falls back to plaintext), plus the keychainStoreFn injection seam next to it.

Proposal: an env var (GCX_KEYCHAIN=off, name open to bikeshedding) checked in defaultKeychainStore that returns the same noop/unavailable store, giving headless machines and CI a deliberate plaintext-fallback mode. The mechanism already exists and behaves exactly right — it's just not reachable outside go test.

Once #1060 is fixed, GRAFANA_TOKEN is a working per-invocation escape hatch, so this is an enhancement rather than a blocker. The case it covers is wanting the stored config to keep working on a box whose keychain is permanently unavailable, without exporting credentials into every shell.

Needs the env-var reference regen and docs alongside the code change.

PR #1118, opened 2026-07-31
feat(telemetry): report http_status, k8s_reason and grafana_auth_method in the usage event

awsome-o:
Closes https://github.com/grafana/gcx-internal/issues/38

Part of https://github.com/grafana/gcx-internal/issues/2, third of three planned PRs. A (#1108) merged as b5ab921e and B (#1109) merged as c7d73b26, so this branch now sits directly on `main` and the diff below is C alone.

## Rebased onto `main`, 8 Sep

B merged, so this is no longer stacked on anything. B squash merged as `c7d73b26`, and its squash is byte identical to the branch diff it came from, so replaying only C's own commits was the right move rather than merging: `git rebase --onto origin/main 35efc007`, which drops B's five commits and leaves C's fourteen. Zero conflicts, and the 34 file diff against `main` is byte for byte the diff that was under review, blob hashes and line offsets aside.

A clean rebase is where this series has been bitten before, so the things that could have gone stale silently were checked one at a time. `internal/config/types.go` is untouched on `main` since B branched, so `GrafanaConfig`, `IsEmpty()` and `Context.Validate` are exactly what the nil or empty auth guard was written against, and `envparse.go` still only nils `TLS` back. `internal/login/login.go` moved under the keychain change (#1271) and the Fleet plugin proxy (#1261), but `resolveGrafanaAuth` still stages `StagedContext.Grafana` at `login.go:723` before the destination gate at `:368` and the cloud auth gate at `:388`, which is what the login capture relies on. No new `HTTPStatusCode() int` implementor and no new `queryerror.New` or `FromBody` call site landed on `main` in that window, so the coverage list on the page is still accurate. `TestFirstRunNoticeTextIsPinnedToItsRevision`, the guard that caught real drift on the last restack, passes at revision 3.

Gates on the rebased head: `mise run lint` 0 issues, `go test ./...` 161 packages green, `go test -race` green on the changed packages including the three real subprocess interrupt tests, `GCX_AGENT_MODE=false mise run reference` a no-op, `mise run docs` builds. CI is green on `0cefd803` (Tests, Linters, Documentation, signed commits), which is also the first time it has run on this branch in weeks: a `CONFLICTING` PR has no merge ref and quietly runs nothing at all.

The staging round trip below was run at `7d291286` and was not repeated, because the receiver gate is already settled and nothing on `main` in this window touches what the event carries. What did get re-run is the local `GCX_TELEMETRY=log` matrix against the rebased binary, since `internal/login/login.go` is the one file in the capture path that `main` actually moved. All three fields still behave as the page says. `token`, `basic` and `anonymous` each report themselves, `GRAFANA_TOKEN` over a credential free context reports `token`, and a command that builds no Grafana connection reports no method. A Kubernetes 404 reports `k8s_reason: NotFound` with no `http_status`; a Kubernetes 403 reports `k8s_reason: Forbidden` with no `http_status`, `error_kind: auth_failure` and exit 3; the same 403 through `gcx api` reports `http_status: 403` with no `k8s_reason` and exit 1. That is the pinned taxonomy unchanged. Nothing left the machine: the events were logged rather than exported, under an isolated `XDG_STATE_HOME`, against a local stub on `127.0.0.1`.

## Restacked again, 25 Aug

B was rebased onto `main` at `fcec55b7`, so this branch moved with it and now carries those 24 commits too. Three conflicts, all resolved in the obvious direction:

- `internal/queryerror/apierror.go` — `main` added `CloudOnly`/`Experimental` and `WithAvailability` to the same struct this PR adds `TransportStatus` to. Both kept; they do not interact, and `FromBody(...).WithAvailability(...)` still preserves the transport status.
- `cmd/gcx/login/command.go` — the import block only. `--oauth-manual` (#1136) touches other parts of the file.
- `docs/sources/anonymous-usage-statistics.md` — B's corrected second-Ctrl-C sentence next to this PR's cross-reference to the failure fields.

Worth re-checking after a move this size, and both still hold: no new `HTTPStatusCode() int` implementor landed on `main`, and the four new `queryerror.New` call sites from the new datasources (Azure Monitor, Cloud Monitoring, Elasticsearch) carry plugin-level statuses through `New`, which by design sets no `TransportStatus` — so they emit no `http_status`, exactly as the contract says.

This round also fixes the auth arm the bot flagged: `PrepareForEnvParse` replaces a nil Grafana block with an empty one, so `grafana == nil` never fired on the tolerant load path and a context with no Grafana block reported `anonymous` — which the page defines as "requests went out unauthenticated". It now guards nil-or-empty like `Context.Validate`, with a test on the real `ParseEnvIntoContext` path that fails against the old arm. The wire changes for exactly that case and no other, so the staging round-trip below still stands. The comments this series adds are also shorter, per review.

## What the earlier restack changed

Nothing about the wire contract, and one real bug caught by A's own guard.

A's review rounds rewrote three of the same doc blocks this PR touches, so those merged by hand and A's corrected claims win: the capture package's own doc no longer says it exists to avoid an import cycle (it doesn't — nothing in `internal/telemetry` writes a capture), the `Event` privacy invariant stays scoped to *volume* rather than to numbers in general, and the architecture map keeps A's description of the capture package while gaining C's vocabularies. The published page keeps A's `--json`/`output_format` caveat next to C's new authentication sentence.

The bug: A landed `TestFirstRunNoticeTextIsPinnedToItsRevision`, which pins the notice byte for byte so a newly disclosed field cannot ship under consent people already gave. It merged cleanly with this branch's bump to revision 3 and then failed, still holding revision 2's text. That is the pin doing its job on drift arriving through a rebase rather than an edit, and the fix is its own commit.

The shared writer walker (`internal/testutils/capturewriters.go`) already carried A's nested-worktree and nested-module skips, so C's version stands unchanged; without those the guard double-counts every writer inside `.claude/worktrees` and fails only locally.

`mise run lint` 0 issues, `go test ./...` green, `GCX_AGENT_MODE=false mise run reference` a no-op on the restacked head. The end-to-end verification below was run at `7d291286`, before the restack; the restack touched doc prose, the notice pin test and the shared walker, and nothing that decides what the event carries.
## Why

When a command fails, `error_kind` collapses almost everything into `error`. The reason is mechanical: `providers.FormatError` interpolates the HTTP status into a message string and discards the int across ~80 call sites, `convertAPIErrors` computes the Kubernetes reason and discards it, and which authentication method a failing invocation used is not recorded at all despite known auth breakage (#1060–#1062).

This adds three optional fields: `http_status`, `k8s_reason`, and `grafana_auth_method`.

## The wire contract

`http_status` is the HTTP transport status of the failing request, only ever 400–599. It is never a status embedded inside a 2xx response body: datasource queries can fail inside an HTTP 200, and that failure sets no `http_status` at all, even though `error_kind` may still say `auth_failure` because the command classifies on the embedded status. That divergence is deliberate on both sides. It is also never a Kubernetes status code.

`k8s_reason` is the Kubernetes status reason, allowlisted to the 19 apimachinery wire values plus an `other` sentinel. The allowlist holds wire values, not Go identifiers — `StatusReasonStoreReadError`'s wire value is `StorageReadError`. An empty reason means the field is omitted; an unrecognized server-supplied reason travels as `other`, never verbatim.

Both are suppressed for exit code 4, because a partial failure has no single causal status — the one that surfaced would stand in for many — and exit code 5, because a canceled run is not a failure.

`grafana_auth_method` is the authentication category selected for the Grafana connection: `oauth`, `token`, `basic`, `mtls`, `anonymous`, or `unknown`. It is eligible on every outcome, including success and cancellation. Absent means no Grafana context was selected, or the invocation selected several different methods (`gcx config check` examines every context concurrently; when they differ the field reports nothing rather than one arbitrary winner). `anonymous` is a valid selection with no credential material. `unknown` is a selection that could not be classified. A selected method whose credential is invalid or rejected reports the method — which authentication category fails is precisely what the field exists to show. It describes Grafana connection authentication only, never Grafana Cloud/GCOM auth.

## What deliberately does not change

Exit codes. The known taxonomy is pinned by new tests rather than corrected: a bare Cloud-provider 401 keeps exiting 1, the SM register/install 401 keeps exiting 3 through its string-matching converter, a raw Kubernetes `*StatusError` Unauthorized keeps exiting 3 while the dynamic client's re-wrap of the same error keeps exiting 1. The typed provider error implements exactly one method, `HTTPStatusCode() int`; implementing `APIServiceName`/`APIUserMessage` as well would satisfy the converter chain's `serviceAPIError` interface and silently flip dozens of provider 401/403 call sites from exit 1 to 3 while shadowing the SM/cloud/stacks/fleet converters behind it. A tripwire test pins the negative shape.

Error messages. All five `FormatError`/`HandleErrorResponse` message forms and both `gcx api` forms are byte-for-byte identical, asserted exactly; the previously untested body-read-failure path now retains its status while the reader error stays reachable through `Unwrap` exactly as the old `%w` exposed it. Converters that string-match on these messages see identical text.

Output contracts, batch telemetry, and cancellation behavior. The exit-4/5 suppression wraps only the two new fields; the batch block's survival on exit 4 and `error_kind`'s presence on every outcome stay pinned by their existing tests.

## Design notes

Capture happens at the top of `reportError`, on the raw error, ahead of the `AlreadyReported` and `EmittedError` short-circuits — an `EmittedError`'s cause chain is where agent-mode in-band failures keep their transport error, and neither short-circuit converts anything. Raw capture is separate from wire filtering: the 400–599 range, both allowlists, and the exit-4/5 suppression live in the usage-event builder, so nothing captured anywhere can reach the wire unfiltered.

`k8s_reason` is extracted via `ReasonForError` rather than inside `convertAPIErrors`, deviating from the original design sketch on purpose: the converter matches only concrete `*StatusError`, and the entire `gcx resources` path re-wraps errors into the dynamic client's value-typed `APIError`, which never reaches it. `ReasonForError` matches the `APIStatus` interface and sees both shapes.

`queryerror.APIError` gains a `TransportStatus` field, set only by `FromBody` before the existing 2xx substitution can replace `StatusCode` with a status embedded in the body. Errors built by `New` carry zero — several call sites synthesize a 400 for absent embedded statuses, and a synthesized status must never masquerade as a transport fact. The telemetry probe is match-wins: a query error with no transport status silences the chain rather than falling through to another carrier.

The auth capture hooks `(*Context).selectGrafanaAuth` itself, the single authority shared by `Context.Validate` on the load path and the validated selection on every transport build. Recording happens before credential validation, which is what lets an invalid credential still report its method. `InferredAuthMethod` is untouched and unused: its error path returns the raw configured string. Login force-records after every `login.Run`, falling back to the staged context's method for runs rejected after auth resolution (destination validation, cloud auth), because login resolves auth by probing and its answer outranks anything a config load captured on the way.

`gcx api` becomes the third `http_status` producer alongside the shared provider helper and the query errors, via the same minimal `gcxerrors.HTTPStatusError` — partly because it is the natural raw-API surface, partly because the process-level usage-event test drives `gcx api`, so the end-to-end wire assertion needed it.

## Disclosure

The first-run notice is bumped to revision 3 and re-shown to existing installs: `grafana_auth_method` is a field derived from how the command ran, and the `Event` doc's own rule requires the notice to change when one is added. The new sentence separates authentication from failure data: the category (never credentials) for Grafana connections, and for some failed commands a 4xx/5xx status or fixed Kubernetes reason category, omitted for partial failures and cancellations. The usage-statistics page gains the full vocabularies and a how-to-read list, including the partial-coverage caveat: `http_status` is recorded wherever the failing client surfaces a typed status (the shared provider error helper, datasource query endpoints, datasource management, Knowledge Graph, Adaptive Logs, `gcx api`), while plain-text clients — GCOM, Fleet, k6, Faro, Adaptive Metrics, the assistant — are not instrumented, so the absence of `http_status` never means no HTTP failure occurred. A drift test ties the page's published vocabularies to the Go allowlists.

## Testing

`mise run lint` 0 issues, `go test ./...` green (targeted packages also run with `-race`), build green, `GCX_AGENT_MODE=false mise run reference` a no-op. `mise run docs` needs mkdocs, which isn't installed locally; CI covers it.

Guards worth calling out:

- The writer set of every new capture slot is pinned by an alias-resolving AST walk over the repository (the existing batch guard is keyed on the literal `SetBatch` and cannot see these setters): `SetHTTPStatus`/`SetK8sReason` only from `fail.CaptureErrorSignals`, `SetGrafanaAuthMethod` only from the config selector, `ForceGrafanaAuthMethod` only from login.
- A `reportError`-level table pins that extraction moves no exit code, including the k8s 3-versus-1 asymmetry.
- A golden privacy sweep feeds every slot content shaped like what must never travel — a resource name and namespace inside a server-controlled reason, a credential in the auth slot — and asserts none of it appears anywhere in the marshaled event while the clamped values do.
- The conflict rule has its `-race` test in `internal/config` next to `config check`'s concurrent pattern, not in the capture package, whose no-concurrency-test comment now names the auth slot as the documented exception.

## Manual verification

Against this repo's docker-compose Grafana 12.4 with `GCX_TELEMETRY=log` (event bytes identical to the export body): `resources get` of a missing dashboard → `k8s_reason: "NotFound"`, no `http_status`, exit 1; a bad `GRAFANA_TOKEN` → `k8s_reason: "Unauthorized"` with `grafana_auth_method: "token"`; `gcx api` against a missing endpoint → `http_status: 404`, exit 1 unchanged; a credential-less context → `"anonymous"`; `GRAFANA_TOKEN` over a persisted oauth context → `"token"`; `gcx config view` → no auth field. The HTTP-200-embedded-error, partial-failure, and cancellation cases are covered by unit and process-level tests (the cancellation process test now also asserts both failure fields are absent while the auth category survives).

## End-to-end verification

Verified against the final head `7d291286`, built and tested from a clean worktree matching upstream. Two synthetic failing events were sent only to the staging receiver (`https://stats.grafana-dev.org/gcx-usage-report`) under a throwaway device ID minted in an isolated `XDG_STATE_HOME`, and both were confirmed read-only in the development BigQuery table `grafanalabs-dev.usage_reports.gcx_raw_reports` via `JSON_VALUE(report, ...)` — exactly two rows for that device ID, carrying exactly what was sent:

- `api` → `http_status: 500`, `grafana_auth_method: "token"`, no `k8s_reason`
- `resources get` → `k8s_reason: "NotFound"`, `grafana_auth_method: "token"`, no `http_status`

Both report `outcome: runtime_error` with `exit_code: 1`. Production was never contacted; the temporary service account and containers were removed, with volumes preserved. Raw-JSON ingestion demonstrably carries all three fields; typed columns for them do not exist yet, which is the follow-up described under the gates below rather than a blocker.

## Gates

1. **Receiver: not a blocker, by reading it.** `usage-stats` stores the whole event JSON in `report` and has no enum or schema validation on the way in (`pkg/routes/gcx.go` sets a fixed list of typed columns and ignores the rest), so nothing here can be rejected or silently dropped — the staging round-trip above proved that end to end. Typed columns for the three fields are still worth having, but they are a query-convenience follow-up, not a merge gate: A merged with its own `batch_*_bucket` and `dry_run` reaching only `report`, which is the same position.
2. **Governance: settled on 25 Aug.** Collecting the authentication *category* — never credential material — and the revision-3 notice wording are both approved. grafana/gcx-internal#2 was closed on 19 Aug, so this thread is where that record lives.
3. Human review on this SHA. B (#1109) merged on 4 Sep, so this is the last thing outstanding.


wardbekker (commented):
## Review — focus on user-visible side effects and data behaviour

I reviewed this PR together with #1108 and #1109, with attention to two questions: can the change interfere with the user, and does it send more than it claims. This comment covers the top commit of #1118, and then the combined effect of all three PRs, because the whole-stack view is where the remaining concerns live.

**Summary for #1118 itself: I found no leakage path, no user-visible behaviour change, no added network or file I/O, and no data race.** The three filters are real, single-path, and independent of the capture side. All findings on this PR alone are low severity.

### What I verified as clean

**Wire contract**
- The `k8s_reason` allowlist is enforced, and `telemetry.K8sReasonLabel` is the only reader-side mapper. `capture.CurrentK8sReason()` has exactly one production caller. No log, output, or agentlog path reads it.
- The allowlist matches the apimachinery **wire** values exactly — all 19 non-empty `StatusReason` constants in `k8s.io/apimachinery@v0.36.0`, including the `StatusReasonStoreReadError` → `"StorageReadError"` trap. Verbatim server strings and long strings collapse to `other`.
- `http_status` can carry a transport status only. All five `HTTPStatusCode() int` implementors derive from `resp.StatusCode`. The one body-derived status in the tree is `queryerror.APIError.StatusCode`, and `CaptureErrorSignals` deliberately reads `TransportStatus` only. Every `queryerror.New` call site passes a body-level status and correctly leaves `TransportStatus` zero. The 400-599 clamp at `cmd/gcx/telemetry.go:134` drops the 2xx and 3xx cases.
- `grafana_auth_method` cannot carry a URL, a username, a token, or a context name. The config side derives from `grafanaAuthMode.String()`, an enum, never from the raw configured string. The login side uses login-authored values, and `StagedContext.Grafana` is nil until `resolveGrafanaAuth` stages it, so the fallback cannot pick up a pre-existing config value. `GrafanaAuthMethodLabel` clamps again as defence in depth.
- No error message text is captured. `dynamic.ParseStatusError` puts `err.Error()` into `Status.Message`, and only `Status.Reason` is read.
- Suppression matches the description: `cmd/gcx/telemetry.go:133-138` suppresses both failure fields for exit 4 and exit 5 only. Exit 0 cannot carry a status, because `CaptureErrorSignals` runs only from `reportError` on a non-nil error.

**User-facing behaviour**
- Error text is byte-identical. `internal/providers/httperror.go` keeps all four format strings, and the only change is `%w` → `%v` for the read error, with the error preserved in `Cause`, so `errors.Is`/`Unwrap` reachability is unchanged. `fallbackWrappedSummary` uses `errors.Unwrap`, which returns nil for `HTTPStatusError{Cause: nil}` exactly as it did for the old `*errors.errorString`, so `DetailedError.Summary`/`Details`/`Parent` are identical on both paths.
- Exit codes cannot reroute. `serviceAPIError` is the only interface-shaped `errors.As` target in the converter chain and it needs three methods; `HTTPStatusError` has one. The two tripwire tests pin this. Every other converter matches a concrete type or a message string, and neither changed.
- No new network calls, no auth work, no token read, no OAuth refresh, no writes to config or credential files. The auth capture sits inside the existing pure `selectGrafanaAuth`, and `grafanaAuthMethodLabel` does no I/O.
- No JSON output change from the new struct fields: nothing in the repo marshals an `error` value.

**Concurrency**
- `gcx config check` genuinely writes the auth slot from up to 10 errgroup goroutines. The CAS loop in `SetGrafanaAuthMethod` is correct in every interleaving I traced: a lost `CAS(nil, …)` re-loads and conflicts, a `Force` racing a `Set` always wins, and conflict is sticky and discards the value, so `Current…` returns `""` deterministically. `checkContext` → `Validate` → `selectGrafanaAuth` is read-only on the shared `Context`.
- `httpStatus` (`atomic.Int64`) and `k8sReason` (`atomic.Pointer[string]`) are written only from `reportError` on the main goroutine.
- Nil-pointer risk in the new assertions: a typed-nil error would panic, but no producer exists. The same shape is already relied on by `convertServiceAPIErrors`/`convertAPIErrors`.
- `internal/gcxerrors` is a leaf, so the new edges create no cycle and violate no CONSTITUTION dependency rule.

### Findings on this PR — all low

**1. `grafana_auth_method: anonymous` is reported for an invocation which never sent a request.** `internal/config/grafana_auth.go:141-142` captures the label before `validateWithAuthSelection`, where the `server is required` check lives. Set `GRAFANA_ORG_ID=1` with no `grafana.server` anywhere: `PrepareForEnvParse` allocates `Grafana`, `parseEnvTags` sets `OrgID`, so `IsEmpty()` is false, `selectGrafanaAuth` finds no credential material and returns mode unknown with a nil error, and the label becomes `anonymous`. Validation then fails with `server is required` and exits 1, with `grafana_auth_method: anonymous` on the event. The published contract says anonymous means "requests went out unauthenticated", and no request went out. Not a leak — it inflates the `anonymous` bucket with misconfigured runs.

**2. The `grafana == nil` non-decision branch is unreachable in production.** `internal/config/envparse.go:21-28` always allocates `Grafana`, and `CleanupAfterEnvParse` only nils `TLS`. The real "no Grafana configuration" case therefore takes the `anonymous` branch, per finding 1, not the `""` branch. `TestGrafanaAuthCaptureUndecidedNeverErasesDecided` builds `config.Context{Name: "cloud-only"}` by hand, so it proves the rule against a shape production never produces.

**3. The shared writer-pin walker resolves the local import name from the directory.** `internal/testutils/capturewriters.go:78` uses `pkgPath[strings.LastIndex(pkgPath, "/")+1:]`. Correct for `internal/telemetry/capture`, but as shared infrastructure it fails open: for a package whose declared name differs from its directory, `calledPackageFuncs` returns nothing and `ElementsMatch` passes vacuously — "no writers found" reads as "no unauthorised writers". Not a regression, since the replaced code hardcoded `"capture"`. It is a latent trap now that two guards depend on it.

**4. `K8sReasonLabels()` returns map-iteration order.** `internal/telemetry/k8sreason.go:60-66`. Harmless today, because both consuming tests are order-insensitive, but any future snapshot or doc generation from it would flake.

---

## Combined effect of #1108 + #1109 + #1118

### Payload inventory

The wire payload is `json.Marshal(telemetry.Event)` with no envelope and no timestamp. Field count goes from **27 on `main` to 34** on the stacked branch. Measured compact sizes:

| Scenario | main | stacked | delta |
|---|---|---|---|
| Plain successful command | 405 B | ~434 B | +7% |
| `resources push`, success | 405 B | 538 B | **+33%** |
| Kubernetes auth failure | 428 B | 501 B | +17% |

Volume and network behaviour are conservative and I confirmed each point: exactly one event per invocation or none, no batching, no queueing, one attempt with no retries, no background goroutine, and the export is fully synchronous before `os.Exit`. The `User-Agent` carries no identifier beyond what the body already holds.

### The strongest remaining concern: `grafana_auth_method` composes with the whois enrichment — medium-high

The field is eligible on **every** outcome, including success and cancellation (`cmd/gcx/telemetry.go:139-144`, outside the failure guard). `anonymous` states that an organisation runs an unauthenticated Grafana. `basic` and `mtls` disclose deployment posture too.

Here is one event shape the code can produce, with nothing invented:

```json
{"command":"resources push","flags":"context,path","exit_code":3,
 "error_kind":"auth_failure","target_kind":"self-hosted","device_id":"3f2504e0-…",
 "batch_succeeded_bucket":"101-1000","batch_failed_bucket":"1","dry_run":false,
 "http_status":403,"k8s_reason":"Forbidden","grafana_auth_method":"anonymous"}
```

The receiver adds the whois network organisation name. A reader then concludes, without any resource name or hostname: *this named company runs a self-hosted Grafana with no authentication on the Grafana connection, an operator pushed 101-1000 resources into it live rather than as a dry run, and exactly one write was rejected 403/Forbidden by RBAC.*

That is a reconnaissance record about a named company's production posture, assembled entirely from fields each of which is defensible on its own. `dry_run: false` plus `command` is what upgrades "they have resources" to "they deploy to production from a laptop". No document in the stack reasons about the composition. Please raise this with whoever gives the governance approval, since #1108 and #1118 were each reasoned about separately.

### The bucket reasoning is sound for one event, but not across many — medium

`internal/telemetry/bucket.go:6-27` is honest, and it names the correct threat model. Two cases it does not cover:

1. **Longitudinal narrowing.** The comment reasons about a single event. With a persistent `device_id` over weeks, a transition from `101-1000` to `1001+` pins a threshold crossing to a date, and the distribution of buckets across many pulls narrows the true value well inside its bucket.
2. **Buckets interacting with the new failure fields.** `batch_failed_bucket: "1"` together with `http_status` and `k8s_reason` from the same event states exactly which single resource class failed and why, for a named organisation. Neither PR's reasoning considers the pair.

Ranges also still carry order-of-magnitude inventory: `gcx resources validate` sets `batch_succeeded_bucket` to the count of resources examined, so `1001+` from a whois'd organisation states that the organisation manages more than 1000 Grafana resources. The bucket hides the digit, not the magnitude.

### `duration_ms` stays exact and unbucketed — medium

`internal/telemetry/event.go:38-42` explicitly exempts it from the "no raw numbers" invariant. With `device_id`, `command`, and the arrival time, it gives a per-install wall-clock activity timeline with millisecond durations. It predates this stack, but the stack is where the "no raw numeric counts" rule is being written down, so the exemption deserves a stated reason.

### A blackholed endpoint costs the full second on every invocation — medium

`internal/telemetry/export.go:26-30` argues that an unreachable endpoint costs one fast failure rather than the full `exportTimeout`. That holds for TCP RST and NXDOMAIN. It does not hold for a silently dropped SYN behind a restrictive corporate egress firewall, or for a stalled DNS resolver — and that is precisely the environment a privacy-wary user runs in. In those cases every `gcx` invocation pays the full second. In a CI loop of hundreds of invocations that is minutes of added wall clock, in exactly the environment which never sees the disclosure. Neither mitigated nor documented.

### Documentation gaps

`docs/sources/anonymous-usage-statistics.md` is unusually thorough, and on field semantics it is accurate. All 34 fields are tabled, the three vocabularies are pinned against the Go source by `TestUsageStatsPageListsTheFullWireVocabularies`, and the awkward truths are stated rather than buried. I spot-checked `dry_run` for `validate` and `pull` against the code and both match. Three gaps:

1. **The first-run notice is not documented, and line 91 cross-references a section which does not describe it.** The Opt-out section describes the three env and config switches and never mentions a notice. So the page never states that the disclosure prints only on an interactive terminal, never in CI or agent mode, at most once per revision, gated by a file under `$XDG_STATE_HOME/gcx/`. A reader cannot learn that a CI runner is enrolled with no notice ever shown. This is also a dangling anchor, so it is mechanically detectable.
2. **The destination is never named.** `https://stats.grafana.org/gcx-usage-report` appears nowhere under `docs/`. Line 151 says only "Grafana's usage-stats service". A privacy-wary user cannot write a firewall rule from the documentation.
3. **The network cost is undocumented** — no mention of the 1-second timeout, that the export is synchronous before exit, or that a blackholed endpoint adds a second to every invocation. "Best-effort" reads as a reliability caveat, not a latency warning.

Smaller: `GCX_TELEMETRY_ENDPOINT` is absent from this page and is unvalidated in `export.go:33`, with no scheme check, so a poisoned environment can redirect the payload over plaintext HTTP. `is_tty` is under-specified — the code reads **stdout** while the notice gate reads **stderr**. The "Invocations that report nothing" list names a `telemetry` command which does not exist in the tree, although `root/telemetry.go:146` suppresses one by name.

### What the stack gets right, and should be preserved

Every server-controlled and user-controlled string on the three new paths is clamped to a fixed vocabulary **at the wire boundary in one place**, independently of the capture side. The `TransportStatus` versus `StatusCode` split keeps a body-level status inside an HTTP 200 off the wire. `noticeRevision` is the correct re-disclosure mechanism and was actually exercised. `docsdrift_test.go` ties the published vocabularies to the Go source, so a new value cannot ship undocumented. `device_id` is a random UUIDv4, mode 0600, and resets when the user deletes the file. The opt-out is genuinely total on the wire across all three new paths, including the new cancellation path, and an unrecognised opt-out value fails toward privacy.

The remaining objections are about defaults, the reach of the disclosure, composition across fields, and synchronous network cost. None of them is about a leaked value.

---

*Reviewed with automated analysis; every finding above was verified by reading the code at the PR head.*

wardbekker:
I pushed one commit to this branch, `f5abc621`, for the medium and higher findings that are documentation or test defects. Revert it if you prefer to fix these another way. Nothing in the commit changes what the code sends.

**1. How the report is sent — new section.** The page never named the destination, never said the export is synchronous, and never gave the one-second limit. A reader could not write a firewall rule from the documentation. The new section names `https://stats.grafana.org/gcx-usage-report`, states one POST with a single attempt and no retries, and points at `GCX_TELEMETRY_ENDPOINT`.

It also states the case the comment at `internal/telemetry/export.go:26-30` does not cover. "An unreachable endpoint costs one fast failure rather than the full `exportTimeout`" holds for TCP RST and NXDOMAIN. It does not hold for a silently dropped SYN, which is what a restrictive corporate firewall does — and that is the environment a privacy-wary user runs in. There, every invocation pays the full second. The section tells that reader to opt out instead of blocking the address.

**2. The one-time notice — new section.** The cancellation section pointed at Opt out "for the one-time notice", and Opt out never described a notice, so the page never stated how the disclosure reaches anyone. The new section gives the state file and the revision behaviour, then states two limits plainly: the notice is printed to an interactive terminal only, so a CI job or a coding agent reports without it ever appearing; and the invocation that prints it has already reported. I also repointed the cross-reference at the new section.

**3. The shared writer walker.** `internal/testutils/capturewriters.go` skipped only `.git`, `vendor`, and `node_modules`, so it walked the git worktrees under `.claude/worktrees` and found every writer twice. It now skips dot- and underscore-prefixed directories and any directory with its own `go.mod`. This is the same fix I pushed to #1108 for the inline walker. Note for the restack: #1108 now edits the inline version that this PR moves into `testutils`, so expect a conflict there — the resolution is to keep this PR's version, which already carries the fix.

`go build` passes, and `go test` passes for `./internal/telemetry/...`, `./cmd/gcx/`, `./cmd/gcx/resources/...`, and `./internal/testutils/...`, including `docsdrift_test.go`. `golangci-lint` reports 0 issues.

Three findings from my review need a decision rather than a patch, so I left them:

- **`grafana_auth_method` is eligible on every outcome, including success.** Combined with the whois organisation name, `anonymous` states that a named organisation runs an unauthenticated Grafana. I gave a worked example in the review.
- **The bucket reasoning covers one event, not a series.** With a persistent `device_id`, a transition from `101-1000` to `1001+` pins a threshold crossing to a date, and `batch_failed_bucket: "1"` beside `http_status` and `k8s_reason` says exactly which single resource class failed and why.
- **CI, agent, and non-TTY installs report with no disclosure.** That predates this stack, but this PR widens what those undisclosed events carry.

Each is a governance question about what may be collected, so it belongs with the approval this series is waiting on rather than in a commit from me.

dafydd-t:
@claude review

internal/telemetry/authmethod.go
```diff
@@ -0,0 +1,48 @@
...
+	AuthMethodAnonymous = "anonymous"
+	// AuthMethodUnknown is a selection that ran and could not classify the
+	// method, such as an unsupported configured auth-method value. A known
+	// method whose credential is merely invalid reports the method instead.
+	AuthMethodUnknown = "unknown"
+
+	authMethodOAuth = "oauth"
+	authMethodToken = "token"
+	authMethodBasic = "basic"
+	authMethodMTLS  = "mtls"
```
dafydd-t, line 20: mtls?

dafydd-t: huh, TIL

cmd/gcx/api/command.go
```diff
@@ -189,10 +190,19 @@ func outputResponse(cmd *cobra.Command, opts *apiOpts, resp *http.Response) erro
 	}
 
 	if resp.StatusCode >= 400 {
+		// Typed so the usage event can carry the status; the rendered text is
+		// unchanged, and the one-method carrier matches no error converter, so
+		// the exit code stays what the fallback path always produced.
```
dafydd-t:
```suggestion
```

awsome-o: Done — dropped the whole comment rather than just that line, since deleting the last line alone would have left the first two dangling. Same pass shortened the rest of the comments this PR adds; the one-method rule on `HTTPStatusError` is the one I kept, because implementing a second and third method there silently moves provider 401/403 call sites from exit 1 to exit 3.

internal/config/grafana_auth.go
```diff
@@ -87,19 +88,61 @@ func selectGrafanaAuth(grafana *GrafanaConfig, stack *StackConfig, contextName s
...
+// telemetry label — every capture write in this file routes through it, so a
+// policy change cannot apply to one branch and miss another. It returns ""
+// for the one state that is not a decision at all: a context with no Grafana
+// block never selected anything, and recording it would erase a decision an
+// earlier load made. An unsupported auth-method IS a decision — "unknown" —
+// and a valid selection with no credential material is "anonymous", a real,
+// distinct answer because requests then go out unauthenticated.
+func grafanaAuthMethodLabel(grafana *GrafanaConfig, selection grafanaAuthSelection, err error) string {
+	switch {
+	case grafana == nil:
```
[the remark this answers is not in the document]
awsome-o: Confirmed and fixed. `PrepareForEnvParse` does install an empty block that `CleanupAfterEnvParse` never nils back, so the arm the comment described was exactly the state being mislabelled as `anonymous`. The arm now reads `grafana == nil || grafana.IsEmpty()`, matching `Context.Validate`.

Two checks before taking it: a block with only `server` set is not empty, so it still reports `anonymous`; and `GRAFANA_TOKEN` still reports `token`, because its env tag populates `APIToken` and the block is no longer empty by the time the label is derived.

Covered by a new test that drives the real tolerant path (`ParseEnvIntoContext` on a context with no Grafana block) plus a row in the label table. Both fail against the old arm — the label row with "Should be empty, but was anonymous".

I did not extend the nil-only guard at `validatedGrafanaAuthSelection`. The label fix is enough for the telemetry behaviour, and changing what that guard rejects is a config-validation change that does not belong in a telemetry PR.

docs/sources/anonymous-usage-statistics.md
```diff
@@ -79,9 +79,47 @@ These fields are easy to misread, so the following constraints are part of the c
...
+- **Both failure fields are omitted for exit codes `4` and `5`.** A partial failure has no single causal status — the one that surfaced would stand in for many — and a canceled run is not a failure.
+- **`grafana_auth_method` describes Grafana connection authentication only**, never Grafana Cloud/GCOM authentication. Absent means no Grafana context was selected, or the invocation selected several different methods (`gcx config check` examines every context, so it reports no single method when they differ). `anonymous` means a valid selection with no credential material — requests went out unauthenticated. `unknown` means the configured method could not be classified. A selected method whose credential is invalid or rejected reports the method: which authentication category fails is precisely what this field exists to show.
+- **`grafana_auth_method` is present on every outcome** — success, failure, partial failure, and cancellation — whenever a Grafana context was selected. `gcx login` reports the method it actually resolved by probing, even when the login fails after authentication was resolved.
```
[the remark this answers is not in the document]
awsome-o: Added, on the page and in the godoc: absence also means the invocation never resolved a Grafana connection, so a command that builds no Grafana transport reports nothing even under a fully configured context — the same "absence proves nothing" the `http_status` bullet already carries.

I left the specific command out. The capture fires from `selectGrafanaAuth`, which a Cloud-only command reaches only if something validates the Grafana block (`configloader.go:521` does that only under `RequireGrafana`), and I would rather the page state the rule than name an example that a later provider change quietly falsifies.

awsome-o:
Restacked onto B's rebased head, so this branch now carries `main` at `fcec55b7` too. All checks pass and `GCX_AGENT_MODE=false mise run all` is green on the result.

Three conflicts from the move, all resolved the obvious way: `queryerror.APIError` keeps both `TransportStatus` and `main`'s new `CloudOnly`/`Experimental`; login's import block; and the usage page holds B's corrected second-Ctrl-C sentence next to this PR's cross-reference. Re-checked the two invariants a move that size could break — no new `HTTPStatusCode() int` implementor landed, and the four new `queryerror.New` sites from the new datasources carry plugin-level statuses through `New`, which sets no `TransportStatus`, so they emit no `http_status` exactly as documented.

Also in this push: the empty-Grafana-block fix (thread above — the arm now guards nil-or-empty like `Context.Validate`, with a test on the real `ParseEnvIntoContext` path that fails against the old one), the third absence cause on the page and in the godoc, your `gcx api` comment dropped, and the rest of the comments this series adds cut down.

Gates now read: receiver is not a blocker (raw JSON ingestion, proven by the staging round-trip), governance settled 25 Aug, and what's left is review plus B merging first.

dafydd-t (approved):

PR #1122, opened 2026-07-31
feat(profiles): add -o dot call-graph output to query

marcsanmi:
Adds `-o dot` to `gcx profiles query` (and `gcx datasources pyroscope query`): the profile as a Graphviz DOT caller-callee graph — plain text carrying full function names, `file:line` locations, self/cumulative values with percentages, and call edges.

## Why

`query`'s existing outputs each fail a consumer class:

- `-o json` returns the raw flame-graph — positional integer arrays meant for renderers. An agent reading it must know an undocumented quad encoding (`[offset, total, self, name-index]`) and decode it before it can answer "which function is hot".
- `-o table` answers that, but caps at 20 rows, truncates names at 60 chars (ambiguous for long Go symbols, unusable with `--stacktrace-selector`), and carries no source locations or call structure.

DOT is the format the **Grafana Assistant** already uses in production for LLM profile analysis (`SelectMergeStacktraces` + `PROFILE_FORMAT_DOT`, noise-stripped, fed straight into the model). This PR brings the same recipe to gcx, including the cleanup pass.

**Measured (blind A/B test, planted CPU bug in the rideshare example, ground truth `utility.mutexLock` at ~74%):** three independent agents, same question, no hints.

| Input | Verdict | Cost |
|---|---|---|
| flame-graph JSON, in-context | ✅ correct | **110k tokens, 131s** (hand-traced the quad arrays) |
| flame-graph JSON, code allowed | ✅ correct | 33k tokens, 82s, **first decoder script was wrong** and needed debugging |
| DOT, in-context | ✅ correct **+ file:line** | **26k tokens, 21s**, zero tooling |

Same answer, ~4× fewer tokens, ~6× faster, no decoding layer to get wrong — and the only format that hands the fixer a `file:line`.

## Behavior by backend

`PROFILE_FORMAT_DOT` is served by pure-v2 read paths (Grafana Cloud, helm-chart defaults). Other backends degrade gracefully — always something useful on stdout, never a hard failure:

| Backend | Result |
|---|---|
| pure v2 (`-architecture.storage=v2`) | cleaned DOT graph |
| `v1-v2-dual` (current OSS binary default; silently downgrades the format server-side) | ranked table + stderr hint, same request |
| v1 legacy (rejects the format) | one retry without it → ranked table + stderr hint |
| no matching samples | empty-window hint + `(no profile data)` |

## Usage

```console
$ gcx profiles query '{service_name="ride-sharing-app", region="eu-north"}' \
    --profile-type process_cpu:cpu:nanoseconds:cpu:nanoseconds --since 30m -o dot

digraph "main" {
...
N8 [label="utility\nmutexLock\nutility.go:32\n11.90s (1.29%)\nof 665.72s (72.13%)" ]
N6 -> N8 [label=" 665.72s" weight=75]
...
```

`--max-nodes` controls graph size (server default: 100-node graph from a 512-node source). Dotted edges mean intermediate frames were elided — raise `--max-nodes` for fuller chains. `--trace-id`, `--profile-id`, and `--stacktrace-selector` compose with `-o dot`; `--span-id` is rejected (it routes via the deprecated span RPC, which has no DOT).

## Notes

- The visual-noise cleanup (fontsize, node ids, tooltips, colors) mirrors the Assistant's and roughly halves the payload; profiling semantics are untouched.
- Verified live against both a pure-v2 and a dual backend (rideshare example); the v1 path is unit-tested.
- Groundwork for follow-ups: the `debug-with-grafana` skill's profiling section will adopt `-o dot` once this merges, and the agent-mode *default* output of `query` (currently the raw flame-graph JSON) is a candidate to revisit separately.


simonswine (dismissed):

wardbekker:
## Review notes

I read the diff against the existing query paths. The problem is real and the
package layout is correct: wire types and cleanup live in
`internal/query/pyroscope`, and the command wiring lives in
`internal/datasources/pyroscope`. `-o table` caps at 20 rows and truncates
names at 60 characters, so it cannot answer the call-graph question. The four
notes below are about the fallback path, not about the feature.

### 1. The v1 retry discards an explicit `--max-nodes`

`internal/datasources/pyroscope/query.go:314` sets `req.MaxNodes =
defaultMaxNodes` (50000) on the retry. `resolveMaxNodes` returned the user's
value on the first request. A user who runs `-o dot --max-nodes 300` against a
v1 backend gets 50000 nodes, not 300.

The retry can keep the resolved value. Only the `Format` field needs a reset.

### 2. The v1 rejection test matches an error string

`internal/datasources/pyroscope/query.go:129`:

```go
return err != nil && strings.Contains(err.Error(), "dot format is only supported")
```

The repository already carries a typed `*queryerror.APIError` with a status
code and a `Message` field. `IsParseError` uses it. An `errors.As` check
survives a change to the server message text.

### 3. `dotCodec.Encode` always returns an error

`internal/datasources/pyroscope/query.go:35-45` defines a codec whose `Encode`
cannot encode. `RunE` then writes the graph with `fmt.Fprintln` at line 327.

`internal/providers/appo11y/services/map.go:483` registers
`serviceMapDOTCodec`, which writes DOT from `Encode`. If `dotCodec.Encode`
calls `CleanDot(resp.Dot)`, the generic `opts.shared.IO.Encode` tail handles
the write, and the special branch in `RunE` goes away.

### 4. The tooltip patterns have no attribute bound

`internal/query/pyroscope/dot.go:10` uses `tooltip=".*?\)" ` and
`labeltooltip=".*?\)" `. Both patterns can cross the closing quote of the
attribute. A tooltip value without a closing parenthesis lets the match run
into the following attributes. `[^"]*` bounds each pattern to one attribute.
`color="#\w{6}"` also misses named colors and 3-digit hex colors.

I did not reproduce a failure for this note. Present Pyroscope output always
puts a parenthesis in the tooltip. This is a durability note.

### Summary

Note 1 is a defect that reproduces today. Notes 2 to 4 are durability and
consistency items. Notes 1 and 4 are small changes. Notes 2 and 3 are larger.

I made no commits on this branch.

marcsanmi:
@wardbekker 

Addressed 1, 2 and 4 in 1115cc7c. Kept the sentinel codec for 3: the response shape is backend-dependent and the fallbacks emit stderr hints, which Encode can't do; reasoning is recorded in a comment on dotCodec, lmkwyt.

simonswine (approved):

dafydd-t (approved):

PR #1125, opened 2026-08-01
fix(config): separate health from resource discovery

amckinley:
## Summary

- use authenticated `/api/health` to report base Grafana connectivity and version before resource discovery
- report Kubernetes-style API discovery separately, preserving its classified error, suggestions, and nonzero exit
- skip discovery after a failed health probe or an incompatible Grafana version
- document the `config check` contract and add ordered, bearer-authenticated `httptest` coverage

Previously, `gcx config check` treated `/apis` discovery as the connectivity check. A Grafana server could be healthy and report a supported version while returning 401 for `/apis`; the command then described the server as offline/unauthorized and hid its version.

After this change, that case reports connectivity online and the detected version, followed by a distinct resource-discovery failure. Discovery remains required, so the command still exits nonzero.

## Verification

- `go test ./cmd/gcx/config ./internal/grafana ./internal/config`
- `go test -race -count=1 ./...`

### DIFF
--- a/cmd/gcx/config/check_test.go
+++ b/cmd/gcx/config/check_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/grafana/gcx/internal/gcxerrors"
@@ -68,6 +69,7 @@ current-context: first
 	assert.Equal(t, 2, strings.Count(output, "Configuration:"), output)
 	assert.Equal(t, 2, strings.Count(output, "Connectivity:"), output)
 	assert.Equal(t, 2, strings.Count(output, "Grafana version:"), output)
+	assert.Equal(t, 2, strings.Count(output, "Resource discovery:"), output)
 }
 
 func TestCheckCommandPreservesCancellation(t *testing.T) {
@@ -95,7 +97,7 @@ current-context: target
 	assert.NotErrorIs(t, err, gcxerrors.ErrAlreadyReported)
 }
 
-func TestCheckCommandExitStatusTracksConnectivityAndVersion(t *testing.T) {
+func TestCheckCommandExitStatusTracksConnectivityVersionAndDiscovery(t *testing.T) {
 	clearConfigCheckEnvironment(t)
 
 	tests := []struct {
@@ -103,6 +105,7 @@ func TestCheckCommandExitStatusTracksConnectivityAndVersion(t *testing.T) {
 		handler    http.HandlerFunc
 		wantErr    bool
 		wantOutput string
+		wantExtra  string
 	}{
 		{
 			name: "success",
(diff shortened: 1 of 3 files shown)


cmd/gcx/config/check_test.go
```diff
@@ -181,3 +188,188 @@ func checkServerHandler(health func(http.ResponseWriter)) http.HandlerFunc {
...
+			wantHealth:    1,
+			wantDiscovery: 1,
+		},
+		{
+			name: "health failure skips discovery",
+			health: func(w http.ResponseWriter) {
+				http.Error(w, `{"message":"unauthorized"}`, http.StatusUnauthorized)
+			},
+			wantErr:  true,
+			wantExit: gcxerrors.ExitGeneralError,
```
dafydd-t, line 316: unauthorized => ExitGeneralError? why not ExitAuthFailure here? I think the same issue (bad credentials) should result in the same exit code for both API endpoints (/api/health and /apis)

docs/sources/configuration.md
```diff
@@ -131,10 +131,12 @@ Use these commands to check the configuration:
...
-The check covers every configured context before returning. It exits non-zero
-when the current context is invalid or any context fails configuration,
-authentication setup, connectivity, or Grafana version checks, so it is safe to
-use as a deployment gate.
+The check covers every configured context before returning. It uses the
+authenticated Grafana health endpoint for connectivity and version, then
+reports Kubernetes-style resource discovery separately. It exits non-zero when
+the current context is invalid or any context fails configuration,
+authentication setup, connectivity, version, or resource discovery checks, so
+it is safe to use as a deployment gate.
```
dafydd-t, line 139: worth noting that for oauth (grafana cloud only), the [client does not attach a token](https://github.com/grafana/gcx/blob/main/internal/grafana/client.go#L80), so its anonymous in that scenario, no authenticated. connectivity would pass even with an invalid credential here. Worth updating the docs here a bit I think

cmd/gcx/config/check_test.go
```diff
@@ -181,3 +188,188 @@ func checkServerHandler(health func(http.ResponseWriter)) http.HandlerFunc {
...
+			wantOutputs: []string{
+				"Connectivity:",
```
dafydd-t, line 318: Do we want a value in this one?

cmd/gcx/config/command.go
```diff
@@ -724,7 +725,8 @@ func checkContext(ctx context.Context, cfg *config.Config, gCtx *config.Context,
 		// future validation/transport drift rather than an expected branch.
 		cmdio.Error(stdout, "Configuration: %s", cmdio.Red(err.Error()))
 		cmdio.Warning(stdout, "Connectivity: %s", cmdio.Yellow("skipped"))
-		cmdio.Warning(stdout, "Grafana version: %s", cmdio.Yellow("skipped")+"\n")
+		cmdio.Warning(stdout, "Grafana version: %s", cmdio.Yellow("skipped"))
+		cmdio.Warning(stdout, "Resource discovery: %s", cmdio.Yellow("skipped")+"\n")
```
dafydd-t, line 729: these are getting repetitive enough that they could deserve a helper function, if you like

dafydd-t (changes requested):
Thanks a lot for this! great idea. I think we need to keep the exit codes consistent and update the doc, and then there are some nits you can decide on.

dafydd-t:
@claude review

dafydd-t:
With https://github.com/grafana/gcx/issues/1222, we will need to remove dependencies on `/api/health`, so I'm going to close this since it also seems inactive. Thanks a lot for making the PR

PR #1126, opened 2026-08-02
feat(skills): add integrate-with-gcx contributor guidance

awsome-o:
## Why

Teams adding a Grafana capability to gcx repeatedly have to rediscover the same decisions: whether a new command is needed, where it belongs, whether the backend is ready, and what contract the command exposes to agents.

This PR adds repo-local contributor guidance for making those decisions consistently before implementation and review.

### 1. New `integrate-with-gcx` contributor skill

Adds `.claude/skills/integrate-with-gcx/`, organized as a small router with progressively loaded references for:

- placement and backend readiness;
- the agent-facing command contract and proportional tests;
- implementation wiring and quality gates;
- diff-triggered self-review.

It treats every new command as an agent tool: contributors are guided through routing metadata, typed and validated inputs, machine-readable output, completeness signals, actionable errors, token-cost hints, and tests of the rendered command surface.

The workflow is:

```text
Place → contract → [add-provider | add-datasource] → Review
```

Terminal outcomes such as “not gcx” or a missing backend prerequisite stop before implementation. Otherwise the workflow is autonomous; it asks only when repository evidence cannot settle a material decision.

This is contributor guidance, not a user-facing gcx skill. It lives in the checkout under `.claude/skills/`, is linked from `AGENTS.md`, and is intentionally excluded from the Claude plugin and `gcx agent skills install` inventory.

### 2. Existing contributor skills aligned with the handoff

`add-provider` and `add-datasource` now accept placement and contract decisions from `integrate-with-gcx` instead of asking for them again. Their direct-entry paths use the same autonomous checkpoints, and stale implementation examples were corrected where they would otherwise teach contributors the wrong gcx pattern.

`migrate-provider` remains a separate, human-driven workflow and is not routed through this v1 skill. This PR only corrects existing migration guidance that was invalid, unsafe, or incompatible with the new repo-local validation.

### 3. Repo-local skills receive the same drift protection

The existing skill gates now cover both trees:

- `claude-plugin/skills/` — portable skills for gcx users;
- `.claude/skills/` — repository-only contributor skills.

Front matter is parsed for tracked skills in both trees, and documented `gcx` commands and flags are checked against the real command tree. The invocation test uses structural fixtures instead of hard-coded volume floors.

### 4. Canonical guidance corrected where the skills depend on it

The audit found several pre-existing documentation errors. The relevant source docs and examples now match the shipped code, including:

- provider registration through `TypedRegistrations()` and the single `providers.Register()` call;
- configuration and credential loading through `providers.ConfigLoader`;
- HTTP transport selection by the actual request URL, preventing Grafana credentials from being attached to a different product host;
- datasource query transport, codec dispatch, and generic-query wiring;
- current OnCall plugin-proxy behavior;
- migration parity checks that fail when either command fails instead of reporting a false match.

## Scope and risk

The diff is larger than one new skill because it also aligns the two skills that receive its handoff, adds validation for the repo-local skill tree, and repairs authoritative guidance those workflows depend on.

Almost all changes are Markdown, tests, or CI/developer validation. This PR does **not** change gcx command behavior, provider behavior, or the installed portable skill inventory.

Out of scope:

- distributing `integrate-with-gcx` to gcx users;
- making provider migration autonomous;
- changing production provider implementations;
- fixing the installer partial-write behavior tracked in #1128.

## Reviewer focus

The highest-value review questions are:

1. Is repo-local placement correct for contributor-only guidance?
2. Does the Place → contract → implementation → Review handoff preserve the right decisions without adding approval gates?
3. Is validating both tracked skill trees against the real command surface the right long-term guardrail?

## Verification

- `GCX_AGENT_MODE=false mise run all` — passed on `72cb7732` (all subtasks, including docs)
- `mise run validate-skills` — passed for both skill trees
- Invocation drift — 952 invocations across 44 portable files; 107 across 17 repo-local files
- `git diff --check` — clean
- Remote status — see GitHub Checks for the canonical result on the current head


annanay25 (approved):
Hey @awsome-o, gave this a pass again and I think this is worth a shot. Honestly its very hard to say how effective this will be unless tested in the field by our contributors when they add integrations/datasources. Happy to approve to get this merged and try it out though!

Issue #1061, opened 2026-07-24
credentials: locked/headless keychain failure is a dead-end error — cause detected (exit 154) but never surfaced


awsome-o:
Sibling of #1060 (the GRAFANA_TOKEN rescue bug). When the macOS login keychain is locked for the invoking session (headless box, or a process tree outside the unlocked security session — /usr/bin/security exits 154), any command needing a keychain-backed credential dies with:

```
configured credential "stack:<stack>" field "grafana-token" was rejected before
network use: the OS keychain could not be read; review the file and
re-authenticate with --config "<path>"
```

No mention of the keychain being locked, no suggestions, no docs link — and the one piece of advice it gives (re-authenticate) can't work, because login would need the same locked keychain to store the new credential. What makes it a proper tail-chaser is that keychain unlock is per security-session: unlocking in an SSH terminal doesn't help a differently-parented process tree, so "I unlocked it and it still fails" is the expected experience.

The cause is actually detected, then flattened three times on the way up:

1. darwinKeychainUnavailableExitCode (internal/credentials/keychain.go:140-153) explicitly documents exit 154 as the headless/locked-session case, but Open()'s probe failure returns unavailableStore{} (keychain.go:36-38, 157-161) whose ops return bare ErrUnavailable — the wrapped cause is dropped
2. resolveSentinelsForOwner maps any non-NotFound store error to the fixed literal "the OS keychain could not be read" (internal/config/keychain.go:972)
3. cmd/gcx/fail/convert.go has no converter for config.CredentialRejectedError, so it falls through to fallbackDetailedError (convert.go:1291) — no Suggestions, no DocsLink

There's already a good remediation hint in the codebase, but only on the write/migration path (keychain.go:1436-1446, "install or unlock your OS credential store...", via WarnUnavailableOnce) — the read path never sees it.

Fix sketch, three small threads:

1. internal/credentials: add an ErrKeychainLocked sentinel wrapped alongside ErrUnavailable for darwin exit 154 (and the Secret Service org.freedesktop.secret.error.islocked signature for Linux symmetry), and have unavailableStore carry the probe error instead of dropping it
2. internal/config/keychain.go:972: cause-aware reject reason — "the OS keychain is locked (headless or locked login session)" when errors.Is(err, ErrKeychainLocked)
[...]

Issue #964, opened 2026-07-13
🐛 gcx hangs indefinitely (silently) on headless Linux when the Secret Service collection is locked — even with a valid token in the config file


SP-graf:
### Description

```shell
# Headless Ubuntu server (SSH-only login), regular user.
# gnome-keyring / D-Bus Secret Service IS running on the user session bus,
# but the default collection is LOCKED — the normal state for headless/SSH logins:
$ busctl --user get-property org.freedesktop.secrets \
    /org/freedesktop/secrets/aliases/default org.freedesktop.Secret.Collection Locked
b true

# ~/.config/gcx/config.yaml contains a valid context with server + plaintext
# service account token (glsa_...), current-context set correctly.

# Every command that loads config hangs FOREVER with ZERO output, even at
# maximum verbosity:
$ gcx -vvv api /api/org      # hangs indefinitely, no output
$ gcx login                  # same
$ gcx datasources list       # same
$ gcx traces query ...       # same

# Argument-parsing errors still print instantly, so the process is alive —
# it never reaches the network (strace shows no connect(2) syscalls at all).
```

### What did you expect to happen?

With a valid plaintext token already present in the config file, commands should work immediately — either by not touching the keyring at all, or by treating a locked collection as "keyring unavailable" and falling back to the plaintext token (as the v0.4.0 changelog's "plaintext fallback when the keychain is unavailable" implies). At minimum, `-vvv` should say the process is waiting on the Secret Service, and the wait should have a timeout.

### How can we reproduce this?

1. Headless Ubuntu server with systemd user session + gnome-keyring installed; log in via SSH only, so the default Secret Service collection is present but **locked** (verify with the `busctl` command above).
2. Put a valid context with a plaintext token in `~/.config/gcx/config.yaml`, set `current-context`.
3. Run `gcx api /api/org` (or any command) — hangs forever, no output at any verbosity.
4. Same box, same command, with the session bus hidden — returns immediately:
   ```shell
   DBUS_SESSION_BUS_ADDRESS=disabled: gcx api /api/org
   ```
5. Control: on another headless box with NO secret service at all (root user), the keyring lookup fails fast and gcx falls back to the config token fine. The hang requires *secret service present + collection locked* — i.e. any systemd-user Ubuntu server with gnome-keyring installed. Likely affects CI runners and servers broadly.

### Root cause (verified via strace + source)

[...]

Issue #955, opened 2026-07-10
🐛 config view hangs ~4 minutes when D-Bus secret-service activation times out (v0.4+)


mikaelol:
### Description

Every command hangs for ~4 minutes before returning when gcx runs in an environment where the D-Bus `org.freedesktop.secrets` service is activatable but never finishes starting (e.g. a headless container with `gnome-keyring` installed but no login session to unlock it). The command eventually succeeds — it just blocks first. This started in v0.4.0. v0.3.0 is instant.

```shell
$ time gcx config view
dbus-daemon: Activating service name='org.freedesktop.secrets' requested by ... comm="gcx config view"
dbus-daemon: Failed to activate service 'org.freedesktop.secrets': timed out (service_start_timeout=120000ms)
contexts:
  a:
    grafana:
      server: https://example.invalid
      token: "**REDACTED**"
      ...
real    4m0.230s
user    0m0.108s
sys     0m0.096s
```

The 4 minutes = 2 × 120s D-Bus service-activation timeouts (gcx makes the secret-service call, D-Bus times out activating the service, gcx retries once, second 120s timeout), after which it falls back to plaintext.

### What did you expect to happen?

`gcx config view` (and other commands) should return immediately. Keyring/secret-service access should use a short timeout and fall back to plaintext quickly when the service is unavailable or slow — as implied by the v0.4.0 changelog ("plaintext fallback when the keychain is unavailable"). A read-only command like `config view` arguably shouldn't touch the keychain at all.

### How can we reproduce this?

1. Build the image below, `docker build -t gcx-repro .` (headless container with an activatable-but-hanging
   org.freedesktop.secrets, mimicking a keyring daemon that never registers)
2. Then:
   ```bash
   # Hangs ~4 min:
   docker run --rm gcx-repro
   
   # Instant — same image, workaround applied:
   docker run --rm gcx-repro bash -lc \
     'export DBUS_SESSION_BUS_ADDRESS=unix:path=/dev/null; time gcx config view'
   ```
   The only difference between the two is whether `DBUS_SESSION_BUS_ADDRESS` is pointed at a dead path, bypassing the broken secrets bus.

<details>
<summary>Dockerfile</summary>

```Dockerfile
FROM ubuntu:24.04

RUN apt-get update && apt-get install --no-install-recommends -y \
    ca-certificates \
    curl \
    dbus-x11 \
    && rm -rf /var/lib/apt/lists/*

RUN GCX_VERSION=0.4.3 \
[...]

PR #1127, opened 2026-08-03
Verify keychain credential writes

Maurice-L-R:
## Summary

- verify a new keychain credential can be read back before replacing plaintext config with its sentinel
- roll back the staged write and leave the YAML unchanged when the keychain reports a successful write but cannot resolve the new entry
- add regression coverage for the successful-`Set`/missing-`Get` failure mode

## Validation

- `go test ./internal/config`

### DIFF
--- a/internal/config/keychain.go
+++ b/internal/config/keychain.go
@@ -1350,6 +1350,20 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding, 
 				account: boundRef.Account, previouslySet: false,
 				owner: owner, field: field,
 			})
+			stored, err := txn.store.Get(boundRef.Account)
+			if err != nil {
+				txn.log.Warn("could not verify keychain entry after write",
+					"owner", owner,
+					"field", string(field),
+					"error", err.Error())
+				return credentials.BoundReference{}, false, fmt.Errorf("verify keychain entry for %q field %q after write: %w", owner, field, err)
+			}
+			if stored != value {
+				txn.log.Warn("keychain entry did not round-trip after write",
+					"owner", owner,
+					"field", string(field))
+				return credentials.BoundReference{}, false, fmt.Errorf("verify keychain entry for %q field %q after write: stored value did not match", owner, field)
+			}
 			return boundRef, true, nil
 		}
 		if errors.Is(err, credentials.ErrUnavailable) {
--- a/internal/config/keychain_bound_internal_test.go
+++ b/internal/config/keychain_bound_internal_test.go
@@ -16,19 +16,20 @@ import (
 )
 
 type boundTestStore struct {
-	entries       map[string]string
-	getErr        error
-	setErr        error
-	setErrValue   string
-	setFailAt     int
-	setCalls      int
-	deleteErr     error
-	deleteThenErr bool
-	deleteFailAt  int
-	deleteCalls   int
-	gets          []string
-	sets          []string
-	deletes       []string
+	entries        map[string]string
+	getErr         error
+	getErrAfterSet error
+	setErr         error
+	setErrValue    string
+	setFailAt      int
+	setCalls       int
+	deleteErr      error
+	deleteThenErr  bool
+	deleteFailAt   int
+	deleteCalls    int
+	gets           []string
+	sets           []string
+	deletes        []string
 }
 
 type boundTestLogger struct {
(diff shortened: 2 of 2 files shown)


internal/config/keychain.go
```diff
@@ -1350,6 +1350,20 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding,
 				account: boundRef.Account, previouslySet: false,
 				owner: owner, field: field,
 			})
+			stored, err := txn.store.Get(boundRef.Account)
+			if err != nil {
```
awsome-o, line 1354: This returns an error for any read failure, including a store that simply isn't reachable. Two branches just above treat that same case as acceptable for a brand new credential and fall back to keeping it in the file ([L1344](https://github.com/grafana/gcx/blob/1f1dca3044aef6a57aaae5af7e6eb8ba7ac0c599/internal/config/keychain.go#L1344) and [L1369](https://github.com/grafana/gcx/blob/1f1dca3044aef6a57aaae5af7e6eb8ba7ac0c599/internal/config/keychain.go#L1369)). So if the store goes away between the save and this read, the whole config write fails, and the user loses any unrelated edits made in the same save.

The write is already recorded at [L1349](https://github.com/grafana/gcx/blob/1f1dca3044aef6a57aaae5af7e6eb8ba7ac0c599/internal/config/keychain.go#L1349) before this check, so failing here goes into `rollback()` at [L1391](https://github.com/grafana/gcx/blob/1f1dca3044aef6a57aaae5af7e6eb8ba7ac0c599/internal/config/keychain.go#L1391), which only logs a failed delete. If the store is unreachable for both, the credential stays in the OS store with nothing pointing at it.

Worth flagging that the obvious fix doesn't work. Treating unreachable as fallback here would leave that recorded write behind, because `commit()` never clears `txn.writes`, only `rollback()` does, so you'd end up with the credential in the OS store and in the file at the same time. Cleanup has to be attempted and its result handled, similar to what `commit()` already does with `rollbackComplete`.

internal/config/keychain.go
```diff
@@ -1350,6 +1350,20 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding,
...
+				txn.log.Warn("could not verify keychain entry after write",
+					"owner", owner,
+					"field", string(field),
+					"error", err.Error())
+				return credentials.BoundReference{}, false, fmt.Errorf("verify keychain entry for %q field %q after write: %w", owner, field, err)
+			}
+			if stored != value {
```
awsome-o, line 1361: This is the part the title is about, and no test reaches it. `boundTestStore.Set` always stores exactly the value it's given ([L71](https://github.com/grafana/gcx/blob/1f1dca3044aef6a57aaae5af7e6eb8ba7ac0c599/internal/config/keychain_bound_internal_test.go#L71)), so nothing can simulate a save that reports success but keeps a different value. Teaching the fake to do that would get this branch covered.

internal/config/keychain_bound_internal_test.go
```diff
@@ -1104,6 +1114,9 @@ func TestBoundKeychainGenericStoreFailuresLeaveDiskAndCallerUntouched(t *testing
 
 			err := Write(context.Background(), ExplicitConfigFile(path), cfg)
 			require.Error(t, err)
+			if tt.name == "write is not retrievable" {
```
awsome-o, line 1117: The assertion is keyed to the case name, so it silently stops running if the case gets renamed. A per case expected error field would match how the rest of the table works. The new case also covers the value going missing after the save, but not the store being unreachable, which is the case that shifts current behaviour the most.

awsome-o (commented):
Thanks for raising the fix, could you share what did you run into that led to this? Curious whether you saw the OS report a credential as saved and then gcx couldn't read it back.

We have a few open issues in the same area (#955, #964, #1060, #1061) around the credential store being slow, locked, or unreadable.

If what you hit is part of that pattern, it probably wants a bigger fix rather than a check on this one path. If your case is specific and solid, happy to merge this once the notes below are sorted.

One thing that makes me unsure: this reads back in the same run, right after saving, while the problems in #1060 and #1061 surface on the next command. So I'm not sure it would catch those.

Also worth updating the write path section in `docs/architecture/config-system.md`, which still describes the old behaviour.

Maurice-L-R:
Superseded by #1130, which uses an upstream branch and incorporates the review feedback on confirmed cleanup, mismatch coverage, table-driven assertions, and architecture documentation.

PR #1130, opened 2026-08-03
Handle unavailable keychain verification safely

Maurice-L-R:
## Summary

- verify new keychain generations before replacing config with a sentinel
- if a new credential is missing or unreadable immediately after `Set`, delete it first and safely retain plaintext rather than persisting an unresolvable sentinel
- fail closed for mismatched or unknown readback failures, and preserve transaction cleanup ownership when deletion cannot be confirmed
- add regression coverage for mismatched writes and unavailable post-write reads; document the transaction behavior

## Validation

- `go test ./internal/config`
- `go test ./...`

This addresses the review feedback on #1127. The fork-based PR will be closed in favor of this upstream-branch PR.


awsome-o (dismissed):

awsome-o (commented):
A few nitpicks inline, none blocking.

internal/config/keychain.go
```diff
@@ -1366,6 +1395,27 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding,
...
+// It is used only before the YAML replacement, when a verification failure
+// makes the new generation unusable. Do not remove the transaction record
+// unless Delete succeeded (or confirmed the account is absent): otherwise the
+// caller's rollback must retain ownership of the cleanup attempt.
+func (txn *keychainWriteTransaction) discardLastStagedWrite() error {
+	if len(txn.writes) == 0 {
+		return errors.New("discard keychain entry after verification failure: no staged write")
+	}
+	write := txn.writes[len(txn.writes)-1]
+	if err := txn.store.Delete(write.account); err != nil && !errors.Is(err, credentials.ErrNotFound) {
```
awsome-o, line 1408: Nitpick: this branch has no test. It plus the two `errors.Join` returns above (L1360 and L1377) never execute in the suite, and that path is the behaviour the third summary bullet describes. A case that makes the verify read and the Delete fail together would cover it, asserting the returned error names both failures and that the staged write stays owned by rollback.

docs/architecture/config-system.md
```diff
@@ -277,7 +277,12 @@ target source before encoding. Its keychain reconcile pass is mutation-aware:
   unrelated writes, while an explicit set/unset can repair or remove it.
 
 Credential writes are a staged transaction. A new or rotated value is written
-under a fresh random generation before the config is replaced. The new config
+under a fresh random generation and read back before the config is replaced. If
+that immediate verification finds a brand-new generation missing or the store
```
awsome-o: Nitpick: "brand-new generation" reads as though rotations are included, since the sentence above says a new or rotated value also gets a fresh random generation. The code only falls back for a credential with no prior keychain reference. Matching the wording further down, "brand-new credential with no prior keychain reference", would remove the ambiguity.

internal/config/keychain.go
```diff
@@ -1350,6 +1350,35 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding,
...
+					"error", err.Error())
+				verifyErr := fmt.Errorf("verify keychain entry for %q field %q after write: %w", owner, field, err)
+				if cleanupErr := txn.discardLastStagedWrite(); cleanupErr != nil {
+					return credentials.BoundReference{}, false, errors.Join(verifyErr, cleanupErr)
+				}
+				// A newly created credential has no prior reference to preserve. If
+				// the store cannot read it back (or has already lost it) and cleanup
+				// is confirmed, keep this one value in the config instead of writing
+				// a sentinel that the next command cannot resolve.
+				if errors.Is(err, credentials.ErrUnavailable) || errors.Is(err, credentials.ErrNotFound) {
```
awsome-o, line 1367: Nitpick: on the `ErrNotFound` arm the caller sets `plaintextFallback`, so the user ends up seeing "credential store unavailable; credentials remain in plaintext on disk". In that case the store was available and just lost the value, so the message points at the wrong cause for the exact scenario this handles.

internal/config/keychain.go
```diff
@@ -1350,6 +1350,35 @@ func (txn *keychainWriteTransaction) stageBoundSet(binding credentials.Binding,
...
+					return credentials.BoundReference{}, false, nil
+				}
+				return credentials.BoundReference{}, false, verifyErr
+			}
+			if stored != value {
+				txn.log.Warn("keychain entry did not round-trip after write",
+					"owner", owner,
+					"field", string(field))
+				verifyErr := fmt.Errorf("verify keychain entry for %q field %q after write: stored value did not match", owner, field)
```
awsome-o, line 1376: Nitpick, optional: this error has no sentinel and no remediation text, so someone who hits it sees "stored value did not match" with nothing to act on. A shared sentinel would also let the two verify messages stop repeating the same prefix.

internal/config/keychain.go
Maurice-L-R: Changed it and added coverage for a verification-read failure plus an initial cleanup failure. The test asserts both errors are returned and that rollback retries cleanup of the staged write.

docs/architecture/config-system.md
Maurice-L-R: Changed it to say “brand-new credential with no prior keychain reference,” so it no longer implies rotations can use the plaintext fallback.

internal/config/keychain.go
Maurice-L-R: Changed it to a cause-neutral warning: the credential store could not securely store the credential. That covers both an unavailable store and a successful write that cannot be read back.

awsome-o (approved):

PR #1132, opened 2026-08-04
feat(skills): add performance-rca skill for profiling-led root cause analysis

marcsanmi:
Adds a profiling RCA workflow for CPU, memory, and contention problems. Agents compare qualified incident and baseline workloads, locate application stacks, and corroborate mechanisms with symptom evidence. The report distinguishes an established cause, suspected cause, cost location, and an inconclusive result.

- Six numbered investigation rules, explicit exits, and a fixed report preserve scope, evidence, and uncertainty. Empty profiles do not imply service health, and missing baselines do not support regression claims.
- Scoped series discovery identifies label combinations; grouped metrics locate workload differences; DOT and pprof comparisons inspect application paths. Guidance covers profile units, self/cumulative cost, normalization, pruning, and trace correlation.
- Supplied anomaly results can seed an investigation. The workflow requires no anomaly detector API and makes no production availability claim.
- Profiles advertise both performance-rca and debug-with-grafana; typed Pyroscope commands advertise performance-rca. Routing tests cover inheritance and the override.

The workflow is read-only. Code changes, rollbacks, load tests, and proactive optimization are outside its scope.

## Validation

- Rebased onto main with the series command included. Both commits retain their original changes; no skill wording changes were needed.
- Ran `GCX_AGENT_MODE=false mise run --continue-on-error all`: the full Go suite, skill validation, build, install/linter tests, reference generation, and documentation build passed. The command drift gate now validates `gcx profiles series` successfully, without an allowlist exception.
- Lint initially reused cached results pointing to a deleted worktree. Rerunning lint with a fresh `GOLANGCI_LINT_CACHE` passed with zero issues. Generated references have no drift, and the working tree is clean.
- Earlier synthetic execution covered 18 documented examples, request mappings, DOT/table fallback, empty profiles, and pprof comparison. These checks do not establish production backend availability.

### DIFF
--- a/claude-plugin/skills/debug-with-grafana/SKILL.md
+++ b/claude-plugin/skills/debug-with-grafana/SKILL.md
@@ -14,7 +14,9 @@ description: >
   cause", "did the rollout cause it", "which endpoint owns the most 5xx",
   "did retries pile up", or any request to analyse an earlier incident
   window using telemetry. For authoring dashboards use create-dashboard;
-  for dashboard inventory use manage-dashboards.
+  for dashboard inventory use manage-dashboards. When the evidence points
+  at the service's own code — CPU, allocations, lock contention — use
+  performance-rca to locate the responsible function.
 ---
 
 # Debug with Grafana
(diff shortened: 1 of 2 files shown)


radiohead (commented):
@marcsanmi I wonder if https://github.com/grafana/skills is a better fit for this skill, since it is broader than just "how to use gcx".

@dafydd-t what do you think?

PR #1136, opened 2026-08-05
feat(login): add --oauth-manual for remote hosts without a reachable callback

### LINKED ISSUE #1135: can't finish browser based oauth if gcx is not running on localhost
I'm running the gcx auth on a remote box. The callback with code cannot reach that gcx instance, as the browser based portion is running on my own mac, not the remote box terminal.

wardbekker:
Fixes #1135

## Problem

The OAuth callback server listens on `127.0.0.1` on the host that runs gcx. A
user who runs `gcx login` over SSH cannot finish the login, because the browser
on their own computer cannot open that address.

`gcx cloud login` is worse: it has no `--oauth-callback-port` flag, so a planned
port forward could not help that step at all.

A different bind address cannot fix this. I checked the `grafana-assistant-app`
plugin source:

- `apps/plugin/src/pages/tunnelUrls.ts` hard-codes host `127.0.0.1` and only
  accepts a port between 1024 and 65535.
- `apps/plugin/src/pages/useAuthFlow.ts` finishes with
  `window.location.href = callbackUrl`.

The second fact is the fix. The browser does a real navigation, so the complete
redirect URL stays in the address bar when the connection fails.

## Change

**The SSH case now needs no flag and no restart.** When gcx detects an SSH
session, has a usable `/dev/tty`, and is not in agent mode, it keeps the
callback server listening **and** reads a pasted redirect URL at the same time.
gcx takes whichever route completes first:

```
Note: gcx runs in an SSH session.
The browser on your computer cannot open the callback address on this host.
Do one of these two steps. gcx accepts the one that completes first.

  1. Add a port forward to this SSH session. Press Enter, then press ~C,
     then type this line:
       -L 54321:127.0.0.1:54321

  2. Approve the login in the browser on your computer. The browser then
     goes to an address that does not load. Copy that address and paste it
     here.

Redirect URL (or wait for the browser):
```

The hint leads with the OpenSSH `~C` escape, which adds a port forward to the
running session. That needs no reconnect and no hostname, and gcx keeps waiting,
so the login completes as soon as the browser reaches the port. A pasted URL
that does not work re-prompts instead of ending the flow, because the callback
server is still up.

**`--oauth-manual` covers the rest.** The flag starts no callback server and
only reads the pasted URL. It serves scripts and terminals that have no
`/dev/tty`. It works on `gcx login` and on `gcx cloud login`. It implies
`--oauth`, and it conflicts with `--oauth-callback-port` and with `--token`.

Where gcx runs over SSH but cannot read a terminal, it falls back to a static
hint with the `ssh -L` command and `--oauth-manual`.

### Files

- `internal/terminal/remote.go` — `IsRemoteSession()` detects an SSH session.
- `internal/auth/callback.go` — one shared callback-parameter handler for the
  HTTP callback and the paste path, so the state check, the PKCE exchange, and
  endpoint validation cannot diverge. This removes about 90 duplicated lines.
- `internal/auth/paste.go` — the watcher that reads pasted URLs while the
  callback server still listens.
- `internal/auth/manual.go` — redirect-URL parsing, a line reader that does not
  over-read the TTY, and the instruction text.
- Manual mode uses a fixed callback port `54321`, the first port of the normal
  auto-pick range `54321-54399`, so the GCOM `redirect_uri` shape does not
  change. The auto-paste path keeps the port that the listener picked, and the
  hint prints that port.
- The interactive login menu offers the manual mode, and puts it first for a
  Cloud target in an SSH session.
- One choice covers the Cloud follow-up step too.

### Design notes on the paste watcher

The watcher reads a separately opened `/dev/tty`, never `os.Stdin`. Go keeps the
standard streams out of its poller, so a blocking read on `os.Stdin` cannot be
cancelled, and a stale reader would compete with the `huh` prompts that run
after login.

Never call `File.Fd` on that file. `Fd` puts the descriptor back into blocking
mode, the read then blocks inside the syscall rather than the poller, and
`Close` can no longer stop it. `TestOpenPasteTerminalReadIsCancellable` is the
regression guard.

The watcher parses and delivers; it never prints. The caller owns the terminal,
so one goroutine writes every message and the two cannot interleave. The watcher
also keeps reading after a delivery, because the caller runs the semantic checks
(state, code, token exchange) and re-prompts when they fail. That re-prompt
needs a live reader.

## Security notes

The pasted URL holds a single-use authorization code and the state value. It
does not hold the PKCE `code_verifier` or a token, so the code alone cannot mint
a token. gcx reads it from a prompt, so it never enters shell history. gcx
prints a reminder to clear the terminal, and a test pins that no error message
echoes the pasted string.

`TestPasteWatcherKeepsReadingAfterCallerRejection` pins the reader that makes
that guarantee hold. A watcher that stopped after the first delivery would leave
the re-prompt with no reader. The next pasted line would stay in the terminal
buffer, and the shell would read it once gcx exited — which writes the
authorization code into the shell history.

## How to test

1. SSH into a remote host and build gcx there.
2. Run `gcx login --server https://<stack>.grafana.net --oauth`.
3. Confirm the two-option SSH prompt appears.
4. Open the printed URL in the browser on your own computer, check the
   verification code, and approve.
5. Confirm the browser fails to load `http://127.0.0.1:54321/callback?...`.
6. Copy the address, paste it into the remote terminal, and confirm gcx signs
   in.
7. Repeat step 2. This time press Enter, press `~C`, then type
   `-L 54321:127.0.0.1:54321`. Confirm gcx completes over the callback with no
   paste.
8. Run `gcx dashboards list` to confirm the token.
9. Repeat with `gcx cloud login`, then run `gcx cloud stacks list`.
10. Run `gcx login --oauth-manual` and confirm gcx starts no callback server.
11. Check the error paths: paste a URL from a second login attempt (state
    mismatch), paste plain text (parse error), and press Ctrl-D (EOF error).
    Confirm each one re-prompts rather than ends the flow.

I ran the whole paste flow end to end against a local fake GCOM server. The
token exchange sent `redirect_uri=http://127.0.0.1:54321/callback`, byte
identical to the authorize request, and gcx reported the authenticated user.

`GCX_AGENT_MODE=false mise run gate` passes (lint, tests, build). `mise run
docs` cannot run locally because `mkdocs` is absent; CI covers it.

### DIFF
--- a/cmd/gcx/cloud/command.go
+++ b/cmd/gcx/cloud/command.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
@@ -19,10 +20,11 @@ import (
 )
 
 type loginOpts struct {
-	oauthURL   string
-	apiURL     string
-	scopes     []string
-	cloudToken string
+	oauthURL    string
+	apiURL      string
+	scopes      []string
+	cloudToken  string
+	oauthManual bool
 }
 
 type gcomOAuthFlow interface {
@@ -39,6 +41,7 @@ func (opts *loginOpts) bindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&opts.oauthURL, "oauth-url", "https://grafana.com", "Base URL for the OAuth login flow (used only by this command)")
 	flags.StringVar(&opts.apiURL, "api-url", "https://grafana.com", "Base URL for Grafana Cloud API resource calls (stacks etc.)")
 	flags.StringSliceVar(&opts.scopes, "scope", auth.DefaultGCOMScopes(), "OAuth2 scopes to request")
+	flags.BoolVar(&opts.oauthManual, "oauth-manual", false, "Complete browser OAuth without a local callback server: gcx prints the URL, then reads the redirect URL that you copy from the browser address bar. Use this when gcx runs on a remote host and the browser runs on your own computer")
 }
 
 func (opts *loginOpts) Validate() error {
@@ -50,6 +53,9 @@ func (opts *loginOpts) Validate() error {
 	}
 	if opts.cloudToken == "" && len(opts.scopes) == 0 {
 		return errors.New("--scope must not be empty for interactive OAuth login")
+	}
+	if opts.cloudToken != "" && opts.oauthManual {
+		return errors.New("--oauth-manual has no effect with --cloud-token")
 	}
 	return nil
 }
@@ -170,7 +176,7 @@ both preserves the explicit OAuth-origin/API-destination pair.`,
 			if opts.cloudToken != "" {
 				return runTokenLogin(mutationCtx, opts, mutationSource, contextName, cloudSafety, mutationGuard)
 			}
-			return runOAuthLogin(mutationCtx, opts, mutationSource, contextName, cloudSafety, mutationGuard)
+			return runOAuthLogin(mutationCtx, opts, mutationSource, contextName, cloudSafety, mutationGuard, cmd.InOrStdin(), cmd.ErrOrStderr())
 		},
 	}
 
@@ -256,15 +262,19 @@ func runOAuthLogin(
 	contextName string,
 	cloudSafety config.CloudMutationSafety,
 	mutationGuard config.LoginMutationGuard,
+	stdin io.Reader,
+	stderr io.Writer,
 ) error {
-	fmt.Fprintln(os.Stderr, "Warning: interactive OAuth login is experimental. It stores an OAuth-issued token in the cloud entry's oauth-token field.")
-	fmt.Fprintln(os.Stderr, "Some commands that talk to grafana.com do not yet work with an OAuth token. For full functionality, use --cloud-token with a Cloud Access Policy token.")
+	fmt.Fprintln(stderr, "Warning: interactive OAuth login is experimental. It stores an OAuth-issued token in the cloud entry's oauth-token field.")
+	fmt.Fprintln(stderr, "Some commands that talk to grafana.com do not yet work with an OAuth token. For full functionality, use --cloud-token with a Cloud Access Policy token.")
 
 	flow := newGCOMOAuthFlow(auth.GCOMOptions{
 		ClientID: defaultClientID,
 		GCOMURL:  opts.oauthURL,
 		Scopes:   opts.scopes,
-		Writer:   os.Stderr,
+		Writer:   stderr,
+		Manual:   opts.oauthManual,
+		Reader:   stdin,
 	})
 
 	result, err := flow.Run(ctx)
@@ -280,15 +290,15 @@ func runOAuthLogin(
 		}
 	}
 
-	fmt.Fprintf(os.Stderr, "Authenticated as %s (%s)\n", result.Info.Login, result.Info.Email)
-	fmt.Fprintf(os.Stderr, "Scopes: %s\n", result.Scope)
+	fmt.Fprintf(stderr, "Authenticated as %s (%s)\n", result.Info.Login, result.Info.Email)
+	fmt.Fprintf(stderr, "Scopes: %s\n", result.Scope)
 
 	entry := cloudEntryFromOAuthResult(opts, result)
 	contextName, entryName, err := config.SaveCloudConfigGuarded(ctx, source, contextName, entry, cloudSafety, mutationGuard)
 	if err != nil {
 		return err
 	}
-	fmt.Fprintf(os.Stderr, "Token saved to cloud entry %q (context %q)\n", entryName, contextName)
+	fmt.Fprintf(stderr, "Token saved to cloud entry %q (context %q)\n", entryName, contextName)
 	return nil
 }
 
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -123,16 +123,13 @@ func (opts *loginOpts) Validate(args []string) error {
 			},
 		}
 	}
-	return nil
-}
-
-// applyOAuthManualImplication makes --oauth-manual a complete selection, so
-// every gate that inspects OAuth sees the fresh-credential intent. It must run
-// before the auto-local credential gates in runLogin.
-func applyOAuthManualImplication(opts *loginOpts) {
+	// --oauth-manual is a complete selection on its own. Fold it into OAuth
+	// here, after the conflict checks above, so the opts are self-consistent for
+	// every gate that reads OAuth later.
 	if opts.OAuthManual {
 		opts.OAuth = true
 	}
+	return nil
 }
 
 // Command returns the `login` Cobra command.
@@ -183,9 +180,6 @@ Auth sources (for non-interactive use):
 //nolint:gocyclo,maintidx // Login deliberately keeps trust preflight, auth selection, and retry setup in one auditable flow.
 func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 	ctx := cmd.Context()
-	// --oauth-manual is a complete selection on its own. Fold it into OAuth
-	// before any gate below reads flags.OAuth.
-	applyOAuthManualImplication(flags)
 	// The auto-discovered-config credential rejection below is the earliest gate
 	// that can end a login, and it runs before any config is read — so there is
 	// nothing but the flags to go on. Record what they ask for here; the
--- a/cmd/gcx/login/command.go
+++ b/cmd/gcx/login/command.go
@@ -19,6 +19,7 @@ import (
 	"github.com/grafana/gcx/internal/gcxerrors"
 	"github.com/grafana/gcx/internal/login"
 	cmdio "github.com/grafana/gcx/internal/output"
+	"github.com/grafana/gcx/internal/terminal"
 	"github.com/grafana/grafana-app-sdk/logging"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -51,6 +52,7 @@ type loginOpts struct {
 	Yes                 bool
 	AllowServerOverride bool
 	OAuthCallbackPort   int
+	OAuthManual         bool
 	OrgID               int
 }
 
@@ -72,6 +74,7 @@ func (opts *loginOpts) setup(flags *pflag.FlagSet) {
 	flags.BoolVar(&opts.Yes, "yes", false, "Non-interactive: skip optional prompts and use defaults")
 	flags.BoolVar(&opts.AllowServerOverride, "allow-server-override", false, "Allow re-pointing an existing context at a different server URL")
 	flags.IntVar(&opts.OAuthCallbackPort, "oauth-callback-port", 0, "Fixed local port for the OAuth callback server (default: auto-pick from 54321-54399). Useful when only specific ports are forwarded between a remote host and your browser")
+	flags.BoolVar(&opts.OAuthManual, "oauth-manual", false, "Complete browser OAuth without a local callback server: gcx prints the URL, then reads the redirect URL that you copy from the browser address bar. Use this when gcx runs on a remote host and the browser runs on your own computer. Implies --oauth")
 	flags.IntVar(&opts.OrgID, "org-id", 0, "Grafana organization ID (defaults to 1 for on-prem)")
 }
 
@@ -92,7 +95,7 @@ func (opts *loginOpts) Validate(args []string) error {
 			},
 		}
 	}
-	if opts.OAuth && opts.Token != "" {
+	if (opts.OAuth || opts.OAuthManual) && opts.Token != "" {
 		return gcxerrors.DetailedError{
 			Summary: "conflicting authentication methods",
 			Details: "--oauth and --token are mutually exclusive. OAuth authenticates via browser; --token uses a service account token.",
@@ -110,7 +113,26 @@ func (opts *loginOpts) Validate(args []string) error {
 			Details: fmt.Sprintf("Port must be between 1 and 65535 (or 0 to auto-pick); got %d.", opts.OAuthCallbackPort),
 		}
 	}
+	if opts.OAuthManual && opts.OAuthCallbackPort != 0 {
+		return gcxerrors.DetailedError{
+			Summary: "conflicting OAuth callback options",
+			Details: "--oauth-manual completes the flow without a callback server, so there is no port to fix with --oauth-callback-port.",
+			Suggestions: []string{
+				"Use --oauth-manual alone and paste the redirect URL",
+				"Or use --oauth-callback-port <port> alone and forward that port with ssh -L",
+			},
+		}
+	}
 	return nil
+}
+
+// applyOAuthManualImplication makes --oauth-manual a complete selection, so
+// every gate that inspects OAuth sees the fresh-credential intent. It must run
+// before the auto-local credential gates in runLogin.
+func applyOAuthManualImplication(opts *loginOpts) {
+	if opts.OAuthManual {
+		opts.OAuth = true
+	}
 }
 
 // Command returns the `login` Cobra command.
@@ -161,6 +183,9 @@ Auth sources (for non-interactive use):
 //nolint:gocyclo,maintidx // Login deliberately keeps trust preflight, auth selection, and retry setup in one auditable flow.
 func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 	ctx := cmd.Context()
+	// --oauth-manual is a complete selection on its own. Fold it into OAuth
+	// before any gate below reads flags.OAuth.
+	applyOAuthManualImplication(flags)
 	// The auto-discovered-config credential rejection below is the earliest gate
 	// that can end a login, and it runs before any config is read — so there is
 	// nothing but the flags to go on. Record what they ask for here; the
@@ -361,6 +386,8 @@ func runLogin(cmd *cobra.Command, flags *loginOpts, args []string) error {
 			CloudOAuthURL:               cloudOAuthURL,
 			UseOAuth:                    flags.OAuth,
 			OAuthCallbackPort:           flags.OAuthCallbackPort,
+			OAuthManual:                 flags.OAuthManual,
+			Reader:                      cmd.InOrStdin(),
 			Yes:                         flags.Yes,
 			OrgID:                       flags.OrgID,
 			Writer:                      cmd.ErrOrStderr(),
@@ -1116,6 +1143,12 @@ func askCloudAuth(ctx context.Context, e *login.ErrNeedInput, opts *login.Option
 		// change requires a freshly supplied credential, never a keep choice.
 		hasUsableExisting = false
 	}
+	// The Cloud step reuses the manual choice made for the stack step, so the
+	// label states where the browser runs instead of adding a fifth entry.
+	oauthLabel := "OAuth (browser)"
+	if opts.OAuthManual {
+		oauthLabel = "OAuth (browser on another computer)"
+	}
 	if hasUsableExisting {
 		label := "Keep the existing Cloud Access Policy token (recommended)"
 		if existingKind == login.CloudCredentialOAuth {
@@ -1123,10 +1156,10 @@ func askCloudAuth(ctx context.Context, e *login.ErrNeedInput, opts *login.Option
 		}
 		options = append(options,
 			huh.NewOption(label, choiceKeep),
-			huh.NewOption("OAuth (browser)", choiceOAuth),
+			huh.NewOption(oauthLabel, choiceOAuth),
 		)
 	} else {
-		options = append(options, huh.NewOption("OAuth (browser) (recommended)", choiceOAuth))
+		options = append(options, huh.NewOption(oauthLabel+" (recommended)", choiceOAuth))
 	}
 	options = append(options,
 		huh.NewOption("Paste a Cloud Access Policy token", choiceToken),
@@ -1197,6 +1230,10 @@ func runCloudOAuth(ctx context.Context, opts *login.Options) error {
 		GCOMURL:  oauthURL,
 		Scopes:   internalauth.DefaultGCOMScopes(),
 		Writer:   opts.Writer,
+		// One manual choice covers both the stack step and the Cloud step:
+		// the browser is on another computer for both.
+		Manual: opts.OAuthManual,
+		Reader: opts.Reader,
 	}
 	var flow login.CloudAuthFlow = internalauth.NewGCOMFlow(flowOpts)
 	if opts.NewCloudAuthFlow != nil {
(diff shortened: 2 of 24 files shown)


wardbekker:
## Update: no flag needed for an interactive SSH login

The first revision told an SSH user to stop gcx and re-run it with `--oauth-manual`. It now does the work for them.

When gcx detects an SSH session and is not in agent mode, it keeps the callback server listening **and** reads a pasted redirect URL. The `select` takes whichever completes first, so both routes stay live and neither needs a restart:

```
Note: gcx runs in an SSH session.
The browser on your computer cannot open the callback address on this host.
Do one of these two steps. gcx accepts the one that completes first.

  1. Add a port forward to this SSH session. Press Enter, then press ~C,
     then type this line:
       -L 54321:127.0.0.1:54321

  2. Approve the login in the browser on your computer. The browser then
     goes to an address that does not load. Copy that address and paste it
     here.

Redirect URL (or wait for the browser):
```

Option 1 uses the OpenSSH `~C` escape, which adds a forward to the **running** session — no reconnect and no hostname. A URL that does not work re-prompts instead of ending the flow, because the callback server is still up.

`--oauth-manual` stays for scripts and for terminals with no `/dev/tty`.

## The trap this uncovered

The watcher reads a separately opened `/dev/tty`, never `os.Stdin`. Go keeps the standard streams out of its poller, so a blocking read on `os.Stdin` cannot be cancelled, and a stale reader would compete with the `huh` prompts that run after login.

The first attempt still hung. `openPasteTerminal` called `term.IsTerminal(int(tty.Fd()))`, and **`File.Fd` puts the descriptor back into blocking mode**. The read then blocks inside the syscall rather than the poller, so `Close` can no longer stop it. Once the browser callback arrived, gcx hung forever in the deferred teardown. A `SIGQUIT` dump under a pty showed the reader parked in `syscall.Syscall(SYS_READ)` instead of `netpollblock`.

The `IsTerminal` check was redundant anyway — `/dev/tty` is the controlling terminal by definition, and the open fails when the process has none.

`TestOpenPasteTerminalReadIsCancellable` guards it. I verified it fails (in 5s, with the diagnostic message) when the `Fd` call is restored, and passes without it. It skips where there is no controlling terminal.

## Verification

Driven under a real pty against a local fake GCOM server:

- **Paste wins**: the exchange used `redirect_uri=http://127.0.0.1:54322/callback`, matching the bound port, and gcx authenticated.
- **Callback wins**: gcx completed in 0.20s. Before the `Fd` fix, the same run hung indefinitely.

`GCX_AGENT_MODE=false mise run gate` passes (lint 0 issues, all tests, build). `mise run docs` cannot run locally because `mkdocs` is absent; CI covers it.

internal/auth/paste.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (p *pasteWatcher) run() {
+	defer close(p.done)
+
+	for {
+		line, err := readLine(p.tty)
+		if err != nil {
+			// The terminal closed, or the flow ended. Stop quietly: the caller
+			// already has a result or an error of its own.
```
dafydd-t: this returns on every `readLine` error, not just the `Close` one the comment describes. Ctrl-D gives EOF → `no input received` → the goroutine ends silently, but `Redirect URL (or wait for the browser):` is still on screen and the select loop is still reading a channel nobody writes to.

could we select on `p.stop` first, then deliver recoverable errors as `pastedInput{Err: ...}` so the existing `Reject` path re-prompts? unless Ctrl-D is meant to abandon the paste route — in which case it should say so before it stops, and step 11 of the test plan needs a tweak

internal/auth/paste.go
```diff
@@ -0,0 +1,189 @@
...
+// unblocks its read at once, so the wait is normally instant. The bound is a
+// backstop: a future regression that makes the file blocking must degrade to a
+// short delay, never to a login that hangs forever.
+const closeGrace = 2 * time.Second
+
+// Close stops the watcher and releases the terminal. Closing a pollable file
+// unblocks the pending read, and Close waits for the reader goroutine to end.
+// The terminal therefore has exactly one reader again before login continues to
+// the prompts that follow.
+func (p *pasteWatcher) Close() {
```
dafydd-t, line 143: `errCh` and `ctx.Done()` in both select loops can fire while this prompt is live. `Close` shuts the tty and gcx exits, so anything typed-but-not-newline-terminated stays in the input queue and the shell reads it next — the same shell-history leak the security notes are about.

narrow window, you'd have to be mid-paste when an unrelated error lands. but a `TCIFLUSH` here before the descriptor goes away closes it, and closes the >8KB residue case in `run` too

internal/auth/flow.go
```diff
@@ -137,6 +186,56 @@ func (f *Flow) Run(ctx context.Context) (*Result, error) {
...
+			if pasted.Err != nil {
+				paste.Reject(pasted.Err)
+				continue
+			}
+			result, cerr := handleCallbackParams(ctx, pasted.Values, state, codeVerifier)
+			if cerr != nil {
+				paste.Reject(pasteRejection(cerr.err))
+				continue
+			}
+			fmt.Fprintln(f.writer, manualCallbackHygieneNotice)
```
dafydd-t: all four `manualCallbackHygieneNotice` calls sit on a success return — here, `flow.go:160`, `gcom.go:145`, `gcom.go:213` — so every failure path skips it.

the case that leaves the most on screen is in your own test plan: step 11, paste a URL from an earlier attempt, state mismatch. under `--oauth-manual` that ends the flow with the code in the scrollback and no warning, then the user re-runs and pastes again. it's also missed here when a port forward wins the race after someone has already pasted.

IMO this wants tying to "a URL was pasted" rather than "the flow succeeded"

internal/auth/flow.go
```diff
@@ -101,11 +112,56 @@ func NewFlow(endpoint string, opts Options) *Flow {
...
+		}
+		return f.runManual(ctx)
+	}
+	return f.runWithCallbackServer(ctx)
+}
+
+// runManual completes the flow without a callback server. The browser cannot
+// reach the callback address, so the user copies the redirect URL out of the
+// address bar and pastes it here.
+func (f *Flow) runManual(ctx context.Context) (*Result, error) {
```
dafydd-t, line 136: `runManual` reads once and returns on the first error, so there's no re-prompt on this path — that only exists on the auto-paste side via `Reject` + `continue`. `TestFlowRun_ManualReportsMissingInput` pins the current behaviour, so the code and tests agree and it's the description that's out of step (step 11 says each error re-prompts).

I'd add the loop rather than fix the description: a typo currently costs a full re-run for a fresh state and code, while the code you were about to paste is still on the clipboard. same in `gcom.go:117`

internal/auth/gcom.go
```diff
@@ -145,44 +185,61 @@ func (f *GCOMFlow) Run(ctx context.Context) (*GCOMResult, error) {
...
-}
 
-func (f *GCOMFlow) startGCOMCallbackServer(ctx context.Context, listener net.Listener, expectedState, codeVerifier, redirectURI string, resultCh chan<- *GCOMResult, errCh chan<- error) *http.Server {
-	return newCallbackServer(listener, errCh, func(w http.ResponseWriter, r *http.Request) {
-		state := r.URL.Query().Get("state")
-		if state != expectedState {
-			errCh <- errors.New("invalid state - possible CSRF attack")
-			renderErrorPage(w, "Invalid state parameter")
-			return
+	for {
```
dafydd-t: this loop and `flow.go:213-233` are the same thing with a different result type, and `runManual` is duplicated the same way — about 45 lines against the ~90 the `callback.go` extraction removed.

the part that came back is the concurrency, and every fix in this review lands twice. a shared helper taking `handle func(url.Values) (any, *callbackError)`, or generics on the result type, collapses both pairs. happy for it to be a followup

internal/auth/manual.go
```diff
@@ -0,0 +1,218 @@
...
+			return "", fmt.Errorf("failed to read the redirect URL: %w", err)
+		}
+	}
+}
+
+// readLineContext runs readLine in a goroutine so a cancelled context ends the
+// wait. A terminal read is not interruptible in Go, so the read goroutine
+// stays blocked until the process exits. The channel is buffered, so that
+// goroutine can never block on the send.
+func readLineContext(ctx context.Context, r io.Reader) (string, error) {
```
dafydd-t, line 127: worth a line on why manual mode is fine reading `os.Stdin` when `paste.go` goes to real lengths to avoid it. I think the answer is that the reader is injectable and may not be a terminal at all, plus `readLine` is byte-at-a-time so it can't over-read — but the comment right above says the goroutine stays blocked until the process exits, and that reads like a contradiction of the watcher's whole argument unless it's spelled out

internal/auth/manual.go
```diff
@@ -0,0 +1,218 @@
...
+		return res.line, res.err
+	case <-ctx.Done():
+		return "", ctx.Err()
+	}
+}
+
+// printRemoteSessionPreamble states why the browser cannot reach the callback
+// address. Every remote-session message opens with it. It prints nothing, and
+// reports false, for a local session.
+func printRemoteSessionPreamble(w io.Writer) bool {
```
dafydd-t: prints and returns a predicate, and the two callers use the return value for different things. `terminal.IsRemoteSession()` already answers the question — could `printRemoteSessionHint` ask it directly and leave this one just printing?

cmd/gcx/login/command.go
```diff
@@ -92,7 +95,7 @@ func (opts *loginOpts) Validate(args []string) error {
 			},
 		}
 	}
-	if opts.OAuth && opts.Token != "" {
+	if (opts.OAuth || opts.OAuthManual) && opts.Token != "" {
 		return gcxerrors.DetailedError{
 			Summary: "conflicting authentication methods",
 			Details: "--oauth and --token are mutually exclusive. OAuth authenticates via browser; --token uses a service account token.",
```
dafydd-t: names a flag the user didn't type when they passed `--oauth-manual --token`

```suggestion
			Details: "--oauth, --oauth-manual and --token are mutually exclusive. OAuth authenticates via browser; --token uses a service account token.",
```

internal/terminal/remote.go
```diff
@@ -0,0 +1,31 @@
...
+
+// IsRemoteSession reports whether gcx runs in an SSH session.
+//
+// The OAuth callback server listens on the loopback address of the host that
+// runs gcx. A browser on a different computer cannot open that address, so the
+// login flow needs either a forwarded port or the manual paste mode.
+//
+// The result is deliberately not cached: callers read it at most twice per
+// process, and tests set the environment with t.Setenv.
+func IsRemoteSession() bool {
```
dafydd-t, line 30: nit: a detached `tmux`/`screen` session inherits stale `SSH_*` from whatever created it, so this can report true on a local terminal. harmless, since the paste route is additive and the callback still works — but worth a line in the doc comment

cmd/gcx/cloud/command.go
```diff
@@ -51,6 +54,9 @@ func (opts *loginOpts) Validate() error {
 	if opts.cloudToken == "" && len(opts.scopes) == 0 {
 		return errors.New("--scope must not be empty for interactive OAuth login")
 	}
+	if opts.cloudToken != "" && opts.oauthManual {
+		return errors.New("--oauth-manual has no effect with --cloud-token")
```
dafydd-t, line 58: nit, and it matches its neighbours so feel free to ignore: `gcx login` returns a `DetailedError` with suggestions for the equivalent conflict

dafydd-t (commented):
the `callback.go` extraction preserves every check in both paths. plugin: state → error param w/ `StripControlChars` → empty code → empty endpoint → `ValidateEndpointURL` → exchange → instance endpoint parse → https scheme → all 8 `Result` fields. GCOM: state → error param → empty code → exchange with `redirectURI`. set and order preserved in both, and the absent endpoint validation on the GCOM side is right since GCOM sends no endpoint. only difference is the old GCOM handler called `StripControlChars` twice for the same value.

two things I'd want before this goes in, both cheap:

- the paste watcher dies silently on any read error, so Ctrl-D leaves a prompt on screen that accepts nothing
- the clear-your-terminal notice only fires on success, so it misses the failure case that leaves the most on screen

`runManual` having no retry loop and `Close` not draining the tty are the same decision applied to the other two paths — I'd do all four together.

wardbekker:
## Shepherd status (iteration 2)

### Continuous integration checks

All checks pass on `7cf5809c`. There are no failures.

### Corrections that the shepherd applied

The corrections answer the review of @dafydd-t, and one defect that the review
did not report.

**Round one.**

- `fix(auth): correct the paste reader response to empty input` — the reader
  treated a bare Enter as a pasted URL. Step 1 of the instructions asks the user
  to press Enter before the SSH escape `~C`, so the recommended route answered
  `That URL did not work: no URL supplied`. The reader now drops an empty line.
  The review did not report this defect.
- The same commit answers the `readLine` report. The reader now sends `Closed`
  when the read ends, which Ctrl-D and a broken terminal both cause. The caller
  states that the paste route ended, and it keeps waiting for the callback. The
  reader checks `stop` before it acts on a read error, so `Close` still ends it
  quietly. A retry loop is wrong here, because a terminal that reports the same
  error forever would make the loop spin.
- `refactor(auth): share one paste flow between the two OAuth flows` —
  `runManualPaste` and `awaitCallbackOrPaste` in `callback.go` now hold the one
  implementation. `checkCallbackBasics` holds the state check, the error
  parameter, and the code check. `flow.go` and `gcom.go` each lose 33 code
  lines. Every later correction to the paste route lands one time.
- The same commit ties `manualCallbackHygieneNotice` to "a URL reached the
  screen", not to "the flow succeeded". All four call sites sat on a success
  return, so every failure path skipped the notice.
  `TestFlowRun_ManualRejectsForeignState` now pins the notice on the state
  mismatch, which is the case that leaves the most on screen.
- `fix(login): name --oauth-manual in the token conflict message` — the message
  named `--oauth` only, and the condition covers `--oauth-manual`.
  `TestOAuthTokenConflictNamesEveryFlag` pins both flags.
- `refactor(auth): remove dead code from the callback helpers` — the `Error` and
  `Unwrap` methods of `callbackError` had no caller. `ErrStateMismatch` is now
  `errStateMismatch`, because only `pasteRejection` reads it and the two sit in
  one package.

**Round two.** `fix(auth): guard the code exchange and flush the terminal input`.

- Only one route may exchange the authorization code. `exchangeGuard` grants the
  exchange with a compare-and-swap. The claim runs after the state check and the
  code check, never before: a URL from an older attempt must not consume the
  claim that the real callback needs.
  `TestGuardClaimFollowsTheStateCheck` pins that order.
- The loser gets `errExchangeClaimed`. The paste branch stays silent. The HTTP
  handler renders the success page, and it sends nothing to `errCh`. The old
  code sent an error there, which could end a flow that had already succeeded.
- `Close` now discards the terminal input queue. A terminal in canonical mode
  delivers a line only after a newline, so a read cannot reach text that the
  user typed without one. The shell reads that text once gcx exits, which puts a
  partly typed redirect URL into the shell history.
- `flushTerminalInput` has one file for each platform family. Linux uses
  `TCFLSH` with `TCIFLUSH`. BSD and darwin use `TIOCFLUSH` with `FREAD`. Every
  other platform does nothing. All of them use `SyscallConn`, never `File.Fd`.
- `Close` ends the pending read with a read deadline in the past, not with a file
  close. A closed descriptor accepts no ioctl.
- `go.mod` moves `golang.org/x/sys` from indirect to direct. The version does not
  change, and `go.sum` does not change.

The gate passed after each commit. `go test ./...` passes with
`GCX_AGENT_MODE=false`. `golangci-lint` reports 0 issues on the changed
packages. The package builds for linux, darwin, and windows.
`mise run reference` produces no drift.

### Items that need a human review

These reports remain open. The author and the reviewer must decide on them.

- `runManual` has no retry loop, so a typo costs a full re-run. The reviewer
  asked for the loop. The shepherd did not add it, because it needs a decision
  about the paste route on a stream that is not a terminal.
- `readLineContext` in `manual.go` reads `os.Stdin`. A cancelled context leaves
  that goroutine blocked until the process exits. This is the hazard that
  `paste.go` removes with `/dev/tty`.
- `printRemoteSessionPreamble` prints and returns a predicate. The two callers
  use the return value for different reasons.
- `terminal.IsRemoteSession` reports true inside a detached `tmux` or `screen`
  session that inherited a stale `SSH_*` variable.
- `gcx cloud login` returns a plain error for the `--oauth-manual` conflict,
  where `gcx login` returns a `DetailedError` with suggestions.

### Items that need a decision from you

None. The pull request needs one approval.

---
*Updated by shepherd-pr*

internal/auth/paste.go
wardbekker: You are right about the old code. It returned on every `readLine` error, and it left the prompt on screen with no reader behind it.

The reader now reports the end of the paste route. `run` sends `pastedInput{Closed: true}`, and `awaitCallbackOrPaste` prints `The paste route ended. gcx still waits for the browser.` Ctrl-D therefore ends the paste route with a message. The callback server stays live, so the port forward route still completes the login.

I did not take the `pastedInput{Err: ...}` variant that you proposed. A terminal that reports the same error on every read would make the loop repeat without an end. The reader checks `stop` before it acts on a read error, so `Close` still ends the goroutine quietly.

Step 11 of the test plan keeps its meaning. A state mismatch is a semantic failure, not a read error, so it still goes through `Reject` and re-prompts.

Tell me if you want the error variant for a named set of recoverable errors. I can add it here.

internal/auth/paste.go
wardbekker: Corrected in `fix(auth): guard the code exchange and flush the terminal input`.

`Close` now discards the terminal input queue before it closes the file. A terminal in canonical mode delivers a line only after a newline, so a read cannot reach text that the user typed without one. The shell reads that text once gcx exits, which is the leak that the security notes describe.

`flushTerminalInput` has one file for each platform family:

- `flush_linux.go` uses `TCFLSH` with `TCIFLUSH`.
- `flush_bsd.go` covers the BSD family and darwin. It uses `TIOCFLUSH` with `FREAD`.
- `flush_other.go` does nothing on every other platform.

All three use `SyscallConn`, never `File.Fd`. `Close` ends the pending read with a read deadline in the past, not with a file close, because a closed descriptor accepts no ioctl.

The same flush covers the remainder of a line longer than 8192 bytes in `run`.

internal/auth/flow.go
wardbekker: Corrected. The notice now depends on "a URL reached the screen", not on "the flow succeeded".

- `runManualPaste` prints the notice with `defer`, directly after the read of the line. Every return below it therefore prints the notice, and the state mismatch of step 11 is one of them.
- `awaitCallbackOrPaste` sets `pasteSeen` and prints the notice in a deferred function. This also covers the case that you name: the callback server delivers a result after the user pasted a URL.

`TestFlowRun_ManualRejectsForeignState` checks the notice on the state mismatch.

internal/auth/flow.go
wardbekker: This report is still open. Your description of the state is correct: `runManualPaste` reads one line, then it returns on the first error. Only the auto-paste side re-prompts.

I did not add the loop yet, for one reason. The manual route reads an injectable reader, which is often not a terminal. `--oauth-manual` exists for scripts, and a loop on a stream that reports the same error forever would repeat without an end.

Two options:

1. Add the loop with a bound of three tries. A read error ends it at once. A parse failure or a semantic failure re-prompts.
2. Correct step 11 of the description, and keep the single read.

I prefer option 1, because a typo now costs a full re-run for a fresh state and a fresh code. Confirm it, and I add it in this pull request. Both flows call `runManualPaste`, so `gcom.go:117` gets the change at the same time.

internal/auth/gcom.go
wardbekker: Corrected in `refactor(auth): share one paste flow between the two OAuth flows`.

`callback.go` now holds the one implementation. `runManualPaste` and `awaitCallbackOrPaste` are generic on the result type, and they take a `paramHandler[T]` for the semantic checks and the token exchange. `checkCallbackBasics` holds the state check, the error parameter check, and the code check.

`flow.go` and `gcom.go` each lose 33 code lines, and both call the shared helpers. Every later correction to the paste route applies one time. The two later commits in this pull request show the benefit: the hygiene notice and the exchange guard each needed one edit.

internal/auth/manual.go
wardbekker: This report is still open, and your reading of the reason is correct. Three facts make the manual route safe:

- `readLineContext` takes an injectable reader. It is often a script pipe, not a terminal.
- `readLine` reads one byte at a time, so it cannot read past the newline. No later prompt loses input to it.
- The manual route runs no callback server, so no second route competes for the same input.

The comment above `readLineContext` states the blocked goroutine, but not the reason why that is acceptable. I will add the three facts to that comment, and I will name the contrast with `openPasteTerminal`.

Tell me if you want the comment in this pull request or in a later one.

internal/auth/manual.go
wardbekker: This report is still open. The function does two things now. `printInstructions` in `paste.go` needs the print and drops the return value. `printRemoteSessionHint` needs the return value as a guard.

I agree that these two tasks must stay separate. `printRemoteSessionHint` can call `terminal.IsRemoteSession()` and return early. `printRemoteSessionPreamble` then only prints, and it returns nothing.

Confirm the change, and I make it in this pull request.

cmd/gcx/login/command.go
wardbekker: Applied in `fix(login): name --oauth-manual in the token conflict message`. The message now names all three flags, and `TestOAuthTokenConflictNamesEveryFlag` checks both OAuth flags in the text. Thank you for the suggestion.

internal/terminal/remote.go
wardbekker: Correct. A detached `tmux` or `screen` session inherits a stale `SSH_CONNECTION` or `SSH_CLIENT` from the session that created it. `IsRemoteSession` then reports true on a local terminal.

The result is harmless, as you say. The paste route only adds a second option, and the callback still works, so the user sees one note too many.

I will add that limit to the doc comment.

cmd/gcx/cloud/command.go
wardbekker: Correct. `gcx login` returns a `DetailedError` with suggestions for the equivalent conflict, and this check returns a plain error.

Every other check in this `Validate` also returns a plain error. A change to this one line therefore makes the function inconsistent inside itself. I plan to convert all four checks in one later change, so `gcx cloud login` matches `gcx login`.

Tell me if you want the conversion in this pull request.

internal/auth/flow.go
wardbekker: Done in `88e4d5e5`. I took option 1.

`runManualPaste` now accepts up to three URLs. `manualPasteTries` holds the bound. A parse failure or a semantic failure re-prompts. A read error ends the flow at once, because the next read reports the same error.

One detail that the loop made visible: a script supplies one URL, so the read after a failed URL reports the end of the input. That message says less about the cause, so the loop returns the failure of the URL instead. A cancelled context still returns the context error.

`printPasteRejection` now holds the reject message for both routes, so the manual route and the watcher print one wording. The state mismatch message drops "run the command again", because the loop asks for another URL in the same run.

Three tests cover the change:

- `TestFlowRun_ManualRetriesAfterABadURL` — a typo, then the correct URL, then success. The exchange runs one time.
- `TestFlowRun_ManualStopsAtTheTryBound` — five bad lines end with an error, and two rejections reach the screen.
- `TestFlowRun_ManualReportsTheURLFailureNotTheEndOfInput` — one bad URL reports the URL.

Both flows call `runManualPaste`, so `gcom.go` gets the same behaviour.

internal/auth/manual.go
wardbekker: Done in `88e4d5e5`. The doc comment of `readLineContext` now states the three facts, and it points at `openPasteTerminal` for the contrast:

```go
// A blocked reader on r is safe here, which is why this path may read os.Stdin
// while the paste watcher goes to /dev/tty for the same job. Three facts make
// the difference:
//
//   - r is injectable, and for --oauth-manual it is often a script pipe rather
//     than a terminal.
//   - readLine reads one byte at a time, so it cannot read past the newline. A
//     later prompt on the same stream loses no input to it.
//   - Manual mode runs no callback server, so no second route competes for the
//     same stream.
```

internal/auth/manual.go
wardbekker: Done in `88e4d5e5`. `printRemoteSessionPreamble` now returns nothing, and it only prints. `printRemoteSessionHint` calls `terminal.IsRemoteSession()` and returns early. `printInstructions` in `paste.go` needs no guard, because `startPasteWatcher` already checked the session.

internal/terminal/remote.go
wardbekker: Done in `88e4d5e5`. The doc comment of `IsRemoteSession` now states the limit:

```go
// The check reads the environment, so it reports true inside a detached tmux or
// screen session that inherited a stale SSH_CONNECTION, SSH_CLIENT, or SSH_TTY
// from the session that created it. That result is harmless: the paste route
// only adds a second way to finish the login, the callback server still works,
// and the user sees one note too many.
```

cmd/gcx/cloud/command.go
wardbekker: This one stays open, and I keep it out of this pull request. All four checks in that `Validate` return a plain error, so one converted check makes the function inconsistent inside itself. I will convert all four in a separate change, which keeps this diff on the paste route.

internal/auth/manual.go
```diff
@@ -0,0 +1,239 @@
...
+
+// printRemoteSessionPreamble states why the browser cannot reach the callback
+// address. Every remote-session message opens with it.
+//
+// It only prints. Each caller decides for itself whether the session is remote:
+// printInstructions runs after startPasteWatcher checked it, and
+// printRemoteSessionHint checks it directly.
+func printRemoteSessionPreamble(w io.Writer) {
+	fmt.Fprintln(w)
+	fmt.Fprintln(w, "Note: gcx runs in an SSH session.")
```
dafydd-t, line 155:
```suggestion
	fmt.Fprintln(w, "Note: gcx is running in an SSH session.")
```

dafydd-t (approved):

Issue #1138, opened 2026-08-05
Provide reusable typed usage errors reachable from internal command packages


awsome-o:
## Problem

Commands own two different kinds of input failure and we only have good support for one of them.

Cobra's own flag and argument errors (unknown flag, missing positional) currently exit 1 rather than 2. That is known, deliberate for now, and recorded as future work in `docs/design/exit-codes.md` §2.3. This issue is not about that.

The gap is the other kind: invalid input that a command validates itself, in its own `Validate()`. Those are usage errors by the taxonomy and should exit 2, but today they come back as ordinary errors and exit 1. We do have the right type for this, `fail.UsageError` at `cmd/gcx/fail/usage.go:12`, which carries a message, the expected format and suggestions. The trouble is where it lives. A command implementation under `internal/datasources/...` importing `cmd/gcx/fail` is the wrong dependency direction, so in practice nobody reaches for it.

## Why now

Five datasource PRs are in flight (#965, #969, #970, #971, #974) and every one of them needs new client-side validation: enum values, durations, positive integers, explicitly-empty and whitespace-only strings. Without a reachable helper, each contributor either returns a plain error and exits 1, or invents their own wrapper. We should not ask five people to solve the same layering problem five different ways.

## What we want

An internal helper, reachable from `internal/...` without depending on `cmd/gcx`, that gives a command:

1. Exit code 2 for input the command itself rejected.
2. An error carrying the rejected value (never echoing secrets), the expected format or allowed values, and a runnable corrected invocation, per `docs/design/errors.md`.
3. A natural place to call it before any loader or server I/O, so bad input never costs a round trip.

Then a short note in the datasource and provider guidance pointing at it, so the next contributor finds it without asking.

## Out of scope

The Cobra flag-error exit code question in `exit-codes.md` §2.3, and the separate drift between `docs/design/errors.md`'s closed summary vocabulary and what `cmd/gcx/fail/convert.go` actually emits. Both deserve their own issues.

PR #1140, opened 2026-08-05
docs(design): rule for datasource query variants (mode flag vs separate path)

### LINKED ISSUE #1139: Codify naming rule for datasource query variants (mode flag vs separate path)
## Problem

When a datasource supports more than one kind of query, we have no written rule for whether that becomes a flag on one command or a separate command path. Two in-flight PRs hit this at the same time and answered it differently.

Azure Monitor (#965) added `logs` and `resource-graph` as bare leaves next to `query`. Elasticsearch (#971) added `logs` as a bare leaf next to `query` and `metrics`.

Neither is covered by what we have written down. `command-naming.md:81` reserves `query` for executing a backend query, and the shorthand set at `:87` is closed at `labels`, `series`, `metrics` and `metadata`. `docs/plans/list-subject-verdicts.md:44` applies the test but only to discovery facets. And `command-naming.md:9` is explicit that a shipped command is not automatically a precedent for new work, so contributors cannot resolve this by copying a neighbour either.

Command paths freeze for the whole major version, so getting this wrong is not fixable later. Right now it lands on review, one PR at a time, which is exactly where we said naming questions should not be decided.

## Proposed rule

Add to `docs/design/command-naming.md`:

> When a datasource or provider supports more than one query variant, choose between a mode flag and a separate path by what the caller has to supply and what they get back.
>
> If the variants take the same expression with substantially the same inputs and the same result schema, they are one `query` leaf with a validated `--mode` enum and a documented default. Two leaves that differ only in presentation give agents and people a coin flip with no rule for choosing.
>
> If a variant targets something with materially different required inputs, or returns a different result schema, it gets its own path as `<target> query`, or an approved shorthand where one already exists. A target that needs its own identity arguments is not a mode of another command.

## What this decides

Azure Monitor becomes `azuremonitor logs query KQL` and `azuremonitor resource-graph query KQL`, keeping `azuremonitor query` for structured metrics. Logs needs a workspace identity and Resource Graph needs a subscription list, so their required inputs genuinely differ from the metrics query and from each other.

Elasticsearch becomes `elasticsearch query [EXPR] --mode logs` with a default of `documents`. Documents and logs share the Lucene expression, the time controls and the implementation, so they are one command with a mode. `metrics` stays its own leaf because its inputs and result schema differ, though note it shares the `finite` protocol class, so protocol class is not the distinguishing test.

Both decisions have been made as owner and communicated on the PRs. This issue is to write the rule down so it outlives those review threads.

## Note

`<target> query` nesting under a datasource kind is new. Everything under `datasources <kind>` is flat today. This is a deliberate new pattern rather than an appeal to precedent, and saying so in the guide matters, given `:9`.

awsome-o:
Closes #1139.

## The problem

When a datasource supports more than one kind of query, we had nothing written down about whether that becomes a flag on one command or a separate command path. Two in-flight PRs hit it in the same week and answered it differently: Azure Monitor (#965) added `logs` and `resource-graph` as bare leaves next to `query`, Elasticsearch (#971) added `logs` as a bare leaf next to `query` and `metrics`.

Nothing covered them. The guide reserves `query` for executing a backend query and closes the shorthand set at `labels`, `series`, `metrics`, `metadata`. `docs/plans/list-subject-verdicts.md` applies its test only to discovery facets. And the guide is explicit that a shipped command is not automatic precedent, so a contributor couldn't resolve it by copying a neighbour either. Command paths freeze for the major version, so this was landing on review one PR at a time, which is exactly where we said naming shouldn't be decided.

## The rule

One new `## Query variants` section, placed right after "Query and discovery operations" since that's the adjacent rule. Same expression operand, substantially the same required inputs and the same success output model means one `query` command with a typed `--mode`. Materially different required identity or inputs, or a different success output model, means its own `<target> query` path, or an approved shorthand where one already fits.

The bit that took the most iterating is the phrase "caller-facing contract". An earlier draft said "same result schema", which reads as the wire payload and gets Elasticsearch wrong: `Search()` and `Logs()` build different requests internally but both return `*querysql.QueryResponse`, which is why they're one command. `Aggregations()` returns a different model, which is why `metrics` isn't. Protocol class doesn't decide it either, and the rule says so, because both are `finite`.

Also written down: `--mode` is a closed validated enum with a documented default, rejecting unknown, whitespace-only and explicitly empty values with an error naming the value and the allowed set; mode picks the query, `--output` picks the presentation; and where the backend already determines the variant there is no flag to add. That last one is the only sentence here that #1139 didn't ask for. Without it the rule reads as a mandate to add `--mode influxql|flux|sql` to InfluxDB, whose language is read off the datasource record today, and that would be a new frozen flag on a shipped command. Easy to strike if you'd rather not have it.

`<target>` names the query surface, not a resource, so the section says explicitly that the identity-placement rule below it doesn't apply and that `query-<target>` is the wrong outcome. Nesting is called out as owner-approved and introduced by this rule rather than inferred from anything shipped, since everything under `gcx datasources <kind>` is flat today. It authorizes the nesting and nothing else, and it binds new work only.

## Alternatives rejected

Recording the rule in `docs/plans/list-subject-verdicts.md`: that file is scoped to the `list-<subject>` family under #387 and is retrospective. New durable rules belong in the naming guide.

Deciding by protocol class: `metrics` is the counterexample and it's in the rule as one.

Promising exit code 2 for a bad `--mode`: the reachable machinery for that is #1138, which isn't built. The rule requires the error's content, not its code.

## Tests

Documentation only. No code, no generated references, no output classes, no ADR table, no navigation. `GCX_AGENT_MODE=false mise run all` passes and regenerated references come back clean.

## One thing to know before merging

There's a competing `### Query variants` section sitting on the #1126 branch as an unpushed local commit (`0fdb8ea2`). It isn't on `main` and isn't on the PR yet. #1139 should own this section, so that commit needs dropping or deferring and #1126 rebasing afterwards. The good parts of it are folded in here: the query-surface framing, the distinction from identity placement, the "not a new top-level command" clause, and the one-flag-name-per-knob rule.

Worth landing this one first either way, because #1126's `references/self-review.md` already refers to "the §Query variants rule in docs/design/command-naming.md". The heading here is spelled to match.

### DIFF
--- a/docs/design/command-naming.md
+++ b/docs/design/command-naming.md
@@ -101,10 +101,10 @@ what they get back, not by counting backend APIs.
 
 Variants that share an expression operand, take substantially the same required
 inputs, and return the same success output model are one `query` command with a
-typed `--mode`. Two leaves that differ only in presentation give a caller, and
-an agent selecting a command, a coin flip with no rule for choosing. Use one
-flag name per knob across modes: the same value must not be `--size` in one mode
-and `--limit` in another.
+typed `--mode`. Separate leaves with the same caller-facing contract give
+callers and agents a coin flip with no rule for choosing. Use one flag name per
+knob across modes: the same value must not be `--size` in one mode and
+`--limit` in another.
 
 Variants that require materially different identities or inputs, or that return
 a different success output model, take their own `<target> query` path, or an
@@ -124,6 +124,12 @@ documented result are one command. Two that return different result models are
 not, even when both are the same output protocol class; protocol class does not
 decide this.
 
+The subject stays positional in every variant, per
+[CONSTITUTION.md § CLI Grammar](../../CONSTITUTION.md#cli-grammar): the metric,
+expression or query text is the operand, and the identity that scopes it is
+flags. `gcx datasources cloudwatch query` is flag-only and shipped in v1.0.0,
+so it is a compatibility exception, not a shape to copy.
+
 `--mode` is a closed, validated enum with a documented default. Reject an
 unknown, whitespace-only, or explicitly empty value with an error naming the
 rejected value and the allowed values, as `--on-error` does in
@@ -131,6 +137,12 @@ rejected value and the allowed values, as `--on-error` does in
 selects how the result is presented. Where the backend already determines the
 variant — InfluxDB's query language is read from the datasource record — the
 caller has nothing to choose and no flag is added.
+
+A kind that gains a `--mode` must keep it reachable from the auto-detecting
+`gcx datasources query`, which already carries kind-specific knobs. Silently
+answering with the default mode there, while the typed command honours the
+flag, splits one contract across two paths. If the generic form cannot carry
+the mode, redirect it to the typed command instead.
 
 `<target>` names the query surface: the expression language and API being
 queried. It nests inside the command's existing area, so for a datasource kind
@@ -163,20 +175,23 @@ today:
 gcx datasources elasticsearch query 'status:500'
 gcx datasources elasticsearch query 'level:error' --mode logs
 
-gcx datasources azuremonitor query --subscription SUB_ID \
-  --resource-group RG --resource NAME --namespace NAMESPACE --metric METRIC
-gcx datasources azuremonitor logs query KQL
-gcx datasources azuremonitor resource-graph query KQL
+gcx datasources azuremonitor query METRIC \
+  --subscription SUB_ID --resource-group RG \
+  --resource NAME --namespace NAMESPACE
+gcx datasources azuremonitor logs query KQL \
+  --subscription SUB_ID --resource-group RG --workspace WORKSPACE
+gcx datasources azuremonitor resource-graph query KQL \
+  --subscription SUB_ID
 ```
 
 Elasticsearch documents and logs share the Lucene expression, the time controls
 and the result model, so they are one command with `--mode` defaulting to
 `documents`; that they build different requests internally is not the test. Its
 `metrics` leaf stays separate: different inputs, a different result model, and
-`metrics` is already in the shorthand set. Azure Monitor keeps the structured
-metrics query as its primary `query`, while logs require a workspace identity
-and Resource Graph requires a subscription list, so each of those is its own
-target path.
+`metrics` is already in the shorthand set. Azure Monitor keeps the metric query
+as its primary `query`, with the metric positional and its resource identity in
+flags, while logs require a workspace identity and Resource Graph requires a
+subscription list, so each of those is its own target path.
 
 ## View verbs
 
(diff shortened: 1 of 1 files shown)


docs/design/command-naming.md
```diff
@@ -93,6 +93,106 @@ vocabulary and intentional aliases are documented in
...
+```
+
+Elasticsearch documents and logs share the Lucene expression, the time controls
+and the result model, so they are one command with `--mode` defaulting to
+`documents`; that they build different requests internally is not the test. Its
+`metrics` leaf stays separate: different inputs, a different result model, and
+`metrics` is already in the shorthand set. Azure Monitor keeps the metric query
+as its primary `query`, with the metric positional and its resource identity in
+flags, while logs require a workspace identity and Resource Graph requires a
+subscription list, so each of those is its own target path.
```
dafydd-t, line 194: this is a bit specific for general guidance, right?

docs/design/command-naming.md
```diff
@@ -93,6 +93,106 @@ vocabulary and intentional aliases are documented in
...
+A datasource or provider may offer more than one way to be queried. Decide
+between one command and several by comparing what the caller must supply and
+what they get back, not by counting backend APIs.
+
+Variants that share an expression operand, take substantially the same required
+inputs, and return the same success output model are one `query` command with a
+typed `--mode`. Separate leaves with the same caller-facing contract give
+callers and agents a coin flip with no rule for choosing. Use one flag name per
+knob across modes: the same value must not be `--size` in one mode and
+`--limit` in another.
```
dafydd-t, line 107:
```suggestion
Variants that share an expression operand, take substantially the same required
inputs, and return the same success output model are one `query` command with a
typed `--mode`. Use consistent flag names across modes: the same value must not be `--size` in one mode and
`--limit` in another.
```

docs/design/command-naming.md
```diff
@@ -93,6 +93,106 @@ vocabulary and intentional aliases are documented in
...
+
+Variants that require materially different identities or inputs, or that return
+a different success output model, take their own `<target> query` path, or an
+approved shorthand from the set above where one already fits. A target that
+needs its own identity arguments is not a mode of another command.
```
dafydd-t, line 112:
```suggestion
Variants that require materially different identities or inputs, or that return
a different success output model, take their own `<target> query` path, or an
approved shorthand from the set above where one already fits.

File: docs/design/command-naming.md [hunk 4]
```diff
@@ -93,6 +93,106 @@ vocabulary and intentional aliases are documented in
...
+`cmd/gcx/resources/onerror.go`. `--mode` selects which query runs; `--output`
+selects how the result is presented. Where the backend already determines the
+variant — InfluxDB's query language is read from the datasource record — the
+caller has nothing to choose and no flag is added.
+
+A kind that gains a `--mode` must keep it reachable from the auto-detecting
+`gcx datasources query`, which already carries kind-specific knobs. Silently
+answering with the default mode there, while the typed command honours the
+flag, splits one contract across two paths. If the generic form cannot carry
+the mode, redirect it to the typed command instead.
```
dafydd-t, line 145: as a british person i need you to stop putting `knobs` into prose please 🙈

PR #1141, opened 2026-08-05
refactor(datasources): make generic query routing complexity-stable

### LINKED ISSUE #1137: Make generic datasource query routing complexity-stable
## Problem

`QueryCmd` in `cmd/gcx/datasources/query.go` dispatches every datasource kind through a hand-maintained switch inside its `RunE` closure, plus a growing set of redirect guards ahead of it. Every new datasource kind adds to the same function, and we have now reached the point where the function fails lint once enough kinds are registered.

This is measured, not predicted. Using the repo's real `.golangci.yaml`, `maintidx` is enabled at its default threshold of `under: 20`:

| tree | `QueryCmd` maintainability index |
|---|---|
| `main` alone | 20 (cyclomatic complexity 27) |
| main + Azure Monitor (#965) | 20 |
| main + PostgreSQL (#969) | 21 |
| main + MySQL (#970, carries #969) | 20 |
| main + PostgreSQL + Azure | 20 |
| main + MySQL + Azure | **19, lint fails** |

```
cmd/gcx/datasources/query.go:26:1: Function name: QueryCmd, Cyclomatic Complexity: 30,
Halstead Volume: 4684.70, Maintainability Index: 19 (maintidx)
```

Note `gocyclo` is not the failing linter. Its threshold is 30 and complexity 30 is not greater than 30. The failure is `maintidx` alone.

## Why this matters now

`main` sits at exactly 20 against a threshold that fails below 20, so it passes by one point. Each of the in-flight datasource PRs is individually clean. The third of #965, #969 and #970 to land breaks CI through no fault of its own diff, and the author who happens to go last inherits a failure caused by the two before them. There is no merge order that avoids this, so it is ours to fix rather than any contributor's.

#971 and #974 do not touch this file and are unaffected.

## What to do

Replace the in-closure switch with a handler registry or table, so registering a kind is data rather than another branch in one function. While doing it:

1. Make the distinction between expression dispatch and typed-command redirect explicit in the structure, rather than implicit in where a guard happens to sit.
2. Preserve the current ordering guarantee that redirects run before `shared.ResolveExpr`, so an argument-less call gets the redirect instead of "expression required".
3. Keep the `supported:` list in the unsupported-type error derived from the registry rather than hand-maintained. It has already drifted; it does not mention `azuremonitor` even though #965 handles that kind via a guard.
4. Add focused routing tests covering dispatch, redirect, redirect-before-expression, and the unsupported default.

Behaviour preserving only. No provider behaviour changes in this PR.

## Acceptance

A tree containing `main` plus all five in-flight datasource PRs passes `mise run lint`.

## Notes for whoever picks this up

Two measurement traps cost real time when this was investigated. `golangci-lint`'s uniq-by-line processor suppresses a `maintidx` finding when another linter reports the same line, so probe `maintidx` on its own to see it. And concurrent `golangci-lint` runs fail with "parallel golangci-lint is running", which is easy to swallow inside a loop and read as a clean result. Use a private `GOLANGCI_LINT_CACHE` when probing several branches.

awsome-o:
Closes #1137.

## The problem

`QueryCmd` dispatched every datasource kind through a hand-maintained switch inside its `RunE` closure, with redirect guards stacked above it. `maintidx` folds a function literal into its enclosing declaration, so every per-kind arm counted against `QueryCmd`, which sat at exactly 20 against a threshold that fails below 20. Each of the five in-flight datasource PRs is lint-clean on its own; whichever landed third broke CI on a diff that hadn't done anything wrong. No merge order avoids it, so it was ours to fix rather than any contributor's.

Reproduced before touching anything, probing `maintidx` on its own because the uniq-by-line processor hides it when another linter reports the same line:

```
main alone:            QueryCmd  CC 27  HV 4499.17  MI 20   (passes by one point)
main + all five PRs:   QueryCmd  CC 30  HV 4684.70  MI 19   (fails)
```

## What this does

Two private, disjoint routing tables in `cmd/gcx/datasources/query_routes.go`: kind to dispatch handler for kinds the generic `<uid> <expr>` form fits, kind to redirect message for kinds whose query is structured. A kind is one or the other or unrouted, never both, and the tests enforce that rather than a struct with two nilable fields making "both" and "neither" constructible. The `RunE` closure becomes `(*genericQueryOpts).run` per the options pattern, and each switch arm becomes a small named handler that returns a value for the command to encode.

The tables are built by `newQueryRoutes()`, called once from `QueryCmd`, deliberately not inlined there. A map literal's size is attributed to the function containing it, so writing the table inside `QueryCmd` would have recreated the exact problem.

It stays in `cmd/gcx/datasources`. Route selection is command dispatch, which `cmd/` owns; the domain HTTP behaviour is still in `internal/query/<kind>`, and there's no reason to export a `Route`/`Dispatcher`/`StructuredQuery` API for one consumer.

## The four commits are meant to be read in order

The first adds characterization tests against the unchanged switch and proves they pass before any production code moves. The second replaces the switch while leaving those tests byte-identical. The third makes the one intentional behaviour change. The fourth is docs. The parity net was written against the old code and the new code never got to edit it, which is the only reason to trust it.

Characterization covers every kind the generic command handles: Prometheus time and step, Loki `--limit`, Pyroscope `--profile-type` and `--max-nodes`, InfluxDB's second datasource GET and the detected language mode plus its `failed to detect influxdb mode` wrapper, ClickHouse limit and interval. Plus reachable query-error wrapping, exactly-one-JSON-value stdout, encode-error propagation, the CloudWatch redirect with and without an expression, and the unknown-kind ordering. Two deliberate mutations, dropping the Loki limit and the Prometheus step wiring, were confirmed to turn them red, so they aren't tests that can't fail.

Structural assertions cover disjoint key sets, keys canonical against `NormalizeKind`, non-nil dispatchers, non-empty redirect messages, the sorted union, and the unknown lookup. No duplicate-key test, because duplicate constant keys in a Go map literal are already a compile error.

## The one behaviour change

The unsupported-kind message is now derived from the tables instead of a hand-written literal:

```
before: (supported: prometheus, loki, pyroscope, influxdb, clickhouse)
after:  (supported: clickhouse, cloudwatch, influxdb, loki, prometheus, pyroscope)
```

Sorted, and redirect kinds included. #1137 names the drift directly: the literal omitted cloudwatch and would have omitted azuremonitor once #965 landed. Redirect kinds belong in it because someone reading that message is asking what the command handles, not which kinds happen to take an expression. Pinned exactly, in human mode and with agent mode on. It's observable, including inside the agent-mode error document, so it's called out here rather than described as internal.

Nothing else moves. `docs/reference/cli/gcx_datasources_query.md` is byte-identical, and so are `datasources query --help`, `help-tree` and `commands --flat -o json`, all diffed against binaries built from `main`.

Ordering is preserved exactly, including the parts that look like accidents and aren't. Redirects still run before `ResolveExpr`, so an argument-less call gets the typed command rather than "expression is required". An unknown kind with no expression still returns "expression is required"; with an expression but a bad `--step` it still returns the parse error; only a fully valid request reaches the unsupported-type verdict. Moving that check up would be more useful in one round trip instead of two, but it's a behaviour change and this PR isn't the place.

## Measurements

Every changed function, not just `QueryCmd`, with a temporary `under: 100` threshold used only to print numbers. Real `mise run lint` is the authority.

```
                       6 kinds (this PR)      11 kinds (all five PRs)
QueryCmd               CC 1  HV 265   MI 50   CC 1  HV 265   MI 50
run                    CC 9  HV 1233  MI 38   CC 9  HV 1299  MI 38
newQueryRoutes         CC 1  HV 147   MI 57   CC 1  HV 258   MI 48
Validate / setup                MI 59 / 64             MI 59 / 64
dispatch handlers               MI 49-54               MI 49-54
```

`QueryCmd` is identical at 6 and 11 kinds. `run` is unchanged. The only thing that moves with kind count is the table constructor, and it moves at constant cyclomatic complexity, so I'd rather not claim the cliff "cannot recur" — it accumulates Halstead volume like anything else holding data. It just does so slowly and without branching, and it's a different function from the one with the budget problem.

## Acceptance

The issue asks for a tree containing `main` plus all five in-flight PRs to pass `mise run lint`. Built in a throwaway worktree from PR head refs, never pushed to anyone's branch:

```
mise run lint  ->  0 issues
```

with all eleven kinds routed: azuremonitor, clickhouse, cloudmonitoring, cloudwatch, elasticsearch, influxdb, loki, mysql, postgres, prometheus, pyroscope.

Two honesty notes on that number. #970 is based on `alyssa/add-postgres-ds`, so it carries #969; I merged #969 then #970 deliberately rather than treating them as independent. And #971 and #974 add no generic routing on their current heads at all, so their entries in that tree are labelled anticipated-post-review shapes, not measurements of merged code. #965, #969 and #970 are real adaptations of hunks that exist today. The only conflicts outside `query.go` were between the contributor PRs themselves, in `codecs.go`, `CODEOWNERS` and `project-structure.md`, and are not caused by this change.

## What the five PRs need on rebase

Each converts to one table entry plus, for dispatch kinds, one handler.

`#965 azuremonitor` — a `redirects` entry built with `structuredQueryRedirect("Azure Monitor", "subscription, resource group, resource, namespace, metric, aggregation", "gcx datasources azuremonitor query --subscription ... ...")`. The guard goes away; ordering is handled by the table.

`#969 postgres` and `#970 mysql` — a `dispatch` entry plus the existing `runPostgresQuery`/`runMySQLQuery` body renamed to a handler taking `genericQueryRequest`. The `cmdio.Warning` call keeps working: `req.warn` is bound to `cmd.ErrOrStderr()` for exactly this. The hand-edit to the supported list goes away, it's derived now.

`#971 elasticsearch` — currently unrouted, so `gcx datasources query <es-uid>` is a dead end. Its query leaf is expression-shaped, so a `dispatch` entry is the natural fix.

`#974 cloudmonitoring` — also currently unrouted, and its query leaf is structured-only, so a `redirects` entry.

## Alternatives rejected

Extracting per-kind helpers and keeping the switch is what #969 already did, and the combined tree still measured 19. Extraction moves volume, it doesn't remove the per-kind branch from the central function.

Driving dispatch off the `DatasourceProvider` registry doesn't work as things stand: the interface has no headless query entry point, no plugin-ID aliases and no capability flags, and `cmd/gcx/datasources` doesn't import `internal/datasources/providers`, so the registry is empty in this package's test binary. It would need a new method on all nine providers, which is the redesign #1137 says not to do.

Raising the threshold or adding a `nolint` was off the table.

## Risks

Blast radius is `gcx datasources query`. Typed datasource commands, CRUD, clients, config and persisted data are untouched. A bug here makes generic queries fail, pick the wrong handler, or format wrongly; it can't touch customer data. Rollback is a plain revert until the first dependent PR merges.

Worth landing this alone, letting main CI and a smoke run confirm it, and giving it a short burn-in before rebasing the five datasource PRs onto it one at a time.

One coordination item: #1126 is not merged and its skill files still describe the switch and "the redirect goes before the switch" in `SKILL.md`, `references/placement-and-readiness.md` §3-§4 and `references/distribution-and-gates.md`. Those files aren't on `main`, so this PR can't fix them. Whichever lands second has to update the other side, and neither should merge while pointing contributors at a switch that no longer exists. The repo-local `.claude/skills/add-datasource/SKILL.md` is updated here.

### DIFF
--- a/.claude/skills/add-datasource/SKILL.md
+++ b/.claude/skills/add-datasource/SKILL.md
@@ -226,8 +226,24 @@ Reference: `internal/datasources/providers/prometheus.go`.
    `gcx datasources list -o json` and add a mapping in
    `internal/datasources/query/resolve.go` if they don't match. Without this,
    auto-discovery and datasource type validation will fail silently.
-3. Optionally add to the auto-detecting `datasources query` switch in
-   `cmd/gcx/datasources/query.go`
+3. **Routing for the auto-detecting `datasources query`** — registration mounts
+   your typed `datasources <kind>` subtree but does not reach the generic
+   command, which routes through the tables in
+   `cmd/gcx/datasources/query_routes.go`. Add exactly one entry, keyed by the
+   normalized kind:
+   - the generic `<uid> <expr>` form can honestly carry your query → add a
+     `dispatch` entry plus a small handler alongside the existing ones;
+   - it cannot, because your query takes structured parameters no single
+     expression represents → add a `redirects` entry built with
+     `structuredQueryRedirect`, naming your typed command. CloudWatch is the
+     worked example.
+
+   Adding neither is also a choice: your kind then reports as unsupported. Make
+   it deliberately — a caller who reasonably reaches for `datasources query`
+   gets a dead end. The two tables must stay disjoint and keyed by normalized
+   kinds; `query_routes_internal_test.go` enforces both, and the supported-kind
+   list in the unsupported-type error is derived, so there is nothing to
+   hand-update.
 
 ### Step 4: Agent Annotations
 
--- a/.claude/skills/add-datasource/SKILL.md
+++ b/.claude/skills/add-datasource/SKILL.md
@@ -402,9 +402,16 @@ Then trace each registration in `RegisterCodecs` to a reachable `Encode`
    Adding neither is also a choice: your kind then reports as unsupported. Make
    it deliberately — a caller who reasonably reaches for `datasources query`
    gets a dead end. The two tables must stay disjoint and keyed by normalized
-   kinds; `query_routes_internal_test.go` enforces both, and the supported-kind
-   list in the unsupported-type error is derived, so there is nothing to
-   hand-update.
+   kinds; `query_routes_internal_test.go` enforces both.
+
+   The supported-kind list in the unsupported-type error is derived from the
+   tables, so you never edit that string. You **do** update the two places that
+   pin its exact value, because it is user-visible text on a GA path and is
+   deliberately not free to drift:
+   `TestQueryRoutesSupportedKindsIsTheSortedUnion` in
+   `query_routes_internal_test.go`, and `wantUnsupportedMessage` in
+   `query_unsupported_test.go`. Both fail with the old and new lists side by
+   side, so the update is mechanical.
 
 ### Step 5: Agent Annotations
 
--- a/.claude/skills/add-datasource/SKILL.md
+++ b/.claude/skills/add-datasource/SKILL.md
@@ -387,19 +387,24 @@ Then trace each registration in `RegisterCodecs` to a reachable `Encode`
    `gcx datasources list -o json` and add a mapping in
    `internal/datasources/query/resolve.go` if they don't match. Without this,
    auto-discovery and datasource type validation will fail silently.
-3. Decide, deliberately, what the auto-detecting `datasources query` should do
-   for your kind — the switch in `cmd/gcx/datasources/query.go` is
-   hand-maintained and no test enforces parity with registration:
-   - the generic `<uid> <expr>` form can carry your query → add the case, so a
-     caller reaching for `datasources query` is not met with the bare
-     "datasource type %q is not supported" default;
-   - it cannot (a structured query with several required parameters) → add an
-     explicit redirect naming your typed command and its flags, the way
-     CloudWatch does. Do not force a lossy generic path. Put the redirect where
-     CloudWatch's is: a guard on the normalized type **before** the switch and
-     before `shared.ResolveExpr`. As a switch case it never fires for
-     `gcx datasources query <uid>` with no expression — that call dies on
-     "expression required" first.
+3. **Routing for the auto-detecting `datasources query`** — registration mounts
+   your typed `datasources <kind>` subtree but does not reach the generic
+   command, which routes through the tables in
+   `cmd/gcx/datasources/query_routes.go`. Add exactly one entry, keyed by the
+   normalized kind:
+   - the generic `<uid> <expr>` form can honestly carry your query → add a
+     `dispatch` entry plus a small handler alongside the existing ones;
+   - it cannot, because your query takes structured parameters no single
+     expression represents → add a `redirects` entry built with
+     `structuredQueryRedirect`, naming your typed command. CloudWatch is the
+     worked example.
+
+   Adding neither is also a choice: your kind then reports as unsupported. Make
+   it deliberately — a caller who reasonably reaches for `datasources query`
+   gets a dead end. The two tables must stay disjoint and keyed by normalized
+   kinds; `query_routes_internal_test.go` enforces both, and the supported-kind
+   list in the unsupported-type error is derived, so there is nothing to
+   hand-update.
 
 ### Step 5: Agent Annotations
 
--- a/.claude/skills/add-datasource/SKILL.md
+++ b/.claude/skills/add-datasource/SKILL.md
@@ -241,9 +241,16 @@ Reference: `internal/datasources/providers/prometheus.go`.
    Adding neither is also a choice: your kind then reports as unsupported. Make
    it deliberately — a caller who reasonably reaches for `datasources query`
    gets a dead end. The two tables must stay disjoint and keyed by normalized
-   kinds; `query_routes_internal_test.go` enforces both, and the supported-kind
-   list in the unsupported-type error is derived, so there is nothing to
-   hand-update.
+   kinds; `query_routes_internal_test.go` enforces both.
+
+   The supported-kind list in the unsupported-type error is derived from the
+   tables, so you never edit that string. You **do** update the two places that
+   pin its exact value, because it is user-visible text on a GA path and is
+   deliberately not free to drift:
+   `TestQueryRoutesSupportedKindsIsTheSortedUnion` in
+   `query_routes_internal_test.go`, and `wantUnsupportedMessage` in
+   `query_unsupported_test.go`. Both fail with the old and new lists side by
+   side, so the update is mechanical.
 
 ### Step 4: Agent Annotations
 
--- a/.claude/skills/integrate-with-gcx/SKILL.md
+++ b/.claude/skills/integrate-with-gcx/SKILL.md
@@ -238,10 +238,10 @@ user it needs a human to drive, and point them at
 Per-leaf wiring CI will not let you skip is tabulated in the reference, along
 with the one gap CI does *not* catch: registration mounts the typed
 `datasources <kind>` subtree, but reaching the generic auto-detecting
-`datasources query` is a separate **judgement, not a checkbox** — a case if the
-`<uid> <expr>` form can honestly carry your query, an explicit redirect if it
-cannot. Both the reasoning and the ordering requirement are in the reference; get
-them from there rather than from memory.
+`datasources query` is a separate **judgement, not a checkbox** — a `dispatch`
+entry if the `<uid> <expr>` form can honestly carry your query, a `redirects`
+entry if it cannot. Both the reasoning and the ordering requirement are in the
+reference; get them from there rather than from memory.
 
 Format the files you touched, then gate:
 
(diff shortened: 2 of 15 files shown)


awsome-o:
Follow-ups filed, none blocking this PR:

- #1144 honor explicit `--limit` for generic SQL queries while preserving per-kind defaults
- #1145 reuse the fetched datasource record when detecting InfluxDB mode
- #1146 require explicit generic-query routing classification for every registered datasource provider

#1146 also carries the scoped error wording — moving from `is not supported (supported: ...)` to `is not handled by \`gcx datasources query\` (routed kinds: ...)`. Owner call recorded here so it does not get relitigated: the derived list stays scoped to kinds this command actually routes, and tempo/athena/infinity are deliberately absent despite having typed query commands. Listing them would contradict the error they appear in.

awsome-o:
Correction to my earlier comment: #1145 is closed as not planned — one redundant datasource GET on one kind was not worth tracking. The remaining follow-ups are #1144 (honor explicit `--limit` for generic SQL queries) and #1146 (routing classification, which carries the scoped `is not handled by` error wording). Neither blocks this PR.

awsome-o:
Rebased onto main (19 commits behind, `query.go` conflicting) and re-verified. CI is green and the PR is mergeable again. Notes on what changed, because the rebase was not mechanical in a couple of places.

#969 landed while this sat, so main now has a postgres arm in the switch this PR deletes. It converts to one `dispatch` entry plus `dispatchPostgres`, which is what the "what the five PRs need on rebase" section predicted, so that part held up. The derived supported-kind list is seven kinds now and both pinned assertions moved with it. That is exactly the mechanical two-test update the add-datasource skill describes, and having to do it myself was a decent check that the wording is right.

The one place the earlier review turned out to be wrong: it removed `genericQueryRequest.warn` as dead state, correctly, because nothing read it at the time. Postgres reads it — it caps an oversized LIMIT and has to say so without polluting the stdout document — so the field is back, with the comment naming its reader instead of describing a seam that might get one later.

That resolution also caught a real bug, and I want to flag how it was caught rather than just that it was. The review commit had removed the `warn: cmd.ErrOrStderr()` line from `query.go` in the same breath as the field. Restoring the field but not the assignment compiles clean, passes lint, and passed the entire existing suite, because an unset `io.Writer` is only a problem for the one kind that reads it. `gcx datasources query <postgres-uid> 'SELECT ... LIMIT 5000'` would have panicked on a nil writer. The new characterization case turns that into a nil-pointer panic in CI instead of a crash in someone's terminal, which is the only reason I trust the seam now.

#1126 merged on the 11th, so the coordination item in the description came due on this side. Three places in the integrate-with-gcx skill described the switch and the hand-placed guard. Worth noting that reference had already written the ordering requirement as durable and the guard placement as "a property of the current hand-written RunE, not policy: if routing becomes declarative the mechanism changes while the ordering requirement does not". That aged well. Routing is declarative now, so the mechanism text moved and the ordering requirement stayed, and it is now free rather than something a contributor has to get right by hand.

Re-measured the thing this PR exists for, since #969 changed the numbers. Main is not where the description says any more:

```
main today, 6 kinds          QueryCmd  CC 27  HV 4331.94  MI 21   passes
main + 1 dispatch kind       QueryCmd  CC 29  HV 4564.69  MI 20   passes by one point
main + 2 dispatch kinds      QueryCmd  CC 31  HV 4798.08  MI 19   fails, confirmed against the real threshold
this branch, 7 kinds routed  QueryCmd  CC 1   HV 265      MI 50
```

So #969 bought a point by extracting a helper, which moves the cliff out by one PR rather than removing it. Four datasource PRs are still open. Two of them add dispatch kinds, and the branch number does not move with kind count at all — `run` is MI 38 and the worst new function is `dispatchPostgres` at 48.

On E2E, since a refactor claim is only worth what was actually run. I stood up Grafana 12.4 and a real Postgres in throwaway containers, created a datasource of every routed kind plus tempo for the unrouted path, and ran 16 cases against binaries built from this branch and from main, diffing stdout, stderr and exit code. All 16 have identical exit codes and byte-identical stdout. Two stderr differences: the unsupported-kind message, which is the one intentional change, and a Loki case where the two runs were 175ms apart and the difference is the `now` timestamp in the URL.

The postgres path is proven against real rows rather than a fake: no LIMIT gives 100, `LIMIT 5000` caps to 1000 and writes the notice to stderr and not stdout, `LIMIT 7` is honoured, and `--step 45s` lands as `intervalMs`. Ordering is intact where it matters. The CloudWatch redirect fires with and without an expression. Tempo with no expression still gets "expression is required", with a bad `--step` still gets the duration error, and only a fully valid request reaches the unsupported verdict. Every failing case left stdout empty. `datasources query --help`, `help-tree` and `commands --flat -o json` are byte-identical to a binary built from main.

One thing E2E turned up that is not mine to fix here. `splitErrorMessage` in `cmd/gcx/fail/convert.go` splits on the first `": "`, so the unsupported-kind error tears across the agent-mode document with unbalanced parens in both halves:

```
"summary": "Datasource type \"tempo\" is not supported (supported",
"details": "clickhouse, cloudwatch, influxdb, loki, postgres, prometheus, pyroscope)"
```

Main does the identical thing with its own list, so this is pre-existing and not a regression, but it does live in the exact string this PR changes. It also survives the rewording #1146 carries, since `(routed kinds: ...)` splits the same way. Flagging it there rather than growing this PR.

dafydd-t (approved):

PR #1149, opened 2026-08-06
fix(auth): scope OAuth callbacks to the login that started them, and honour the chosen port on every leg

### LINKED ISSUE #1147: bug(auth): wrong-state OAuth callback consumes and aborts the active login
## Description

The loopback OAuth callback handler claims its `sync.Once` guard before validating the HTTP method and the OAuth `state` parameter.

Consequently, a stale or unrelated request to `/callback` with missing or mismatched state:

1. consumes the one-shot handler;
2. produces `invalid state - possible CSRF attack`;
3. terminates the active login; and
4. prevents a later correct callback from succeeding, via `410 Gone` or a closed listener after shutdown.

This is a callback robustness / local denial-of-service bug. Any local process able to reach the loopback listener can abort an in-progress login. The error message alone is not evidence that a CSRF attack occurred, and the wording has caused this failure to be misread as a security incident.

In unified Cloud login, stack OAuth and GCOM OAuth run sequentially and independently scan ports 54321-54399. Reusing the same port is common, although not guaranteed, so replaying or refreshing the first leg's callback can reach and abort the second flow.

Implementation note: the two halves of the fix have to land together. Because the wrong-state request consumes `sync.Once` today, removing only the flow-terminating error would turn this failure into a hang — the legitimate callback would then be answered with `410 Gone` while `Run` keeps waiting. The `once` claim has to move after validation.

Relevant code:

- State validation: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/flow.go#L181-L188
- One-shot callback handler: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/flow.go#L252-L268
- Callback port selection: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/flow.go#L433-L449
- GCOM state validation: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/gcom.go#L160-L167

## What did you expect to happen?

- Validate the request method and state before claiming the one-shot handler.
- Return 400 for a missing or mismatched state without consuming `sync.Once` and without terminating the active flow.
- A correct callback received after one or more rejected callbacks must still succeed.
- A replay of an already accepted callback should continue to return `410 Gone`.
- Bound the overall wait so a flow that never receives a valid callback cannot block indefinitely (today the only bound is context cancellation).
- Distinguish the user-facing message for a stale or unmatched callback from an actual attack, so this failure mode is not reported as a suspected CSRF attack.
- Never log raw OAuth state, code, or tokens.

## How can we reproduce this?

1. Start an OAuth login and note its callback port.
2. Request `/callback?state=<wrong-state>` on that port.
3. Send the legitimate callback with the correct state.
4. Observe that the first request has already consumed or terminated the flow, so the legitimate callback cannot succeed.

## gcx Version

Observed in the v1.0.0 implementation.

## Environment Details

Platform-independent.

## Anything else we should know?

There is currently no test in `internal/auth` that issues a request to the `/callback` handler, so none of this behaviour is covered. Suggested regression coverage:

- missing state;
- incorrect state;
- non-GET request;
- wrong callback followed by correct callback;
- concurrent valid callbacks;
- replay after success; and
- a stale callback from the first OAuth leg reaching the second leg.

A unique callback path per OAuth leg would provide additional defence in depth, but it cannot be done in gcx alone: the instance leg's callback URL is built by the Assistant plugin with a fixed `/callback` path (only the port is parameterised), and the GCOM leg's `redirect_uri` must match the registered value. It is not required for the initial fix.

Related:

- https://github.com/grafana/gcx/issues/1135
- https://github.com/grafana/gcx/pull/1136
- https://github.com/grafana/gcx/issues/814
- https://github.com/grafana/gcx/pull/887
- https://github.com/grafana/gcx/pull/904

### LINKED ISSUE #1148: bug(login): --oauth-callback-port is ignored by GCOM/Cloud OAuth
## Description

`gcx login --oauth-callback-port <port>` propagates the requested fixed port only to the Grafana stack OAuth flow.

When unified Cloud login starts its optional GCOM OAuth step, `GCOMOptions` has no callback-port field and `GCOMFlow.Run` passes `0`, which falls back to automatic selection from ports 54321-54399. Standalone `gcx cloud login` also has no equivalent option — it exposes only `--cloud-token`, `--oauth-url`, `--api-url` and `--scope`.

This blocks users who depend on a fixed forwarded or firewall-approved port from completing the second OAuth leg, which is the case the flag was added for: only specific ports are reachable between a remote host and the browser.

Relevant code:

- Stack OAuth receives the configured port: https://github.com/grafana/gcx/blob/v1.0.0/internal/login/login.go#L647-L663
- `GCOMOptions` has no port field: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/gcom.go#L54-L67
- GCOM uses automatic selection: https://github.com/grafana/gcx/blob/v1.0.0/internal/auth/gcom.go#L87-L115
- Unified login constructs GCOM options without a port: https://github.com/grafana/gcx/blob/v1.0.0/cmd/gcx/login/command.go#L1132-L1149

## What did you expect to happen?

A fixed callback port selected for unified login should apply consistently to every browser OAuth leg.

If callback-server mode remains supported by `gcx cloud login`, that command should expose the equivalent option. The GCOM authorization request and the token exchange must use byte-identical redirect URIs.

## How can we reproduce this?

1. Run `gcx login <context> --server https://<stack>.grafana.net --cloud --oauth --oauth-callback-port 8250`.
2. Complete the stack OAuth leg.
3. Continue into the Grafana Cloud (GCOM) OAuth leg.
4. Observe that the second leg listens on an automatically selected port from 54321-54399 rather than 8250.

## gcx Version

Observed in the v1.0.0 implementation.

## Environment Details

Platform-independent. Most visible on remote hosts where only specific ports are forwarded to the browser.

## Anything else we should know?

Acceptance criteria:

- Add a callback-port option to `GCOMOptions`.
- Pass it through to `listenOnCallbackPort`.
- Propagate it through unified login and, where applicable, standalone `gcx cloud login`.
- Apply consistent port validation across both legs.
- Add tests proving both OAuth legs use the requested port.
- Coordinate with the manual callback mode in #1136: a fixed port should either be rejected or explicitly documented as ignored when manual mode is active.

Related:

- https://github.com/grafana/gcx/issues/611
- https://github.com/grafana/gcx/pull/612
- https://github.com/grafana/gcx/issues/1135
- https://github.com/grafana/gcx/pull/1136
- https://github.com/grafana/gcx/pull/847
- https://github.com/grafana/gcx/pull/904

awsome-o:
Closes #1147
Closes #1148

## #1147 — a wrong-state callback could end somebody else's login

The loopback callback handler claimed its `sync.Once` guard on arrival, before
it checked the HTTP method or the OAuth `state`. So any request at all to
`/callback` consumed the one-shot handler, pushed an error onto the flow's error
channel, and ended the login. The genuine callback that followed got `410 Gone`,
and the user saw `invalid state - possible CSRF attack`.

Any local process can reach a loopback listener, so anything on the machine
could abort a login in progress. It also happened by accident: in a unified
Cloud login the stack leg and the grafana.com leg run back to back and both scan
54321-54399, so reloading the first leg's callback tab routinely killed the
second. And the wording got an ordinary stale tab reported as a security
incident.

Ownership now decides who may consume a flow, not arrival order:

| Request | Response | Effect on the flow |
|---|---|---|
| Not `GET` | `405` + `Allow: GET` | none |
| `state` missing or not ours | `400` + a browser page | none |
| Ours, already claimed | `410 Gone` | none — replay stays rejected |
| Ours, deadline passed | `410 Gone` | none |
| Ours, first to arrive | success or error page | terminal |

The method check is explicit rather than a `"GET /callback"` ServeMux pattern.
That pattern also matches `HEAD`, which would have let a HEAD request run the
token exchange as a side effect — a hole in the gate and a side-effecting HEAD.

`callbackArbiter` replaces the `sync.Once`. It is shared rather than per-path
because over SSH gcx keeps the callback server listening *while* it reads a
pasted redirect URL, so the two genuinely race. Both take the same claim, so a
pasted URL and a browser callback can never both exchange the same authorization
code.

The flow is claimed **as late as possible** — only once the code is about to go
on the wire. `handleCallbackParams` splits into `validateCallbackParams` (no
side effect, every rejection retryable) and `completeCallback` (spends the code,
every error terminal), and the paste path validates before it claims. Claiming
earlier let an unusable paste — the authorize URL pasted by mistake, which
carries our state but no code — hold the flow just long enough for the real
browser callback to be answered `410 Gone` and lost.

A failed token exchange is **terminal on both paths**: the code is spent, the
browser callback carries the same one, and reopening would exchange it twice.
Only retryable failures re-prompt, which keeps #1136's behaviour where it still
makes sense.

The arbiter also bounds the wait. Previously the only limit was context
cancellation, so a login that never received a valid callback waited forever.
The bound is **30 minutes** — it exists to stop an unattended hang, not to hurry
anyone, and the authorization code is minted at the *end* of the consent flow,
so a tighter bound rejects logins that were about to succeed. A granted claim
stops the clock, so an exchange in flight is never cut off. Rejected requests
deliberately do not extend it.

**Manual mode is deliberately unbounded** apart from Ctrl-C. `readLineContext`
cannot cancel the read it wraps, so a deadline there returns while an abandoned
goroutine still owns the terminal, and the user's half-typed redirect URL lands
in the shell with a live authorization code in it. A hang is the lesser evil.

The state comparison moved to one shared constant-time helper used by the HTTP
gate, the paste gate, and both param handlers, so they cannot drift on what
counts as ours.

Deliberately **not** done: no rejection notice is printed from the handler
goroutine. That would race the flow goroutine that owns the terminal.

## #1148 — the chosen callback port only reached the first leg

`--oauth-callback-port` reached only the stack OAuth leg. `GCOMOptions` had no
port field and `GCOMFlow` always passed `0`, so the Cloud leg auto-picked.
`gcx cloud login` had no such flag at all.

That defeats the case the flag exists for: on a remote host only specific ports
are forwarded to the browser, so a user who gets past the stack leg on the port
they forwarded then hits a second leg listening somewhere unreachable.

A fixed port describes the machine gcx runs on, not one leg, so it now applies
to all of them. `gcx cloud login` gains the same flag. Auto-pick stays the
default and unflagged logins are unchanged.

The conflicts are scoped to where they are real. `gcx cloud login` rejects a
port alongside `--cloud-token`, because that token skips OAuth entirely. Unified
login does **not** — there a Cloud token still coexists with a port the stack leg
needs, and a stack token with a port the Cloud leg needs. Both reject a port
with `--oauth-manual`, which starts no listener. Both validate the number
through one shared `auth.ValidateCallbackPort`, and every check runs before any
browser or network side effect.

GCOM's `redirect_uri` is still built once from the bound port and shared by the
authorize request and the exchange; a test now pins that the two are
byte-identical, since GCOM rejects the exchange otherwise.

## Relationship to #1136

This is **stacked on #1136** (`ward/oauth-manual-remote`), not on main. #1136
rewrites the same functions — it extracts `handleCallbackParams` and
`handleGCOMCallbackParams` into `internal/auth/callback.go` and turns both `Run`
bodies into select loops — so branching from main would have conflicted hard and
risked regressing the paste path. Both issues also say to coordinate with it.

Nothing here duplicates, absorbs, or changes #1136's manual/SSH functionality.
It builds on it: the shared param handlers are exactly the right seam, and the
paste path already treated a state mismatch as non-terminal, so this change
makes the HTTP path agree and puts both behind one arbiter. #1136's re-prompt
behaviour is preserved exactly — parse errors, foreign state, and exchange
failures all still re-prompt with the server still listening.

Please review and merge #1136 first. Once it merges the base should retarget to
main automatically; I will verify that rather than assume it, then rebase,
re-inspect the diff, and rerun the gates.

## Tests

Nothing in `internal/auth` previously issued a request to the live `/callback`
handler, which is why none of this was caught.

- `arbiter_test.go` — the ownership contract, deterministically: concurrent
  claims grant exactly one, HTTP-vs-paste cannot both claim, a claim racing the
  deadline yields one coherent outcome, release reopens but expires after the
  deadline.
- `wiring_internal_test.go` — package-internal, proves the real HTTP handler and
  the real paste gate arbitrate through the *same* arbiter instance. The unit
  tests prove the contract; this proves the wiring.
- `callback_server_test.go` — the live listener end to end: wrong state, missing
  state, HEAD and four other methods, replay while the first exchange is held
  open, denial with and without a matching state, cancellation, the deadline,
  and that a rejected caller's own values are never echoed into the page, the
  output, or the error.
- `gcom_port_test.go` — fixed port honoured, auto-pick unchanged, authorize and
  exchange `redirect_uri` byte-identical, invalid port and manual+port rejected
  before any side effect.
- Flag propagation for both legs and both commands.

Success-path tests bind an automatic port and recover it from the flow's own
output; only the deliberate conflict cases reserve one. The flow's writer is
mutex-guarded, since the flow writes while the test reads.

I checked these fail against the old ordering before relying on them: restoring
the claim-before-validation ordering fails exactly the #1147 tests, HEAD
included.

## Verification

`GCX_AGENT_MODE=false mise run all` passes — lint 0 issues, full suite, docs,
build, no reference drift. Affected packages also pass under `-race`.

Also driven end to end against the built binary with `--oauth-callback-port
8250`: the leg bound 8250; wrong state, missing state and a foreign denial each
returned 400; POST and HEAD each returned 405 with `Allow: GET`; the login
survived all of it; and the legitimate callback that followed completed
normally. Nothing supplied by the rejected callers appeared in the output.

## Risks

- The 30-minute wait bound is a judgement call; easy to retune.
- **Normal CI has never run on this branch** — GitHub creates no workflow runs
  for it, apparently because the base is a non-default branch. Gates were run
  locally. This must not be marked ready until CI runs after the retarget.
- The ownership gate is strict, so if a provider ever omitted `state` from an
  `error=` redirect (RFC 6749 §4.1.2.1 says it MUST echo it), that denial would
  become non-terminal and the user would wait out the deadline instead of
  failing fast. Relaxing the gate would reinstate the abort, so the deadline is
  the mitigation.


awsome-o:
Reopening to trigger CI — no runs were created on open.

awsome-o:
Note on CI: GitHub created no workflow runs for this PR — only the org-wide Socket and Semgrep scanners appear. Zero runs exist for the branch via the Actions API, and closing/reopening did not create any. I suspect the CI workflow is not firing because the base is a non-default branch; I could not confirm that from here. It should run normally once #1136 merges and this retargets to main.

In the meantime the gates were run locally on the exact pushed tree:

- `GCX_AGENT_MODE=false mise run all` — lint 0 issues, full test suite, docs build, binary build, no reference drift
- `go test -race` on internal/auth, internal/login, cmd/gcx/login, cmd/gcx/cloud
- end-to-end against the built binary, described in the PR body

I will rerun everything after the rebase onto main.

awsome-o:
Pushed a third commit fixing what an adversarial review of the first two turned up. Fifteen findings; I fixed fourteen and left one open below.

The three that actually mattered were all self-inflicted, and two of them contradicted claims in this PR's description:

**`release()` broke the very invariant this PR asserts.** It reopened the flow after *any* failed pasted attempt, including one where the authorization code had already reached the token endpoint. The browser callback carries that same code, so it could claim and exchange it a second time — reproduced against the real handler under `-race`. `callbackError` now records whether the code was spent: a malformed URL still re-prompts, a failed exchange ends the flow. That also resolves an inconsistency where the HTTP leg treated an exchange failure as terminal and the paste leg did not.

**A panicking handler hung the login forever.** A granted claim disarms the deadline and net/http swallows the panic, so nothing reached either channel and the select had no reachable case. The `defer arb.settle()` I added as a late "robustness improvement" was the thing that made it unrecoverable. The handler now recovers into an error; the panic value is not reported since it can carry request data.

**The deadline was fighting the flow it was meant to protect.** It ran from listener bind and covered the whole cross-machine consent round trip — but the authorization code is minted at the *end* of that, so ten minutes rejected logins that were about to succeed. Now 30 minutes. Manual mode drops its bound entirely: `readLineContext` cannot cancel the read it wraps, so a deadline there returned while an abandoned goroutine still owned the terminal, and a half-typed redirect URL would land in the shell with a live code in it — precisely what #1136's `/dev/tty` design exists to prevent.

Also fixed: superseded and late pastes now get an answer instead of silence; a stray callback renders as a notice rather than "Authentication Failed — please try again"; the flow says once that it ignored something, so a probed login isn't silent for 30 minutes; and `gcx cloud login` gives the same `DetailedError` guidance as `gcx login` for the same flag conflicts.

Root cause for four of these: the paste arm had **no test coverage at all**. Every flow-level test forces agent mode, which nils the paste watcher, and no test can turn agent mode off without opening a real browser. It is now extracted as `resolvePaste` and driven directly across all seven branches. I checked the new tests fail without the fixes.

**Left open — worth a second opinion before merge.** Threading `--oauth-callback-port` into the Cloud leg changes the `redirect_uri` sent to grafana.com from the historical 54321-54399 range to an arbitrary user-chosen port. That is fine if GCOM applies the RFC 8252 §7.3 loopback-port exemption, and broken if it validates registered redirect URIs by exact match. I cannot verify that from here. Whoever owns the gcx OAuth client registration in GCOM should confirm before this merges; if it is exact-match, the Cloud leg needs to stay inside the registered range.

`GCX_AGENT_MODE=false mise run all` green, `-race` green on all four affected packages.

awsome-o:
A verification pass over the last two commits (two agents, adversarial, pointed at this branch) checked whether the fixes actually do what they claim. It found one that did not and one test that could not fail. Both fixed in `b7a72e61`.

**The test-leak guard was backwards.** `t.Cleanup` is LIFO, and I registered the release *before* `startStackFlow` — so the httptest server's `Close` ran first, blocked on the still-parked exchange handler, and the release never ran. Precisely the hang it was added to prevent. Reproduced with an injected early `t.Fatal`: 45s timeout and a goroutine dump. Registering after reverses the order; the same injection now reports in 0.01s.

**One test could not fail.** `TestResolvePaste_DeadlineDuringExchangeSkipsTheReprompt` never called `resolvePaste` — it drove the arbiter directly, then asserted on a buffer nothing in the test wrote to. Removed rather than patched; the state machine it meant to cover is already pinned elsewhere.

That combination left `case !arb.release():` as the single uncovered statement in both legs, with no test failing when it is reverted — and it is the arm my previous commit message headlined. Having looked at it properly: after the spent-code rule, every failure that can reach it is a parameter check with no I/O, so the window is a few instructions, not the minutes I implied. It is a genuine guard for a narrow race, not for the scenario I described. I have said so in the comment and left it deliberately untested rather than covering it with something that only looks like a test.

Also closed: the Cloud leg's manual mode had no no-deadline test (reverting that half failed nothing); nothing pinned the 30-minute constant the docs promise; a bool was read across goroutines unsynchronised in `TestHTTPAndPasteGatesAgreeOnOwnership`; the cloud port rejections were pinned by message wording rather than `errors.As` on `DetailedError`; and the GCOM success case did not assert the token its stack-leg twin asserts.

**And the open question from my earlier comment is now answered — favourably.** grafana.com applies the RFC 8252 §7.3 loopback exemption: `matchesRedirectUri` (added in grafana-com#19063, the PR that introduced this PKCE flow) compares scheme, host and path and ignores the port for `127.0.0.1`/`localhost`, pinned there by a test named *allows any port for a 127.0.0.1 loopback redirect*. So any `--oauth-callback-port` value is accepted on the Cloud leg and **#1148 is safe on any port** — no need to stay inside 54321-54399. I was wrong to call it unverifiable; the trail was two links from this branch's own git history. The hedging in `gcom.go` and `auth-system.md` is replaced with the actual rule, with the source cited.

No open items from the review remain. `GCX_AGENT_MODE=false mise run all` green (lint 0 issues after clearing a stale cross-worktree golangci-lint cache), `-race` green, both `resolvePaste` legs at 94.7% with the uncovered statement documented above.

awsome-o:
Agreed on all three, and I reproduced each before fixing rather than taking them on reading. Fixed in `b329fadf`; PR body corrected too — thanks, it was stale on three separate counts.

**[P2] The unusable paste is the serious one, and it is the same shape as #1147 itself.** Deterministic repro: a paste carrying our state but no code claimed the flow, a concurrent browser callback got `410 Gone`, then the paste released — real callback unrecoverable, login waiting out its deadline.

The fix is structural rather than a guard: **claim as late as possible**. `handleCallbackParams` splits into `validateCallbackParams` (no side effect, every rejection retryable) and `completeCallback` (spends the code, every error terminal). The paste path validates before it claims. The HTTP path still claims first, which is correct there — a callback carrying our state is ours whatever is wrong with it.

That has a payoff beyond the bug: every post-claim failure is now spent, so nothing can hand a flow back, so `release()` and `lateDeadline` became unreachable. Both removed — along with the one `resolvePaste` arm that had no test and could not sensibly get one, which you may recall was the loose end from the previous round. The arbiter is now `idle → exchanging → done` plus `expired`. `claimPastedCallback` went too; the HTTP-versus-paste race is proven through the real `resolvePaste` now, which is a stronger claim than the helper was.

**[P3] Confirmed exactly as described** — recovering the panic means the handler returns without writing, so Go sent an empty `200`. Measured it at 200, now 500, pinned in the panic test.

**[P2] Confirmed.** `grafanaAuthOptions` never saw the port, and in a remote session manual OAuth is the *promoted default* — so it was the likely choice, not a corner. Dropped from the menu when a port is fixed, with the remote promotion suppressed too, across all three target shapes.

All three are negative-controlled: reverting each individually fails a test.

**Docs and PR body**: `login.md` now says a Cloud login *can* involve two browser steps and that `--yes` skips the Cloud one. The body no longer claims manual mode is bounded, the timeout is ten minutes, or that exchange failures re-prompt — all three were stale, and the last one was the reason the earlier round had a bug.

**Agreed on process too.** Staying draft until #1136 merges, then retarget, rebase onto main, re-inspect and rerun everything. I have added the CI gap to the PB body's risk list so it cannot be missed at ready-for-review time — no normal CI has run on any head of this branch.

`GCX_AGENT_MODE=false mise run all` green (lint 0 issues after clearing a stale cross-worktree golangci-lint cache), `-race` green on all four packages.

PR #1158, opened 2026-08-11
feat(irm): add escalation-chain and date-window filters to alert-groups list

### LINKED ISSUE #1156: irm oncall alert-groups list: no --escalation-chain filter (API supports it)
## Problem

`gcx irm oncall alert-groups list` exposes `--integration` and `--team`, but there is
no way to filter by **escalation chain**, even though the OnCall API supports it.

The API advertises the filter in its own filter-discovery endpoint:

```
GET /api/plugins/grafana-irm-app/resources/alertgroups/filters/

search             | search
team               | team_select
integration        | options
escalation_chain   | options      <-- not exposed by gcx
acknowledged_by    | options      <-- not exposed by gcx
resolved_by        | options      <-- not exposed by gcx
started_at         | daterange    <-- not exposed by gcx
resolved_at        | daterange    <-- not exposed by gcx
label              | alert_group_labels
...
```

and it works when passed straight through:

```
GET /api/plugins/grafana-irm-app/resources/alertgroups/?escalation_chain=<CHAIN_ID>
```

## Why this filter in particular

The escalation chain is what determines **who actually gets paged**, so it is the
filter you need in order to attribute alert load to a rotation.

Filtering by integration is not a substitute, because integration -> schedule is
many-to-many. A single Alertmanager integration commonly fans out through several
routes to several different chains, each notifying a different schedule: one rota
for `"severity": "critical"`, a different rota for deploy/rollout-labelled traffic,
plus escalation tiers behind both.

Alert group records themselves contain **no `route` or `escalation_chain` field**, so
without this filter there is no way to recover which rotation an alert was routed
to. Chain-filtered queries over the same integration return cleanly disjoint
subsets, which is exactly the disambiguation needed.

## Suggested fix

Add `--escalation-chain strings` (repeatable, comma-separated) to
`irm oncall alert-groups list`, following the existing `--integration` / `--team`
pattern.

Worth considering at the same time, since they are all in the API's filter list and
none are currently reachable: `--acknowledged-by`, `--resolved-by`, `--label`, and a
`--started-at` / `--resolved-at` date range. Note `--max-age` is not a substitute for
a date range because it only anchors to now, so a historical window cannot be
expressed.

## Related

- #207 covers list filters for the OnCall provider in the other direction
  (grafanactl reaching gcx parity). Its alert-groups list mentions `--route-id` and
  `--started-at` but not `escalation_chain`.

---

Filed by **Claude Code**, on behalf of @tomwilkie, while using gcx to analyse
OnCall rotation shapes and alert load. gcx v1.0.0 (homebrew).

tomwilkie:
Fixes #1156.

## What

`gcx irm oncall alert-groups list` gains six filters that the OnCall internal API already advertises on its own filter-discovery endpoint but that gcx could not reach:

| Flag | Query param |
|---|---|
| `--escalation-chain` | `escalation_chain` (repeated) |
| `--acknowledged-by` | `acknowledged_by` (repeated) |
| `--resolved-by` | `resolved_by` (repeated) |
| `--from` / `--to` | `started_at=<from>_<to>` |
| `--resolved-from` / `--resolved-to` | `resolved_at=<from>_<to>` |

`--escalation-chain` is the one the issue is really about. Alert group records carry no `route` or `escalation_chain` field, so there was previously no way to attribute alert load to a rotation. `--integration` is not a substitute: one integration fans out through several routes to several chains, each notifying a different schedule.

`--max-age` only ever anchors to now, so a historical window could not be expressed at all. `--from`/`--to` accept RFC3339, a unix timestamp, or a relative expression like `now-30d` via `shared.ParseTime`, matching the nearest sibling `gcx irm incidents list`. `--max-age` and `--from`/`--to` are mutually exclusive because both compile into `started_at`.

## Notable decisions

- **IDs, not names**, matching the shipped `--team` / `--integration` "PK" convention. Help text points at `escalation-chains list` / `users list` for discovery.
- **One-sided windows are completed, not rejected.** The API's daterange params need both ends, so an omitted end becomes now and an omitted start becomes the unix epoch.
- **All time expressions now resolve to absolute windows in `resolveAlertGroupListFilters`**, which removes the duplicated duration parsing and layout formatting that previously sat in the transport layer.
- **`list` only.** The bulk action verbs are unchanged. ADR 001 § 3.4's mirror rule enumerates `--team`, `--integration`, `--status`, `--max-age`, `--mine`, and that set still matches; these are analysis filters rather than bulk-mutation targets.
- **`--label` deliberately left out.** OnCall serializes labels as `{key:{id,name},value:{id,name}}`, so the wire filter format is very likely an id pair rather than `key:value`. Worth a follow-up issue once that is confirmed.

The SA-token fallback path speaks the public API, which has none of these filters and only a lower bound on `started_at`. `--from` maps onto that bound; every other dropped filter is named in the existing unsupported-filter note.

## Verification

`GCX_AGENT_MODE=false mise run all` equivalents: `golangci-lint run ./internal/providers/irm/...` clean, `GCX_AGENT_MODE=false go test -race ./...` green, CLI reference regenerated.

New tests: wire-encoding table over the query params (repeated, never comma-joined; naive-UTC `<from>_<to>` pairs; `--to` alone defaulting the start to the epoch), validation table for the mutual exclusion / unparseable / inverted-window errors, and a legacy-path test asserting each dropped filter is named.

Smoke-tested against a live stack, which confirms the params are honored server-side rather than silently ignored:

- a bogus chain ID returns `HTTP 400 {"escalation_chain":["Select a valid choice. ..."]}` — same for `acknowledged_by` and `resolved_by`
- a real chain returns 68 of 200+ groups, and two chains over the same set have zero overlap — the disjointness the issue asks for
- `--from 2026-06-01 --to 2026-06-02` returns only groups started on June 1, proving a historical window that `--max-age` cannot express
- `--resolved-from now-2d --resolved-to now` returns only `resolved` groups with resolved timestamps inside the window

### DIFF
--- a/claude-plugin/skills/oncall-triage/SKILL.md
+++ b/claude-plugin/skills/oncall-triage/SKILL.md
@@ -27,13 +27,18 @@ gcx configured with an active context that targets a Grafana stack with OnCall e
 gcx irm oncall alert-groups list
 ```
 
-Default filter: `state in {firing, acknowledged, silenced}` and `is_root=true` (matches the OnCall UI). See `--help` for the full flag set; common narrowers are `--state`, `--team <PK>`, `--integration <PK>`, `--mine`, `--max-age 24h`. Team / integration filters take PKs, not names — resolve first:
+Default filter: `state in {firing, acknowledged, silenced}` and `is_root=true` (matches the OnCall UI). See `--help` for the full flag set; common narrowers are `--state`, `--team <PK>`, `--integration <PK>`, `--escalation-chain <ID>`, `--mine`, `--max-age 24h`. Team / integration / chain filters take IDs, not names — resolve first:
 
 ```bash
 gcx irm oncall teams list -o json | jq -r '.[] | "\(.metadata.name): \(.spec.name)"' | grep -i "my team"
+gcx irm oncall escalation-chains list
 ```
 
+To attribute alert load to a rotation, filter by escalation chain, not integration: one integration routes through several chains to several schedules, and the alert group record carries no chain field of its own. Chain-filtered queries over the same integration return disjoint subsets.
+
 Escape hatches: `--all` (drops both defaults — returns resolved + child groups), `--include-child-groups`, `--state resolved`.
+
+For a historical window use `--from` / `--to` (RFC3339, unix timestamp, or `now-30d`); `--max-age` only anchors to now, and the two cannot be combined. `--resolved-from` / `--resolved-to` bound resolved_at instead. `--acknowledged-by <user-id>` / `--resolved-by <user-id>` attribute handling to a person.
 
 Default table: `ID TITLE SEVERITY STATE TEAM SUBJECT AGE`. `-o wide` adds `RULE` (Grafana rule URL) and `ALERTS` (group-wide count). Use `-o wide` when the user needs the rule pivot.
 
--- a/docs/reference/cli/gcx_irm_oncall_alert-groups_list.md
+++ b/docs/reference/cli/gcx_irm_oncall_alert-groups_list.md
@@ -13,27 +13,62 @@ Use --all to bypass these defaults entirely (returns resolved and child groups t
 Use --state to override the status filter (e.g. --state firing,acknowledged).
 Use --include-child-groups to keep the status default but include child groups.
 
+Alert group records carry no escalation chain field, so --escalation-chain is the
+only way to attribute alert load to the rotation that was actually paged. It is not
+interchangeable with --integration: one integration routes to several chains.
+
+--max-age anchors to now; use --from/--to for a historical started-at window, and
+--resolved-from/--resolved-to for a resolved-at window. They accept RFC3339, a unix
+timestamp, or a relative expression like now-30d. --max-age and --from/--to cannot
+be combined.
+
 ```
 gcx irm oncall alert-groups list [flags]
+```
+
+### Examples
+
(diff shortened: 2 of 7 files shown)


willgallego-grafana (approved):
lgtm!

wardbekker (commented):
@tomwilkie The change looks good. All checks pass. I found no defect. Three optional items remove duplicate work:

1. `oncall_actions.go:855` and `oncall_commands_extra.go:278` convert `--max-age` the same way. One helper replaces both. It removes about 8 lines.

2. `resolveTimeWindow` (`oncall_commands_extra.go:155`) returns a `bool` that a `*timeWindow` result makes unnecessary. Each of the two call sites then loses three lines.

3. `Validate` (`oncall_commands_extra.go:118`) parses both windows. It then discards both results. `resolveAlertGroupListFilters` runs before `loader.LoadOnCallClient`, so it already returns the same errors. Keep the mutual-exclusion check. Move the two window test cases across.

`oncall_commands_extra.go:732` reads `opts.To` and not `filters`. That is correct, because only `opts` records which flag the user typed.

Careful: a push removes the approval of @willgallego-grafana. Group the changes into one push.

dafydd-t (approved):
Thanks! we'll address the followups later

tomwilkie:
Thanks all! Sorry I missed the review feedback.

PR #1162, opened 2026-08-12
Set AllowsEdits on gcx-managed resources so dashboards stay UI-editable

### LINKED ISSUE #837: After switching from grafanactl to gcx my dashboards have become provisioned and cannot be updated anymore
### Description

My workflow for a single Grafana instance (v13.0.2):

    create, edit, update dashboards in Grafana UI
    backup dashboards to git repository
    restore only rarely if necessary.

Now after switching from grafanactl to gcx and running a gcx resources push this has turned all my dashboards into provisioned and non-updatable.

How can I set my existing dashboards to updatable and get back the original workflow?

The dashboards should have been changeable/editable through the Grafana UI.

1. create a Grafana dashboard
2. pull it using `grafanactl resources pull`
3. push it using `gcx resources push`
4. edit dashboard and then try to save it, this will show the notice that the dashboard cannot be overwritten and only a copy created or saved.

### gcx Version

gcx version 0.4.0 848bffbe3304c59120a9a372cd66ffa5cfd501c4 (installed from repo binaries)

### Environment Details

WSL2 Docker host, Ubuntu 26.04 devcontainer, Grafana server v13.0.2 (3fcdbc5a)

- [x] I checked my configuration with `gcx config view`
- [x] I verified I'm using the correct context with `config current-context`
- [x] I verified my authentication token/credentials are valid

dashboard metadata excerpt (* = redacted):

```json
  "metadata": {
    "annotations": {
      "grafana.app/createdBy": "user:*",
      "grafana.app/folder": "*",
      "grafana.app/managedBy": "kubectl",
      "grafana.app/managerId": "gcx",
      "grafana.app/sourcePath": "file://resources/dashboards.v1.dashboard.grafana.app/*",
      "grafana.app/updatedBy": "service-account:*",
      "grafana.app/updatedTimestamp": "*"
    },
    "creationTimestamp": "*",
    "generation": 56,
    "labels": {
      "grafana.app/deprecatedInternalID": "1468"
    },
    "name": "*",
    "namespace": "*",
    "resourceVersion": "*",
    "uid": "*"
  },
```

stephaniehingtgen:
If you create a dashboard with gcx, it is not allowing you to update in the UI afterwards because it has managed by fields without the allows edit, so it fails the check [here](https://github.com/grafana/grafana/blob/b7a238f355d155542fc6eabcbda96b48a218d17c/public/app/features/dashboard-scene/scene/DashboardScene.tsx#L1477-L1491). With [ManagerProperties.AllowsEdits](https://github.com/grafana/grafana/blob/b7a238f355d155542fc6eabcbda96b48a218d17c/pkg/apimachinery/utils/manager.go#L1-L20) set, users can update the dashboard after gcx creates it.

Fixes #837

### DIFF
--- a/internal/resources/process/managerfields.go
+++ b/internal/resources/process/managerfields.go
@@ -22,8 +22,8 @@ func (m *ManagerFieldsAppender) Process(r *resources.Resource) error {
 	}
 
 	r.Raw.SetManagerProperties(utils.ManagerProperties{
-		Kind:     resources.ResourceManagerKind,
-		Identity: "gcx", // TODO: use version information to set the identity.
+		Kind:        resources.ResourceManagerKind,
+		Identity:    "gcx", // TODO: use version information to set the identity.
 		AllowsEdits: true,
 	})
 
--- a/internal/resources/process/managerfields.go
+++ b/internal/resources/process/managerfields.go
@@ -24,6 +24,7 @@ func (m *ManagerFieldsAppender) Process(r *resources.Resource) error {
 	r.Raw.SetManagerProperties(utils.ManagerProperties{
 		Kind:     resources.ResourceManagerKind,
 		Identity: "gcx", // TODO: use version information to set the identity.
+		AllowsEdits: true,
 	})
 
 	// TODO: should we set timestamp & checksum as well?
--- a/internal/resources/process/managerfields_test.go
+++ b/internal/resources/process/managerfields_test.go
@@ -49,9 +49,10 @@ func TestManagerFieldsAppender(t *testing.T) {
 						"name":      "example",
 						"namespace": "default",
 						"annotations": map[string]any{
-							utils.AnnoKeyManagerKind:     string(resources.ResourceManagerKind),
-							utils.AnnoKeyManagerIdentity: "gcx",
-							utils.AnnoKeySourcePath:      "file://some/test/path.json",
+							utils.AnnoKeyManagerKind:        string(resources.ResourceManagerKind),
+							utils.AnnoKeyManagerIdentity:    "gcx",
+							utils.AnnoKeyManagerAllowsEdits: "true",
+							utils.AnnoKeySourcePath:         "file://some/test/path.json",
 						},
 					},
 					"spec": map[string]any{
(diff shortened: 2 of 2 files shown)


wardbekker:
Both TODO comments in `internal/resources/process/managerfields.go` come from `main`, not from this pull request. Here is a suggestion for each, if you want to remove them.

### 1. Set `Identity` from the version (line 26)

The package `internal/version` already holds the value. `version.Get()` returns `SNAPSHOT` when the linker does not set a version.

```go
import (
	"github.com/grafana/gcx/internal/resources"
	"github.com/grafana/gcx/internal/version"
	"github.com/grafana/grafana/pkg/apimachinery/utils"
)

r.Raw.SetManagerProperties(utils.ManagerProperties{
	Kind:        resources.ResourceManagerKind,
	Identity:    "gcx/" + version.Get(),
	AllowsEdits: true,
})
```

This change is safe for two reasons:

- `IsManaged()` compares the `Kind` only (`internal/resources/resources.go:166`). No gcx code reads `Identity`.
- `ServerFieldsStripper` deletes `AnnoKeyManagerIdentity` on pull (`internal/resources/process/serverfields.go:34`). An identity that changes per release does not change the pull output.

Three test assertions need the new value: `managerfields_test.go:53`, `:113`, and `:135`. The tests do not call `version.Set`, so the value is `gcx/SNAPSHOT`. Build the expected string with `"gcx/" + version.Get()` in the test. A release build then keeps the test green.

### 2. Checksum and timestamp (line 30)

I recommend that you delete this TODO, or that you move it to an issue. There are two reasons:

- `SourceInfo` carries the `Path` and the `Format` only (`internal/resources/resources.go:378`). The processor cannot compute a checksum or a timestamp from that data.
- No gcx code reads `Checksum` or `TimestampMillis`. Upstream describes both fields as reconciliation data for a manager that syncs a source into Grafana. gcx pushes one time per command.

If you do want the two fields, the reader is the correct place to compute them. `ReadFile` already holds the bytes:

```go
// internal/resources/resources.go
type SourceInfo struct {
	Path       string
	Format     format.Format
	Checksum   string    // hex SHA-256 of the file bytes
	ModifiedAt time.Time // file modification time
}
```

```go
// internal/resources/local/reader.go — ReadFile already has `data`
sum := sha256.Sum256(data)
// pass hex.EncodeToString(sum[:]) and the os.Stat modification time into readRaw
```

The processor then copies them:

```go
src := utils.SourceProperties{Path: r.Source.String()}
if r.Source.Checksum != "" {
	src.Checksum = r.Source.Checksum
}
if !r.Source.ModifiedAt.IsZero() {
	src.TimestampMillis = r.Source.ModifiedAt.UnixMilli()
}
r.Raw.SetSourceProperties(src)
```

`ReadBytes` has no file path and no modification time. The two guards keep both annotations empty in that case.

dafydd-t (approved):
Thank you!

PR #1163, opened 2026-08-12
feat(irm): resolve PIR document links from incident hook runs

musculman:
## Summary

The post-incident review (PIR) link is not carried in the IRM incident payload. That is by design: PIRs are optional and, as of today, only the Google Workspace integration creates them. The link is recorded on the *integration hook run* that copied the PIR template, not on the incident.

This adds `gcx irm incidents get-pir <incident-id>`, which reads an incident's hook runs and resolves the PIR document URL from them.

- **Client**: `GetHookRuns` and `GetPIRURL` on the incidents client, plus `HookRun` / `HookRunMetadata` / `HookRunField` types. Only the fields the lookup reads are modelled.
- **Resolution**: finds a hook run with `hookID` of `grate.google.copyFile` or `grate.googleworkspace.copyTemplate`, then pulls the `fileURL` entry from `metadata.fields`.
- **Path**: `IntegrationService` is served from the unversioned base path alongside `SeveritiesService` and `IncidentContextService`, not the `/v1` path the other incident calls use. A test pins this so it does not silently drift.
- **Output**: prints the bare URL so it pipes cleanly (`| xargs open`). An incident without a PIR is a normal outcome — stdout stays empty, exit code is 0, and the explanation goes to stderr. Agent/JSON mode returns `{"incidentID": "...", "pirURL": ""}`.

### DIFF
--- a/internal/providers/irm/incidents_client.go
+++ b/internal/providers/irm/incidents_client.go
@@ -38,7 +38,41 @@ const (
 	actAddPath        = incidentBasePath + "/ActivityService.AddActivity"
 	sevGetPath        = incidentLegacyBasePath + "/SeveritiesService.GetOrgSeverities"
 	ctxQueryPath      = incidentLegacyBasePath + "/IncidentContextService.QueryIncidentContext"
+	// IntegrationService is likewise not part of the documented v1 API and
+	// is served from the unversioned base path.
+	hookRunsPath = incidentLegacyBasePath + "/IntegrationService.GetHookRuns"
 )
+
+// pirFileURLField is the hook-run metadata field holding the document URL.
+const pirFileURLField = "fileURL"
+
+// isPIRHookID reports whether the hook copies a post-incident review template
+// into a document. PIRs are optional and only the Google Workspace integration
+// creates them, so these are the only hook runs that can carry a PIR link.
+func isPIRHookID(hookID string) bool {
+	switch hookID {
+	case "grate.google.copyFile", "grate.googleworkspace.copyTemplate":
+		return true
+	default:
+		return false
+	}
+}
+
+// pirURLFromHookRuns returns the PIR document URL recorded on a PIR-creating
+// hook run, or "" when the incident has no PIR document.
+func pirURLFromHookRuns(runs []HookRun) string {
+	for _, run := range runs {
+		if !isPIRHookID(run.HookID) || run.Metadata == nil {
+			continue
+		}
+		for _, f := range run.Metadata.Fields {
+			if f.Key == pirFileURLField && f.Value != "" {
+				return f.Value
+			}
+		}
+	}
+	return ""
+}
 
 // Client is an HTTP client for the Grafana IRM Incidents API.
 type IncidentClient struct {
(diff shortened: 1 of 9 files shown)


internal/providers/irm/incidents_client.go
```diff
@@ -38,8 +38,42 @@ const (
...
+// pirURLFromHookRuns returns the PIR document URL recorded on a PIR-creating
+// hook run, or "" when the incident has no PIR document.
+func pirURLFromHookRuns(runs []HookRun) string {
+	for _, run := range runs {
+		if !isPIRHookID(run.HookID) || run.Metadata == nil {
+			continue
+		}
+		for _, f := range run.Metadata.Fields {
+			if f.Key == pirFileURLField && f.Value != "" {
+				return f.Value
```
olegbespalov: `pirURLFromHookRuns` returns the first match, but `GetByOrgIDAndIncidentID` has no `ORDER BY`. So if both `copyFile` and `copyTemplate` ran (or the hook was re-run), which URL we print can flip between calls for the same incident. IRM shows all of them as bookmarks; we silently pick one.

Worth preferring latest by `lastRun`, or at least calling out that the pick is arbitrary.

internal/providers/irm/incidents_commands_impl.go
```diff
@@ -290,6 +290,102 @@ func NewGetCommand(loader GrafanaConfigLoader) *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+
+			// Absence is a normal outcome, not an error — the note is a
+			// diagnostic so it stays out of the stdout result.
+			if url == "" {
+				cmdio.Info(cmd.ErrOrStderr(),
+					"Incident %s has no PIR document (only the Google Workspace integration creates them).",
+					incidentID)
```
olegbespalov, line 362: Empty stdout + exit 0 for no PIR is the right pipe contract. The `cmdio.Info` on stderr is a bit awkward though — especially in agent mode, where `"pirURL":""` already says the same thing and DESIGN wants stderr quiet unless something actually went wrong.

Maybe drop the note under agent mode, or only print it with `-v`.

internal/providers/irm/incidents_client.go
```diff
@@ -38,8 +38,42 @@ const (
...
+
+// pirURLFromHookRuns returns the PIR document URL recorded on a PIR-creating
+// hook run, or "" when the incident has no PIR document.
+func pirURLFromHookRuns(runs []HookRun) string {
+	for _, run := range runs {
+		if !isPIRHookID(run.HookID) || run.Metadata == nil {
+			continue
+		}
+		for _, f := range run.Metadata.Fields {
+			if f.Key == pirFileURLField && f.Value != "" {
```
olegbespalov: Nit: `copyTemplate` sets both `metadata.URL` and `fields[fileURL]`. Fields-only matches what IRM bookmarks do today, so this is fine — but a URL-only / partial run would currently look like “no PIR”. A `metadata.URL` fallback would be a cheap safety net if you feel like it.

olegbespalov (dismissed):
Looks good — resolution matches what IRM does for PIR bookmarks (`copyFile` / `copyTemplate` + `fileURL` on the unversioned IntegrationService path). A couple of small notes below, nothing blocking.

internal/providers/irm/incidents_client.go
musculman: Good point, fixed by collecting every PIR link on the incident and choosing by comparison instead of arrival order: most recent lastRun first, then hook rank (copyTemplate over copyFile, as the PIR-specific hook), then the URL itself.

internal/providers/irm/incidents_commands_impl.go
musculman: You are right, and after a short conversation with Cursor it agreed that it should be stricter than that. It is now gated on !agent.IsAgentMode() — humans keep the explanation, agents get the empty pirURL and a silent stderr. I don't think this is too useful for a `-v` flag so did not add it

internal/providers/irm/incidents_client.go
musculman: Fixed. HookRunMetadata now has a URL field, read only when there's no non-empty fileURL entry, so normal runs resolve exactly as they did before and only the URL-only partial case changes behaviour.

olegbespalov (dismissed):
LGTM 👍

wardbekker (commented):
The pull request models only the fields that the command reads, so the wire types stay small.

`internal/providers/irm/incidents_client.go:639` — `GetPIRURL` makes a second request. `Incident.IncidentHookRuns` already exists on main, so `gcx irm incidents get -o json` can carry the same hook runs. The field type is `[]any`, so I cannot check the contents from the code. Does `IncidentsService.GetIncident` return the `fileURL` entry? One request is enough if it does.

`internal/providers/irm/incidents_client.go:54-127` — `pirHookRank`, `isPIRHookID`, `pirCandidate`, `beats`, `pirURLFromHookRun`, and `pirURLFromHookRuns` select one URL in about 80 lines. One loop over a rank map gives the same deterministic result in about half the lines.

`internal/providers/irm/incidents_types.go:45` — `LenientTime.MarshalJSON` is dead code. No code marshals a `HookRun`. The `omitzero` tag at line 143 has no effect for the same reason.

musculman:
> The pull request models only the fields that the command reads, so the wire types stay small.
> 
> `internal/providers/irm/incidents_client.go:639` — `GetPIRURL` makes a second request. `Incident.IncidentHookRuns` already exists on main, so `gcx irm incidents get -o json` can carry the same hook runs. The field type is `[]any`, so I cannot check the contents from the code. Does `IncidentsService.GetIncident` return the `fileURL` entry? One request is enough if it does.
> 
> `internal/providers/irm/incidents_client.go:54-127` — `pirHookRank`, `isPIRHookID`, `pirCandidate`, `beats`, `pirURLFromHookRun`, and `pirURLFromHookRuns` select one URL in about 80 lines. One loop over a rank map gives the same deterministic result in about half the lines.
> 
> `internal/providers/irm/incidents_types.go:45` — `LenientTime.MarshalJSON` is dead code. No code marshals a `HookRun`. The `omitzero` tag at line 143 has no effect for the same reason.

Ok this is excellent feedback on many things I was not aware of. I also have dumped a full payload of an incident to cursor and asked to optimize the code so an update is coming that should take care of all of this and hopefully make it easier. thanks again!

musculman:
On IncidentHookRuns:

Checked against a real incident with a PIR: GetIncident returns incidentMembership, taskList and refs, but no incidentHookRuns at all, so the dedicated call is needed, not a second one. Caveat: we only send {"incidentID": id}

On the selection code:

Agreed. It's now one comparator over the hook runs (lastRun, hook rank, URL) pirCandidate, beats and isPIRHookID are gone, about half the code. A package-level rank map isn't an option though: gochecknoglobals rejected exactly that earlier in this PR.

On the dead code:

Right on both. LenientTime is deleted entirely, the real payload has lastRun as an RFC3339 string, so it's now FlexTime like every other timestamp in this client, and omitzero went with it.

thanks again @wardbekker

dafydd-t (approved):
Great idea! thank you @musculman

PR #1166, opened 2026-08-12
refactor(resources): set the manager identity from the gcx version

wardbekker:
Follows up on the review comment in #1162, which reported two TODO comments in `internal/resources/process/managerfields.go`.

## 1. The manager identity now carries the version

The annotation `grafana.app/manager-identity` held the constant `gcx`. An operator could not see which gcx build pushed a resource.

`process.ManagerIdentity()` now returns `gcx/<version>`, for example `gcx/3.0.0`. A build without version information reports `gcx/SNAPSHOT`.

The change is safe for two reasons:

- `IsManaged()` compares the manager kind only (`internal/resources/resources.go:166`). No gcx code reads the identity.
- `ServerFieldsStripper` deletes the identity annotation on pull (`internal/resources/process/serverfields.go:34`). An identity that changes per release does not change the pull output.

## 2. The checksum and the timestamp stay empty

This pull request removes the second TODO comment. It records the decision in the code and in `docs/architecture/resource-model.md`.

Grafana uses `SourceProperties.Checksum` and `SourceProperties.TimestampMillis` to reconcile a resource from a source over time. gcx pushes one time per command. No gcx code reads the two fields back. `SourceInfo` also carries the path and the format only, so the processor has no data to compute either field.

## Tests

- `TestManagerIdentity` covers the snapshot value and a set version.
- `TestManagerFieldsAppender` expects `process.ManagerIdentity()`. One test case keeps a stale `gcx` identity in the input, which proves that the processor overwrites the old value.

### DIFF
--- a/docs/architecture/resource-model.md
+++ b/docs/architecture/resource-model.md
@@ -65,8 +65,11 @@ Every `Resource` carries a `SourceInfo` (line 374) recording where it came from.
 
 Resources carry manager metadata in annotations (via `GrafanaMetaAccessor`):
 - `grafana.app/manager-kind` — which tool manages the resource (gcx uses `utils.ManagerKindKubectl` as placeholder, line 19)
-- `grafana.app/manager-identity` — identity string ("gcx")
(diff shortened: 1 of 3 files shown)


javiermolinar (approved):
LGTM

PR #1167, opened 2026-08-12
docs(review): share one review process between local runs and CI

dafydd-t:
## Summary

Make /review-pr, /integrate-with-gcx and CI reviews all use the same material for assessing the diff.

- Add a  `review-pr` skill to guide the review process: It describes the intention, runs a typical review, and then reviews the diff against the agent-facing docs in the repo, as well as some nudges to surface over-engineering. It attempts to distil its findings into actionable results that consider all the findings, not just one at a time.
- Routes `@claude review` to use claude-code-review workflow instead of the general-purpose one, so we get all these extra checks on re-runs
- Points `claude-code-review.yml` at that skill, so a PR gets the same treatment whether a developer reviews it locally or the workflow does
- Adds `T12: Over-engineering` to `self-review.md`, widens `T5` to cover any duplicated utility helper, and corrects the exported-for-tests rule so `export_test.go` is never a finding

## Why

The automated reviewer triggered on a new PR was OK. but running "@/claude review" would trigger a generic review prompt, with nothing particularly related to this repo. This consolidates that, and also makes the pre-PR, CI and PR-review guidance consistent.

## Notes

- `@claude review` will not work until this is on `main`: `issue_comment` workflows always run from the default branch.


.claude/skills/integrate-with-gcx/references/self-review.md
```diff
@@ -401,3 +411,34 @@ Where a style guide and the entire surrounding command family disagree — a
...
+not this trigger.
+
+1. **Count the call sites now.** Aim for three before an abstraction — an
+  interface, a wrapper type, a factory, a config option, a plugin seam — earns
+  its place. One caller means write it inline and extract when the second
+  arrives; the second caller's needs are what should choose the signature. Read
+  every field of a new options or request struct and name the caller that sets
+  it. A field nothing sets is the strongest evidence in this whole document,
+  and it usually drags dead branches behind it: a `Continue` flag nobody sets
+  keeps a disk read, an error path, and a test alive.
```
dafydd-t: this contradicts itself: stating 3 required, but the blurb only talks about 2

.claude/skills/integrate-with-gcx/references/self-review.md
```diff
@@ -401,3 +411,34 @@ Where a style guide and the entire surrounding command family disagree — a
...
+2. **Check your own PR body for the admission.** If you wrote that the new entry
+  point exists "so other command trees can…", or "to avoid duplicating X in
+  future", that is this check failing in your own words. Name the second caller
+  and link it, or cut the generality and say what you cut.
```
dafydd-t: something like "check the PR description, or any attached comments from authors if its an existing GH PR, for any times there are violations that might have been on purpose"

.claude/skills/integrate-with-gcx/references/self-review.md
```diff
@@ -401,3 +411,34 @@ Where a style guide and the entire surrounding command family disagree — a
...
+  caller that builds the struct, the validator that already rejected it. A
+  tolerant parser for a heading that every bundled doc already carries is a
+  branch that can never run, plus the tests that "cover" it. This is a different
+  question from [T8](#t8-tests-that-cannot-fail) — T8 asks whether a test would
+  catch a real defect, T12 asks whether the input exists at all. A test whose
+  fixture the code will never see is a finding under both.
+Close by stating the smaller version in the PR body: the deletions and merges
+that would resolve every finding above and in T5, and the resulting size change.
+Reviewers compute this anyway. Writing it yourself turns a review round into a
+decision.
```
dafydd-t:
```suggestion

Close by stating the smaller version in the PR body: the deletions and merges
that would resolve every finding above and in T5, and the resulting size change.
Reviewers compute this anyway. Writing it yourself turns a review round into a
decision.
```

.claude/skills/review-pr/SKILL.md
```diff
@@ -0,0 +1,258 @@
...
+   shipped behaviour, safety problems, and violations of CONSTITUTION.md or
+   DESIGN.md. Behaviour that changes silently inside a change described as a
+   refactor belongs here, as does a call to a function, field, or option that
+   does not exist.
+3. **Other** — should be fixed, does not block. Includes documentation that
+   contradicts the code it describes.
+4. **Over-engineering** — findings from T12 and T5, ranked and capped below.
+5. **The smaller version** — one consolidated remedy. Omit when section 4 is
+   empty.
+6. **Smaller points** — one line each, no discussion.
```
dafydd-t: combine "other" and "Smaller points"

.claude/skills/integrate-with-gcx/references/self-review.md
dafydd-t: accepted. it said "aim for three" but every sentence under it argued one-versus-two, and the skill's lock-in ranking was already using two. settled on two everywhere: an abstraction earns its place at the second real caller, because that caller's needs are what should pick the signature. dropped the three, which was folklore I brought in rather than anything this repo showed.

.claude/skills/integrate-with-gcx/references/self-review.md
dafydd-t: accepted, added as the other half of check 2. it only mined the description for a self-incriminating admission; it now also looks for a deviation the author explained on purpose, in the body or the existing thread, so a reviewer doesn't re-raise something already answered. and where the reasoning is there but wrong, argue with the reasoning rather than reporting the deviation as if nobody had noticed.

.claude/skills/integrate-with-gcx/references/self-review.md
dafydd-t: applied — missing blank line from when I edited T12. while checking I found a stated count of four above a list of five in the skill, same class of thing, fixed too.

.claude/skills/review-pr/SKILL.md
dafydd-t: accepted. merged into one "Other findings" list, worst first. the label already separates a nit from a worth-fixing defect, so the two headings were carrying nothing the labels weren't. section 4 overflow now lands there as one-liners instead of in a section that no longer exists.

javiermolinar (approved):
LGTM

PR #1168, opened 2026-08-12
feat: add experimental gcx traces baseline command

javiermolinar:
## Summary

Stacked on #1165 (`traces-diff-command`). Adds `gcx traces baseline TRACE_ID`: given a seed trace, retrieve same-operation candidate traces to compare against with `gcx traces diff`.

Having a command to get baseline traces is integral to the investigation workflow we are building. The idea is that, given a faulty trace, we can retrieve healthy traces based on it. To do that, we can simply leverage the power of TraceQL and build a query based on the structural aspects of the original trace.

Once we have the baselines, we can compare them with the faulty trace using the /diff endpoint.

### DIFF
--- a/internal/datasources/tempo/baseline.go
+++ b/internal/datasources/tempo/baseline.go
@@ -310,13 +310,15 @@ func buildBaselineResult(seedID string, profile seedProfile, resp *tempo.SearchR
 			DurationMs:        t.DurationMs,
 		}
 		if t.ServiceStats != nil {
-			spans := 0
+			spans, errorSpans := 0, 0
 			for _, s := range t.ServiceStats {
 				spans += s.SpanCount
+				errorSpans += s.ErrorCount
 			}
 			services := len(t.ServiceStats)
 			candidate.SpanCount = &spans
 			candidate.ServiceCount = &services
+			candidate.ErrorCount = &errorSpans
 		}
 		result.Candidates = append(result.Candidates, candidate)
 	}
--- a/internal/datasources/tempo/baseline.go
+++ b/internal/datasources/tempo/baseline.go
@@ -211,7 +211,7 @@ func (opts *baselineOpts) resolveWindow(profile seedProfile, now time.Time) (tim
 // health is not filtered here — 'gcx traces diff' surfaces downstream errors.
 func buildBaselineQuery(service, operation string, downstream []string) string {
 	var b strings.Builder
-	fmt.Fprintf(&b, "{ trace:rootService = %s && trace:rootName = %s } && { name = %s && status != error }",
+	fmt.Fprintf(&b, "{ trace:rootService = %s && trace:rootName = %s } && { name = %s && span:status != error && nestedSetParent = -1 }",
 		strconv.Quote(service), strconv.Quote(operation), strconv.Quote(operation))
 	for _, svc := range downstream {
 		fmt.Fprintf(&b, " && { resource.service.name = %s }", strconv.Quote(svc))
@@ -333,7 +333,7 @@ func parseSeedTrace(trace map[string]any) seedProfile {
 
 		// Root = the earliest-starting parentless span.
 		if parent, _ := span["parentSpanId"].(string); strings.TrimSpace(parent) == "" {
-			if !rootFound || start < rootStart {
+			if !rootFound || (start != 0 && (rootStart == 0 || start < rootStart)) {
 				rootFound, rootStart = true, start
 				profile.RootService = service
 				profile.RootOperation, _ = span["name"].(string)
--- a/internal/datasources/tempo/baseline.go
+++ b/internal/datasources/tempo/baseline.go
@@ -219,7 +219,7 @@ change.`,
 
 	cmd.Annotations = map[string]string{
 		agent.AnnotationTokenCost: "medium",
-		agent.AnnotationLLMHint:   "gcx traces baseline -d UID <trace-id> -o json",
+		agent.AnnotationLLMHint:   "gcx datasources tempo baseline -d UID <trace-id> -o json",
 		agent.AnnotationStability: agent.StabilityExperimental,
 	}
 
--- a/internal/datasources/tempo/baseline.go
+++ b/internal/datasources/tempo/baseline.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
@@ -46,7 +47,7 @@ func (opts *baselineOpts) setup(flags *pflag.FlagSet) {
 
 	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.tempo is configured)")
 	flags.StringArrayVar(&opts.Filters, "filter", nil, "Raw TraceQL spanset expression to refine candidates when unfiltered results are not valid comparisons (repeatable; ANDed with the generated query)")
-	flags.IntVar(&opts.Limit, "limit", 20, "Maximum number of candidates to return")
+	flags.IntVar(&opts.Limit, "limit", 20, "Maximum number of candidates to return; must be at least 1")
 	flags.StringVar(&opts.Window, "window", defaultBaselineWindow, "Search window padding applied before and after the seed trace's time range, so candidates from before or after the seed are eligible (e.g., 30m, 6h, 7d). Ignored when --from/--to are set")
 	// --from/--to give an absolute window override anchored on explicit times
 	// rather than the seed. --since is intentionally omitted: it anchors on now,
@@ -63,6 +64,12 @@ func (opts *baselineOpts) Validate() error {
 		if strings.TrimSpace(filter) == "" {
 			return errors.New("--filter must not be empty")
 		}
+	}
+	if opts.Limit < 1 {
+		return errors.New("--limit must be at least 1")
+	}
+	if strings.TrimSpace(opts.Window) == "" {
+		return errors.New("--window must not be empty")
 	}
 	pad, err := dsquery.ParseDuration(opts.Window)
 	if err != nil {
@@ -162,12 +169,10 @@ change.`,
 				return err
 			}
 
-			// The seed can match its own retrieval query, so fetch one extra and
-			// drop it, keeping up to --limit candidates.
-			fetchLimit := opts.Limit
-			if fetchLimit > 0 {
-				fetchLimit++
-			}
+			// The seed can match its own retrieval query. Fetch one slot for it
+			// plus one spare candidate so truncation remains detectable after the
+			// seed is excluded.
+			fetchLimit := opts.Limit + 2
 			downstream := downstreamServices(profile.ServiceSpans, profile.RootService, topDownstreamServices)
 			req := tempo.SearchRequest{
 				Query: buildBaselineQuery(profile.RootService, profile.RootOperation, downstream, opts.Filters),
@@ -182,11 +187,23 @@ change.`,
 			}
 
 			resp = excludeTrace(resp, seedID)
+			serverHasMore := resp != nil && len(resp.Traces) > opts.Limit
 			resp = limitTraces(resp, opts.Limit)
+			returned := 0
+			if resp != nil {
+				returned = len(resp.Traces)
+			}
+			meta := cmdio.AttachListMeta(
+				cmdio.PagedListMeta(returned, opts.Limit, serverHasMore, 0),
+				os.Args,
+			)
 
 			result := buildBaselineResult(seedID, profile.ServiceSpans, resp, req.Query)
-
-			return opts.IO.Encode(cmd.OutOrStdout(), result)
+			if err := opts.IO.Encode(cmd.OutOrStdout(), result); err != nil {
+				return err
+			}
+			cmdio.EmitListTruncationHint(cmd.ErrOrStderr(), meta)
+			return nil
 		},
 	}
 
@@ -276,6 +293,7 @@ func buildBaselineResult(seedID string, seedCounts map[string]int, resp *tempo.S
 		SeedSpanCount:    seedSpans,
 		SeedServiceCount: len(seedCounts),
 		Query:            query,
+		Candidates:       make([]tempo.BaselineCandidate, 0),
 	}
 	if resp == nil {
 		return result
--- a/internal/datasources/tempo/baseline.go
+++ b/internal/datasources/tempo/baseline.go
@@ -198,7 +198,8 @@ change.`,
 				os.Args,
 			)
 
-			result := buildBaselineResult(seedID, profile.ServiceSpans, resp, req.Query)
+			result := buildBaselineResult(seedID, profile, resp, req.Query)
+			result.ListMeta = meta
 			if err := opts.IO.Encode(cmd.OutOrStdout(), result); err != nil {
 				return err
 			}
@@ -283,15 +284,11 @@ func downstreamServices(serviceSpans map[string]int, rootService string, n int) 
 // buildBaselineResult enriches the candidates with the structural context
 // (per-candidate span/service counts) and attaches the seed profile. Candidate
 // order is preserved as returned by search.
-func buildBaselineResult(seedID string, seedCounts map[string]int, resp *tempo.SearchResponse, query string) *tempo.BaselineResult {
-	seedSpans := 0
-	for _, n := range seedCounts {
-		seedSpans += n
-	}
+func buildBaselineResult(seedID string, profile seedProfile, resp *tempo.SearchResponse, query string) *tempo.BaselineResult {
 	result := &tempo.BaselineResult{
 		SeedTraceID:      seedID,
-		SeedSpanCount:    seedSpans,
-		SeedServiceCount: len(seedCounts),
+		SeedSpanCount:    profile.SpanCount,
+		SeedServiceCount: len(profile.ServiceSpans),
 		Query:            query,
 		Candidates:       make([]tempo.BaselineCandidate, 0),
 	}
@@ -299,19 +296,23 @@ func buildBaselineResult(seedID string, seedCounts map[string]int, resp *tempo.S
 		return result
 	}
 	for _, t := range resp.Traces {
-		spans := 0
-		for _, s := range t.ServiceStats {
-			spans += s.SpanCount
-		}
-		result.Candidates = append(result.Candidates, tempo.BaselineCandidate{
+		candidate := tempo.BaselineCandidate{
 			TraceID:           t.TraceID,
 			RootServiceName:   t.RootServiceName,
 			RootTraceName:     t.RootTraceName,
 			StartTimeUnixNano: t.StartTimeUnixNano,
 			DurationMs:        t.DurationMs,
-			SpanCount:         spans,
-			ServiceCount:      len(t.ServiceStats),
-		})
+		}
+		if t.ServiceStats != nil {
+			spans := 0
+			for _, s := range t.ServiceStats {
+				spans += s.SpanCount
+			}
+			services := len(t.ServiceStats)
+			candidate.SpanCount = &spans
+			candidate.ServiceCount = &services
+		}
+		result.Candidates = append(result.Candidates, candidate)
 	}
 	return result
 }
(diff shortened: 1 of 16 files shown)


dafydd-t:
@claude review

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+// buildBaselineQuery constructs the TraceQL retrieval query per the design:
+// same root identity, the operation span constrained to a successful
+// (non-error) status, and a topology fingerprint pinning the seed's top
+// downstream services so candidates stay on the same execution path. Whole-trace
+// health is not filtered here — 'gcx traces diff' surfaces downstream errors.
+func buildBaselineQuery(service, operation string, downstream []string) string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "{ trace:rootService = %s && trace:rootName = %s } && { name = %s && status != error }",
```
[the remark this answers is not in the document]
javiermolinar: Fixed on a betterway

 {
     resource.service.name = "checkout"
     && name = "POST /checkout"
     && span:status != error
     && nestedSetParent = -1
   }

This will return only traces where the parent root span is faulty

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,426 @@
...
+				return err
+			}
+
+			// The seed can match its own retrieval query, so fetch one extra and
+			// drop it, keeping up to --limit candidates.
+			fetchLimit := opts.Limit
+			if fetchLimit > 0 {
+				fetchLimit++
+			}
+			downstream := downstreamServices(profile.ServiceSpans, profile.RootService, topDownstreamServices)
```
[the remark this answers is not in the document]
javiermolinar: I don't like this idea since it leaks the implementation of the baseline candidates. I think a better one would be

 1. Search using the top-three-service topology fingerprint.
 2. If that returns no candidates after excluding the seed, retry using only:
     - root service
     - root operation
     - healthy root status
 3. Emit a semantic hint:

 │ No exact-path candidates found; showing healthy same-operation candidates.

internal/datasources/tempo/baseline_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	first := result.Candidates[0]
+	assert.Equal(t, "exact", first.TraceID)
+	assert.Equal(t, 6, first.SpanCount)
+	assert.Equal(t, 2, first.ServiceCount)
+
+	second := result.Candidates[1]
+	assert.Equal(t, 40, second.SpanCount)
+	assert.Equal(t, 1, second.ServiceCount)
+}
```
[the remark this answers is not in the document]
javiermolinar: Fixed in dba7b2cd. Added a command-level `httptest` that executes `BaselineCmd` and captures the real Tempo proxy requests, asserting datasource routing, generated `q`, default `start`/`end`, over-fetch `limit`, seed exclusion, result capping, and the truncation hint.

internal/query/tempo/types.go
```diff
@@ -39,11 +39,41 @@ type SearchResponse struct {
...
+}
+
+// BaselineResult is the baseline-candidate list for a seed trace, in the order
+// returned by search, including the seed's structural profile for context.
+type BaselineResult struct {
+	SeedTraceID      string              `json:"seedTraceID"`
+	SeedSpanCount    int                 `json:"seedSpanCount"`
+	SeedServiceCount int                 `json:"seedServiceCount"`
+	Query            string              `json:"query"`
+	Candidates       []BaselineCandidate `json:"candidates"`
```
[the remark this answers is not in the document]
javiermolinar: Fixed in dba7b2cd. Empty baseline results now serialize `candidates` as `[]`. Retrieval reserves one slot for possible seed exclusion plus one truncation probe, and emits the standard list truncation hint when more candidates are available.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+func (opts *baselineOpts) setup(flags *pflag.FlagSet) {
+	dsquery.RegisterCodecs(&opts.IO, false)
+	opts.IO.DefaultFormat("table")
+	opts.IO.BindFlags(flags)
+
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.tempo is configured)")
+	flags.IntVar(&opts.Limit, "limit", 20, "Maximum number of candidates to return")
+	flags.StringVar(&opts.Window, "window", defaultBaselineWindow, "Search window padding applied before and after the seed trace's time range, so candidates from before or after the seed are eligible (e.g., 30m, 6h, 7d). Ignored when --from/--to are set")
```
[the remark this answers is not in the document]
javiermolinar: Fixed in dba7b2cd. Validation now rejects `--limit < 1` and blank/whitespace-only `--window` values, with focused tests for both cases.

javiermolinar:
@claude review

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+// 'gcx traces diff' surfaces downstream errors.
+func buildBaselineQuery(service, operation string, downstream, filters []string) string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "{ trace:rootService = %s && trace:rootName = %s } && { name = %s && span:status != error && nestedSetParent = -1 }",
+		strconv.Quote(service), strconv.Quote(operation), strconv.Quote(operation))
+	for _, filter := range filters {
+		fmt.Fprintf(&b, " && (%s)", strings.TrimSpace(filter))
+	}
+	for _, svc := range downstream {
+		fmt.Fprintf(&b, " && { resource.service.name = %s }", strconv.Quote(svc))
```
carles-grafana: I think this would return and empty set if we use traceql service name intersection. As this is querying spans that belong to multiple services.

internal/query/tempo/types.go
```diff
@@ -39,11 +41,43 @@ type SearchResponse struct {
...
+	ListMeta         *cmdio.ListMeta     `json:"list_meta,omitempty" yaml:"list_meta,omitempty"`
```
carles-grafana, line 83: I think this doesn't work well with gcx traces
 baseline seed --json traceID, because it won't look inside candidates

carles-grafana (changes requested):

internal/datasources/tempo/baseline.go
javiermolinar: I don't think so since we use the && between span sets:

This is what you mean
```
{ service.name = "a" && service.name = "b" }
```
Which returns empty

The generated query is:

```
{ service.name = "a" } && { service.name = "b"}
```

This means give me a trace that has spans from a and from b

javiermolinar: of course the optional filter could be wrong but that's out of scope

internal/query/tempo/types.go
javiermolinar: good catch

carles-grafana (dismissed):

awsome-o (commented):
Nice command, and the Long is unusually clear about what it deliberately doesn't do.

Two I'd worry about. The seed can come back as its own baseline candidate, which would leave the caller diffing a trace against itself. And a truncated seed goes unnoticed, so the query gets built off an incomplete picture.

The `llm_hint` points at the wrong path, worth a look since that's the line an agent actually reads. One open question on `nestedSetParent`, and one thought on `ErrorCount` that's more suggestion than finding.

Everything else is small and grouped into a single comment so it doesn't drown the rest.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+	return opts.ValidateTimeRange()
+}
+
+// BaselineCmd returns the `baseline` subcommand: given a seed trace, it finds
+// healthy same-operation candidate traces to diff against.
+func BaselineCmd(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &baselineOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "baseline TRACE_ID",
```
awsome-o: Correcting myself: I first wrote that every sibling under `traces` holds this slot with a verb. That's wrong — `labels`, `metrics` (tempo) and `series` (loki) are all nouns in the same slot, so a noun here isn't off-pattern on its face.

The narrower point still stands, though. docs/design/command-naming.md:87 makes those a **closed list**: "`labels`, `series`, `metrics`, and `metadata` are established shorthand operations in signal and datasource command families." `baseline` would be a fifth member of that list, and the guide is deliberate about not letting these grow by analogy.

Two ways to land it, and as maintainer I'll take the call rather than leave it hanging:

**`gcx traces baselines <trace-id>`** is my preference. Plural, so it reads as "enumerate these" exactly like `labels`/`metrics`/`series`, and it fixes the thing that actually bothers me about the current name: singular `baseline` with a positional reads like "make this trace the baseline", when the positional is the faulty seed and the baselines are the output. Cost is one line added to the shorthand list in the naming guide, in this PR.

**`gcx traces list-baselines <trace-id>`** is the zero-amendment alternative — it's the `<operation>-<subject>` compound from case 2 of the placement rule, since the required positional identity is the seed's. Guide-compliant with nothing to extend, but clunkier and no sibling in this family uses a compound.

Either is fine by me; `baseline` singular is the one I'd rather not ship. Experimental gives us room to change it later, but the rename is free today and isn't once people script against it.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+			seedID := args[0]
+
+			client, err := tempo.NewClient(cfg)
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			// Fetch the seed trace (no time range → full lookup) and read its
+			// root identity and span time range.
+			seed, err := client.GetTrace(ctx, datasourceUID, tempo.GetTraceRequest{TraceID: seedID})
```
awsome-o, line 155: `GetTraceResponse` has `Partial` and nothing reads it.

If Tempo truncated the seed, a lot of spans lose their parent, so root election picks an arbitrary middle span and the service pins come from an incomplete trace. The query then matches nothing and the user sees an empty list with no idea the seed was cut short. A stderr warning when `*seed.Partial` is true covers it.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+				return err
+			}
+			cmdio.EmitListTruncationHint(cmd.ErrOrStderr(), meta)
+			return nil
+		},
+	}
+
+	cmd.Annotations = map[string]string{
+		agent.AnnotationTokenCost: "medium",
+		agent.AnnotationLLMHint:   "gcx traces baseline -d UID <trace-id> -o json",
```
awsome-o: This hint is what the `gcx datasources tempo baseline` mount serves, but it says `gcx traces baseline`. `ApplyCommandMetadata` only rewrites the copy under `gcx traces`, because provider.go supplies its own.

So an agent that found the command under `datasources` gets told to run a different path. diff.go:114 and the other tempo constructors all use `gcx datasources tempo <verb>` here.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+
+// buildBaselineQuery constructs the TraceQL retrieval query per the design:
+// same root identity, the operation span constrained to a successful
+// (non-error) status, optional user-supplied candidate filters, and a topology
+// fingerprint pinning the seed's top downstream services so candidates stay on
+// the same execution path. Whole-trace health is not filtered here —
+// 'gcx traces diff' surfaces downstream errors.
+func buildBaselineQuery(service, operation string, downstream, filters []string) string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "{ trace:rootService = %s && trace:rootName = %s } && { name = %s && span:status != error && nestedSetParent = -1 }",
```
awsome-o: Real question, you're closer to Tempo than I am. Does `nestedSetParent` work on every block format we support, or does it need vParquet4?

Asking because the failure is silent. An unsupported intrinsic just returns nothing, which looks identical to "no healthy baselines exist". If it does need vParquet4 I'd set `agent.AnnotationAvailability` the way `DiffCmd` does.

Separately: the first spanset already pins `trace:rootName`. Is `nestedSetParent = -1` adding anything over `name = <op>`? If not, dropping it kills the version question too.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+			candidate.SpanCount = &spans
+			candidate.ServiceCount = &services
+		}
+		result.Candidates = append(result.Candidates, candidate)
+	}
+	return result
+}
+
+// limitTraces trims resp to at most n traces, preserving order. n <= 0 means no
+// cap. It returns a new response so the caller's slice is not aliased.
```
awsome-o: Nitpick, two small things.

`cmdio.TruncatePagedList(resp.Traces, opts.Limit)` is the documented helper for exactly this over-fetch shape and returns the same ListMeta, so it'd replace this function plus the manual `serverHasMore` / `PagedListMeta` pair up at line 190.

And the comment isn't quite right: the response struct is new, but `resp.Traces[:n]` shares the backing array, so the caller's slice *is* aliased. Someone appending to the result later would clobber the original.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+// excludeTrace returns a copy of resp without the seed trace.
+func excludeTrace(resp *tempo.SearchResponse, seedID string) *tempo.SearchResponse {
+	if resp == nil {
+		return resp
+	}
+	filtered := make([]tempo.SearchTrace, 0, len(resp.Traces))
+	for _, t := range resp.Traces {
+		if strings.EqualFold(t.TraceID, seedID) {
```
awsome-o: I think the seed can survive this and come back as its own top candidate, which would leave someone diffing a trace against itself for an empty result. The slot you reserved for it in `fetchLimit` goes too.

Reason I think so: search trims leading zeros off trace IDs, while `GetTrace` takes the padded 32 character form, and that padded form is what people copy out of the trace view. Haven't checked it against a live Tempo though, so tell me if the two agree in practice. Lowercasing and stripping leading zeros before the compare would settle it cheaply either way.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+			p.MaxEndNanos = ts
+		}
+		p.HasTimeRange = true
+	}
+}
+
+// forEachSpan calls fn for every span in the OTLP-shaped trace, passing the
+// resource's service.name ("" when absent). It isolates the resourceSpans ->
+// scopeSpans -> spans descent so callers stay flat.
+func forEachSpan(trace map[string]any, fn func(service string, span map[string]any)) {
```
awsome-o: `forEachSpan` and `resourceServiceName` are the same resourceSpans → scopeSpans → spans descent and the same `service.name` attribute lookup that `extractTraceSpans` / `extractServiceName` already do in internal/query/tempo/formatter.go:431.

The two copies already disagree — that one routes IDs through `decodeIDField`, this one doesn't — which is the parent-span thing above. So the divergence isn't hypothetical, and the next OTLP shape fix lands in one copy only.

Exporting the walker from `internal/query/tempo` and calling it from here would collapse both.

internal/query/tempo/types.go
```diff
@@ -39,13 +41,47 @@ type SearchResponse struct {
 
 // SearchTrace represents a single trace in the search results.
 type SearchTrace struct {
+	TraceID           string                  `json:"traceID"`
+	RootServiceName   string                  `json:"rootServiceName"`
+	RootTraceName     string                  `json:"rootTraceName"`
+	StartTimeUnixNano string                  `json:"startTimeUnixNano"`
+	DurationMs        int                     `json:"durationMs"`
+	ServiceStats      map[string]ServiceStats `json:"serviceStats,omitempty"`
```
awsome-o: This lands on `SearchTrace`, which is the response type for the already-shipped `gcx traces query` / `gcx datasources tempo query`. So those commands now emit a `serviceStats` object per trace in JSON/YAML and expose new `--json` discovery paths.

Additive and harmless, but it's a behavior change to a shipped surface and it's not in the PR body or the CHANGELOG line, which only mention baseline. Worth disclosing. Also worth a thought about token cost for agents running `-o json` over a wide search.

internal/query/tempo/types.go
```diff
@@ -39,13 +41,47 @@ type SearchResponse struct {
...
+}
+
+// ServiceStats holds per-service span and error counts returned in Tempo search
+// result metadata.
+type ServiceStats struct {
+	SpanCount  int `json:"spanCount,omitempty"`
+	ErrorCount int `json:"errorCount,omitempty"`
```
awsome-o, line 56: Now that an LLM is going to be picking which candidate to diff, this feels like the field it most wants. What it gets today is span and service counts, which tell it how big a trace is, not whether the trace went well.

`ErrorCount` is already decoded per service on every candidate and then dropped on the floor. Surfacing it, an ERRORS column and the same in the JSON, would give the model something to rank on instead of guessing.

internal/query/tempo/types.go
```diff
@@ -39,13 +41,47 @@ type SearchResponse struct {
...
 	RootTraceName     string `json:"rootTraceName"`
 	StartTimeUnixNano string `json:"startTimeUnixNano"`
 	DurationMs        int    `json:"durationMs"`
+	SpanCount         *int   `json:"spanCount,omitempty"`
+	ServiceCount      *int   `json:"serviceCount,omitempty"`
+}
+
+// BaselineResult is the baseline-candidate list for a seed trace, in the order
+// returned by search, including the seed's structural profile for context.
+type BaselineResult struct {
```
awsome-o: `BaselineResult` is a command output envelope rather than a Tempo wire type — nothing decodes it from a response, and it carries `list_meta` and `ListItemsKey`.

Putting it here makes `internal/query/tempo` the first `internal/query/*` package that imports `internal/output`; grep across all of them on main comes back empty. Every other list_meta envelope in the repo lives in the command or provider layer. I'd move it next to its command in `internal/datasources/tempo` so the wire-types package doesn't start dragging codec concerns into its consumers.

internal/datasources/tempo/baseline.go
awsome-o: Small naming thought before this path settles. `labels`, `metrics` and `series` are nouns in this same slot so a noun is fine here, but command-naming.md:87 reads those four as a closed list and `baseline` would be a fifth.

The bit that nags me more is the singular. `gcx traces baseline <id>` sounds like "treat this trace as the baseline", when the positional is actually the broken trace you want baselines for. `baselines` reads the right way round and matches its neighbours, at the cost of one line in the naming guide. `list-baselines` gets there without touching the guide at all.

Not a blocker, and if you read it the other way I'm happy to leave it. Just cheaper to settle now than after people script it.

internal/datasources/tempo/baseline.go
```diff
@@ -0,0 +1,469 @@
...
+// the seed are captured. Overridable with --window.
+const defaultBaselineWindow = "30m"
+
+// topDownstreamServices is how many of the seed's busiest downstream services
+// are pinned into the retrieval query as a topology fingerprint. The design
+// recommends ~3-4: enough to keep candidates on the same execution path without
+// over-constraining and losing recall.
+const topDownstreamServices = 3
+
+type baselineOpts struct {
```
awsome-o, line 33: Rest of what I noticed, all small and all take or leave. Grouping them so they don't drown the ones above.

`--filter` is the one I'd read twice: `&&` between spansets applies across the whole trace, so a tenant filter matches any trace that touched that tenant somewhere, not traces whose root operation ran for it. Worth a line in the flag help.

`--limit 0` returns everything on `gcx datasources tempo query` and errors here; output.md §15.1 has a shared binder for this.

`parseSeedTrace` reads `parentSpanId` as a plain string, but Tempo emits these as base64 and `decodeIDField` in formatter.go exists for exactly that. Same function is also a second copy of `extractTraceSpans` and `extractServiceName`, so reusing the existing walk covers both at once.

`BaselineResult` carries `list_meta` and nothing decodes it off the wire, so it reads as command output living in the wire types package. It also makes `internal/query/tempo` the only `internal/query/*` that imports `internal/output`.

`ServiceStats` on `SearchTrace` means `gcx traces query` now emits `serviceStats` as well, which isn't in the PR body or the CHANGELOG.

`limitTraces` duplicates `cmdio.TruncatePagedList`, and its comment says the caller's slice isn't aliased when `resp.Traces[:n]` shares the backing array.

When nothing comes back, `result.Query` would explain why but the table never prints it.

And a naming thought: `baselines` plural reads the right way round, since `baseline <id>` sounds like "treat this as the baseline" when the positional is the broken trace. Cheaper now than once the skill hardcodes the path.

internal/datasources/tempo/baseline.go
javiermolinar: We support only vparquet4 and 5 and nestedParent has been an intrinsic property for some time already. `nestedSetParent = -1 ` is the "trick" to select only the root parent span. We are telling Tempo that we only want traces where the root parent span has an status different than error

internal/datasources/tempo/baseline.go
javiermolinar: This is not that absolute

> If Tempo truncated the seed, a lot of spans lose their parent

A partial trace means that Tempo refuses to return more parts of a trace but it will return it in order. Thus:

- Truncation preserves parentSpanID
- If the real root is absent, we already have an existing validation error
- Available child services can still provide useful fingerprint pins. We only need a few of them to craft the query.


What I will do is create a warning, so the user or agent can know that its working with partial results and act on consequence.

internal/datasources/tempo/baseline.go
javiermolinar: This is something I discussed here https://github.com/grafana/gcx/issues/1218

I kinda dislike having 2 different mount points. I will change it for consistency with the others

internal/datasources/tempo/baseline.go
javiermolinar: Trimming the zeroes is the wrong solution to this problem, nice catch. In Tempo this is configurable so we cannot rely on that to remove the seed from the candidates. Instead we can add a new clause to our traceQL expression to skip the seed trace entirely

internal/query/tempo/types.go
javiermolinar: Agree, easy change

internal/datasources/tempo/baseline.go
javiermolinar: Most of these are false positives

### Fix 

 1. Use cmdio.TruncatePagedList. Fair enough
 2. Document serviceStats output. Fair

 ### Reject

 - Explain --filter spanset semantics
   “Raw TraceQL spanset expression” is already accurate.

 - Support --limit 0
   Tempo treats zero as its configured default, not unlimited.

 - Decode base64 parentSpanId
   Its encoding is irrelevant.

 - Print the generated query for empty tables
   Adds noise.

 - Rename the command to baselines
   Nope, it doesn't change anything

 ### Defer

 - Move BaselineResult out of internal/query/tempo
  This is a big refactor

awsome-o (approved):

PR #1173, opened 2026-08-13
feat(agent): detect Codex CLI via CODEX_THREAD_ID

dafydd-t:
## Summary

- adds `CODEX_THREAD_ID` → `codex` to the harness detection list in `internal/agent`
- matches on set-and-non-empty rather than truthy: Codex sets the var to a thread ID (an opaque UUID-like string), unlike the boolean marker vars other harnesses use — harness entries now carry a `nonEmpty` predicate and both `detectFromEnv()` and `Name()` share it
- table-driven tests for the new var, test helpers that clear harness env vars updated, and reference docs regenerated (the `--agent` help text now lists `CODEX_THREAD_ID`)

Codex CLI sets no boolean marker var; `CODEX_THREAD_ID` is set for every Codex session, so it's the reliable signal. Currently Codex sessions are invisible in usage telemetry's Top Agents / Agent-vs-Direct breakdowns.


internal/agent/agent.go
```diff
@@ -11,16 +11,36 @@ import (
...
 )
 
+// harnessEnvVar describes an environment variable that signals a specific
+// agent harness.
+type harnessEnvVar struct {
+	envVar, name string
+	// nonEmpty matches on any non-empty value instead of a truthy one, for
+	// harnesses that set an identifier (e.g. a thread ID) rather than a flag.
+	nonEmpty bool
+}
```
dafydd-t: i dont think we needed to add this type really.

dafydd-t: addressed - back to the anonymous struct with a nonEmpty field, the check now lives in a package-level harnessDetected helper shared by detectFromEnv and Name

internal/agent/agent_test.go
```diff
@@ -43,6 +43,16 @@ func TestIsAgentMode(t *testing.T) {
 			envVars:  map[string]string{"OPENCODE": "1"},
 			wantMode: true,
 		},
+		{
+			name:     "CODEX_THREAD_ID set to a thread ID",
+			envVars:  map[string]string{"CODEX_THREAD_ID": "0198a213-9f2c-7e33-b7c5-3a1d4e5f6a7b"},
```
dafydd-t:
```suggestion
			envVars:  map[string]string{"CODEX_THREAD_ID": "some-thread-id"},
```

make this more obviously generated/irrelevant

dafydd-t: addressed - suggestion accepted, swapped both occurrences (IsAgentMode and Name cases) to `some-thread-id`

PR #1187, opened 2026-08-14
fix(irm): accept envelope manifests and send a valid timezone from list-final-shifts

### LINKED ISSUE #1185: irm oncall: schedules/shifts create drop spec fields typed as any; list-final-shifts always returns Invalid timezone
Found while I provisioned a Grafana IRM environment end to end with gcx. Three
problems blocked the work. Every one of them is reproducible on a current Grafana
Cloud stack. All three concern `gcx irm oncall`.

Version: `gcx version` reports `(devel)`, go1.26.4, darwin/arm64.

## 1. `schedules create` and `shifts create` drop integer valued spec fields

`gcx irm oncall schedules create -f` never sends `spec.type`. The server then
rejects the request, and the message contradicts the manifest, which does set the
field.

```console
$ cat schedule.yaml
apiVersion: oncall.ext.grafana.app/v1alpha1
kind: Schedule
metadata:
  name: probe
spec:
  name: my schedule
  type: 2
  time_zone: Europe/Amsterdam

$ gcx irm oncall schedules create -f schedule.yaml
{"summary":"Request failed with status 400","details":"{\"type\":\"This field is required\"}"}
```

The same manifest fails with `type: "web"`, `type: "calendar"`, and `type: "2"`,
and it fails as JSON as well as YAML. The direct call succeeds with the same
value:

```console
$ gcx api /api/plugins/grafana-irm-app/resources/schedules/ \
    -d '{"name":"my schedule","type":2,"time_zone":"Europe/Amsterdam"}'
{"id":"ST846IQ3N3AXE","name":"my schedule","type":2, ...}
```

`gcx resources list-types` declares these fields as bare `true` in the generated
schema, which means any type:

```json
"Schedule": { "spec": { "properties": {
  "type": true, "team": true, "on_call_now": true, "enable_web_overrides": true
}}}
```

Every field that the schema declares as `true` looks affected. `Shift` declares
`type`, `frequency`, `shift_end`, and `rolling_users` the same way, so
`shifts create` cannot create a rotation either. Those four fields are exactly
the ones a rotation needs.

Expected: the create commands send every field that the manifest sets.

## 2. `schedules list-final-shifts` always fails with "Invalid timezone"

The command has no flag to supply a timezone, so there is no way to make it
succeed.

```console
$ gcx irm oncall schedules list-final-shifts SE4LY5V8K3BRU
{"summary":"Request failed with status 400","details":"{\"detail\":\"Invalid timezone\"}"}

$ gcx irm oncall schedules list-final-shifts --help
Flags:
      --end string      End date (YYYY-MM-DD) (default "2026-08-21")
      --start string    Start date (YYYY-MM-DD) (default "2026-08-14")
```

The schedule itself has a valid timezone (`Europe/Amsterdam`), and the web user
interface renders its final schedule correctly. Expected: the command either
sends the timezone of the schedule, or it accepts a `--timezone` flag.

## 3. `--insecure-log-http-payload` does not log request bodies

The flag help promises the request body:

> Log full HTTP request/response bodies including raw credentials, authorization
> tokens, cookies, and OAuth refresh tokens.

With `-vvv` the debug output holds the method, the URL, and the response status,
but no body in either direction:

```console
$ gcx irm oncall schedules create -f schedule.yaml --insecure-log-http-payload -vvv 2>&1 | grep -i body
(no output)
```

This matters for problem 1. The flag is the tool that would show a caller that
gcx dropped a field, and it stays silent.

## Suggested priority

Problem 1 is the important one. It makes `schedules create` and `shifts create`
unusable, and it pushes a caller to `gcx api`, which loses the manifest workflow
that the rest of gcx offers. Problems 2 and 3 are smaller, and problem 3 makes
problem 1 hard to diagnose.

wardbekker:
## Summary

Corrects problems 1 and 2 of #1185. Problem 3 stays open — see the note at the end.

- Manifest input accepts the K8s envelope, so `create -f` and `update -f` no longer discard every field of the file.
- `irm oncall schedules list-final-shifts` takes a `--timezone` flag and sends a name that the API accepts.

## Problem 1 — create and update lost every field of the manifest

The report describes this as "spec fields typed as `any` are dropped". The type of the field is not involved. `providers.ReadFileOrStdin` decoded the whole document into the bare domain struct. The keys `apiVersion`, `kind`, `metadata`, and `spec` match no field of that struct, so the decode produced an empty object and the command posted it. The 400 answer named `type` because `type` is the first required field that the server checks.

The envelope is the shape that gcx itself prints. `gcx resources list-examples` returns it, `pull` writes it, and the provider `list` commands emit it. A caller who follows the documented workflow always hits this.

`ReadFileOrStdin` now decodes from `spec` when the document carries a non-null object-valued `spec` together with `apiVersion` or `kind`. Every other document decodes whole, so a bare object with its own `spec` field behaves as before. All seven call sites take bare domain structs with no `kind` or `apiVersion` field, so the test cannot misfire on them.

This repairs the round trip for `gcx irm oncall`, where `resources list-examples` prints the envelope. `AlertProvider.TypedRegistrations()` returns `nil`, so `gcx alert` never had that round trip; there the change adds tolerance for a hand-written envelope.

An envelope that sets `apiVersion` or `kind` but carries no object-valued `spec` now reports an error. That document produced the same empty object as the defect above.

## Problem 2 — list-final-shifts always answered "Invalid timezone"

The command sent `time.Now().Location().String()` as `user_tz`. On a host that reads its zone from the clock, that call returns the literal string `Local`, which the API rejects. The command had no flag, so no caller could make it succeed.

The command now takes `--timezone`. Without the flag it reads the timezone of the schedule, then uses UTC. `Validate` rejects only `Local`; the binary embeds no tz database, so a `time.LoadLocation` check would reject every real IANA name on Windows and inside a minimal container image. The API adjudicates every other name.

The API renders each shift boundary in the zone that `user_tz` names, so a fallback to UTC moves the timestamps that the caller reads. The command writes a warning to stderr on each fallback.

This is not a change for every host: Go names `time.Local` from `TZ`, so a host that sets `TZ` to an IANA name already sent a value that the API accepts. For those hosts the default now changes from the zone of the host to the zone of the schedule.

## Changes

| File | Change |
| --- | --- |
| `internal/providers/fileinput.go` | Unwrap a K8s envelope before the decode |
| `internal/providers/fileinput_test.go` | New — table test for both manifest shapes and the pass-through cases |
| `internal/providers/irm/oncall_commands_extra.go` | `--timezone` flag, `Validate`, `resolveTimezone` |
| `internal/providers/irm/oncall_crud_commands_test.go` | Schedule and shift create accept both manifest shapes |
| `internal/providers/irm/oncall_commands_extra_test.go` | Timezone resolution and rejection table |
| `internal/providers/irm/oncall_output_contract_test.go` | Fake serves `GetSchedule` |
| `docs/reference/cli/…list-final-shifts.md` | Regenerated |
| `CHANGELOG.md` | Entries under Unreleased |

## Test plan

- [x] `GCX_AGENT_MODE=false mise run all` passes (lint, tests, build, docs).
- [x] New tests cover both fixes. Each fails on the parent commit.
- [x] Reference docs regenerated; no drift.
- [ ] Verify against a live Grafana Cloud stack.

## Not in this pull request

Problem 3 of the issue — `--insecure-log-http-payload` logs no request body — stays open. Do not close #1185 when this merges.

Refs #1185


wardbekker:
## Shepherd status

The branch head is `6f3a04fa`. The pull request is mergeable.

### Current state

- Rebased all six change commits onto `main` at `8d10a674`.
- Resolved the changelog conflict. The new entries are in the current `Unreleased` section. The historical `v1.2.0` section is unchanged.
- Ran `GCX_AGENT_MODE=false mise run all`. It passed.
- All GitHub checks pass on the new commit.
- Ran a live read/write smoke test on `<my-context>`. It passed.
- Corrected the OnCall schedule example to use numeric type `2`, as required by the live API.
- Changed the envelope regression test to run the generated schedule example through the create command.

### Live smoke test

- Confirmed that the context is online.
- Called `list-final-shifts` without `--timezone`. The command used the schedule timezone and returned data.
- Called `list-final-shifts` with `--timezone UTC`. The command returned data.
- Called `list-final-shifts` with `--timezone Local`. Validation rejected the value before the request.
- Created `<smoke-schedule>` from a Kubernetes envelope manifest.
- Read the schedule and confirmed its name and `Europe/Amsterdam` timezone.
- Updated the schedule from a second envelope manifest.
- Read the schedule and confirmed its new name and `UTC` timezone.
- Called `list-final-shifts` without `--timezone` on the updated schedule. The schedule timezone lookup succeeded.
- Deleted the temporary schedule and confirmed that it no longer exists.

The first create attempt used the former `type: web` value from `gcx resources list-examples schedules`. The live API rejected this value as `Invalid type`. A retry with numeric value `2` passed. Commit `6f3a04fa` corrects the example and adds regression coverage. The API error also confirms that the command sent the envelope `spec` instead of an empty object.

### Scope decisions

- Keep `resolveTimezone` and its second `GetSchedule` request.
- Keep the `--timezone` flag name.
- Keep validation that rejects the literal timezone name `Local`.
- Cover the equivalent k6 manifest issue in a separate change.

### Remaining gate

The pull request needs an approving review.

---
*Updated by shepherd-pr*

dafydd-t:
@claude review

dafydd-t (approved):

Issue #1186, opened 2026-08-14
irm: eight IRM provisioning operations have no gcx command and require gcx api (templates, maintenance, step order, incident severity, teams, plugin sync, SCIM, service accounts)


wardbekker:
I provisioned a complete Grafana IRM environment with gcx: teams, users,
schedules, rotations, overrides, escalation chains, routes, integrations,
templates, outgoing webhooks, alert traffic, incidents, and a reporting
dashboard. This issue lists the operations where **no gcx command exists**, so I
had to fall back to `gcx api`.

This is separate from #1185, which is about existing commands that drop their
payload. Every item below is a missing capability, not a defect.

Ordered by how much the gap hurt.

## 1. Alert templates per integration — no command

Templates decide what a responder actually reads and hears. They set the alert
title, the message, the grouping identifier, the resolve condition, and a separate
rendering per channel for web, phone call, Short Message Service, email, Slack,
and Microsoft Teams.

`gcx irm oncall integrations` offers `create`, `delete`, `get`, `list`, and
`update` only. The `Integration` spec has `additionalProperties: false` and no
template field, so a template cannot travel with the integration either.

I had to use:

```bash
gcx api "/api/plugins/grafana-irm-app/resources/alert_receive_channel_templates/<id>/" \
  -X PUT -d @templates.json
```

That endpoint carries about 20 template fields. Nothing in gcx reaches them. For
my case this was a hard requirement: the customer needs Dutch text-to-speech on
an incoming call, which is `phone_call_title_template`.

Suggested shape: `gcx irm oncall integrations templates get|set <id>`.

## 2. Maintenance mode on an integration — no command

Maintenance mode suppresses escalation during planned work. The `Integration`
spec exposes `maintenance_mode` and `maintenance_till` for reading, but starting
maintenance is a separate action:

```bash
gcx api "/api/plugins/grafana-irm-app/resources/alert_receive_channels/<id>/start_maintenance/" \
  -d '{"mode":1,"duration":10800}'
```

There is no `gcx irm oncall integrations start-maintenance` or
`stop-maintenance`. Sending `maintenance_mode` through `update` does not start
it.

## 3. Escalation step order and route order cannot be expressed

An escalation chain is an **ordered** sequence, and routes match top to bottom
with first-match semantics. Order is the meaning of these objects.

Neither `EscalationPolicy` nor `Route` has an `order` or `position` field in its
spec. The underlying API accepts `position` on create, and it has a
[...]

PR #1191, opened 2026-08-14
fix(irm): honour incident severity, and add an incidents update command

wardbekker:
## Summary

Item 4 of #1186, and the highest-severity item in it: **`gcx irm incidents create` silently dropped the severity.** Every incident was created at the default severity, whichever severity the manifest set. `gcx resources list-types` advertised both `spec.severity` and `spec.severityID`, which made the silence worse. No command could change the severity afterwards.

```console
$ gcx irm incidents create -f inc.json    # spec.severity: Critical
✔ Created incident probe (id=1)
$ gcx irm incidents get 1
severity: Pending
```

## Changes

```bash
# create now applies the severity
gcx irm incidents create -f incident.yaml    # spec.severity: Critical

# and it can be changed afterwards
gcx irm incidents update <id> --severity Critical
gcx irm incidents update <id> --title "Checkout latency above the objective"
```

| File | Change |
| --- | --- |
| `internal/providers/irm/incidents_client.go` | Adds the title and severity update operations; validates severity labels and IDs before writes; adds consistent versioned/unversioned fallback |
| `internal/providers/irm/incidents_types.go` | Adds the update request bodies and internal mutation metadata |
| `internal/providers/irm/incidents_update_command.go` | Adds `incidents update` through the shared `TypedCRUD` path and emits a structured mutation result |
| `internal/providers/irm/incidents_resource_adapter.go` | Applies status, title, and severity; uses one shared read-field strip list |
| `internal/providers/irm/incidents_commands.go` | Wires `update` onto the incidents noun |
| `internal/agent/command_annotations.go`, `cmd/gcx/root/testdata/output_classes.json` | Adds agent metadata and the output class |
| `docs/reference/cli/`, `docs/architecture/architecture.md` | Regenerates and updates the documentation |

## Compatibility note

Incident reads no longer emit `spec.severityID`. The field is write-only: a hand-written manifest can still use it, but `get`, `list`, and `pull` return `spec.severity` only. This prevents an organization-specific identifier from overriding an edited severity label or breaking a cross-stack push. Existing callers that read `.spec.severityID` must use `.spec.severity` instead.

The new `incidents update` command emits a `gcx.mutation` result for JSON, YAML, and agent output. Its `changed` field distinguishes a write from an idempotent no-op. Text output remains one status line.

## Design notes

**Why `Create` makes a second write.** `IncidentsService.CreateIncident` ignores the severity in the request body. `UpdateSeverity` is the only route. `Create` validates the requested label or identifier against the organization severity list before it creates the incident, and then applies the severity when the created incident does not already carry it. A manifest with no severity still costs one call. A manifest with a severity also reads the severity list before it writes.

**Why the fix lives in the shared CRUD path.** The provider command and the `resources push` pipeline both reach the same `TypedCRUD` update function. The create and update fixes therefore apply to both access paths.

**How severity input resolves.** `UpdateSeverity` takes the display label, not the identifier. Both `spec.severity` and `spec.severityID` resolve through `SeveritiesService.GetOrgSeverities` before any write. An unknown value is an error that names `gcx irm incidents severities list`.

**What `Update` changes.** `resources push` previously applied the status and dropped the title and severity. `Update` now applies all three and skips fields that already match. The IRM API has no update operation for the other incident fields.

**How the fallback works.** `GetIncident` and the field update operations try the versioned path first and the unversioned path after a 404. If both update paths return 404, `GetIncident` distinguishes a missing incident from an unavailable update operation.

## Compliance

| Level | Result |
| --- | --- |
| CONSTITUTION | Pass. The provider update command uses `TypedCRUD`; both access paths share the same update logic and serialized resource shape. |
| VISION | Pass. IRM Incidents is an existing product surface. |
| DESIGN | Pass. Positional subject, flag modifiers, `finite` output class, text status output, and a structured mutation result for machine formats. |
| ARCHITECTURE | Pass. The new file sits inside the existing IRM provider; the architecture file table records it. |

## Test plan

- [x] `GCX_AGENT_MODE=false mise run all` passes
- [x] Tests cover label and identifier validation before writes
- [x] Tests cover versioned/unversioned fallback and missing-operation classification
- [x] Tests cover all changed and unchanged fields on the shared update path
- [x] Tests cover request ID binding and text, JSON, YAML, and agent output
- [x] Reference documentation regenerated

Refs #1186 (item 4)

### DIFF
--- a/cmd/gcx/root/testdata/output_classes.json
+++ b/cmd/gcx/root/testdata/output_classes.json
@@ -293,6 +293,7 @@
  "gcx irm incidents list-contexts": "finite",
  "gcx irm incidents open": "interactive",
  "gcx irm incidents severities list": "finite",
+ "gcx irm incidents update": "finite",
  "gcx irm oncall alert-groups acknowledge": "finite",
  "gcx irm oncall alert-groups delete": "finite",
  "gcx irm oncall alert-groups get": "finite",
--- a/cmd/gcx/root/testdata/output_classes.json
+++ b/cmd/gcx/root/testdata/output_classes.json
@@ -270,6 +270,7 @@
  "gcx irm incidents list-contexts": "finite",
  "gcx irm incidents open": "interactive",
  "gcx irm incidents severities list": "finite",
+ "gcx irm incidents update": "finite",
  "gcx irm oncall alert-groups acknowledge": "finite",
  "gcx irm oncall alert-groups delete": "finite",
  "gcx irm oncall alert-groups get": "finite",
--- a/docs/architecture/architecture.md
+++ b/docs/architecture/architecture.md
@@ -802,7 +802,8 @@ Provider command tree backed by fleet-management `Set/Get` + observed-state RPCs
 | `internal/providers/irm/oncall_commands.go` | OnCall CLI commands (schedules, integrations, escalation chains) |
 | `internal/providers/irm/oncall_adapter.go` | Resource adapter for OnCall resources |
 | `internal/providers/irm/incidents_client.go` | Incidents REST client |
-| `internal/providers/irm/incidents_commands.go` | IRM Incidents CLI commands (list, get, create, close, open, list-activity, list-contexts, activity add, severities) |
+| `internal/providers/irm/incidents_commands.go` | IRM Incidents CLI commands (list, get, create, close, open, list-activity, list-contexts, activity add, severities, update) |
+| `internal/providers/irm/incidents_update_command.go` | The `incidents update` command (severity, title) |
 
 ### Faro Provider
 
--- a/docs/reference/cli/gcx_dbo11y.md
+++ b/docs/reference/cli/gcx_dbo11y.md
@@ -12,7 +12,7 @@ Inspect Grafana Database Observability instances and query performance
 ### Options inherited from parent commands
 
 ```
-      --agent                       Enable agent mode (JSON output, no color). Auto-detected from CLAUDECODE, CLAUDE_CODE, CURSOR_AGENT, GITHUB_COPILOT, AMAZON_Q, or GCX_AGENT_MODE env vars.
+      --agent                       Enable agent mode (JSON output, no color). Auto-detected from CLAUDECODE, CLAUDE_CODE, CURSOR_AGENT, GITHUB_COPILOT, AMAZON_Q, OPENCODE, PI_CODING_AGENT, or GCX_AGENT_MODE env vars.
       --context string              Name of the context to use (overrides current-context in config)
       --insecure-log-http-payload   Log full HTTP request/response bodies including raw credentials, authorization tokens, cookies, and OAuth refresh tokens. Do not ship these logs.
       --no-color                    Disable color output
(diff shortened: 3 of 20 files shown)


wardbekker:
## Shepherd status

This branch went through three review rounds. All items are applied.

### Maintainer decisions

- A pulled manifest carries the severity label alone. The first round made `severityID` win over `severity`. The second review showed that this reintroduced the same silent drop in the other direction, and that it broke cross-stack push, because a severity identifier is organisation-scoped. A hand-authored `severityID` still resolves.
- `Update` sends a status only when the manifest sets one, and only when it differs from the server value.
- An empty `--title` or an empty `--severity` is an error. The schema marks `title` as required, and an empty label matches no entry in the severity list.
- `update` prints one success line, like `close`.

### Corrections

1. `Create` resolves the severity label before it posts `CreateIncident`. A wrong `spec.severityID` now creates no incident.
2. A failed severity update after a create returns the created incident next to an error. The error names the identifier, and it gives the command that repairs the state. The command no longer wraps that error with `failed to create incident:`, which contradicted it.
3. One list, `incidentStripFields`, feeds both `TypedCRUD.StripFields` and `ToResource`. So `irm incidents get|create|update` and `resources get|list|pull` emit the same spec keys, as CONSTITUTION.md requires for a dual-path resource. A test compares the two key sets. Both paths remove `severityID`, so a pulled manifest cannot carry two fields that disagree.
4. The registered schema describes `spec.severityID` as write-only: a push accepts it, and a read never returns it.
5. `SeverityID` is gone from `createIncidentRequest`. The comment said that `CreateIncident` ignores the field, and the request body still sent it.
6. `Update` resolves the severity label before it writes anything. An unknown `severityID` now costs no write, and the error text no longer says that the severity update failed when gcx attempted none.
7. The doc comment of `Update` names the three fields that it applies, and the fields that it drops: `labels`, `incidentType`, `isDrill`, `description`, and `fieldGroupUUID`.
8. A half-applied update returns nil next to an error that names the applied fields. The incident that an earlier round returned reached no caller.
9. `Update` reads the incident with `Get`, then skips each field that the caller left empty or that already matches the server.
10. `UpdateStatus` goes through `updateIncidentField`, which removes about 25 lines.
11. `NewUpdateCommand` calls `client.Update`. CONSTITUTION.md § TypedCRUD requires an adapter-backed provider command to reach data through the shared CRUD methods.
12. `updateIncidentField` maps a 404 on both base paths to the not-found error, and names the incident identifier.
13. `Validate` uses `flags.Changed`, so it separates an explicit empty value from an omitted flag.
14. The `--severity` usage string uses single quotes, so cobra prints `--severity string` again.
15. The inline example of `create` shows `severity`.

### A cost worth knowing

On the push path gcx reads the incident twice: `pusher.go` reads it to choose between create and update, and `Update` reads it again to skip an unchanged field.

---
*Updated by shepherd-pr*

wardbekker:
Commit `847943f7` resolves all review comments.

The client now validates a severity label before it sends a write request. It now uses the versioned and unversioned paths for read and update operations. Machine output now includes the `changed` value. Tests now verify the incident ID. The PR description now identifies `spec.severityID` as a write-only field and describes the compatibility change.

`GCX_AGENT_MODE=false mise run all` passes.

dafydd-t (approved):

PR #1194, opened 2026-08-14
fix(output): warn about invalid --json field paths

wardbekker:
## Summary

The `--json` flag accepted a field path that did not exist at the selected level. The command then returned `null` values without a diagnostic. A script could read this as a valid empty result.

gcx now writes an advisory warning when a declared item type denies the path. The warning gives a valid dotted path when gcx can find one. The command still returns the selected JSON and exits successfully. This keeps output transformation problems from reporting a completed mutation as failed.

For example, OnCall users store their identity at `metadata.name` in the emitted resource. These invalid selections now suggest that path:

```console
gcx irm oncall users list --json pk
gcx irm oncall users list --json spec.pk
```

This path returns the ID:

```console
gcx irm oncall users list --json metadata.name
```

## Behavior

- A declared path stays valid when its value is `null` or when `omitempty` removes it from all rows.
- A path stays valid when at least one emitted object contains it.
- An unstructured object or a dynamic map has no declared field set. gcx does not warn for these values.
- A type that implements `json.Marshaler` controls its JSON shape. gcx does not use reflection to warn for this shape.
- A path that enters an array warns that the caller must use `--jq`.
- An empty typed result gives the same validation result as a populated typed result.
- Field discovery for an empty typed result includes nested paths.
- Invalid field diagnostics go to stderr. The JSON result stays on stdout.

## Changes

- `FieldSelectCodec` distinguishes a missing path, a declared path, and a path that enters an array.
- The OnCall list commands validate the emitted resource shape. Provider identity fields map to `metadata.name` instead of the removed `spec` field.
- Flags-only OnCall list commands and `instrumentation clusters list` reject positional arguments.
- The Tempo and alerting custom JSON types do not cause false path warnings.
- `resources list-types` applies field selection to each descriptor.
- The agent skills use `--jq` for Prometheus samples below `data.result[]`.
- The help text uses `--json <path>,<path>`.
- The design and architecture documents describe the current behavior.
- The generated CLI reference includes the current flag text.

## Compatibility

This change affects each command that accepts `--json`.

A path that a declared item type denies now writes a warning. It returned the same `null` value without a warning before this change. Dynamic JSON shapes keep the previous behavior because gcx cannot safely deny their paths.

## Test plan

- Command-path tests for `gcx irm oncall users list`
- Command-path tests for flags-only list argument validation
- Codec tests for `json.Marshaler`, embedded fields, `omitempty` fields, empty typed results, and array paths

Refs #1186


wardbekker:
## Shepherd status

This branch went through three review rounds. All items are applied.

### The rule

gcx rejects a field path only when a declared item type denies it. Where gcx has no declared type for the value, it rejects nothing, and the path keeps its `null`. The first round tested the emitted data instead, which rejected correct invocations whenever an `omitempty` field held its zero value in every row.

A path that continues past an array gets its own error:

    --json cannot reach a value inside an array: data.result.metric. Use --jq to read a value inside an array.

### Corrections

1. `selectFields` takes the item type. `Encode` reads that type from the value: a slice element, the items key of a `ListEnvelope`, the `items` field of an envelope, or the single object. `resolvePath` answers absent, present, or inside-array, and `MakeFieldValidator` uses the same walk.
2. `structTypeOf` answers nil for an `unstructured.Unstructured` element type, so a bare slice of unstructured items rejects no path. The single-object route already did this. Without the change, `gcx irm oncall users list --json spec.username` exited 2 when the organisation held no user.
3. `jsonFieldByName` and `typePaths` recurse into an anonymous field whose json tag gives no name. `encoding/json` promotes such a field, so the earlier code recorded a key that the output never carries, and missed every key that it does. Without the change, `gcx frontend apps list --json kind` failed on an empty page. A depth bound of 8 stops a cycle.
4. `resolvePath` stops at an array, so `gcx metrics query -d <uid> up --json data.result.metric` reports the array error instead of a silent `null`. `gcx metrics cardinality label-names --json cardinality.label_name` does the same.
5. `gcx resources list-types` encodes a typed `descriptorEntry`, so both discovery and selection run through `Options.Encode`. That command also never called `opts.Validate()`, so `--json` was dead on it before this branch.
6. `claude-plugin/skills/debug-with-grafana/SKILL.md` and its `references/query-patterns.md` use `--jq '.data.result[] | {metric, value}'`. `QueryResponse` cannot implement `ListEnvelope`, because that interface exposes one top-level key and the samples live at `data.result[]`.
7. The did-you-mean clause groups its candidates per absent name, and `MakeFieldValidator` supplies candidates from the declared type.
8. An empty object rejects nothing.
9. `internal/style/help.go` and every other place print `--json <path>,<path>`.
10. `docs/architecture/cli-layer.md` lists the converter chain in run order, with the three converters that it had lost.

### Three behaviour changes that the description does not list

- `ArrayPathSelectionError` is a second rejection class. It takes `gcx metrics query --json data.result.metric` and `gcx metrics cardinality label-names --json cardinality.label_name` from exit 0 to exit 2.
- `gcx resources list-types` now honours `--json` and `--jq` for the first time, and its `--json` payload changes from whole-envelope selection to per-item selection.
- The `debug-with-grafana` agent workflow no longer carries the broken invocation.

### A correction to the description

The description shows `Did you mean spec.username, spec.email?`. The code emits one of these three forms:

- one absent field: `Did you mean spec.username?`
- two or more absent fields: `Did you mean spec.username for username; spec.email for email?`
- one name with two paths: `Did you mean metadata.name, spec.name?`

---
*Updated by shepherd-pr*

wardbekker:
## Review update

I addressed the five current review comments in commit `87209346`.

- The OnCall users list command uses its declared resource shape for field validation. `--json username,email` now returns the suggested `spec` paths.
- Field validation does not reject paths that a custom JSON marshaler controls. This change covers the Tempo tag-values response and alert contact-point settings.
- Field discovery for an empty typed result includes nested paths.
- I removed the unused object-sampling code from candidate generation.
- I updated the pull request description. It now lists the behavior changes and the custom JSON rule.

I also rebased the branch on the latest `origin/main`. I regenerated the CLI reference after the rebase.

Verification completed successfully:

- `GCX_AGENT_MODE=false mise run reference-drift`
- Signed commit verification

Please review the new head commit.

internal/providers/metrics/cardinality_json_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// execCardinalityLabelNames runs `label-names -d prom-uid` with the given
+// extra arguments against a server that answers one label name, and returns
+// stdout and the run error.
+func execCardinalityLabelNames(t *testing.T, args ...string) (string, error) {
+	t.Helper()
+
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
dafydd-t: I know there are other examples in the codebase, but do we need `NewTLSServer` here? why not the more obvious `NewTestServer`?

internal/providers/faro/list_json_test.go
dafydd-t, line 1: what does this test cover that isn't covered by the output tests?

docs/design/output.md
```diff
@@ -136,7 +136,7 @@ gcx resources get dashboards/my-dash --json ?
 
 | Value | Behavior |
 |-------|----------|
-| `--json field1,field2` | Emit JSON with only those fields; missing fields produce `null` |
+| `--json <path>,<path>` | Emit JSON with only those field paths. A path that exists and holds null renders as `null`. A path that the declared item type denies, and that no emitted object carries, is a usage error that names the real dotted paths ending with the same leaf name. A path that continues past an array is a usage error that names `--jq` |
```
dafydd-t: This content for `behaviour` is too long

internal/providers/metrics/cardinality_json_test.go
wardbekker: I removed this command-level test. The shared output tests cover the array-path rule, so this HTTP server is no longer needed.

internal/providers/faro/list_json_test.go
wardbekker: This test did not cover a command-specific boundary. I removed the file. The shared output tests cover empty typed lists and promoted JSON fields.

docs/design/output.md
wardbekker: I shortened the table cell to three short statements. The detailed rules remain in the text below the table.

javiermolinar (commented):
I see this introduces a lot of new machinery to achieve that rejection. We are also conflating that error response with the responses from the command action ie:

`gcx irm oncall routes create -f route.json --json bogus`

Returns an error even though the route was created. We could validate the transformation before mutation but I think a better solution would be emit a warning

jalopez (dismissed):
LGTM from grafana-com-maintainers side of things

domasx2 (dismissed):

mderynck:
Running `gcx irm oncall users list --json pk` prints “Did you mean spec.pk?” Following that suggestion still returns `null`. gcx puts the user ID in `metadata.name`, so the validator should suggest `metadata.name` instead of `spec.pk`.

wardbekker:
Updated in signed commit `3a635192`.

- `--json pk` and `--json spec.pk` now suggest `metadata.name` for OnCall list resources.
- Command-path tests verify the warning and the valid identity path.
- Flags-only OnCall list commands and `instrumentation clusters list` now reject positional arguments.
- The branch is rebased on current `main`.
- `GCX_AGENT_MODE=false mise run all` passes.
- I resolved the outdated review threads that the current branch fully addresses.

@claude review

radiohead (dismissed):
Looks good, I've ran a one-off test between current stable and this PR builds and the agents used fewer calls to course-correct with this fix.

I did find a couple of minor issues:

1. Docs (`docs/architecture/cli-layer.md`) are now outdated, listing invalid paths as usage errors instead of warnings that they are now.
2. Because we don't load schemas eagerly, there's different output in `resources list-types -o json` and `resources list-types --json group,version,kind,plural`. The former filters away resources unsupported by the APIs (based on current context), the latter does not.

Neither one of these is blocking, so I'm approving this PR – feel free to address the issues here or separately.

mderynck (dismissed):

PR #1195, opened 2026-08-14
docs(extensions): add ADR-023 for third-party extension design

dafydd-t:
## Summary

- Proposes ADR-023: a third-party extension mechanism for gcx (`gcx ext install/list/uninstall/update/<name>`), manifest-driven install (krew-inspired declarative platform selector), no PATH scanning, no forced hosting location, and no credential handoff — extensions reach Grafana/Cloud APIs by shelling out to `gcx` itself via a `GCX_EXT_GCX_BIN` env var.
- Background research comparing gh CLI extensions, kubectl/krew, cargo, Docker CLI plugins, and Helm, plus a PR-by-PR implementation plan, lives in `docs/research/2026-08-14-extensions-design.md`.
- Adds the ADR row to `ARCHITECTURE.md`'s ADR index.

Docs only, no code changes. Status is `proposed` — opening as draft since ADRs need explicit sign-off per CONSTITUTION.md before this becomes `accepted`.


docs/adrs/extensions/001-third-party-extensions-design.md
```diff
@@ -0,0 +1,56 @@
...
+- **No separate `run` verb.** `ext` registers `install`/`list`/`uninstall`/`update` as real subcommands; anything else in the first argument position is treated as an extension name and looked up in `index.json`. This is the same "known verbs win, anything else is a name" pattern git uses at its own top level.
+- **No credential handoff.** Extensions never receive a Grafana or Cloud token directly, in any form. Running an extension sets `GCX_EXT_GCX_BIN` on its environment (the path to the invoking `gcx` binary, the same idea as Docker's `DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND`); an extension that needs Grafana/Cloud data shells out to that binary and consumes its JSON output (`gcx api`, `gcx resources get`, any provider command with `--output json`), inheriting whatever auth and refresh logic gcx already handles. This mirrors how gh extensions commonly call `gh api` rather than handling GitHub auth themselves.
+- **No sandboxing.** The extension subprocess inherits the parent's full environment, the same posture every surveyed tool ships. Checksum verification is mandatory and fails closed for any non-local source; there is no signature/provenance check beyond the checksum.
```
dafydd-t, line 45: Add a note for public docs: we'll need to make it very clear to users that extensions are not audited, and it is their responsibility to check them before they install them.

docs/adrs/extensions/001-third-party-extensions-design.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Created**: 2026-08-14
+**Status**: proposed
+**Supersedes**: none
+
+## Context
+
+gcx has no way for third parties to add their own subcommands today. The CLI tree is built entirely at compile time (`cmd/gcx/root/command.go`), and providers self-register through a single `init()` call each.
+
+The goal is a third-party extension mechanism with four high level goals: easy to author, must run on every OS/arch gcx itself ships for (`linux/darwin/windows` × `amd64/arm64`), must not force extensions to be hosted in one particular place, and must not require a second binary to find or manage them (thinking about [Krew](https://krew.sigs.k8s.io/)).
```
awsome-o, line 11: Could we use one or two representative third party cases here to make the v1 contract and tradeoffs tangible? A partner integration, customer specific workflow, or migration tool would let us check the proposed manifest and exec design against real user and developer needs. Stable Grafana capabilities are useful as a boundary case rather than another extension target.

docs/adrs/extensions/001-third-party-extensions-design.md
```diff
@@ -0,0 +1,56 @@
...
+- **A "compile from source" fallback** when no manifest platform row matches — install just fails instead, keeping one predictable install code path.
+- **`gcx ext scaffold`** (a Go-only starter-repo generator) — pure authoring convenience, not required for install/run to work. We can build this later if we need it.
+- **A central, official discovery index, or `gcx ext search`** — discovery works however authors tell people about their extension. We can build this later if we need it.
+- **A metadata-handshake protocol plus a synthesized per-extension Cobra tree**, so extensions show up in `gcx commands`/`gcx help-tree` with full parity to built-ins — real engineering weight (a new public `extsdk` package, runtime output-validation) that none of the four constraints actually call for.
+
+## Consequences
+
+**Easier:** third parties can extend gcx without needing gcx's approval, review, or hosting. A Go author's release process is "tag, push, `goreleaser release`, copy the checksums into the manifest" — gcx's own build matrix and release pipeline map directly onto what the manifest needs. Nothing about the mechanism is GitHub-specific, satisfying the hosting constraint concretely.
```
awsome-o, line 61: I like the independence this gives external authors. One accepted tradeoff is that the same route may become attractive to internal teams shipping stable product capabilities. If that happens, commands under the same gcx name could differ in authentication, context, output, errors, safety, and support ownership, and experiments could remain outside core indefinitely. I think that risk is worth stating here.

docs/adrs/extensions/001-third-party-extensions-design.md
```diff
@@ -0,0 +1,56 @@
...
+- **Direct credential handoff to extensions** (a `gcx config resolve-extension-credentials` command, manifest-declared permissions, tagging the caller) — designed, then cut, because "shell out to `gcx` itself" already solves the same problem with no new credential-resolution code path at all.
+- **A "compile from source" fallback** when no manifest platform row matches — install just fails instead, keeping one predictable install code path.
+- **`gcx ext scaffold`** (a Go-only starter-repo generator) — pure authoring convenience, not required for install/run to work. We can build this later if we need it.
+- **A central, official discovery index, or `gcx ext search`** — discovery works however authors tell people about their extension. We can build this later if we need it.
+- **A metadata-handshake protocol plus a synthesized per-extension Cobra tree**, so extensions show up in `gcx commands`/`gcx help-tree` with full parity to built-ins — real engineering weight (a new public `extsdk` package, runtime output-validation) that none of the four constraints actually call for.
```
awsome-o, line 57: Deferring full parity makes sense for v1, but it creates a clear boundary. These extensions will not appear in `gcx commands` or `help-tree` and will not follow gcx output conformance, so agents cannot treat them like native gcx capabilities. For core Grafana commands, gcx needs to control those behaviors so agents learn one reliable pattern. I think the ADR should state that limitation directly.

docs/research/2026-08-14-extensions-design.md
```diff
@@ -0,0 +1,284 @@
...
+- gcx's own build rules some options out. The CLI tree is built at compile time. Providers register themselves in code. The build has no CGO and targets 6 platforms. So Go's native plugin system (`.so` files) can't work here. Extensions have to be separate programs.
+- The plan: a manifest-driven design, close to krew's. Install with `gcx ext install <source>` (a local path, a direct URL, or any git host - no GitHub required). Run with `gcx ext <name>`. Extensions never get Grafana or Cloud credentials directly - if they need that access, they shell out to the `gcx` binary itself, which is already logged in.
+- `gcx ext scaffold` (a Go starter-repo generator) is cut from v1. It's a nice-to-have, not something the mechanism needs to work.
+- A bigger, more complex design (a handshake step, extensions showing up in `gcx commands`/`gcx help-tree` just like built-in commands, a public SDK package) was worked out too, but it's not part of v1. It solves a problem nobody asked to solve.
```
awsome-o: The survey covers the installation choices well. I would add the product level conclusion: GitHub CLI and kubectl show executable extensions working for optional commands, Krew separates distribution and discovery, Terraform shows that deep external integrations need a stronger shared contract, and MCP shows that agent tools need structured discovery and behavior. That would make it clearer what this first version is testing and what it does not solve yet.

docs/research/2026-08-14-extensions-design.md
```diff
@@ -0,0 +1,284 @@
...
+- Test a git-URL install and a direct-manifest-URL install against a scratch repo, to confirm nothing GitHub-specific gets hit.
+- Test that a checksum mismatch, and a missing checksum, both fail hard.
+- Test that `GCX_EXT_GCX_BIN` is set correctly when an extension runs, and that a script extension calling `$GCX_EXT_GCX_BIN resources get ... --output json` (or similar) works using the user's own logged-in session.
+- Test that running an installed extension records a telemetry event with the extension's name; test that running an unmatched/typo'd name (or one whose manifest sets `telemetry.reportUsage: false`) records nothing.
+- `go test ./cmd/gcx/ext/... ./internal/ext/...`, then a full `mise run gate`, before calling this done.
+
+## Implementation Plan
+
+A sequence of small PRs, each one buildable and testable on its own. Nothing gets mounted into the live CLI (so no doc-regen or CI surface changes) until PR 6, which keeps PRs 1-5 small and low-risk. Each PR after PR 0 only depends on the one right before it.
```
awsome-o: The technical sequence is nicely incremental. I would add a product learning checkpoint before broader platform investment: try one independently owned extension, optionally try one limited Grafana experiment, then review author speed, user and support clarity, agent usability, and whether expected product capabilities are drifting out of core. That gives us evidence before committing to broader distribution or discovery.

awsome-o (commented):
I like the simple v1 approach and the visible `gcx ext` boundary. The mechanism seems useful for customer specific, partner, community, migration, and other independently owned commands that might otherwise become separate CLIs.

Two consequences feel important to capture. First, this should not accidentally become the easier path for stable Grafana capabilities that need gcx's shared quality and behavior contract. Second, because command tree, output, and metadata parity are deferred, v1 extensions will be runnable through gcx but will not automatically behave like first class gcx capabilities for users or agents.

I think the broader ecosystem and rollout questions can remain follow up work once we have validated the mechanism with a real independently owned extension.

docs/adrs/extensions/001-third-party-extensions-design.md
awsome-o: or just representative use case or pattern we want to support

docs/adrs/extensions/001-third-party-extensions-design.md
dafydd-t: the handshake adds a fair bit of work for the author, and we'd need to introduce a library/sdk to help. So i think its worth deferring, and we could even explore how to avoid it (i dunno, maybe listing extensions on `help`? spitballing...)

docs/adrs/extensions/001-third-party-extensions-design.md
dafydd-t: thats a good idea. Will have a look

dafydd-t: I might try [this stack](https://github.com/grafana/gcx/pull/1105) of features as a PoC. see what we get

dafydd-t: [vibed a PoC here](https://github.com/grafana/gcx/pull/1211), with a couple of extensions examples. Will digest the result of this later.

dafydd-t: good shout — done. built the Azure onboarding from #1105 as an extension to test it: https://github.com/grafana/gcx/pull/1211

the Context section now names three cases. two are what the mechanism is actually for (a partner integration, a customer migration tool), and Azure onboarding is the boundary case, which turned out to be the more useful one to write down: it fits the mechanism perfectly on technical criteria and still belongs in core. "implementable as an extension" is not the test.

most of the changes further down came out of building it.

docs/adrs/extensions/001-third-party-extensions-design.md
dafydd-t: agreed, and stated. new paragraph in Consequences: the mechanism cannot tell a partner integration from a Grafana team shipping a feature outside the release process, and if that happens commands under one gcx name diverge on auth, context, output, errors, safety and support ownership.

mitigation is editorial, not technical — the ADR now says who this is for, and the Azure boundary case says what it is not for.

docs/adrs/extensions/001-third-party-extensions-design.md
dafydd-t: accepted, stated directly in Consequences now.

the PoC made it concrete: gcx ext classifies as `raw` (the class already means "bytes gcx does not own", so it fits), which exempts it from the one-JSON-value rule the rest of the CLI follows, and installed extensions do not appear in `gcx commands` or `gcx help-tree` at all. an agent that finds `gcx setup datasources azure` today would find nothing for the same capability as an extension.

that is now written up as the direct cost of deferring the handshake, and as a reason a capability being implementable as an extension does not make it a good one.

docs/adrs/extensions/001-third-party-extensions-design.md
aangelisc: Maybe that could be a part of the `install` step? It'd be cool if, in future, they were reachable and audited like the plugins catalog but food for future thought.

docs/adrs/extensions/001-third-party-extensions-design.md
```diff
@@ -0,0 +1,77 @@
...
+- **No separate `run` verb.** `ext` registers `install`/`list`/`uninstall`/`update` as real subcommands; anything else in the first argument position is treated as an extension name and looked up in `index.json`. This is the same "known verbs win, anything else is a name" pattern git uses at its own top level. The boundary that follows is part of the CLI grammar and must be documented as such: **gcx's own global flags go before `ext`, and every argument after the extension's name is passed to it verbatim** (`gcx --context prod ext my-ext --dry-run`). Extensions must therefore not define flags that shadow gcx's globals.
+- **No credential handoff.** Extensions never receive a Grafana or Cloud token directly, in any form. An extension that needs Grafana/Cloud data shells out to the invoking `gcx` binary and consumes its JSON output (`gcx api`, `gcx resources get`, any provider command with `--output json`), inheriting whatever auth and refresh logic gcx already handles. This mirrors how gh extensions commonly call `gh api` rather than handling GitHub auth themselves.
```
aangelisc, line 38: This is interesting. I agree with this in principle but it then means that core Grafana components should still be built in the `gcx` repository. Although if Grafana exposes a public API for that component, then maybe this isn't as much of a concern.

PR #1213, opened 2026-08-19
fix(irm): make alert-groups list --limit 0 return every alert group

### LINKED ISSUE #1157: irm oncall alert-groups list: --limit 0 silently capped at 1000, no way to paginate further
## Problem

`gcx irm oncall alert-groups list --limit 0` documents `0` as "return all", but the
result is silently capped at 1000 rows:

```
$ gcx irm oncall alert-groups list --all --max-age 24h --limit 0 -o json
...
"list_meta": {"truncated": true, "returned": 1000, "cap": 1000}
```

On a busy stack this means **a single day cannot be retrieved**. In the instance I was
working with, one day is roughly 1,200 alert groups, so `--max-age 24h --limit 0` is
already truncated. There is no flag to raise the cap and no cursor/continuation flag
to fetch the next page, so the command cannot produce a complete result set for any
window longer than a few hours.

The cap is at least honestly reported in `list_meta`, which is good. The blocker is
that there is no way to act on it.

## Why it matters

Any analysis over alert history (load per rotation, per-hour distribution, ack rates,
noise ratios) needs more than 1000 rows. Today the only way through is to bypass the
CLI and drive the underlying endpoint directly:

```
GET /api/plugins/grafana-irm-app/resources/alertgroups/?perpage=100
-> {"next": ".../alertgroups?cursor=...", "results": [...]}
```

That endpoint paginates by **opaque cursor** (not page number, unlike the
page-numbered OnCall endpoints), and has no such ceiling. Having to reimplement
cursor-following outside gcx is the part that feels wrong, since gcx is otherwise the
right tool for this.

## Suggested fix

Any one of these would unblock it:

1. Make `--limit 0` genuinely mean all, following `next` cursors internally.
2. Add an explicit `--all-pages` / `--paginate` flag that does the cursor walk.
3. Surface the cursor: emit `list_meta.continue` as a usable `--cursor` value so
   callers can page themselves. (`list_meta.continue` currently returns a suggested
   *command string* with a higher `--limit`, which does not help once the 1000 cap is
   the binding constraint.)

Option 1 or 2 would be most useful, since the cursor walk is mechanical.

## Note on the workaround

For anyone hitting this before it is fixed: the cursor chain is sequential and cannot
be parallelised, but the endpoint accepts a `started_at=<from>_<to>` range, so the
window can be partitioned (e.g. by day) and the partitions fetched concurrently.

---

Filed by **Claude Code**, on behalf of @tomwilkie, while using gcx to analyse
OnCall rotation shapes and alert load. gcx v1.0.0 (homebrew).

tomwilkie:
Fixes #1157

## Problem

The rich list path bounded **every** fetch at `alertGroupListHardCap = 1000`, so:

- `--limit 0` returned at most 1000 alert groups. On a busy stack that is under a day of history, so no complete result set was retrievable for any useful window.
- `--limit N` for `N > 1000` was *also* silently cut down to 1000 — `effectiveCap` only ever shrank, never grew — so there was no way to raise the ceiling either.

The truncation was reported honestly via `list_meta.cap`, but the cap variant deliberately carries **no continuation**, because no larger `--limit` could beat the cap. So a caller had no action to take: the only escape was to bypass gcx and follow the endpoint's opaque `next` cursors by hand — exactly the mechanical work the client already does internally.

## Fix

Remove the cap. `--limit 0` follows `next` cursors to exhaustion, and `--limit N` is honored exactly.

That lets the command adopt the shared `BindListLimit` binder and its uniform *"0 means all results are returned"* wording, which is now an honest promise — so this **removes** the documented capped-source special case rather than adding a flag. `alert-groups list` was the only command taking that exception.

A drained page carries no `list_meta`, which per the truncation contract is the signal that the output **is** the complete set. A user-limited page is unchanged: `truncated`/`returned`, plus the observed `total` and a `--limit 0` continuation when the source happened to drain, or a doubled-limit continuation when it did not. The old honesty guard that withheld a large observed `total` is gone — `--limit 0` genuinely retrieves it now.

Two safety details:

- **Bulk-by-filter mutations** (`acknowledge`/`resolve`/`delete --team ...`) have no user-supplied limit to bound them, so `resolveBulkTargets` now passes its own `alertGroupBulkTargetCap = 1000`. Same value, same `perpage`, same trimming — that path is byte-for-byte unchanged.
- The unbounded walk **stops if a `next` cursor stops advancing**, so a misbehaving server cannot spin the loop forever.

`internal/output` is untouched: the cap machinery stays as documented general contract with its own tests, it simply has no caller now.

## Verification

Live against a busy stack — the case from the issue:

| invocation | before | after |
|---|---|---|
| `--all --max-age 24h --limit 0` | 1000 items, `list_meta.cap: 1000` | **3412 items, no `list_meta`** |
| `--all --max-age 24h --limit 2500` | 1000 items | 2500 items, truncated + `--limit 5000` continuation |
| `--all --max-age 24h --limit 10` | unchanged | 10 items, truncated + `--limit 20` continuation |

New tests: `--limit 0` drains past 1200 items, `--limit 1200` is honored in full, a non-advancing cursor terminates, and an end-to-end `--limit 0` returns everything with no `list_meta`. The two cap-dependent tests were rewritten and the `--limit` validation test retargeted at the real command path (the check moved to the shared `Options.Validate`).

`GCX_AGENT_MODE=false go test ./...` passes, `-race` passes on `irm` and `output`, `golangci-lint run ./...` reports only 3 gosec findings that pre-exist on `main`, and all four reference generators re-run with no drift beyond the expected `--limit` help-text change.

## Compliance

- **CONSTITUTION § frozen public surface** — no command path, alias, flag or positional syntax changes. `--limit 0` moves toward what the flag help already promised ("0 for all"), so this is a defect fix rather than a flag change. Anyone who wants the old ceiling can pass `--limit 1000`.
- **Output contract** — still exactly one JSON value on stdout (`finite`); hints stay on stderr.
- **docs/design/output.md §15** — absence of `list_meta` still means "complete set", and it is now *true* for `--limit 0`, where it previously could not be. No new hint strings; all truncation still flows through the shared helpers. §15.1/§15.3/§15.6 and the research doc are updated to record that no capped source remains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- irm: `gcx irm oncall alert-groups list --limit 0` returns every matching alert group instead of stopping at 1000, and `--limit N` above 1000 is honored (#1157).
 - instrumentation: app and service writes (`clusters apps configure`/`remove`, `services include`/`exclude`/`clear`) no longer fail with `otlp_url is required`.
 
 ## v1.1.0 (2026-08-14)
--- a/docs/design/output.md
+++ b/docs/design/output.md
@@ -342,11 +342,15 @@ Maximum number of <subject> to return. 0 means all results are returned
 ```
 
 **Capped-source exception:** commands whose fetch is bounded by a client-side
-safety cap (e.g. `irm oncall alert-groups list`, capped at 1000) must NOT use
-the binder — "0 means all" would be dishonest there. They keep a bespoke flag
-description that discloses the cap, and disclose the cap at runtime via
-`ListMeta.Cap` plus the cap-variant hint (below). `--limit 0` on a capped
-source means "as much as the cap allows", and the output must say so.
+safety cap must NOT use the binder — "0 means all" would be dishonest there.
+They keep a bespoke flag description that discloses the cap, and disclose the
+cap at runtime via `ListMeta.Cap` plus the cap-variant hint (below).
+`--limit 0` on a capped source means "as much as the cap allows", and the
+output must say so. No command currently takes this exception: the last one,
+`irm oncall alert-groups list`, dropped its 1000-item cap in favour of a real
+cursor drain (grafana/gcx#1157). Prefer draining — a cap that a caller cannot
+raise makes the complete set unreachable, which is the defect that issue
+reported.
 
 The binder is deliberately minimal: commands still pass the limit to their
 clients for server-side pushdown where the API supports it.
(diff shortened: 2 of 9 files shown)


willgallego-grafana (approved):
lgtm!

dafydd-t (approved):

PR #1226, opened 2026-08-21
feat(agento11y): add rules list-scores for online evaluation results

ivanahuckova:
## Summary
- Add `gcx agento11y rules list-scores <rule-id>` to fetch online evaluation score rows from `GET /eval/rules/{rule_id}/scores`
- Support full filter parity (evaluator, pass/fail, time range, agent/model/provider, score value, min/max, sort, limit) and table/wide/json/yaml output, with the list envelope (`{"items": [...], "list_meta": {...}}`) so truncation rides in-band, and a 1000-row client-side safety cap on `--limit 0`
- Extend score types with rule-scores metadata (conversation title, agent, model, provider) and document the command in CLI reference and the agento11y skill

## Scope note
- This PR is **purely additive** — it adds the new `rules list-scores` command and leaves the shipped `generations list-scores` command untouched. An earlier revision changed `generations list-scores`'s `--limit` from page-size-with-drain to a real cap; that is a v1.x frozen-surface change per CONSTITUTION.md and was pulled out per review. It will land separately with a release note (see Follow-ups). `generations list-scores` in this branch is byte-identical to `main`.

## Test plan
- [x] `go test ./internal/providers/agento11y/scores/...`
- [x] `go test ./cmd/gcx/root/... -run 'Consistency|SkillsGcx'`
- [x] `mise run lint`
- [x] `mise run gate`
- [ ] Manual: `./bin/gcx agento11y rules list-scores <rule-id> --passed=false -o json` against a stack with online eval traffic
- [ ] Manual: confirm the real `/eval/rules/{id}/scores` wire keys for `gen_model` and `gen_provider` (no sibling precedent — `experiments.ScoreItem` carries `source_kind`/`source_id`/`agent_name` but no model/provider). If the keys differ, the `MODEL` column in `-o wide` renders `-` and `--model`/`--provider` silently filter nothing.

## Deliberately not fixed here (with rationale)
This came up in review and is intentionally left out of this PR — it is shared infrastructure, and folding it in would widen the blast radius beyond "add `rules list-scores`". Filing as a follow-up keeps the decision reviewable on its own.

- **Truncation continuation can suggest a `--limit` above the 1000-row safety cap.** For a `list-scores --limit N` with `N` roughly in (500, 1000) and more than 1000 matching rows, the "see more with `--limit 2N`" hint proposes a limit the cap silently reduces back to 1000 (a two-step discovery; the data is never wrong). The doubling is produced by the shared `internal/output` helpers (`BuildListLimitCommand` / `listContinueCommand`), which don't know about a per-command cap — `irm oncall alert-groups list` uses the same helper and has the same behavior. Fixing it means teaching the shared helper to clamp the continuation to the cap, which affects every command that uses it; that belongs in its own change, not a feature PR.

## Follow-ups (out of scope)
- Give `generations list-scores` a real `--limit` cap (default page behavior currently drains all pages) as its own PR with a release note, since it changes a v1.0.0 flag behavior — pulled out of this PR per review.
- `experiments list-scores` and `experiments trials list-scores` still cap silently and treat negative `--limit` as "no limit"; converging the whole `list-scores` family onto the standard limit/truncation contract is worth a separate change.
- Clamp the shared list-truncation continuation hint to a command's safety cap (see "Deliberately not fixed here").

### DIFF
--- a/internal/providers/agento11y/agento11yhttp/client.go
+++ b/internal/providers/agento11y/agento11yhttp/client.go
@@ -68,11 +68,20 @@ func (c *Client) DoRequest(ctx context.Context, method, path string, body io.Rea
 // using the common { "items": [...], "next_cursor": "..." } envelope.
 // Pass maxItems <= 0 for no limit (fetch all pages).
 func ListAll[T any](ctx context.Context, c *Client, basePath string, query url.Values, maxItems ...int) ([]T, error) {
+	items, _, err := ListAllWithHasMore[T](ctx, c, basePath, query, maxItems...)
+	return items, err
+}
+
+// ListAllWithHasMore is like [ListAll] but reports whether the result was
+// truncated by maxItems. hasMore is true when a next page exists or the final
+// page overshot the bound — not when the source is genuinely complete.
+func ListAllWithHasMore[T any](ctx context.Context, c *Client, basePath string, query url.Values, maxItems ...int) ([]T, bool, error) {
 	limit := 0
 	if len(maxItems) > 0 {
 		limit = maxItems[0]
 	}
 	var all []T
+	var hasMore bool
 
 	// Copy to avoid mutating the caller's map during pagination.
 	q := make(url.Values, len(query))
@@ -86,25 +95,26 @@ func ListAll[T any](ctx context.Context, c *Client, basePath string, query url.V
 
 		resp, err := c.DoRequest(ctx, http.MethodGet, path, nil)
 		if err != nil {
-			return nil, err
+			return nil, false, err
 		}
 
 		if resp.StatusCode != http.StatusOK {
 			err := HandleErrorResponse(resp)
 			resp.Body.Close()
-			return nil, err
+			return nil, false, err
 		}
 
 		var page listResponse[T]
 		if err := json.NewDecoder(resp.Body).Decode(&page); err != nil {
 			resp.Body.Close()
-			return nil, fmt.Errorf("failed to decode response: %w", err)
+			return nil, false, fmt.Errorf("failed to decode response: %w", err)
 		}
 		resp.Body.Close()
 
 		all = append(all, page.Items...)
 
 		if limit > 0 && len(all) >= limit {
+			hasMore = page.NextCursor != "" || len(all) > limit
 			all = all[:limit]
 			break
 		}
@@ -116,9 +126,9 @@ func ListAll[T any](ctx context.Context, c *Client, basePath string, query url.V
 	}
 
 	if all == nil {
-		return []T{}, nil
+		return []T{}, hasMore, nil
 	}
-	return all, nil
+	return all, hasMore, nil
 }
 
 // DoJSON executes an HTTP request against the plugin API, optionally
--- a/internal/providers/agento11y/scores/client.go
+++ b/internal/providers/agento11y/scores/client.go
@@ -9,7 +9,24 @@ import (
 	"github.com/grafana/gcx/internal/providers/agento11y/agento11yhttp"
 )
 
-const generationScoresPathFmt = "/query/generations/%s/scores"
+const (
+	generationScoresPathFmt = "/query/generations/%s/scores"
+	ruleScoresPathFmt       = "/eval/rules/%s/scores"
+
+	// ruleScoreDefaultLimit is higher than the generation default (50) because
+	// rule scores are the failure-theme-analysis surface (the llm_hint suggests
+	// --limit 100); a single generation's score set is small, so 50 suffices.
+	ruleScoreDefaultLimit = 100
+	ruleScoreMaxValues    = 20
+
+	// scoreMaxPageLimit is the largest per-request page size sent to the API.
+	scoreMaxPageLimit = 500
+	// scoreHardCap bounds the total rows fetched when no smaller user --limit
+	// applies. Both rule and generation score tables can grow large, so an
+	// unbounded --limit 0 could load an unbounded set into memory; the cap keeps
+	// that bounded and is disclosed to the caller via ListMeta.Cap.
+	scoreHardCap = 1000
+)
 
 // Client is an HTTP client for Agent Observability generation score endpoints.
 type Client struct {
@@ -28,4 +45,98 @@ func (c *Client) ListByGeneration(ctx context.Context, generationID string, limi
 		query.Set("limit", strconv.Itoa(limit))
 	}
 	return agento11yhttp.ListAll[Score](ctx, c.base, fmt.Sprintf(generationScoresPathFmt, url.PathEscape(generationID)), query)
+}
+
+// ListScoresOptions holds filters for ListByRule.
+type ListScoresOptions struct {
+	Limit       int
+	EvaluatorID string
+	Passed      *bool
+	From        string
+	To          string
+	AgentNames  []string
+	Models      []string
+	Providers   []string
+	ScoreValues []string
+	MinValue    *float64
+	MaxValue    *float64
+	SortBy      string
+	SortDir     string
+}
+
+// ListByRule returns online evaluation score rows for a rule, paginated. The
+// bool is true when more rows exist beyond the returned page. opts is assumed
+// valid: the sole caller (the CLI command) validates it before calling.
+func (c *Client) ListByRule(ctx context.Context, ruleID string, opts ListScoresOptions) ([]Score, bool, error) {
+	query := buildRuleScoreQuery(opts)
+	items, hasMore, err := agento11yhttp.ListAllWithHasMore[Score](ctx, c.base, fmt.Sprintf(ruleScoresPathFmt, url.PathEscape(ruleID)), query, ruleScoreFetchCap(opts.Limit))
+	return items, hasMore, err
+}
+
+// ruleScoreFetchCap is the total-rows bound passed to ListAll as maxItems for the
+// rule command. A user --limit within (0, scoreHardCap] is honored verbatim; 0
+// ("no limit") or a limit above the cap is bounded by scoreHardCap — a rule's
+// score table grows with traffic and filters, not a larger --limit, are the route
+// beyond the cap. Validation rejects negatives upstream.
+func ruleScoreFetchCap(limit int) int {
+	if limit <= 0 || limit > scoreHardCap {
+		return scoreHardCap
+	}
+	return limit
+}
+
+// ScoreHardCap reports the safety cap for use as PagedListMeta's safetyCap and as
+// a test assertion hook.
+func ScoreHardCap() int { return scoreHardCap }
+
+// scorePageLimit is the per-request limit query param sent to the API.
+// The server caps page size at scoreMaxPageLimit; use that when fetching all pages.
+func scorePageLimit(limit int) int {
+	if limit <= 0 || limit > scoreMaxPageLimit {
+		return scoreMaxPageLimit
+	}
+	return limit
+}
+
+// buildRuleScoreQuery renders validated options into query params. It does not
+// validate; the CLI command validates before ListByRule is reached, so sort
+// fields are non-empty and filter slices carry no empty entries (guaranteed by
+// listRuleOpts.Validate).
+func buildRuleScoreQuery(opts ListScoresOptions) url.Values {
+	params := url.Values{}
+	params.Set("sort_by", opts.SortBy)
+	params.Set("sort_dir", opts.SortDir)
+	params.Set("limit", strconv.Itoa(scorePageLimit(opts.Limit)))
+
+	if opts.EvaluatorID != "" {
+		params.Set("evaluator_id", opts.EvaluatorID)
+	}
+	if opts.Passed != nil {
+		params.Set("passed", strconv.FormatBool(*opts.Passed))
+	}
+	if opts.From != "" {
+		params.Set("from", opts.From)
+	}
+	if opts.To != "" {
+		params.Set("to", opts.To)
+	}
+	if opts.MinValue != nil {
+		params.Set("min_value", strconv.FormatFloat(*opts.MinValue, 'f', -1, 64))
+	}
+	if opts.MaxValue != nil {
+		params.Set("max_value", strconv.FormatFloat(*opts.MaxValue, 'f', -1, 64))
+	}
+	for _, v := range opts.AgentNames {
+		params.Add("agent_name", v)
+	}
+	for _, v := range opts.Models {
+		params.Add("model", v)
+	}
+	for _, v := range opts.Providers {
+		params.Add("provider", v)
+	}
+	for _, v := range opts.ScoreValues {
+		params.Add("score_value", v)
+	}
+	return params
 }
--- a/internal/providers/agento11y/scores/client_test.go
+++ b/internal/providers/agento11y/scores/client_test.go
@@ -91,3 +91,183 @@ func TestClient_ListByGeneration_Empty(t *testing.T) {
 	require.NoError(t, err)
 	assert.Empty(t, items)
 }
+
+func TestClient_ListByRule(t *testing.T) {
+	now := time.Now().Truncate(time.Second)
+	client := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method)
+		assert.Contains(t, r.URL.Path, "/eval/rules/rule-1/scores")
+		assert.Equal(t, "eval-1", r.URL.Query().Get("evaluator_id"))
+		assert.Equal(t, "false", r.URL.Query().Get("passed"))
+		assert.Equal(t, "2026-04-01T00:00:00Z", r.URL.Query().Get("from"))
+		assert.Equal(t, "2026-04-02T00:00:00Z", r.URL.Query().Get("to"))
+		assert.Equal(t, []string{"agent-a", "agent-b"}, r.URL.Query()["agent_name"])
+		assert.Equal(t, []string{"gpt-4"}, r.URL.Query()["model"])
+		assert.Equal(t, []string{"openai"}, r.URL.Query()["provider"])
+		assert.Equal(t, []string{"fail"}, r.URL.Query()["score_value"])
+		assert.Equal(t, "0.1", r.URL.Query().Get("min_value"))
+		assert.Equal(t, "0.9", r.URL.Query().Get("max_value"))
+		assert.Equal(t, "created_at", r.URL.Query().Get("sort_by"))
+		assert.Equal(t, "desc", r.URL.Query().Get("sort_dir"))
+		assert.Equal(t, "50", r.URL.Query().Get("limit"))
+
+		w.Header().Set("Content-Type", "application/json")
+		writeJSON(w, map[string]any{
+			"items": []scores.Score{
+				{
+					ScoreID:           "s-1",
+					GenerationID:      "gen-1",
+					ConversationID:    "conv-1",
+					ConversationTitle: "Help with billing",
+					EvaluatorID:       "eval-1",
+					EvaluatorVersion:  "1",
+					RuleID:            "rule-1",
+					ScoreKey:          "relevance",
+					ScoreType:         "number",
+					Value:             scores.ScoreValue{Number: new(0.2)},
+					Passed:            new(false),
+					Explanation:       "Off topic",
+					AgentName:         "agent-a",
+					GenModel:          "gpt-4",
+					GenProvider:       "openai",
+					CreatedAt:         now,
+				},
+			},
+		})
+	}))
+
+	passed := false
+	minV, maxV := 0.1, 0.9
+	items, hasMore, err := client.ListByRule(context.Background(), "rule-1", scores.ListScoresOptions{
+		Limit:       50,
+		EvaluatorID: "eval-1",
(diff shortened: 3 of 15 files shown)


ivanahuckova:
@claude review

ivanahuckova:
@claude review

ivanahuckova:
@claude review

ivanahuckova:
@claude review

internal/providers/agento11y/scores/commands.go
```diff
@@ -48,18 +75,295 @@ func NewListScoresCommand(loader *providers.ConfigLoader) *cobra.Command {
...
 				return err
 			}
-			return opts.IO.Encode(cmd.OutOrStdout(), scores)
+			if err := opts.IO.Encode(cmd.OutOrStdout(), scores); err != nil {
+				return err
+			}
+			// Output stays a bare array (a shipped shape), so the truncation hint
+			// is stderr-only per output.md §15.2; it reports the same client-side
+			// safety cap the rule command uses.
+			meta := scoreListMeta(len(scores), opts.Limit, hasMore, ScoreHardCap())
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

A generation has no filter flags, so the cap-variant "refine filters" hint stranded rows 1001+. Generation now **honors an explicit `--limit` verbatim** (the real route to more rows) and caps **only `--limit 0`** at the 1000-row backstop (`ListByGeneration` fetches `limit` when > 0, else `scoreHardCap`). The truncation hint is built with `safetyCap = 0`, so it emits `showing first N; more results are available. See more with: … --limit 2N` rather than the filters-based advice generation can't act on. Covered by `TestNewListScoresCommand_ZeroLimitCappedWithRoute`.

internal/providers/agento11y/scores/types.go
```diff
@@ -1,31 +1,63 @@
...
+	// (/query/generations/<id>/scores) returns a nested {source: {kind, id}}
+	// envelope; the eval endpoints (/eval/rules/<id>/scores, experiments) return
+	// a flat source_kind/source_id pair. SourceKind/SourceID capture the flat
+	// wire form on decode; UnmarshalJSON folds them into Source so callers see a
+	// single shape regardless of which endpoint produced the row.
+	Source      *ScoreSource `json:"source,omitempty"`
+	SourceKind  string       `json:"source_kind,omitempty"`
+	SourceID    string       `json:"source_id,omitempty"`
+	AgentName   string       `json:"agent_name,omitempty"`
+	GenModel    string       `json:"gen_model,omitempty"`
```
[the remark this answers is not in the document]
ivanahuckova: Not changed in code, flagged instead. `gen_model`/`gen_provider` were added from the API shape; removing them if they're correct would gut the model/provider surface, and I can't confirm the wire keys without a live stack. So [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27) adds an explicit manual-verification item to the PR test plan to confirm both keys against a real `/eval/rules/{id}/scores` response **before merge** — if they differ we rename (and the `MODEL` column / `--model`/`--provider` filters get fixed with them). Prefer flagging over guessing here.

internal/providers/agento11y/scores/commands.go
```diff
@@ -48,18 +75,295 @@ func NewListScoresCommand(loader *providers.ConfigLoader) *cobra.Command {
 		Long:  `List evaluation scores produced by online rules for a generation.`,
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := opts.IO.Validate(); err != nil {
+			if strings.TrimSpace(args[0]) == "" {
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Both commands now bind the trimmed value once (`ruleID` / `genID`) and use it for both the empty-check and the request, so `list-scores "  gen-1 "` no longer sends a `%20`-padded path segment. Covered by `TestListScoresCommands_TrimIDBeforeRequest`, which asserts the request path contains the clean ID and no `%20`.

internal/providers/agento11y/scores/commands.go
```diff
@@ -48,18 +75,295 @@ func NewListScoresCommand(loader *providers.ConfigLoader) *cobra.Command {
...
+	}{
+		{"evaluator-id", o.EvaluatorID},
+		{"from", o.From},
+		{"to", o.To},
+		{"sort-by", o.SortBy},
+		{"sort-dir", o.SortDir},
+	}
+	for _, s := range scalars {
+		if flags.Changed(s.name) && s.val == "" {
+			return fmt.Errorf("--%s was set to an empty value; omit the flag to skip this filter", s.name)
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Removed `sort-by`/`sort-dir` from `validateNonEmptyFilters`' `scalars` (they're not filters). `--sort-by ""` now falls through to the switch in `validateScoreOptions`, giving the better error `--sort-by must be "created_at" or "value" (got "")`. Test updated to assert the allowed-values message.

internal/providers/agento11y/scores/client_test.go
```diff
@@ -87,7 +88,193 @@ func TestClient_ListByGeneration_Empty(t *testing.T) {
...
+	hardCap := scores.ScoreHardCap()
+	assert.Equal(t, hardCap, scores.ScoreFetchCap(0), "no user limit is bounded by the hard cap")
+	assert.Equal(t, hardCap, scores.ScoreFetchCap(hardCap+1), "a limit above the cap is bounded by it")
+	assert.Equal(t, 50, scores.ScoreFetchCap(50), "a limit within the cap is honored verbatim")
+	assert.Equal(t, hardCap, scores.ScoreFetchCap(hardCap), "a limit equal to the cap is honored")
+}
+
+func TestClient_ListByRule_ZeroLimitBoundedByHardCap(t *testing.T) {
+	hardCap := scores.ScoreHardCap()
+	// A server that always returns a full page + cursor would page forever
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Reworded the comment to say the cap bounds total **rows**, not requests: a stream of full non-empty pages + cursor is stopped once accumulated rows reach the cap, and an empty-page-with-cursor loop is pre-existing `ListAll` behavior that's explicitly out of scope. No longer overclaims.

internal/providers/agento11y/scores/commands.go
```diff
@@ -48,18 +75,295 @@ func NewListScoresCommand(loader *providers.ConfigLoader) *cobra.Command {
...
+func (o *listRuleOpts) setup(flags *pflag.FlagSet) {
+	// GenMeta only affects the wide layout (AGENT/MODEL columns), so the plain
+	// table codec is left at its default.
+	o.IO.RegisterCustomCodec("table", &TableCodec{})
+	o.IO.RegisterCustomCodec("wide", &TableCodec{Wide: true, GenMeta: true})
+	o.IO.DefaultFormat("table")
+	o.IO.BindFlags(flags)
+	// A client-side safety cap bounds --limit 0, so the standard "0 means all"
+	// binder does not apply; disclose the cap in the help instead.
+	flags.IntVar(&o.Limit, "limit", ruleScoreDefaultLimit, fmt.Sprintf("Maximum number of scores to return. 0, or a value above %d, returns up to the %d-row safety cap; narrow with filters to see beyond it", scoreHardCap, scoreHardCap))
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Added a rationale comment on `ruleScoreDefaultLimit`: rule scores are the failure-theme-analysis surface (the `llm_hint` suggests `--limit 100`), whereas a single generation's score set is small, so 50 suffices. Deliberate divergence, now recorded.

internal/providers/agento11y/scores/client.go
```diff
@@ -21,11 +35,106 @@ func NewClient(base *agento11yhttp.Client) *Client {
...
+	items, hasMore, err := agento11yhttp.ListAllWithHasMore[Score](ctx, c.base, fmt.Sprintf(ruleScoresPathFmt, url.PathEscape(ruleID)), query, ScoreFetchCap(opts.Limit))
+	return items, hasMore, err
+}
+
+// ScoreFetchCap is the total-rows bound passed to ListAll as maxItems, shared by
+// the rule and generation score commands. A user --limit within (0, scoreHardCap]
+// is honored verbatim; 0 ("no limit") or a limit above the cap is bounded by
+// scoreHardCap. Validation rejects negatives upstream. Exported so callers can
+// pass the same value as PagedListMeta's safetyCap.
+func ScoreFetchCap(limit int) int {
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Unexported to `ruleScoreFetchCap` — after generation moved to its own inline `--limit 0` backstop, `ListByRule` is the only caller. Dropped the redundant direct unit test; the cap behavior is covered through `ListByRule`. `ScoreHardCap()` stays exported as the test assertion hook, as you noted.

internal/providers/agento11y/scores/commands.go
```diff
@@ -81,45 +389,69 @@ func (c *TableCodec) Format() format.Format {
...
+	// Accept both the bare slice (generation list-scores) and the rule
+	// list-scores envelope, so the codec renders the rows either way.
+	var items []Score
+	switch t := v.(type) {
+	case []Score:
+		items = t
+	case ruleScoresEnvelope:
+		items = t.Items
+	case *ruleScoresEnvelope:
+		items = t.Items
```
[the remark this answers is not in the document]
ivanahuckova: Fixed in [`1d8d33e1`](https://github.com/grafana/gcx/commit/1d8d33e1738d1d65cf322952a68f375c062dda27).

Removed the `*ruleScoresEnvelope` case from the codec type switch — only the value form is ever passed (`RunE` encodes `ruleScoresEnvelope{…}`).

awsome-o (commented):
Went through this with the review-pr and integrate-with-gcx guidance. Five comments below. One I think needs sorting before merge, one worth a look, three you can take or leave.

The approach looks right to me. `rules list-scores` reuses `ListAll` for paging and the shared `PagedListMeta` / `AttachListMeta` / `EmitListTruncationHint` helpers, and it goes with the envelope for the new command while leaving the old bare array alone on the generation side. That's what output.md §15.2 asks for in both cases. Registration, output class, agent annotation and the skill update are all where they should be, and the bit in the skill explaining how the two commands differ in shape is the sort of thing that normally gets left out. 1467 lines looks like a lot until you notice 763 of it is tests and generated docs, so the size is mostly fair. I agree with the "deliberately not fixed" note too, the shared continuation helper belongs in its own PR.

The one that needs sorting is the `generations list-scores --limit` change. You explain it clearly in the description, which helps, but it still changes how a flag behaves on a command that's been out since v1.0.0, and I don't think a feature PR should carry that. More in the comment on `commands.go`.

The unverified wire keys are what I'd actually fix first though. Those can hand someone the wrong answer, where the flag change only hands them fewer rows.

If you wanted to trim: dropping the 20-entry `--score-value` cap and turning `ScoreHardCap()` into an exported const is about 15 lines and one error path. Short list, which is the right answer for a diff this size. Pulling the `generations --limit` change out is the bigger one, roughly 60 lines plus its tests.

Three things I checked and decided not to raise, so nobody brings them up next round. `validateNonEmptyFilters` is exactly what T2 of the self-review asks for when an empty flag value widens the query, so I'd keep it. `checkFinite` guards a value `strconv.ParseFloat` really does return, so it isn't dead code. And `safetyCap: 0` on the generation side gives an accurate hint, because an explicit `--limit` is honoured as-is there.

One suggestion: if you haven't run `integrate-with-gcx` on this branch, it's worth a go. T9 is the trigger behind both of the API questions below, and T3 has the "don't fold a compatibility change into an unrelated feature PR" rule that the flag change bumps into.

`go test ./internal/providers/agento11y/...` passes on the branch.

internal/providers/agento11y/scores/commands.go
```diff
@@ -33,7 +47,20 @@ func (o *listOpts) setup(flags *pflag.FlagSet) {
 	o.IO.RegisterCustomCodec("wide", &TableCodec{Wide: true})
 	o.IO.DefaultFormat("table")
 	o.IO.BindFlags(flags)
-	flags.IntVar(&o.Limit, "limit", 50, "Maximum number of scores to return")
+	// A client-side safety cap bounds --limit 0 (a generation's score set can be
+	// large), so the standard "0 means all" binder does not apply; an explicit
+	// --limit is honored verbatim, so it stays the route to more rows.
+	flags.IntVar(&o.Limit, "limit", 50, fmt.Sprintf("Maximum number of scores to return. 0 returns up to a %d-row safety cap; set a higher --limit to fetch more", scoreHardCap))
```
awsome-o: I think this one needs sorting out before merge.

It changes what `--limit` does on a command that's been out since v1.0.0. It used to just be the page size, telling the API how many rows to send back per request, and the code kept asking for more pages until they ran out. So `--limit 50` on a generation with 200 scores gave you all 200.

Now it's a real cap. The same command gives you 50 and exits 0. Anyone with a script doing `... -o json | jq 'length'` gets 50 instead of 200 and no way to tell, because the warning goes to stderr. And `--limit -1` used to mean "give me everything", now it's an error.

I get the reasoning in the description, that the old behaviour didn't match its own help text. But CONSTITUTION.md speaks to this directly. Flags stay stable within a major version, and the v1.0.0 surface stays supported through all of v1.x "including invocations that predate or deviate from the rules above". So the help text mismatch isn't enough on its own.

The new rules command doesn't need this change either. Could it go out on its own with a release note? Or if you'd rather keep it here, is someone happy to sign off on the break?

internal/providers/agento11y/scores/client.go
```diff
@@ -21,11 +38,111 @@ func NewClient(base *agento11yhttp.Client) *Client {
...
+		params.Set("min_value", strconv.FormatFloat(*opts.MinValue, 'f', -1, 64))
+	}
+	if opts.MaxValue != nil {
+		params.Set("max_value", strconv.FormatFloat(*opts.MaxValue, 'f', -1, 64))
+	}
+	for _, v := range opts.AgentNames {
+		params.Add("agent_name", v)
+	}
+	for _, v := range opts.Models {
+		params.Add("model", v)
```
awsome-o, line 133: This is the one I'd look at first.

Eleven filter params go out on the wire here, and as far as I can tell nothing has confirmed the API actually reads them. Your test plan has that box unticked too. I went looking for something similar elsewhere in the repo and came up empty. `experiments.ScoreItem` has `source_kind`, `source_id` and `agent_name`, but nothing in agento11y sends `model`, `provider`, `score_value`, `min_value`, `max_value`, `sort_by` or `sort_dir`.

Most APIs just ignore query params they don't recognise. If this one does, then `--model gpt-4 --passed=false` quietly returns failures from every model, and the table and the JSON both look like a filtered result. Someone digging into why gpt-4 keeps failing would end up blaming it for another model's failures.

A `MODEL` column showing `-` is no big deal. A filter that does nothing is. Same question for `gen_model` and `gen_provider` on the way back in.

Any chance you can run that manual check before merging? Or move the params you can't confirm into a follow-up?

internal/providers/agento11y/scores/client.go
```diff
@@ -9,7 +9,24 @@ import (
...
-const generationScoresPathFmt = "/query/generations/%s/scores"
+const (
+	generationScoresPathFmt = "/query/generations/%s/scores"
+	ruleScoresPathFmt       = "/eval/rules/%s/scores"
+
+	// ruleScoreDefaultLimit is higher than the generation default (50) because
+	// rule scores are the failure-theme-analysis surface (the llm_hint suggests
+	// --limit 100); a single generation's score set is small, so 50 suffices.
+	ruleScoreDefaultLimit = 100
+	ruleScoreMaxValues    = 20
```
awsome-o, line 20: Where does 20 come from?

Right now `--score-value` with 21 entries fails inside gcx rather than at the API, and the error tells the user to split the request to get under a limit the server might not even have.

If it's there to stop the URL getting too long, worth saying so in the comment so the number has a reason behind it. Otherwise I'd let the API reject it and pass on whatever it says. Fine as a follow-up either way.

internal/providers/agento11y/agento11yhttp/client.go
```diff
@@ -86,25 +95,26 @@ func ListAll[T any](ctx context.Context, c *Client, basePath string, query url.V
...
 			resp.Body.Close()
-			return nil, fmt.Errorf("failed to decode response: %w", err)
+			return nil, false, fmt.Errorf("failed to decode response: %w", err)
 		}
 		resp.Body.Close()
 
 		all = append(all, page.Items...)
 
 		if limit > 0 && len(all) >= limit {
+			hasMore = page.NextCursor != "" || len(all) > limit
```
awsome-o, line 117: Small thing, and I couldn't confirm it from here.

Some paginated APIs hand back a cursor any time a page comes back full, not just when there's genuinely more to fetch. If this one does that, a rule with exactly 100 matching rows on the default `--limit 100` would come back with `list_meta.truncated: true` and a "more results are available" hint, when actually you've got the lot.

Erring this way is the safe direction and it only costs one wasted query, so I wouldn't hold the PR for it. But since §15.2 treats `list_meta` turning up as the signal there's more, worth a quick check while you're testing the wire keys.

internal/providers/agento11y/scores/client.go
```diff
@@ -21,11 +38,111 @@ func NewClient(base *agento11yhttp.Client) *Client {
...
+func ruleScoreFetchCap(limit int) int {
+	if limit <= 0 || limit > scoreHardCap {
+		return scoreHardCap
+	}
+	return limit
+}
+
+// ScoreHardCap reports the safety cap for use as PagedListMeta's safetyCap and as
+// a test assertion hook.
+func ScoreHardCap() int { return scoreHardCap }
```
awsome-o, line 90: Take or leave this one. The comment says it's here as a test hook, and the only non-test caller is `commands.go` in the same package. The tests are in `scores_test`, so an exported const would do the same job without adding a function to the package surface.

internal/providers/agento11y/scores/commands.go
ivanahuckova: Agreed, and pulled out. This PR is now purely additive — it only adds the new `rules list-scores` command, and `generations list-scores` is byte-identical to `main` again (reverted `ListByGeneration`, the flag help, and the command handler; verified with a diff against `origin/main`). So `--limit` on `generations list-scores` keeps its v1.0.0 behavior: page size, drains all pages, `-1`/`0` mean "everything". No frozen-surface change rides in this PR.

The real `--limit` cap for `generations list-scores` will land as its own PR with a release note — tracked in the Follow-ups section of the description. Force-pushed as `0bed6034`.

ivanahuckova: https://github.com/grafana/gcx/pull/1226/commits/0bed6034ea39e0448ac43648edaa293815031150

internal/providers/agento11y/scores/client.go
ivanahuckova: image
 
I tested this manually and verified all works as expcted. There is issue mentioned for ` --score-value` param: `│ 8   │ --score-value        │ numeric: 8/10→0 (values exist!); bool: false→1000 [only false], true→1000 [only true] │ ⚠️ type-dependent │` , but this is an agento11y server-side issue and I will follow-up and add fix to eval API.

internal/providers/agento11y/agento11yhttp/client.go
ivanahuckova: image

ivanahuckova:
Thank you @awsome-o for review. I updated and tested all things you mentioned and re-requested your review again. :)

awsome-o (approved):
Approving. The generations `--limit` fix looks right.

One thing we noticed: `--score-value` comes back empty for numeric scores, even when rows with that value exist. You might expect `--score-value 8` to return the rows scored 8, like it does for the string values. Not sure if an empty result looks the same as a genuine "no matches".

Not a blocker. Up to you whether to fix it now or just log it for later.

The two open threads on `ruleScoreMaxValues = 20` and `ScoreHardCap()` were take-or-leave, so feel free to resolve them as they are.

ivanahuckova:
>One thing we noticed: --score-value comes back empty for numeric scores, even when rows with that value exist. You might expect --score-value 8 to return the rows scored 8, like it does for the string values. Not sure if an empty result looks the same as a genuine "no matches".

this is a bug in evaluator api that i need to fix. `--score-value 8` command is correct and should work once evaluator api is fixed.

PR #1231, opened 2026-08-24
feat(install): support main branch builds

wardbekker:
## Summary

- support `GCX_VERSION=main` by building `github.com/grafana/gcx/cmd/gcx@main` with Go
- preserve checksum-verified release downloads for tagged versions
- add installer coverage and document the new install option

## Verification

- real `GCX_VERSION=main` smoke install
- `mise run tests`
- `GOLANGCI_LINT_CACHE=<isolated-cache> mise run lint`
- `GCX_AGENT_MODE=false mise run docs`


dafydd-t:
I think, if folks are willing to run the bleeding edge, we can expect the to know how to clone & build from source? What is the target use case here?

PR #1244, opened 2026-08-25
test: simplify and profile the test suite

wardbekker:
## Measured impact

| Measure | Original PR | Final PR | Change |
|---|---:|---:|---:|
| CI test critical path | 4m 32s | 56s | 3m 36s faster, or 79.4% |
| CI Go test step | 3m 57s | 22s | 3m 35s faster, or 90.7% |
| Local uncached Go tests | 57.77s | 55.66s | 2.11s faster, or 3.7% |

The final CI run forced all tests to run with `GOFLAGS=-count=1`. It reused only compiled dependencies. The result does not depend on cached test results.

The first run had no per-shard build caches. Its slowest test job took 2m 48s. Later PRs can restore the per-shard caches from the default branch. A Go version or dependency change can cause another cold run.

The full workflow still waits for lint and documentation. They took 3m 29s and 3m 28s in the final run. This PR reduces the test check wait. It does not reduce the full workflow wait by half.

## Summary

- Split the Go package set into six stable, weighted CI shards.
- Give each CI shard an independent Go build cache.
- Force CI to run all tests instead of using cached test results.
- Run three package shards in parallel during local `mise run tests`.
- Keep one aggregate `Tests` job for the branch protection check.
- Reuse the current test binary for agent conformance checks.
- Replace fixed delays and retry-based test cases with deterministic local synchronization.
- Use the system `true` executable instead of a temporary unsigned editor executable on macOS.
- Document the local shard controls.

This PR does not add JSON results, JUnit results, or a slow-test report. These files are not necessary for the wait-time goal.

## Expected impact

Users will not see a product change. The PR does not change CLI behavior, output, APIs, or production dependencies.

Maintainers can see these costs:

- CI uses six test runners instead of one.
- CI stores one Go build cache for each shard.
- The first run after a Go version or dependency change can be slower than a warm run.
- Test logs are split across six jobs.
- Local sharding can use more CPU and memory. It gives only a small speed increase on the measured workstation.

Set `GCX_TEST_SHARD_COUNT=1` to run one local shard.

## Validation

- `GOFLAGS=-count=1 GCX_TEST_SHARD_COUNT=6 mise run tests:cli`
- All final GitHub Actions checks passed.
- All six CI shards passed with test result caching disabled.
- Lint reported 0 issues.
- The build, generated references, skill validation, and documentation build passed.
- The documentation build reported only existing link warnings.

## Compliance

- The change does not modify a product command or public API.
- The change does not modify a CONSTITUTION invariant or a DESIGN rule.
- The change does not add a production dependency.
- The change does not add or move a package, command group, data model, or pipeline.
- CONTRIBUTING and the project structure document describe the test workflow.

### DIFF
--- a/internal/login/detect_test.go
+++ b/internal/login/detect_test.go
@@ -141,7 +141,7 @@ func TestProbeTarget(t *testing.T) {
 func TestProbeTargetTimeout(t *testing.T) {
 	// AC-019: probe timeout → TargetUnknown
 	slowSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		time.Sleep(5 * time.Second)
+		<-r.Context().Done()
 	}))
 	defer slowSrv.Close()
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -73,6 +73,14 @@ jobs:
       - name: Tests
         run: mise run tests
 
+      - name: Upload Go test results
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: go-test-results
+          path: .test-results/
+          if-no-files-found: warn
+
   docs:
     name: Documentation
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
         run: mise run lint
 
   test-shards:
-    name: Go tests (${{ matrix.shard_number }}/3)
+    name: Go tests (${{ matrix.shard_number }}/6)
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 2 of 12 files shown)


dafydd-t:
I have to say, I am not convinced this extra complexity on top of go's test tooling is worth the benefit, or at least, I'd prefer to find optimisations that don't build so heavily on top of the test tooling.

- Go already run each package's tests in parallel, and `-parallelism` is set by default to the number of cores on your machine (going above your number of cores might cause issues)
- We can run tests for specific packages with `-p`
- We should look instead at making individual tests _within_ a package be runnable in a parallel way.

We have 596 test files, 3543 test functions. only 30 files use t.Parallel().

I think instead we should see if we can make more of these test functions use `t.Parallel()`.

internal/login/detect_test.go
```diff
@@ -141,7 +141,7 @@ func TestProbeTarget(t *testing.T) {
 func TestProbeTargetTimeout(t *testing.T) {
 	// AC-019: probe timeout → TargetUnknown
 	slowSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		time.Sleep(5 * time.Second)
+		<-r.Context().Done()
 	}))
```
dafydd-t, line 145: This is probably worth keeping. 5 seconds is way too long

wardbekker:
as per discussion, let's use standard go test functionality to speed up test execution during development, as this only impacts CI on GH and it already runs pretty efficiently today. Don't want to overcomplicate.

PR #1249, opened 2026-08-26
Update OAuth sign-in configuration instructions

marinnedea:
Added note about custom roles and access permissions for gcx CLI.

The current documentation doesn't take into account custom roles where the `plugins.app:access` permissions are stripped down.


marinnedea:
I'm not 100% sure I put that comment in the best place.. though.

dafydd-t (approved):
Thank you!

PR #1252, opened 2026-08-26
feat(datasources): add OpenSearch datasource support

ktw4071:
## Summary

Adds an **OpenSearch** datasource type to gcx, part of the datasource-coverage expansion initiative. Users can now search documents, tail logs, run metric aggregations, and browse index/field mappings directly from the CLI.

`grafana-opensearch-datasource` is a fork of the core Elasticsearch plugin, and I confirmed **live** against a real OpenSearch datasource that its query wire protocol is (almost) identical — so this implementation is a faithful sibling of `internal/query/elasticsearch` / `internal/datasources/elasticsearch`, not a new design.

Commands (under `gcx datasources opensearch`):
- `query [EXPR]` — Lucene document search, `--mode documents` (default, raw source documents) or `--mode logs` (newest-first, plugin-internal fields trimmed)
- `metrics [EXPR]` — metric aggregation over a time histogram, `--agg count|avg|sum|min|max|cardinality`, optional `--group-by` split into named series
- `list-indices` — indices visible to the datasource, with mapped field counts
- `list-fields` — mapped fields per index, nested/multi-fields flattened to dotted names

Also reachable via the auto-detecting `gcx datasources query <uid> <expr>`, which now returns a structured-query redirect naming `opensearch query`/`metrics` (see Design below).

## Design

- **Follows the Elasticsearch pattern exactly**: same client shape (`internal/query/grafanaquery` + `internal/query/dataframe` for transport, `querysql.QueryResponse` reused for `query`'s output), same command layout, same Explore-link builders mirroring the client's own query models so the two can never drift apart.
- **Two real plugin divergences found and fixed via live testing**, not assumed from reading Elasticsearch's code:
  1. The `logs` metric's row-count setting is named `size` in OpenSearch's plugin, not `limit` as in Elasticsearch's — `limit` was silently ignored, with the backend falling back to its own default of 500 rows regardless of the requested value. Confirmed via raw `/api/ds/query` probes at several sizes (1, 2, 3, 7) before and after.
  2. `metrics --group-by`'s grouped series arrive with an empty `schema.name` (unlike Elasticsearch's, which apparently names the frame) — OpenSearch instead labels the value field, e.g. `{"app.keyword": "grafana"}`. `parseAggsResponse` now tries the frame name first and falls back to the value field's labels, so the same parser handles either plugin's shape.
  3. Every other query shape — `raw_data` documents, aggregations' metric/bucket fields, and `_mapping` metadata discovery — uses identical field names and response shapes on both plugins, confirmed live.
- **Fixed a pre-existing gap alongside this**: Elasticsearch itself was never registered in the generic query dispatcher's routing tables (`cmd/gcx/datasources/query_routes.go`) — neither `dispatch` nor `redirects` — so `gcx datasources query <elasticsearch-uid> ...` reported "unsupported" instead of pointing at the typed command. Both `elasticsearch` and `opensearch` now get a `redirects` entry, since both have structured queries (mode/agg/field/group-by) that can't fit a single generic expression.
- **`query --limit` and `metrics --group-size` now disclose their own cap**, closing a completeness gap present in Elasticsearch too (silent truncation, no signal that more results exist) — the same defect class fixed on every SQL-dialect datasource this cycle. The mechanism is the sentinel half of that pattern without the SQL-specific "bail" half: there's no raw-SQL-injection-safety concern to opt out of (Lucene search is always read-only), but the same "fetch cap+1, trim, disclose" idea applies to a JSON `size`/`group-size` field exactly as it does to a `LIMIT` clause. A clean (non-sentinel) request is kept for the Explore link so it never leaks the +1.
- **PPL is out of scope for this PR.** OpenSearch also supports a SQL-like piped query language (verified working live via `queryType: "PPL"`) — deliberately deferred as a separate feature, not a Lucene-DSL limitation.

## Sample Output

All output below is real, live against the dev stack's `opensearch-ds-m` datasource (`grafana-opensearch-datasource` v2.34.3, OpenSearch 2.19.0 backend), two seeded indices (`grafana-logs`, `grafana-metrics`).

```
$ gcx datasources opensearch list-indices -d opensearch-ds-m
INDEX                      FIELDS
.kibana_1                  129
.opensearch-observability  17
.plugins-ml-config         6
.ql-datasources            5
grafana-logs               20
grafana-metrics            18

$ gcx datasources opensearch list-fields -d opensearch-ds-m --index grafana-logs
INDEX         FIELD               TYPE
grafana-logs  @timestamp          date
grafana-logs  app                 text
grafana-logs  app.keyword         keyword
grafana-logs  datacenter          text
grafana-logs  datacenter.keyword  keyword
grafana-logs  duration_ms         long
grafana-logs  host                text
...

$ gcx datasources opensearch query -d opensearch-ds-m 'level:error' --since 1h --limit 3
# stderr:
⚠ showing the first 3 rows; more rows match — raise --limit (max 1000) to see more
# stdout (exactly 3 rows — the sentinel row is trimmed before output; a query
# with no more matches beyond --limit stays silent, no false warning):
@TIMESTAMP            _ID                   _INDEX        _TYPE  APP    DATACENTER  DURATION_MS  HOST   LEVEL  LOGGER       MESSAGE                                  TYPE  USER_ID
2026-08-26T21:11:21Z  ynXpP6ABTuYPniJnyFTF  grafana-logs  -      faro   apac0       -            host4  ERROR  metrics      Rate limit exceeded                      log   user_1403
2026-08-26T21:11:21Z  y3XpP6ABTuYPniJnyFTF  grafana-logs  -      mimir  eu0         1245         host2  ERROR  metrics      Email delivery failed                    log   -
2026-08-26T21:11:21Z  yXXpP6ABTuYPniJnyFTF  grafana-logs  -      mimir  apac0       -            host1  ERROR  http.server  OPTIONS /api/v1/auth/logout - 422 - 0ms  log   -

$ gcx datasources opensearch query -d opensearch-ds-m 'level:error' --mode logs --since 1h --limit 3
@TIMESTAMP            _ID                   _INDEX        APP    DATACENTER  DURATION_MS  HOST   LEVEL  LOGGER       MESSAGE                                  TYPE  USER_ID
2026-08-26T21:11:21Z  yXXpP6ABTuYPniJnyFTF  grafana-logs  mimir  apac0       -            host1  ERROR  http.server  OPTIONS /api/v1/auth/logout - 422 - 0ms  log   -
2026-08-26T21:11:21Z  ynXpP6ABTuYPniJnyFTF  grafana-logs  faro   apac0       -            host4  ERROR  metrics      Rate limit exceeded                      log   user_1403
2026-08-26T21:11:21Z  y3XpP6ABTuYPniJnyFTF  grafana-logs  mimir  eu0         1245         host2  ERROR  metrics      Email delivery failed                    log   -

$ gcx datasources opensearch metrics -d opensearch-ds-m --since 10m
TIME                  VALUE  SERIES
2026-08-26T21:01:30Z  40
2026-08-26T21:01:40Z  80
2026-08-26T21:01:50Z  80
2026-08-26T21:02:00Z  80

$ gcx datasources opensearch metrics -d opensearch-ds-m 'level:error' --group-by app.keyword --group-size 3 --since 10m
# stderr:
⚠ showing the top 3 groups by count; more groups match — raise --group-size to see more
# stdout (6 distinct apps matched — beyla, faro, grafana, loki, mimir, tempo —
# only the top 3 by count are kept and disclosed):
TIME                  VALUE  SERIES
2026-08-26T21:01:40Z  2      faro
2026-08-26T21:02:00Z  1      faro
2026-08-26T21:02:20Z  1      faro
2026-08-26T21:02:50Z  2      faro

$ gcx datasources opensearch metrics -d opensearch-ds-m --since 15m --step 1m
TIME                  VALUE  SERIES
2026-08-26T21:00:00Z  40
2026-08-26T21:01:00Z  480
2026-08-26T21:02:00Z  480
...
2026-08-26T21:15:00Z  440

$ gcx datasources opensearch metrics -d opensearch-ds-m --since 15m --step 5m
TIME                  VALUE  SERIES
2026-08-26T21:00:00Z  1960
2026-08-26T21:05:00Z  2400
2026-08-26T21:10:00Z  2400
2026-08-26T21:15:00Z  440
# --step controls the date-histogram bucket width for `metrics` (1m -> 16 buckets, 5m -> 4 buckets,
# same 15-minute range). It's registered on `query` too but is a no-op there — confirmed live,
# `--step 1m` vs `--step 10m` produce byte-identical document/log search output, since there's no
# histogram for it to size. Inherited from Elasticsearch's shared query-model assembly, not new here.

$ gcx datasources opensearch metrics -d opensearch-ds-m --agg avg --field duration_ms --since 10m
TIME                  VALUE               SERIES
2026-08-26T21:01:30Z  3050.3333333333335
2026-08-26T21:01:40Z  2661.125
2026-08-26T21:01:50Z  1909.7272727272727

$ gcx datasources opensearch query -d opensearch-ds-m 'level:error' --since 1h --share-link
🛈 Explore link: https://datasourcedev.grafana-dev.net/explore?panes=%7B%22gcx%22%3A%7B%22datasource%22%3A%22opensearch-ds-m%22%2C%22queries%22%3A%5B%7B...

$ gcx datasources query opensearch-ds-m 'anything'
{"error":{"summary":"OpenSearch queries are structured (Lucene query, mode (documents/logs), or metric aggregation); the generic `gcx datasources query <uid> <expr>` form can't carry them — use `gcx datasources opensearch query --mode documents|logs ...` instead"}}

$ gcx datasources opensearch query -d opensearch-ds-m 'level:error' --since 1h --limit 5000
{"type":"gcx.error","schema_version":"1","error":{"summary":"--limit must be between 1 and 1000, got 5000","exitCode":1}}

$ gcx datasources opensearch query -d opensearch-ds-m 'level:error' --since 1h --limit -5
{"type":"gcx.error","schema_version":"1","error":{"summary":"--limit must be between 1 and 1000, got -5","exitCode":1}}

$ gcx datasources opensearch metrics -d opensearch-ds-m --since 1h --group-size -5
{"type":"gcx.error","schema_version":"1","error":{"summary":"--group-size must be at least 1, got -5","exitCode":1}}
# all three rejected before any datasource I/O — no network round-trip, no partial output

$ gcx datasources opensearch metrics -d opensearch-ds-m --since 10m -o graph
  │
  │⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀
  │
80└─────────────────────────────────────────────────────────────────────────
  17:35:10    17:36:46    17:38:23    17:40:00    17:41:37    17:43:14
# confirms the codec-wiring risk the plan called out up front: *opensearch.MetricsResponse
# needed its own table/wide/graph arms (FromOpenSearchResponse) — this renders correctly

$ gcx datasources opensearch metrics -d opensearch-ds-m --agg percentiles --since 10m
{"type":"gcx.error","schema_version":"1","error":{"summary":"Invalid --agg \"percentiles\" (supported: avg, cardinality, count, max, min, sum)","exitCode":1}}

$ gcx datasources opensearch metrics -d opensearch-ds-m --agg avg --since 10m
{"type":"gcx.error","schema_version":"1","error":{"summary":"--field is required for --agg avg","exitCode":1}}

$ gcx datasources opensearch list-fields -d opensearch-ds-m --index ""
{"type":"gcx.error","schema_version":"1","error":{"summary":"--index must not be empty","exitCode":1}}

$ gcx datasources opensearch query -d does-not-exist --since 1h
{"type":"gcx.error","schema_version":"1","error":{"summary":"Datasource \"does-not-exist\" not found","exitCode":1,"details":"failed to get datasource \"does-not-exist\"\n\nData source not found","suggestions":["List available datasources: gcx datasources list"]}}

$ gcx datasources opensearch query -d opensearch-ds-m 'level:this_never_matches_anything_xyz' --since 1h --limit 3
No data
# zero matches, well under --limit — no false truncation warning on stderr
```

## Testing

- [x] Unit tests for the query client (request/response wire-shape pinning via `httptest`, `ParseMapping` flattening, `ValidateAgg`, `TruncateRows`/`TruncateSeries` boundary cases), the CLI commands (validation-before-I/O, Explore URL model-matching), and the graph codec adapter.
- [x] Mutation-tested the two plugin-divergence fixes, the metrics validation coverage, and both truncation boundary checks — reverted each, confirmed the corresponding test goes red, restored.
- [x] `GCX_AGENT_MODE=false mise run all` passes (lint 0 issues, tests, build, docs).
- [x] **Verified live** against `opensearch-ds-m` on the dev stack — all four commands, both `query` modes, all three `metrics` shapes (`count`, `--group-by`, `--agg avg`), `--share-link`, and the generic-dispatcher redirect (see Sample Output above).

Part of the datasource-coverage effort (7 datasources shipped this cycle: postgres, mysql, azuremonitor, elasticsearch, cloudmonitoring, mssql, bigquery; opensearch is the 8th).


awsome-o (commented):
Went through this against the compliance hierarchy and the self-review triggers. Suite and lint are green on the head, and I mutation-tested a few of the claims rather than eyeballing them.

Nothing I found blocks. No regression to a shipped surface, no CONSTITUTION or DESIGN violation, and the registration side is complete: provider, `NormalizeKind`, both routing tables, `output_classes.json`, regenerated reference. All four commands are correctly `finite`.

Two things I want to say are fine, since they look wrong at first glance:

* I thought `query` was pinning the range filter to `@timestamp`, since it has no `--time-field` and `queryModel` defaults one. It isn't. The plugin reads `GetConfiguredFields().TimeField` off the datasource settings for both `AddDateRangeFilter` and the `date_histogram` fallback, so the model's `timeField` never gets looked at and your comment at `client.go:118` is right.
* Truncation showing up only on stderr is what bigquery, mysql and postgres already do, so I'm not counting that against this.

The two test gaps below are the only ones I'd want dealt with before this goes in, and both are small. The rest you can take or leave.

internal/datasources/opensearch/metrics.go
```diff
@@ -0,0 +1,140 @@
...
+	flags.StringVar(&opts.TimeField, "time-field", "", "Time field for the date histogram bucket (defaults to the datasource's configured time field if omitted)")
+}
+
+func (opts *metricsOpts) Validate() error {
+	if err := opts.SharedOpts.Validate(); err != nil {
+		return err
+	}
+	if opts.GroupSize < 1 {
+		return fmt.Errorf("--group-size must be at least 1, got %d", opts.GroupSize)
+	}
```
awsome-o: **Worth capping.** `--limit` is bounded at 1000 but `--group-size` only has a floor, so `--group-size 100000` passes validation and goes out as `terms.settings.size: "100001"`. OpenSearch's default `search.max_buckets` is 65536, so that either comes back as `too_many_buckets_exception` or runs a genuinely expensive aggregation.

What makes me raise it here rather than on elasticsearch, which has the identical check, is the new warning on line 115: it tells people to raise `--group-size` and doesn't say how far they can go. Could it get a ceiling the way `--limit` has one (and get named in the warning)? Needs a `maxGroupSize` const alongside `maxLimit`.

```suggestion
	if opts.GroupSize < 1 || opts.GroupSize > maxGroupSize {
		return fmt.Errorf("--group-size must be between 1 and %d, got %d", maxGroupSize, opts.GroupSize)
	}
```

internal/datasources/opensearch/query.go
```diff
@@ -0,0 +1,167 @@
...
+	search := resolved.Client.Search
+	explore := QueryExploreURL
+	exploreSubject := "query"
+	if opts.Mode == modeLogs {
+		search = resolved.Client.Logs
+		explore = LogsExploreURL
+		exploreSubject = "logs query"
+	}
+
+	resp, err := search(cmd.Context(), resolved.DatasourceUID, sentinelReq)
```
awsome-o, line 158: **Test gap.** I swapped `sentinelReq` for `req` on this line and ran the full suite: green. Same swap at `metrics.go:110`: green. And the other direction, passing `sentinelReq` into the explore builder on line 157 so every shared link leaks the +1: also green.

The commit message is straight about this half being live-verified rather than unit-tested, so nothing's oversold, but it does mean neither mistake would be caught again. `TruncateRows` on its own doesn't reach it, since the wiring is what decides whether it ever sees an extra row. Would a command-level test asserting the wire request carries `limit+1` while the Explore URL carries `limit` be worth adding? Seems like the one thing that would pin both halves.

internal/datasources/query/codecs.go
```diff
@@ -69,6 +70,8 @@ func (c *queryTableCodec) Encode(w io.Writer, data any) error {
 		return cloudmonitoring.FormatTable(w, resp)
 	case *elasticsearch.MetricsResponse:
 		return elasticsearch.FormatMetricsTable(w, resp)
+	case *opensearch.MetricsResponse:
```
awsome-o, line 73: **Test gap.** Deleted this arm, ran `go test ./internal/datasources/query/...`, green, while `gcx datasources opensearch metrics` on the default `-o table` would fail at encode time.

The commit message calls this "the exact failure mode tests can't catch, since Go's type switch has no way to assert a missing arm", but `TestQueryCodecsElasticsearchMetrics` at `codecs_test.go:101` is doing that for the sibling type today. Dropping `*opensearch.MetricsResponse` into that test looks like a couple of lines, and it would cover the wide and graph arms at 122 and 180 too.

internal/datasources/opensearch/query.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return fmt.Errorf("query failed: %w", err)
+	}
+	if opensearch.TruncateRows(resp, opts.Limit) {
+		cmdio.Warning(cmd.ErrOrStderr(), "showing the first %d rows; more rows match — raise --limit (max %d) to see more", opts.Limit, maxLimit)
```
awsome-o, line 163: **Follow-up.** Good fix, and I'd like it on elasticsearch as well. `internal/datasources/elasticsearch` has no `Warning` or `Truncate` anywhere, and its `--limit` help string is byte-identical to this one, so the same flag documented the same way silently caps on one command and discloses on the other. Anyone who learns the behaviour from opensearch gets it wrong on elasticsearch.

You call it out in the description, so I'm reading it as deliberate scope rather than an oversight. Can we get an issue filed for it so it doesn't get lost?

internal/query/opensearch/client.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// SearchQueryModel returns the OpenSearch plugin query model for a document
+// search (the raw_data metric type).
+//
+// Client.Search and the Explore link builder both call this, so the query the
+// CLI sends and the query Explore opens cannot drift apart. The time range is
+// not part of the model: the request body carries it for the client, and the
+// pane range carries it for Explore.
+func SearchQueryModel(dsUID string, req SearchRequest) map[string]any {
```
awsome-o, line 61: **Direction, not a blocker.** Normalizing the opensearch/elasticsearch names away, `list_fields.go`, `list_indices.go`, `mapping_list.go`, `run.go` and `formatter.go` come out identical, and `client.go`, `types.go`, `query.go`, `metrics.go`, `explore.go` differ by 53, 29, 22, 16 and 4 lines. Not asking for a rewrite: per-kind packages are the house pattern, and you're using the shared infra properly throughout (`grafanaquery`, `dataframe`, `querysql`, `dsquery`).

But the SQL dialects have `internal/query/sql` for exactly this, and the Lucene family now has two members and nothing equivalent. The truncation comment above is the first bill for that, since the fix has to be written twice. Is a shared Lucene layer something you'd pick up as a follow-up, or do you expect the two plugins to drift far enough apart that it wouldn't hold?

internal/datasources/opensearch/query.go
```diff
@@ -0,0 +1,167 @@
...
+type queryOpts struct {
+	dsquery.SharedOpts
+
+	Datasource string
+	Mode       string
+	Limit      int
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	opts.Setup(flags, false)
```
awsome-o, line 38: **Nitpicks**, three of them, none blocking.

`opts.Setup(flags, false)` registers `--step` (that bool is `enableGraph`, not step), and you've confirmed live that it does nothing on `query`. Its help still reads "Query step (e.g., '15s', '1m')" and the generated reference lists it as a real flag, so an agent will reach for it to control resolution. Looks like a `SharedOpts` thing rather than yours to fix here, but flagging it.

`url.PathEscape(index)` at `internal/query/opensearch/client.go:308` percent-encodes `*` and `,`, which are the two characters that make an index pattern a pattern, and both `--index` help strings advertise "index or index pattern". The live run used a literal index name. Not sure whether Grafana's resource proxy hands the plugin the decoded segment or the raw one, so it may well be fine, but one probe with `--index 'grafana-*'` would settle it.

`TestListIndicesCmd_HasIndexFlag` only asserts the flag exists with an empty shorthand and default. If `opts.Index` stopped reaching `client.Mapping` it would still pass.

internal/datasources/opensearch/metrics.go
ktw4071: Applied and completed — your suggested diff referenced `maxGroupSize`, which didn't
exist yet, so the package didn't build after accepting it. Added the const (1000,
same margin-under-max_buckets reasoning as `--limit`'s cap), updated `--group-size`'s
help text to show the range, named the ceiling in the truncation warning ("raise
--group-size (max 1000)"), and added test coverage for the new upper bound. Live-
verified `--group-size 100000` now rejects with the updated message.

internal/datasources/opensearch/query.go
ktw4071: Added. Split `runQuery` into `runQuery` (config/datasource resolution) + a new
`executeQuery(cmd, opts, resolved, share)` taking an already-resolved query, so a
test can drive the sentinel wiring directly against a fake HTTP server without
faking config loading and datasource resolution too (nothing else in the repo has
that end-to-end scaffolding, so I didn't want to invent it just for this).

New `TestExecuteQuery_SentinelWiring` in query_internal_test.go builds a
`resolvedQuery` by hand against an httptest server and asserts both destinations in
one test. Mutation-tested exactly the two swaps you described by hand: `sentinelReq`
→ `req` at the client call fails on the captured size (expected 6, got 5); the
reverse at the Explore URL fails there instead (expected 5, got 6).

internal/datasources/query/codecs.go
ktw4071: Added `TestQueryCodecsOpenSearchMetrics` in codecs_test.go, covering table, wide,
and graph. Mutation-tested by deleting the wide-codec arm: failed with
"invalid data type for query wide codec", exactly the failure mode you called out
as untestable by the type switch itself.

internal/datasources/opensearch/query.go
ktw4071: Agreed, this is deliberate scope, not an oversight — opensearch got the fix first
since it's the newer sibling. Filed #1259 to track porting the sentinel/disclosure
pattern onto elasticsearch so it doesn't get lost.

internal/query/opensearch/client.go
ktw4071: Good question — I'd lean toward picking this up, but only once there's a concrete
second bill beyond the truncation fix, rather than speculatively now with just two
members. Filed #1260 as an open design question rather than a task, so it's tracked
without committing to it yet. Curious if you have a threshold in mind (third
Lucene-family datasource? N duplicated fixes?) for when it'd clearly be worth it.

internal/datasources/opensearch/query.go
ktw4071: Taking these one at a time:

- `--step` help text: agreed this is `SharedOpts`, shared by every datasource kind
  that embeds it — out of scope for this PR. Leaving as-is unless you'd rather see
  a separate follow-up filed for it too.
- `--index` percent-encoding: live-tested against opensearch-ds-m with both
  `--index 'grafana-*'` and `--index 'grafana-logs,grafana-metrics'` — both return
  correct, complete results. `url.PathEscape` does encode `*`/`,`, but Grafana's
  resource proxy decodes the path segment before forwarding to the plugin, so this
  isn't a live bug. No code change.
- Weak `TestListIndicesCmd_HasIndexFlag`: extracted `fetchMappingResult` out of
  `newMappingListCmd`'s RunE and added `TestFetchMappingResult_IndexReachesClient`,
  which captures the actual outgoing `_mapping` request path against a fake server.
  Mutation-tested by hardcoding the index to `""` in the call site — failed exactly
  the way you described ("opts.Index stopped reaching client.Mapping").

internal/datasources/opensearch/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+				Start:     resolved.Start,
+				End:       resolved.End,
+				StepMs:    resolved.StepMs,
+			}
+			sentinelReq := req
+			if opts.GroupBy != "" {
+				sentinelReq.GroupSize = opts.GroupSize + 1
+			}
+
+			resp, err := resolved.Client.Aggregations(cmd.Context(), resolved.DatasourceUID, sentinelReq)
```
awsome-o: The sentinel test you added covers `query.go`, but this is the second site that comment pointed at and it's still unpinned. I swapped `sentinelReq` back to `req` here, and separately passed `sentinelReq` into `MetricsExploreURL` on line 128, and both times `internal/datasources/opensearch` and `internal/query/opensearch` stayed green. The first one means `--group-size 10` quietly returns 10 groups and never warns when an 11th exists; the second puts group size plus one into every shared Explore link.

Looks like a copy of `TestExecuteQuery_SentinelWiring` aimed at the metrics path, 30 lines or so? The wiring is correct as written, so I'm fine with it landing after this merges if you'd rather not reopen the PR.

internal/datasources/opensearch/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+			if err != nil {
+				return fmt.Errorf("query failed: %w", err)
+			}
+			if opts.GroupBy != "" && opensearch.TruncateSeries(resp, opts.GroupSize) {
+				cmdio.Warning(cmd.ErrOrStderr(), "showing the top %d groups by count; more groups match — raise --group-size (max %d) to see more", opts.GroupSize, maxGroupSize)
```
awsome-o: Someone who hits this warning and follows it up to `--group-size 1000` gets a `too_many_buckets_exception`, which is the failure the cap at line 22 says it's there to prevent. The total is terms times date_histogram buckets rather than terms alone, and `intervalMsFor` aims at roughly 1000 histogram buckets, so `--group-size 100` is already near 100k against a default `search.max_buckets` of 65536.

Would it read better to drop the ceiling from this string and just say more groups match? The cap still does its job as a sanity bound, without naming a number a default cluster won't reach. I haven't tried this against a cluster with `search.max_buckets` raised, so say if it behaves better in practice than the arithmetic suggests.

internal/query/opensearch/client.go
```diff
@@ -0,0 +1,364 @@
...
+	if bucketAggs == nil {
+		bucketAggs = []any{}
+	}
+	return map[string]any{
+		"refId":      "A",
+		"datasource": map[string]any{"type": pluginID, "uid": dsUID},
+		"query":      query,
+		"metrics":    []any{metric},
+		"bucketAggs": bucketAggs,
+		"timeField":  orDefault(timeField, DefaultTimeField),
```
awsome-o: This defaults the model's `timeField` to `@timestamp`, but the comment at line 118 argues the opposite for the bucket agg: an unset time field has to reach the plugin empty so it uses the datasource's own configured field. Both can't be right, and I can't tell from the code which one the plugin actually reads.

If it reads this root field, then against a datasource configured on something like `event_time`, `gcx datasources opensearch query` comes back empty with no way out, since `--time-field` only exists on `metrics`. If it ignores the root field, this is dead weight and the line 118 comment is the accurate one. Do you know which way it goes? I don't have a cluster on a non default time field to try it against.

awsome-o (dismissed):
Approving. Gates come back clean for me too on 3e43290b (build, `go test -race ./...`, `mise run lint`, `mise run reference-drift`), and the threads from last round hold up when I check them against the code rather than the replies.

Three comments below, none of them blocking, and one is the other half of the sentinel thread. The `--group-size` one is the only thing a user would actually notice.

The cap probably wants porting to elasticsearch at some point, otherwise the same flag accepts different ranges on the two commands. That reads like #1259 scope rather than a reason to hold this.

internal/datasources/opensearch/metrics.go
ktw4071: Added — copy of TestExecuteQuery_SentinelWiring aimed at the metrics path, as
suggested. Split RunE into RunE (config/datasource resolution) + a new
executeMetrics(cmd, opts, resolved, share), mirroring query.go's split.
TestExecuteMetrics_SentinelWiring builds a resolvedQuery by hand and asserts
both destinations. Mutation-tested both swaps by hand exactly as you described
them — both fail as expected.

Extracted the shared httptest server scaffolding between this test and
TestExecuteQuery_SentinelWiring into newSentinelCaptureServer to keep dupl
happy, since the two now only differ in which field of the query they dig the
sentinel value out of.

internal/datasources/opensearch/metrics.go
ktw4071: Reworded to drop the ceiling, per your suggestion.

```
$ gcx datasources opensearch metrics -d opensearch-ds-m 'level:error' --group-by app.keyword --group-size 3 --since 10m
⚠ showing the top 3 groups by count; more groups match — raise --group-size to see more
```

Tried to reproduce the failure live first, against a genuinely high-cardinality
field over a wide range:

```
$ gcx datasources opensearch metrics -d opensearch-ds-m --group-by user_id.keyword --group-size 1000 --since 30d
⚠ showing the top 1000 groups by count; more groups match — raise --group-size to see more
# succeeded, no too_many_buckets_exception — 1000 series returned, each with
# only 2-3 populated histogram buckets (min/max/avg: 2, 3, 2.98)
```

No error — min_doc_count:1 only creates buckets with actual matching
documents, and this dataset's high-cardinality fields are also sparse (2-3
populated histogram buckets per group), so it never hit the multiplicative
worst case. So this is fixing the misleading claim on your math rather than a
reproduced live failure — the arithmetic still holds for denser real-world
dimensions (many concurrently-active hosts, say), I just couldn't construct
that shape against this sandbox dataset.

internal/query/opensearch/client.go
ktw4071: grafana/opensearch-datasource's pkg/opensearch/query_request.go:80,
verbatim: https://github.com/grafana/opensearch-datasource/blob/e77d88f5a7759dc4ff9ef7116b5b5192d266c7fd/pkg/opensearch/query_request.go#L80

    // we had a string-field named `timeField` in the past. we do not use it anymore.
    // please do not create a new field with that name, to avoid potential problems with old, persisted queries.

Cross-checked lucene_handler.go: AddDateRangeFilter and the date_histogram
fallback both pull exclusively from client.GetConfiguredFields().TimeField
(the datasource's own setting) or bucketAgg.Field — never from the request
body's top-level timeField.

Live-verified too, via `gcx api /api/ds/query` against opensearch-ds-m — same
request body, only the field under test changed:

```
# A: top-level "timeField": "@timestamp" (gcx's real default)
$ gcx api /api/ds/query -X POST -d @query.json
frames: 1 | docs: 5 | first ts: 1788903469000

# B: identical request, "timeField": "this_field_does_not_exist_anywhere"
$ gcx api /api/ds/query -X POST -d @query-bogus-timefield.json
frames: 1 | docs: 5 | first ts: 1788903469000   <- byte-identical to A
```

Contrast check — same swap, but on bucketAggs' date_histogram field instead
(the one AggsQueryModel deliberately leaves empty):

```
# C: bucketAggs[0].field: "" (gcx's real behavior — empty, correct fallback)
$ gcx api /api/ds/query -X POST -d @aggs.json
frames: 1 | histogram buckets: 61

# D: identical request, bucketAggs[0].field: "this_field_does_not_exist_anywhere"
$ gcx api /api/ds/query -X POST -d @aggs-bogus-field.json
frames: 1 | histogram buckets: 0   <- breaks completely
```

So client.go:118's comment is correct, client.go:149's default is genuinely
inert everywhere, and there's no correctness risk from query/logs lacking a
--time-field flag. Added a comment at 149 citing this so it doesn't get
re-litigated a third time.

awsome-o (approved):

PR #1255, opened 2026-08-27
feat(datasources): add StarTree Pinot query command

yduartep:
## Summary
- Adds `gcx datasources pinot query` for StarTree PinotQL (`startree-pinot-datasource`) through Grafana's unified query API.
- Same shape as ClickHouse (`-d`, `--since`/`--from`/`--to`, `--limit` with a stderr cap warning, `--share-link` / `--open`), but a custom request body: `queryType: PinotQL`, `editorMode: Code`, `displayType: TABLE`, `pinotQlCode`, and `tableName`.
- StarTree requires `tableName` (schema + macros) before it runs `pinotQlCode`. gcx derives it from the first confident `FROM`, including inside a subquery (`SELECT … FROM (SELECT col FROM events)` → `events`). `--table` overrides that, and is required when the SQL names no table (`SELECT 1`). The same name is used in the Explore URL.
- Registers the kind as `pinot` so the generic command can auto-detect it. The generic form takes the UID as a positional argument (no `-d`) and also accepts `--table`.

**Out of scope:** `list-tables` and `describe-table` (schema discovery). Other SQL datasources ship those so callers can find table and column names before writing a query. Pinot is query-only in this PR; those commands will land in a follow-up.

## How to test
Build the branch (`mise run build` or `go build -o bin/gcx ./cmd/gcx`) against a logged-in stack that has a StarTree Pinot datasource.

Typed command (`-d` is the datasource UID):

```bash
bin/gcx datasources pinot query -d UID --since 7d \
  'SELECT count(*) FROM events LIMIT 10'
```

Subquery (no `--table`; Pinot nested queries need the multistage engine):

```bash
bin/gcx datasources pinot query -d UID --since 1h --limit 0 \
  'SET useMultistageEngine = true;
   SELECT count(*) FROM (SELECT col FROM events) sub'
```

SQL with no extractable table fails in gcx:

```bash
bin/gcx datasources pinot query -d UID 'SELECT 1'
```

`--table` is the override:

```bash
bin/gcx datasources pinot query -d UID --table events 'SELECT 1'
```

Generic command (UID is positional; `-d` is not a flag here):

```bash
bin/gcx datasources query UID --since 7d \
  'SELECT count(*) FROM events LIMIT 10'
```

- A successful query prints a table (try `-o json` for the raw payload).
- `--share-link` is only on the typed command; it prints an Explore URL on stderr. Opening it should show the same SQL and the same rows.
- A query the plugin cannot decode (for example `SELECT *` on a table with `STRING_ARRAY` columns) should return the plugin error, not a gcx crash.


yduartep:
@claude review

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+// the statement. Appending "LIMIT n" as a bare suffix after one would land
+// inside the comment and be dropped, silently leaving the query unbounded.
+var trailingLineCommentRe = regexp.MustCompile(`--[^\n]*$`)
+
+func selectBody(sql string) string {
+	return leadingSetRe.ReplaceAllString(sql, "")
+}
+
+func bail(sql string) bool {
+	return limitBailRe.MatchString(sql) || trailingLineCommentRe.MatchString(strings.TrimRight(sql, "; \t\n"))
```
aranhave: `--limit` is silently bypassed for `UNION` and `OFFSET` queries here. For example, `--limit 100` leaves `LIMIT 5000 OFFSET 0` unchanged with no warning, even though the flag describes a maximum. Can we enforce the bound for these forms or return a clear warning/error?

awsome-o (commented):
Read this against `clickhouse` and `postgres`, since it borrows from both. Build, lint and `go test ./...` are all green on b0139e91, so everything below is behavior rather than a broken gate.

The two helpers in `internal/query/pinot/types.go` are where I'd want another pass: the appended LIMIT breaks `LIMIT 10, 20` and a trailing `OPTION(...)`, enforcement silently drops out when a literal contains `delete` or `--`, and `ExtractTableName` returns the wrong table often enough to matter now that it feeds `--share-link`. The rest are small.

One process note: `docs/architecture/project-structure.md` lists every sibling in both trees (lines 96 and 117) and the two new packages aren't there. CLAUDE.md's PR checklist item 4 asks for that one.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+// limitStatementRe matches statements that can legally take a trailing LIMIT
+// clause. Pinot only allows LIMIT on SELECT-shaped statements.
+var limitStatementRe = regexp.MustCompile(`(?is)^\s*(SELECT|WITH)\b`)
+
+// UNION bails so a trailing LIMIT on a multi-leg query is left alone — Faro
+// user-journey dumps already carry LIMIT 2000 and must not be capped silently.
+// OFFSET / LIMIT OFFSET would double up into a syntax error. DML keywords
+// anywhere fail safe (no LIMIT added) rather than corrupting a write. A
+// SELECT mentioning those words in a string literal also skips enforcement.
+var limitBailRe = regexp.MustCompile(`(?i)(\bUNION\b|\bLIMIT\s+\d+\s+OFFSET\b|\bOFFSET\s+\d+\b|\b(INSERT|UPDATE|DELETE|MERGE)\b)`)
```
awsome-o: Someone paginating with `--limit 100` gets a parse error back from the plugin. `EnforceLimit("SELECT * FROM t LIMIT 10, 20", 100, 1000)` returns `SELECT * FROM t LIMIT 10, 20 LIMIT 100` with capped=false: `limitClauseRe` in `internal/query/sql/types.go:29` wants `LIMIT <digits>` at the very end, so the `, 20` stops it matching, and the bail list here only guards `LIMIT n OFFSET`. ClickHouse carries `\bLIMIT\s+\d+\s*,` in its own bail regex at `internal/query/clickhouse/types.go:31`, so it looks like that alternative got dropped in the copy.

Same shape with a trailing `OPTION(...)`: `SELECT * FROM t OPTION(timeoutMs=5000)` comes back as `... OPTION(timeoutMs=5000) LIMIT 100`, and Pinot wants OPTION as the last element of the statement. `SELECT * FROM t LIMIT 5000 OPTION(timeoutMs=5000)` is the worst of the pair, the 5000 is neither capped nor reported and a second LIMIT still gets grafted on. I drove these through the real helper rather than a live StarTree, so tell me if the plugin is more forgiving than I'm assuming.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+// trailingLineCommentRe matches a `--` line comment that runs to the end of
+// the statement. Appending "LIMIT n" as a bare suffix after one would land
+// inside the comment and be dropped, silently leaving the query unbounded.
+var trailingLineCommentRe = regexp.MustCompile(`--[^\n]*$`)
+
+func selectBody(sql string) string {
+	return leadingSetRe.ReplaceAllString(sql, "")
+}
+
+func bail(sql string) bool {
```
awsome-o: A DML word or `--` inside a string literal skips enforcement altogether, so no LIMIT is emitted at all. `SELECT * FROM t WHERE action = 'delete'` and `SELECT '--' FROM t` both come back unchanged with capped=false, which also means the warning at `internal/datasources/pinot/query.go:97` never fires. Pinot's default for a selection query with no LIMIT is 10 rows as far as I know, so someone who passed `--limit 100` reads 10 rows as the whole answer.

The comment above frames this as failing safe, but for a SELECT the safe direction looks like the other one: emit the cap, since nothing reachable here can turn a SELECT into a write. The flag help at `query.go:31` has the matching gap, it promises capping at 1000 with a warning and never says that UNION, OFFSET and these literal cases opt out entirely. A UNION query carrying `LIMIT 2000` returns 2000 rows while the help still claims a 1000 ceiling. That's the T3 honesty rule in `integrate-with-gcx/references/self-review.md`, partiality has to be disclosed somewhere the caller sees it.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+	if !limitStatementRe.MatchString(selectBody(sql)) {
+		return sql, false
+	}
+	return querysql.EnforceLimit(sql, limit, maxLimit, bail)
+}
+
+// fromTableRe captures the first unquoted or double-quoted table after FROM.
+// Subqueries (`FROM (SELECT ...)`) do not match. Used only to fill StarTree's
+// tableName editor field; Pinot executes pinotQlCode regardless.
+var fromTableRe = regexp.MustCompile(`(?is)\bFROM\s+"?([a-zA-Z_][a-zA-Z0-9_.]*)"?`)
```
awsome-o: This returns the wrong table for several ordinary queries, and the value isn't cosmetic any more: `client.go:57` puts it in the request body sent to the plugin, and `explore.go:22` puts it in the share link, so `--share-link` can hand someone a URL opening on a table their query never touched.

Against the real helper: `SELECT EXTRACT(YEAR FROM ts) FROM events` gives `ts`, `SELECT * FROM (SELECT x FROM inner_t) a` gives `inner_t`, `SELECT 'FROM x' AS a FROM events` gives `x`, and `SELECT * FROM "my-table"` gives `my` because the character class stops at the hyphen and the closing quote is never matched. The second one also contradicts line 65: subqueries do match, the regex just walks past the outer `FROM (` to the inner one. Would it be happier matching only the shapes you can be confident about and sending an empty `tableName` otherwise, given Pinot executes `pinotQlCode` regardless? Separately, `(?s)` is inert on this pattern and on `leadingSetRe` and `limitStatementRe`, none of the three contains a `.` metacharacter.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestExtractTableName(t *testing.T) {
+	tests := []struct {
+		name string
+		sql  string
+		want string
+	}{
+		{"simple from", `SELECT count(*) FROM faro_pinot_events_v2`, "faro_pinot_events_v2"},
+		{"quoted from", `SELECT 1 FROM "faro_pinot_events_v2"`, "faro_pinot_events_v2"},
+		{"set prefix", "SET useMultistageEngine = true;\nSELECT * FROM faro_pinot_logs_v1", "faro_pinot_logs_v1"},
+		{"subquery from skipped", "SELECT * FROM (SELECT 1)", ""},
```
awsome-o: These two pass for reasons unrelated to what their names claim. `SELECT * FROM (SELECT 1)` has no inner FROM, so it can't tell "skips subqueries" apart from "falls through to the inner table", and the real subquery case does return the inner table. `bail on EXPLAIN` at line 52 passes because `limitStatementRe` requires `^\s*(SELECT|WITH)`, not because EXPLAIN is in `limitBailRe`; loosen that allow list later and the case still goes green while its name promises a guard that was never there.

Worth swapping in cases that would actually go red under the defect? `SELECT * FROM (SELECT x FROM inner_t) a` covers the first one.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+	querysql "github.com/grafana/gcx/internal/query/sql"
+)
+
+const (
+	// DatasourceType is the Grafana plugin ID for StarTree Pinot datasources.
+	DatasourceType = "startree-pinot-datasource"
+)
+
+// EscapeSQLString escapes single quotes for use in SQL string literals.
+func EscapeSQLString(s string) string {
```
awsome-o: Nothing calls this outside `types_test.go:11`. Every other dialect's copy has a real caller in its schema discovery commands (`internal/datasources/clickhouse/list_tables.go:76`, `postgres/describe_table.go:108`, `mysql/list_tables.go:45`), and this PR ships no `list-tables` or `describe-table`, so no identifier ever gets interpolated here.

The test is what would keep it alive, since it makes the helper look covered. I'd drop both, unless the schema commands are already queued behind this?

internal/datasources/pinot/explore.go
```diff
@@ -0,0 +1,28 @@
+package pinot
+
+import (
+	"strings"
+
+	dsquery "github.com/grafana/gcx/internal/datasources/query"
+)
+
+// QueryExploreURL builds a Grafana Explore URL for a StarTree PinotQL query.
+func QueryExploreURL(host string, query dsquery.ExploreQuery, tableName string) string {
```
awsome-o: Every sibling is `QueryExploreURL(host string, query dsquery.ExploreQuery) string` (`clickhouse/explore.go:10`, `bigquery/explore.go:11`, `athena/explore.go:10`). The third parameter here means the caller at `query.go:126` runs `pinot.ExtractTableName(sql)` a second time, since `client.Query` already derives the same value from the same `sql` at `internal/query/pinot/client.go:57`. Two derivations of one value from one input, with nothing keeping them in step.

Deriving it inside `QueryExploreURL` from `query.Expr` puts the signature back in the family and drops the duplicate. Related: the doc comment at `client.go:33` says `tableName` is copied "so Explore can open", but that field goes into the plugin request body; the Explore link computes its own separately, so a reader trusting the comment concludes the field is cosmetic.

internal/datasources/pinot/query.go
```diff
@@ -0,0 +1,155 @@
...
+	share := &dsquery.ExploreLinkOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "query [EXPR]",
+		Short: "Execute a PinotQL query against a StarTree Pinot datasource",
+		Long: `Execute a PinotQL query against a StarTree Pinot datasource.
+
+EXPR is the SQL query to execute, passed as a positional argument or via --expr.
+Datasource is resolved from -d flag or datasources.pinot in your context.
+Server-side macros ($__timeFilter, $__timeGroup, etc.) are supported.
```
awsome-o, line 51: Is this line true for StarTree? It's verbatim from `internal/datasources/postgres/query.go:52`, and postgres gets those macros from the sqlds engine, which reads `rawSql` and `format`. The body this client builds at `internal/query/pinot/client.go:49` sends `queryType`/`editorMode`/`displayType`/`tableName`/`pinotQlCode` and no `rawSql` at all, and `$__timeGroup` in particular is an sqlds macro rather than a Pinot one.

Nothing in the examples or the tests exercises a macro, so I couldn't settle it from the branch; one probe against a real StarTree datasource would. This is agent facing help, so if the macros aren't expanded then an agent reading it emits `WHERE $__timeFilter(ts)` and gets a parse error, and I'd delete the line rather than soften it.

internal/datasources/pinot/query.go
```diff
@@ -0,0 +1,155 @@
...
+	Limit      int
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	opts.Setup(flags, false)
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.pinot is configured)")
+	flags.IntVar(&opts.Limit, "limit", defaultLimit, fmt.Sprintf("Max rows to return; requests above %d are capped, with a warning (0 disables enforcement)", maxLimit))
+}
+
+func (opts *queryOpts) Validate() error {
```
awsome-o, line 32: Small one: there's no `query_test.go` in this package, so nothing pins that `-d` and `--limit` are bound to the fields `Validate` reads, or that `--limit=-5` fails before any config or network call. `postgres/query_test.go`, `mysql/query_test.go` and `bigquery/query_test.go` each run that through `cmd.Execute()` on a zero value `ConfigLoader`.

ClickHouse skipped it too, so this isn't a house rule so much as a suggestion. The flag binding is just the kind of thing that would still pass every test in this PR if it were wrong.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,84 @@
...
+// query (e.g. SET useMultistageEngine = true;). Those prefixes are stripped
+// before the SELECT-shaped allow-list so EnforceLimit can still bound the
+// SELECT that follows.
+var leadingSetRe = regexp.MustCompile(`(?is)^(?:\s*SET\b[^;]*;\s*)+`)
+
+// limitStatementRe matches statements that can legally take a trailing LIMIT
+// clause. Pinot only allows LIMIT on SELECT-shaped statements.
+var limitStatementRe = regexp.MustCompile(`(?is)^\s*(SELECT|WITH)\b`)
+
+// UNION bails so a trailing LIMIT on a multi-leg query is left alone — Faro
```
awsome-o: This pins one team's present day query and one literal into the rationale for a rule that governs every Pinot query gcx will ever send. Once that dump changes its limit or goes away, the comment still reads as authoritative and the reason behind the bail has quietly gone.

The invariant underneath is that a caller's LIMIT on a statement with several legs is left intact, because an appended clause would bind to the last leg only. Stating that instead survives the Faro query changing.

awsome-o (commented):
Ran the branch: build, `go test ./...`, lint and `reference-drift` are all green.

Two things I would want sorted before this goes in. `ExtractTableName` returns the wrong table for `EXTRACT(... FROM ts)` and for a string literal holding the word from, and `limitBailRe` misses `LIMIT a, b` and `OPTION(...)`, both of which get a second `LIMIT` appended. The `EXPLAIN` and `INSERT` test cases also never reach `bail`, so the DML branch has no coverage.

The rest is cleanup and help text. Each comment says what I ran and what I could not check.

internal/query/pinot/types.go
awsome-o: `ExtractTableName` takes the first `FROM` anywhere in the statement, so `SELECT EXTRACT(YEAR FROM ts) FROM events` comes back as `ts`, and `SELECT 'from admin' AS x FROM t` comes back as `admin`. I ran both against this branch. That string is what goes into the request body at client.go:57 and into the Explore pane at explore.go:22, so `--share-link` on either query opens on the wrong table.

The comment here says Pinot executes `pinotQlCode` regardless, which I could not check against a live plugin. If that holds, the body field is harmless and only the link is wrong; if the plugin routes or validates on `tableName`, the query itself fails. Do you know which? Either way the reason given at client.go:33 ("so Explore can open") does not apply, since explore.go derives its own copy. `SELECT * FROM my_db.events` also returns `my_db.events`, because the character class allows dots.

internal/query/pinot/types.go
awsome-o: Two Pinot spellings of a limit are missing here, and appending after either one is a syntax error the caller never wrote. On this branch `EnforceLimit("SELECT * FROM t LIMIT 10, 20", 100, 1000)` gives back `SELECT * FROM t LIMIT 10, 20 LIMIT 100`, and `SELECT * FROM t OPTION(timeoutMs=5000)` gives back `SELECT * FROM t OPTION(timeoutMs=5000) LIMIT 100`.

`LIMIT a, b` is the case the raw SQL reference calls out by name (`.claude/skills/add-datasource/references/raw-sql.md`, the bail predicate table), and mysql (`\bLIMIT\s+\d+\s*,`) and clickhouse both carry it. `OPTION(...)` is the older form of what the `SET ...;` prefix does, and types.go:25 already handles the prefix. I checked the string that comes out, not that Pinot rejects it, so if its parser happens to tolerate a trailing `LIMIT` after `OPTION` then only the first case matters.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,70 @@
...
+		{"appends LIMIT when missing", "SELECT 1", 100, "SELECT 1 LIMIT 100", false},
+		{"appends LIMIT after SET prefix", "SET useMultistageEngine = true;\nSELECT 1", 100, "SET useMultistageEngine = true;\nSELECT 1 LIMIT 100", false},
+		{"keeps existing LIMIT if under max", "SELECT 1 LIMIT 50", 100, "SELECT 1 LIMIT 50", false},
+		{"caps existing LIMIT exceeding max", "SELECT 1 LIMIT 5000", 100, "SELECT 1 LIMIT 1000", true},
+		{"requested limit above max is capped and reported", "SELECT 1", 5000, "SELECT 1 LIMIT 1000", true},
+		{"limit 0 disables enforcement", "SELECT 1", 0, "SELECT 1", false},
+		{"bail on UNION", "SELECT 1 FROM a UNION SELECT 2 FROM b LIMIT 2000", 100, "SELECT 1 FROM a UNION SELECT 2 FROM b LIMIT 2000", false},
+		{"bail on SET plus UNION", "SET useMultistageEngine = true;\nSELECT 1 FROM a\nUNION ALL\nSELECT 2 FROM b LIMIT 2000", 100, "SET useMultistageEngine = true;\nSELECT 1 FROM a\nUNION ALL\nSELECT 2 FROM b LIMIT 2000", false},
+		{"bail on LIMIT OFFSET", "SELECT * FROM t LIMIT 100 OFFSET 10", 100, "SELECT * FROM t LIMIT 100 OFFSET 10", false},
+		{"bail on EXPLAIN", "EXPLAIN SELECT * FROM t", 100, "EXPLAIN SELECT * FROM t", false},
```
awsome-o: These two cases never reach `bail()`, so they pass for a different reason than their names claim. `limitStatementRe` at types.go:29 only matches statements starting with `SELECT` or `WITH`, so `EXPLAIN ...` and `INSERT ...` both return at types.go:58 before `querysql.EnforceLimit` runs. The upshot is that the `(INSERT|UPDATE|DELETE|MERGE)` branch of `limitBailRe`, the one branch with a stated safety rationale, has no coverage at all. Same for the standalone `\bOFFSET\s+\d+\b` branch, since line 51 is already caught by the `LIMIT n OFFSET` branch.

`raw-sql.md` flags this exact trap under the bail predicate section: a statement that comes back unchanged proves nothing if the allow list rejected it before `bail` ran. Going through a SELECT reaches them, `SELECT * FROM t WHERE op = 'delete'` and `SELECT * FROM t OFFSET 10` both come back unchanged via `bail` on this branch.

internal/agent/command_annotations.go
```diff
@@ -79,6 +79,9 @@ var commandAnnotations = map[string]annotation{
 	"gcx datasources clickhouse list-tables":    {Cost: "small"},
 	"gcx datasources clickhouse describe-table": {Cost: "small", Hint: "TABLE -d UID --database default -o json"},
 
+	// datasources pinot
+	"gcx datasources pinot query": {Cost: "medium", Hint: "-d UID 'SELECT count(*) FROM faro_pinot_events_v2' -o json"},
```
awsome-o: This entry cannot reach the command tree. `ApplyAnnotations` only writes a key that is missing (lines 666 and 669), and query.go:146 already sets both `AnnotationTokenCost` and `AnnotationLLMHint` inline, so the registry copy never wins. I built the branch and `gcx commands --flat -o json` emits only the inline hint.

The two strings have already drifted apart, the registry one has no command prefix. bigquery, mysql and postgres all set their annotations inline and add no registry entry; clickhouse and athena predate that. Dropping lines 82 and 83 leaves one source.

internal/query/pinot/types.go
awsome-o: `EscapeSQLString` has no caller. In the other dialects it exists because `list-tables` and `describe-table` interpolate identifiers into SQL (clickhouse/list_tables.go:76, postgres/describe_table.go:108 and so on), and this kind ships neither command, so nothing can call it. types_test.go:10 is the only reference, and it exercises `strings.ReplaceAll` and nothing else.

Drop both until the command that needs it turns up?

internal/datasources/pinot/query.go
awsome-o, line 51: Is this confirmed against the StarTree plugin? The How to test section only covers plain SELECTs and nothing in the tests touches interpolation. It matters more than the usual help text claim, because the `from`/`to` in the body (client.go:41) only do anything if the plugin interpolates them, so if it does not, the macros and the time flags are both inert and someone reading this help goes down a dead end. Happy to be wrong if you have run one.

internal/datasources/pinot/query.go
```diff
@@ -0,0 +1,155 @@
...
+		Long: `Execute a PinotQL query against a StarTree Pinot datasource.
+
+EXPR is the SQL query to execute, passed as a positional argument or via --expr.
+Datasource is resolved from -d flag or datasources.pinot in your context.
+Server-side macros ($__timeFilter, $__timeGroup, etc.) are supported.
+Use --share-link to print the equivalent Grafana Explore URL, or --open to
+open it in your browser after the query succeeds.`,
+		Example: `
+  # Simple query
+  gcx datasources pinot query -d UID 'SELECT count(*) FROM faro_pinot_events_v2'
```
awsome-o: `faro_pinot_events_v2` is a Grafana internal table, and it is the example in all five `Example` lines, in the `Long` text above, and in the llm_hint at line 148. Anyone outside Grafana who runs the first thing the help suggests gets a table that does not exist on their stack, and an agent routing on the hint does the same. `WHERE appId = 66` at line 62 has the same problem.

clickhouse uses `events` and `logs`, bigquery uses `dataset.events`, athena uses `events`. Would a neutral placeholder work here?

internal/datasources/pinot/explore.go
awsome-o: Would this be happier deriving the table itself? It is the only three argument form among the raw SQL dialects; athena, bigquery, clickhouse and prometheus are all `QueryExploreURL(host string, query dsquery.ExploreQuery) string`, and the ones that do take a third argument pass a typed request rather than a string derived from a field already inside `query`.

`tableName` is fully determined by `query.Expr`, and query.go:126 passes both views of the same input while client.go:57 derives it a third time. Calling `pinot.ExtractTableName(query.Expr)` in here drops the duplication at the call site and makes a mismatched pair impossible to construct.

internal/datasources/providers/pinot.go
```diff
@@ -0,0 +1,25 @@
...
+type pinotDSProvider struct{}
+
+func (p *pinotDSProvider) Kind() string      { return "pinot" }
+func (p *pinotDSProvider) ShortDesc() string { return "Query StarTree Pinot datasources" }
+
+func (p *pinotDSProvider) QueryCmd(loader *providers.ConfigLoader) *cobra.Command {
+	return pinot.QueryCmd(loader)
+}
+
+func (p *pinotDSProvider) ExtraCommands(_ *providers.ConfigLoader) []*cobra.Command {
```
awsome-o, line 23: Returning nil means pinot ships `query` and nothing else, so someone has to already know the table and column names before they can write one. Every other SQL dialect here ships `list-tables` and `describe-table`, and the PR description points at why that stings: `SELECT *` on a table with `STRING_ARRAY` columns comes back as a plugin error, so the usual way of finding out what is in a table does not work either.

This reads like a deliberate scope call rather than something missed, and I am fine with that. Could you say so in the description or file a follow up, so the gap is on the record?

internal/query/pinot/types.go
yduartep: Kept the bail: UNION and OFFSET SQL is still sent unchanged, because appending a second LIMIT would be a parse error (or would bind only the last UNION leg).

`--limit` now warns on stderr when that skip happens, and the flag help discloses it. `--limit 0` stays quiet. Covered in `TestWarnLimitEnforcement`.

internal/query/pinot/types.go
yduartep: `LIMIT offset,count` is now left as written (it already bounds the result; no warning). `OPTION` must stay last, so we still skip rewriting and warn on stderr — including `LIMIT 5000 OPTION(...)`, which is no longer given a second LIMIT. Covered in `TestEnforceLimit` and `TestWarnLimitEnforcement`.

yduartep: `LIMIT offset,count` is now left as written (it already bounds the result; no warning). `OPTION` must stay last, so we still skip rewriting and warn on stderr — including `LIMIT 5000 OPTION(...)`, which is no longer given a second LIMIT. Covered in `TestEnforceLimit` and `TestWarnLimitEnforcement`.

internal/query/pinot/types.go
yduartep: `dmlBailRe` is gone. Real `INSERT`/`DELETE` still never reach `bail()` (only `SELECT`/`WITH` do). `--` now has to look like a comment (`whitespace` then `--`), so `WHERE action = 'delete'` and `SELECT '--' FROM t` get `LIMIT` appended. A real trailing `--` comment still skips. Covered in `TestEnforceLimit`.

internal/query/pinot/types.go
yduartep: `ExtractTableName` no longer takes the first `FROM` anywhere in the raw string.

String literals and `EXTRACT(...)` are stripped first, then we take only the first remaining `FROM` we can be confident about: a quoted identifier (including hyphens) or an unquoted/schema-qualified name. `FROM (` is a subquery and returns empty — we do not walk into the inner table. Pinot still executes `pinotQlCode`; `tableName` is only the editor/share-link hint.

Covered in `TestExtractTableName` (including `EXTRACT(YEAR FROM ts)`, `'FROM x'` / `'from admin'`, `"my-table"`, `my_db.events`, and `FROM (SELECT x FROM inner_t)`).

yduartep: `ExtractTableName` no longer takes the first `FROM` anywhere in the raw string.

String literals and `EXTRACT(...)` are stripped first, then we take only the first remaining `FROM` we can be confident about: a quoted identifier (including hyphens) or an unquoted/schema-qualified name. `FROM (` is a subquery and returns empty — we do not walk into the inner table. Pinot still executes `pinotQlCode`; `tableName` is only the editor/share-link hint.

Covered in `TestExtractTableName` (including `EXTRACT(YEAR FROM ts)`, `'FROM x'` / `'from admin'`, `"my-table"`, `my_db.events`, and `FROM (SELECT x FROM inner_t)`).

internal/query/pinot/types_test.go
yduartep: The subquery half is already covered: `SELECT * FROM (SELECT x FROM inner_t) a` is in `TestExtractTableName` and returns empty.

`EXPLAIN` and `INSERT` never reach `bail()` — `limitStatementRe` rejects them first. Those `TestEnforceLimit` cases are now named `explain is not SELECT or WITH` / `insert is not SELECT or WITH`. `TestLimitNotEnforced` already asserts both are false.

The old DML bail regex is gone, so `WHERE action = 'delete'` appends LIMIT. Bare `OFFSET 10` is covered via SELECT in `TestLimitNotEnforced`.

internal/query/pinot/types_test.go
yduartep: The subquery half is already covered: `SELECT * FROM (SELECT x FROM inner_t) a` is in `TestExtractTableName` and returns empty.

`EXPLAIN` and `INSERT` never reach `bail()` — `limitStatementRe` rejects them first. Those `TestEnforceLimit` cases are now named `explain is not SELECT or WITH` / `insert is not SELECT or WITH`. `TestLimitNotEnforced` already asserts both are false.

The old DML bail regex is gone, so `WHERE action = 'delete'` appends LIMIT. Bare `OFFSET 10` is covered via SELECT in `TestLimitNotEnforced`.

internal/query/pinot/types.go
yduartep: Removed. The only caller was `TestEscapeSQLString`. Pinot has no `list-tables` / `describe-table` that interpolates identifiers, so the helper is unused. We can add it back when those commands land.

yduartep: Removed. The only caller was `TestEscapeSQLString`. Pinot has no `list-tables` / `describe-table` that interpolates identifiers, so the helper is unused. We can add it back when those commands land.

internal/datasources/pinot/explore.go
yduartep: `QueryExploreURL` is now `(host, query)` like the other raw-SQL dialects. `tableName` comes from `ExtractTableName(query.Expr)`, so the call site cannot pass a mismatched name.

The plugin request body still derives its own `tableName` from the SQL. The `client.go` comment no longer says that field is so Explore can open — it is the editor hint; Pinot runs `pinotQlCode`.

yduartep: `QueryExploreURL` is now `(host, query)` like the other raw-SQL dialects. `tableName` comes from `ExtractTableName(query.Expr)`, so the call site cannot pass a mismatched name.

The plugin request body still derives its own `tableName` from the SQL. The `client.go` comment no longer says that field is so Explore can open — it is the editor hint; Pinot runs `pinotQlCode`.

internal/datasources/pinot/query.go
yduartep: Confirmed against a live StarTree datasource. gcx leaves `$__timeFilter("timestamp")` in `pinotQlCode` and sends `--since` as `from`/`to`; the plugin expands the macro.

```
gcx datasources pinot query -d UID --since 1d \
  'SELECT count(*) FROM faro_pinot_events_v2 WHERE $__timeFilter("timestamp")'
```

That returned a count (`453778704`). The same query with `--since 1h` returned a much smaller count (`~28M`), so the window is applied, not ignored.

A follow-up `SELECT $__timeFromMillis, $__timeToMillis, min("timestamp"), max("timestamp")` on `--since 1d` showed a 24h millis range (`86400000`), with row timestamps inside that window (minute-aligned, which is the plugin's `$__timeFilter` behavior).

`$__timeGroup` is a StarTree plugin macro as well (not sqlds-only). Help line stays.

yduartep: Confirmed against a live StarTree datasource. gcx leaves `$__timeFilter("timestamp")` in `pinotQlCode` and sends `--since` as `from`/`to`; the plugin expands the macro.

```
gcx datasources pinot query -d UID --since 1d \
  'SELECT count(*) FROM faro_pinot_events_v2 WHERE $__timeFilter("timestamp")'
```

That returned a count (`453778704`). The same query with `--since 1h` returned a much smaller count (`~28M`), so the window is applied, not ignored.

A follow-up `SELECT $__timeFromMillis, $__timeToMillis, min("timestamp"), max("timestamp")` on `--since 1d` showed a 24h millis range (`86400000`), with row timestamps inside that window (minute-aligned, which is the plugin's `$__timeFilter` behavior).

`$__timeGroup` is a StarTree plugin macro as well (not sqlds-only). Help line stays.

yduartep: Correction — the live probe used this exact SQL (`-d` omitted here):

```
gcx datasources pinot query -d UID --since 1d \
  'SELECT count(*) FROM faro_pinot_events_v2 WHERE $__timeFilter("timestamp")'
```

Result: `453778704`.

internal/datasources/pinot/query.go
yduartep: Added `internal/datasources/pinot/query_test.go`, same `--limit=-5` case as postgres/mysql/bigquery. A zero `ConfigLoader` proves `Validate` runs first (`--limit must be >= 0`).

Pushed on https://github.com/grafana/gcx/commit/9d44f4e6211369edd189f7e8b8a43882365e7584

internal/query/pinot/types.go
yduartep: This is already addressed. The Faro / `LIMIT 2000` rationale was removed when I rewrote the UNION bail for the warning work.

`unionOrOffsetRe` now states the invariant: appending LIMIT on UNION would bind only the last leg (or is invalid with OFFSET), so the SQL is left unchanged.

https://github.com/grafana/gcx/commit/ba422711

internal/agent/command_annotations.go
yduartep: Removed the registry row. `ApplyAnnotations` only fills missing keys; `QueryCmd` already sets cost and hint inline, so that entry never won (and the hint had already drifted).

https://github.com/grafana/gcx/commit/6441e4bb3f35d73feddd9a2f80b19014dbc09d3e

internal/datasources/pinot/query.go
yduartep: Help, the agent hint, fixtures, and the CLI reference now use generic `events` / `logs`. The time-range example uses `$__timeFilter("timestamp")` instead of an internal app id.

Pushed on commit https://github.com/grafana/gcx/commit/bdb41217082ef3bb7013340a6a242524dc02245d

internal/datasources/providers/pinot.go
yduartep: Intentional for this PR: Pinot ships `query` only. `list-tables` / `describe-table` need Pinot catalog queries and would grow this change further; they will be a follow-up PR.

Noted in the PR description under **Out of scope**.

yduartep:
Hi @awsome-o thanks for the comments. I already addressed all of them. When you have a chance, can you do a new review? thanks

yduartep:
@claude review

internal/query/pinot/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+	bodyMap := map[string]any{
+		"queries": []any{
+			map[string]any{
+				"refId":       "A",
+				"datasource":  map[string]any{"type": DatasourceType, "uid": datasourceUID},
+				"queryType":   "PinotQL",
+				"editorMode":  "Code",
+				"displayType": "TABLE",
+				"tableName":   ExtractTableName(req.RawSQL),
```
aranhave: `SELECT count(*) FROM (SELECT col FROM events) sub` sends an empty `tableName`, which [StarTree rejects before executing SQL](https://github.com/startreedata/startree-grafana-pinot-datasource/blob/v1.0.24/pkg/plugin/dataquery/pinot_ql_code_query.go#L26-L43). Can we expose a table override for these queries? #1257 already needs one for the session query, but neither datasource command lets callers set it.

yduartep: `ExtractTableName` now walks into `FROM (` and takes the first real table, so `SELECT count(*) FROM (SELECT col FROM events) sub` sends `events`. No `--table` needed for that shape.

SQL that names no table (`SELECT 1`, `FROM (SELECT 1)`) fails in gcx before the POST: StarTree needs a table name for schema and macros. Pass `--table <name>`. Same flag on the generic command. `QueryRequest.TableName` is the override for programmatic callers (#1257).

Verified against a live StarTree datasource: the subquery ran (Pinot still needs `SET useMultistageEngine = true` for nested SQL); `SELECT 1` without `--table` returned the new error.

https://github.com/grafana/gcx/commit/02602933

internal/query/pinot/types.go
```diff
@@ -0,0 +1,155 @@
...
+	if rest == "" || rest[0] == '(' {
+		return ""
+	}
+	if rest[0] == '"' {
+		m := quotedTableRe.FindStringSubmatch(rest)
+		if len(m) < 2 {
+			return ""
+		}
+		return m[1]
+	}
```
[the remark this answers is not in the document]
yduartep: ExtractTableName now walks quoted and unquoted identifier segments joined by dots, so `FROM "db"."events"` (also `db."events"`, `"db".events`, `"cat"."db"."events"`) yields `db.events` instead of stopping at the schema. A trailing `"db".` still returns empty so we do not send a partial name.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,131 @@
...
+		{"subquery count from events", "SELECT count(*) FROM (SELECT col FROM events) sub", "events"},
+		{"union wrapper uses first inner table", "SELECT * FROM (\n  SELECT 1 FROM measurements\n  UNION ALL\n  SELECT 1 FROM events\n) journey", "measurements"},
+		{"extract year from is not the table", `SELECT EXTRACT(YEAR FROM ts) FROM events`, "events"},
+		{"from inside string is not the table", `SELECT 'FROM x' AS a FROM events`, "events"},
+		{"from inside string lowercase", `SELECT 'from admin' AS x FROM t`, "t"},
+		{"hyphenated quoted table", `SELECT * FROM "my-table"`, "my-table"},
+		{"table alias", `SELECT * FROM events AS e`, "events"},
+		{"join uses left table", `SELECT * FROM orders AS o JOIN customers AS c ON o.id = c.id`, "orders"},
+		{"no from", "SELECT 1", ""},
+	}
```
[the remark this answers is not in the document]
yduartep: Added those shapes to TestExtractTableName in internal/query/pinot/types_test.go, including mixed quoting, a three-part name, spaces around dots, and a trailing-dot miss.

cmd/gcx/datasources/query.go
```diff
@@ -30,7 +31,8 @@ func (o *genericQueryOpts) setup(flags *pflag.FlagSet) {
 	o.shared.Setup(flags, true)
 	flags.StringVar(&o.profileType, "profile-type", "", "Profile type ID for pyroscope queries (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds')")
 	flags.Int64Var(&o.maxNodes, "max-nodes", 1024, "Maximum nodes in flame graph (pyroscope only)")
-	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, "Maximum number of log lines to return for loki queries (0 means no limit)")
+	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, "Maximum log lines for loki, or max rows for pinot (0 means no limit)")
+	flags.StringVar(&o.table, "table", "", "StarTree table name for pinot queries when the SQL has no extractable FROM")
```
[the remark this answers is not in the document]
yduartep: Generic `Validate()` now rejects `--limit < 0` before any HTTP, same as the typed Pinot command. `gcx datasources query UID --limit -5 'SELECT 1'` fails with `--limit must be >= 0`. Covered by `TestGenericQueryFlagValidationPrecedence`.

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

cmd/gcx/datasources/query_routes.go
```diff
@@ -273,6 +276,35 @@ func dispatchClickHouse(ctx context.Context, req genericQueryRequest) (any, erro
 	return resp, nil
 }
 
+func dispatchPinot(ctx context.Context, req genericQueryRequest) (any, error) {
+	sql, capped := pinot.EnforceLimit(req.expr, req.limit, 1000)
+	if capped {
+		cmdio.Warning(req.warn, "LIMIT in query exceeds the maximum of 1000 and was capped")
+	}
+
+	client, err := pinot.NewClient(req.cfg)
```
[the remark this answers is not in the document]
yduartep: `dispatchPinot` now warns on stderr when `LimitNotEnforced` is true and `--limit` is not 0, matching the typed command. UNION/OFFSET/OPTION/trailing `--` comments stay unbounded and say so. `TestGenericQueryCharacterization_PinotSkipLimitWarnsOnStderr` pins a UNION query.

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

internal/datasources/pinot/query.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func warnLimitEnforcement(w io.Writer, expr string, capped bool, limit int) {
+	if capped {
+		cmdio.Warning(w, "LIMIT in query exceeds the maximum of %d and was capped; use --limit 0 to disable enforcement", maxLimit)
+		return
+	}
+	if limit != 0 && pinot.LimitNotEnforced(expr) {
+		cmdio.Warning(w, "query uses UNION, OFFSET, or OPTION, so --limit was not applied; the SQL was sent unchanged. Use --limit 0 to disable this warning")
+	}
```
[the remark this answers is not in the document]
yduartep: The skip warning now lists UNION, OFFSET, OPTION, and a trailing line comment so it matches every `LimitNotEnforced` shape.

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,181 @@
...
+// LimitNotEnforced reports whether EnforceLimit will leave sql unchanged
+// because a LIMIT suffix would bind only the last UNION leg, collide with
+// OFFSET, or land after OPTION. LIMIT offset,count is not included: that form
+// already bounds the result, so it is left alone without a warning.
+func LimitNotEnforced(sql string) bool {
+	if !limitStatementRe.MatchString(selectBody(sql)) {
+		return false
+	}
+	return unionOrOffsetRe.MatchString(sql) || optionClauseRe.MatchString(sql)
+}
```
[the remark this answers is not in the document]
yduartep: `LimitNotEnforced` now uses `bail` except `LIMIT offset,count` (already bounded). A trailing `--` comment reports true so callers warn instead of staying silent.

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

internal/datasources/pinot/query_internal_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestWarnLimitEnforcement(t *testing.T) {
+	const (
+		unionSQL   = "SELECT 1 FROM a UNION SELECT 2 FROM b LIMIT 2000"
+		offsetSQL  = "SELECT * FROM t LIMIT 5000 OFFSET 0" //nolint:unqueryvet // LIMIT/OFFSET shape fixture
+		plainSQL   = "SELECT 1 LIMIT 50"
+		commaSQL   = "SELECT * FROM t LIMIT 10, 20"           //nolint:unqueryvet // LIMIT offset,count shape fixture
+		optionSQL  = "SELECT * FROM t OPTION(timeoutMs=5000)" //nolint:unqueryvet // OPTION shape fixture
+		cappedWarn = "LIMIT in query exceeds the maximum of 1000 and was capped; use --limit 0 to disable enforcement"
+		skipWarn   = "query uses UNION, OFFSET, or OPTION, so --limit was not applied; the SQL was sent unchanged. Use --limit 0 to disable this warning"
+	)
```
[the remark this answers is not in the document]
yduartep: `skipWarn` matches the new user-facing sentence. Added cases for a trailing `--` comment (warns) and `--limit 0` on that shape (quiet).

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,137 @@
...
+		{"option cannot take a trailing limit", "SELECT * FROM t OPTION(timeoutMs=5000)", true},
+		{"limit then option cannot be rewritten", "SELECT * FROM t LIMIT 5000 OPTION(timeoutMs=5000)", true},
+		{"union", "SELECT 1 FROM a UNION SELECT 2 FROM b LIMIT 2000", true},
+		{"union all after SET", "SET useMultistageEngine = true;\nSELECT 1 FROM a\nUNION ALL\nSELECT 2 FROM b LIMIT 2000", true},
+		{"limit offset", "SELECT * FROM t LIMIT 5000 OFFSET 0", true},
+		{"bare offset", "SELECT * FROM t OFFSET 10", true},
+		{"explain never reaches bail", "EXPLAIN SELECT * FROM t", false},
+		{"insert never reaches bail", "INSERT INTO t VALUES (1)", false},
+		{"trailing comment is not union/offset", "SELECT 1 -- keep going", false},
+		{"dml word in literal is not union/offset", "SELECT * FROM t WHERE action = 'delete'", false},
```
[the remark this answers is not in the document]
yduartep: `SELECT 1 -- keep going` is now expected to be `LimitNotEnforced` (true), matching `EnforceLimit`'s bail.

Addressed in 61c3101c69a03ac3c7f9f9f43dec54fc1d754de3.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,184 @@
...
+// that form already bounds the result.
+func LimitNotEnforced(sql string) bool {
+	if !limitStatementRe.MatchString(selectBody(sql)) {
+		return false
+	}
+	if limitCommaRe.MatchString(sql) {
+		return false
+	}
+	return bail(sql)
+}
```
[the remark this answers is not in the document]
yduartep: `bail` and `LimitNotEnforced` now run on SQL with single-quoted literals stripped (same helper `ExtractTableName` already used). `SELECT 'UNION' FROM t`, `'OPTION(...)'`, and `'LIMIT 10, 20'` get `--limit` applied and do not warn. Also return an empty Explore URL when `tableName` is still empty after extract/override, so `--share-link` does not open StarTree with `tableName: ""`.

Addressed in a2670a587e299e8de898b27b3b59b5fc9bb865f9.

internal/datasources/pinot/query.go
```diff
@@ -0,0 +1,169 @@
...
+	Datasource string
+	Table      string
+	Limit      int
+}
+
+func (opts *queryOpts) setup(flags *pflag.FlagSet) {
+	opts.Setup(flags, false)
+	flags.StringVarP(&opts.Datasource, "datasource", "d", "", "Datasource UID (required unless datasources.pinot is configured)")
+	flags.StringVar(&opts.Table, "table", "", "StarTree table name when the SQL has no extractable FROM (required in that case)")
+	flags.IntVar(&opts.Limit, "limit", pinot.DefaultLimit, pinot.LimitFlagUsage(pinot.MaxLimit))
```
yduartep, line 29: The `--limit` help no longer lists skip shapes by hand. `LimitFlagUsage` / `LimitSkipWarning` / the generic flag all read `LimitSkipShapes` (`UNION, OFFSET, OPTION, or a trailing comment`), so help, stderr, and `LimitNotEnforced` cannot drift. Trailing `--` and unclosed `/*` are both trailing comments. Keyword matches inside quotes or comments no longer skip `--limit`.

Addressed in 4e7e7441d17348857e7a5a5d38b6b668f7371acd.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,278 @@
...
+var ErrTableNameRequired = errors.New("could not derive a table name from the SQL. StarTree needs one to load schema and expand macros before it runs the query. Pass --table <name> (a real table the query uses)")
+
+// ExtractTableName returns the first FROM table we can be confident about, or
+// empty if the shape is unclear. FROM inside strings, quoted identifiers,
+// comments, and EXTRACT/TRIM/SUBSTRING/OVERLAY calls is ignored. FROM ( is
+// skipped so a subquery or UNION wrapper yields the first inner table
+// (SELECT … FROM (SELECT col FROM events) → events). An unclosed /* blanks
+// only that tail, so a real FROM before it is still used. Used to fill
+// StarTree's required tableName field; Pinot still executes pinotQlCode.
+func ExtractTableName(sql string) string {
```
yduartep, line 421: `ExtractTableName` now ignores `FROM` inside comments, quoted identifiers, strings, and EXTRACT/TRIM/SUBSTRING/OVERLAY. `SELECT 1 -- FROM events` / `FROM t` yields `t` on the plugin body and Explore URL.

An unclosed `/*` only blanks that tail, so `FROM events /* keep` still yields `events` (returning empty for any unclosed block would regress that).

Addressed in 79eb4e25cf198b84266d12486fe28d3c68938727.

yduartep:
> ### 🔵 Needs a closer look
> Pinot LIMIT enforcement can silently fail for valid trailing `--` comments that don’t have whitespace before `--`, because the trailing-comment regex is too strict and can allow `LIMIT` to be appended inside the comment.
> 
> Review details

Re: trailing `--` comment detection (`internal/query/pinot/types.go:56`):

`trailingLineCommentRe` was too strict — it required whitespace or `;` before `--`, so `SELECT 1 FROM t/*note*/-- comment` was not treated as ending in a comment. `EnforceLimit` could append `LIMIT n` inside the comment (no real cap, no skip warning).

Loosened the regex from `(^|[\s;])--[^\n]*$` to `--[^\n]*$`. `hasTrailingComment` still runs on `stripQuoted()` first, so `SELECT '--' FROM t` does not false-positive.

Regression cases in `TestEnforceLimit` / `TestLimitNotEnforced`:
- `SELECT 1 FROM t/*note*/-- comment` → unchanged, `LimitNotEnforced` true

Addressed in 9e3ee8e6.

cmd/gcx/datasources/query.go
```diff
@@ -30,7 +32,8 @@ func (o *genericQueryOpts) setup(flags *pflag.FlagSet) {
 	o.shared.Setup(flags, true)
 	flags.StringVar(&o.profileType, "profile-type", "", "Profile type ID for pyroscope queries (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds')")
 	flags.Int64Var(&o.maxNodes, "max-nodes", 1024, "Maximum nodes in flame graph (pyroscope only)")
-	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, "Maximum number of log lines to return for loki queries (0 means no limit)")
+	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, fmt.Sprintf("Maximum log lines for loki, or max rows for pinot (0 means no limit). Pinot: not applied to %s (warned on stderr)", pinot.LimitSkipShapes))
+	flags.StringVar(&o.table, "table", "", "StarTree table name for pinot queries when the SQL has no extractable FROM")
```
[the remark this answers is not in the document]
yduartep: Addressed in 9392515c / 4b0fc44f: `--limit` help and generated docs now disclose Pinot's implicit default of 100 when the flag is omitted. `limitSet` keeps explicit `--limit 50` distinct from omitted.

cmd/gcx/datasources/query.go
```diff
@@ -30,7 +32,8 @@ func (o *genericQueryOpts) setup(flags *pflag.FlagSet) {
 	o.shared.Setup(flags, true)
 	flags.StringVar(&o.profileType, "profile-type", "", "Profile type ID for pyroscope queries (e.g., 'process_cpu:cpu:nanoseconds:cpu:nanoseconds')")
 	flags.Int64Var(&o.maxNodes, "max-nodes", 1024, "Maximum nodes in flame graph (pyroscope only)")
-	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, "Maximum number of log lines to return for loki queries (0 means no limit)")
+	flags.IntVar(&o.limit, "limit", dsquery.DefaultLokiLimit, fmt.Sprintf("Maximum log lines for loki, or max rows for pinot (0 means no limit). Pinot uses %d when --limit is omitted. Pinot: not applied to %s (warned on stderr)", pinot.DefaultLimit, pinot.LimitSkipShapes))
+	flags.StringVar(&o.table, "table", "", "StarTree table name for pinot queries when the SQL has no extractable FROM")
```
yduartep, line 36: Addressed in 5a8a0696: registered `--table` on `gcx datasources query`; value flows through to `dispatchPinot` as intended. Covered by `TestGenericQueryCharacterization_PinotTableOverride`.

internal/providers/agento11y/eval/experiments/export_test.go
```diff
@@ -65,7 +65,7 @@ func runExperimentExportCommand(t *testing.T, serverURL string, args ...string)
 func writeRawJSON(t *testing.T, w http.ResponseWriter, body string) {
 	t.Helper()
 	w.Header().Set("Content-Type", "application/json")
-	_, err := io.WriteString(w, body) //nolint:gosec // Raw test fixtures intentionally verify byte-for-byte response fidelity.
+	_, err := io.WriteString(w, body) // #nosec G705 -- raw test fixtures verify byte-for-byte response fidelity
```
yduartep: Addressed in 53bd1497: reverted the suite-wide G705 exclusion in .golangci.yaml. This call site now uses #nosec G705 instead of widening _test.go gosec exclusions. mise run lint passes locally (0 issues).

internal/query/pinot/types.go
```diff
@@ -0,0 +1,278 @@
...
+
+// EnforceLimit ensures the SQL has a LIMIT clause within bounds and reports
+// whether an explicit trailing LIMIT was capped to maxLimit, so callers can
+// warn instead of truncating silently.
+// If limit is 0, enforcement is disabled (pass-through).
+// SET prefixes are ignored for the SELECT-shaped allow-list. UNION, OFFSET,
+// LIMIT offset,count, OPTION(...), and statements ending in a comment
+// pass through unchanged. Keywords inside quotes or comments do not trigger a
+// bail. Real DML never reaches bail: only SELECT/WITH do.
+func EnforceLimit(sql string, limit, maxLimit int) (string, bool) {
```
awsome-o, line 395: With the default limit, `SELECT 1 FROM events LIMIT 5 /* note */` becomes `SELECT 1 FROM events LIMIT 5 /* note */ LIMIT 100`, turning valid SQL into invalid SQL. A leading comment also causes `/* note */ SELECT 1 FROM events` to bypass both limit enforcement and its warning.

Could the statement and trailing LIMIT checks account for comments, with regression coverage for these cases? Preserving the query and warning when rewriting is unsafe would also work.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,278 @@
...
+// maskNonTableFrom blanks regions where FROM is not a table clause, keeping
+// the string the same length so FROM positions still index the parse copy.
+// Quoted identifiers are blanked only on the search copy; the parse copy
+// keeps them so FROM "my-table" still yields my-table.
+func maskNonTableFrom(sql string, maskQuotedIdents bool) string {
+	s := replaceSameLen(sqlStringRe, sql)
+	if maskQuotedIdents {
+		s = replaceSameLen(sqlQuotedIdentRe, s)
+	}
+	s = replaceSameLen(sqlLineCommentRe, s)
```
awsome-o: `SELECT 1 /* -- note */ FROM events` is rejected with “Could not derive a table name” before any request is sent. The line comment mask interprets `--` inside the block comment and removes the real `FROM`. An apostrophe inside a block comment can similarly confuse the string mask.

Could comment and quote handling respect their surrounding context, and add these cases to the extraction tests?

awsome-o (commented):
The generic `--table` binding and its request test are now present. Two SQL comment handling cases remain in the unchanged Pinot helpers; I reproduced both locally. Details inline.

internal/query/pinot/types.go
yduartep: Addressed in 583ecc48. Comment and quote handling now scans in source order via `maskLexical`, so `--` and quotes inside block comments no longer mask a real `FROM`. Regression cases: `SELECT 1 /* -- note */ FROM events` and `SELECT 1 /* don't stop */ FROM events` in `TestExtractTableName`.

internal/query/pinot/types.go
yduartep: Addressed in 583ecc48. `LIMIT n` followed by a trailing block comment now bails (no second LIMIT appended). Leading block comments are stripped before the SELECT-shaped check so `/* note */ SELECT … FROM events` still gets a default LIMIT. Trailing block comments without a preceding LIMIT still allow appending (e.g. `SELECT 1 FROM t /* UNION */`). Covered in `TestEnforceLimit` / `TestLimitNotEnforced`.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,388 @@
...
+		part, rest, ok = tableIdentSegment(rest)
+		if !ok {
+			return ""
+		}
+		parts = append(parts, part)
+	}
+	return strings.Join(parts, ".")
+}
+
+// parseQuotedIdent reads a SQL double-quoted identifier, unescaping "" to ".
```
yduartep, line 457: Re: quoted identifiers with escaped quotes (`quotedTableRe` / `FROM "my""table"`):

Replaced the `^"([^"]+)"` regex with `parseQuotedIdent`, which scans double-quoted identifiers and unescapes SQL-standard `""` → `"`. `ExtractTableName` now returns the full logical name (`my"table`, `db.my"table`) instead of truncating at the first inner quote.

Regression cases in `TestExtractTableName`:
- `SELECT * FROM "my""table"` → `my"table`
- `SELECT * FROM "db"."my""table"` → `db.my"table`

Addressed in 9ff14bef.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,390 @@
...
+				s = s[end+2:]
+				continue
+			}
+			return s
+		}
+		return s
+	}
+}
+
+func limitStatementBody(sql string) string {
```
yduartep: Re: `limitStatementBody()` SET vs leading comment order (`types.go:200`):

`limitStatementBody` now strips leading comments first, then SET prefixes, then leading noise again. `/* note */ SET useMultistageEngine = true;\nSELECT 1` is treated as SELECT-shaped and gets LIMIT enforcement instead of passing through silently.

Regression cases in `TestEnforceLimit` / `TestLimitNotEnforced`:
- `/* note */ SET …; SELECT 1` → `… LIMIT 100`, `LimitNotEnforced` false

Addressed in 97956e5a.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,390 @@
...
+func limitAppendUnsafe(sql string) bool {
+	s := strings.TrimRight(stripQuoted(sql), "; \t\n")
+	if trailingLineCommentRe.MatchString(s) {
+		return true
+	}
+	if limitBeforeTrailingCommentRe.MatchString(s) {
+		return true
+	}
+	return sqlUnclosedBlockRe.MatchString(sqlBlockCommentRe.ReplaceAllString(s, " "))
+}
```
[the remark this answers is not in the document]
yduartep: `limitAppendUnsafe` now runs `stripLineComments` before unclosed-block detection. `--` line comments are blanked while block comments are skipped as units, so `/*` inside a line comment no longer triggers a false bail.

Regression: `SELECT 1 -- /* note\nFROM t` → LIMIT appended, `LimitNotEnforced` false.

Addressed in 5fcb4ef1.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,182 @@
...
+		{"appends LIMIT when -- is only a literal", "SELECT '--' FROM t", 100, "SELECT '--' FROM t LIMIT 100", false},
+		{"appends LIMIT when UNION is only a literal", "SELECT 'UNION' FROM t", 100, "SELECT 'UNION' FROM t LIMIT 100", false},
+		{"appends LIMIT when OPTION is only a literal", "SELECT * FROM t WHERE note = 'OPTION(timeoutMs=1)'", 100, "SELECT * FROM t WHERE note = 'OPTION(timeoutMs=1)' LIMIT 100", false},
+		{"appends LIMIT when LIMIT offset,count is only a literal", "SELECT * FROM t WHERE hint = 'LIMIT 10, 20'", 100, "SELECT * FROM t WHERE hint = 'LIMIT 10, 20' LIMIT 100", false},
+		{"appends LIMIT when UNION is only a quoted identifier", `SELECT "UNION" FROM t`, 100, `SELECT "UNION" FROM t LIMIT 100`, false},
+		{"appends LIMIT when UNION is only in a line comment", "SELECT 1 -- UNION\nFROM t", 100, "SELECT 1 -- UNION\nFROM t LIMIT 100", false},
+		{"appends LIMIT when UNION is only in a block comment", "SELECT 1 FROM t /* UNION */", 100, "SELECT 1 FROM t /* UNION */ LIMIT 100", false},
+		{"bail on unclosed block comment", "SELECT 1 FROM t /* keep", 100, "SELECT 1 FROM t /* keep", false},
```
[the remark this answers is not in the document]
yduartep: Regression added in `5fcb4ef1`:

- `TestEnforceLimit`: `appends LIMIT when block open is only in a line comment` — `SELECT 1 -- /* note\nFROM t` gets `LIMIT 100` appended
- `TestLimitNotEnforced`: `block open in line comment is not unclosed block` — same SQL, `LimitNotEnforced` false

Paired with `stripLineComments` in `limitAppendUnsafe` (see reply on types.go).

internal/providers/agento11y/eval/experiments/export_test.go
yduartep: Re: gosec suppression style on this call site:

Switched from `#nosec G705` to `//nolint:gosec` to match the dominant test convention (`main_test.go`, `usageevent_process_test.go`). Scoped suppression stays on this line; the suite-wide G705 exclusion in `.golangci.yaml` remains reverted.

`mise run lint` passes (0 issues).

Addressed in d0e1587c.

internal/providers/agento11y/eval/experiments/export_test.go
```diff
@@ -65,7 +65,7 @@ func runExperimentExportCommand(t *testing.T, serverURL string, args ...string)
 func writeRawJSON(t *testing.T, w http.ResponseWriter, body string) {
 	t.Helper()
 	w.Header().Set("Content-Type", "application/json")
-	_, err := io.WriteString(w, body) //nolint:gosec // Raw test fixtures intentionally verify byte-for-byte response fidelity.
+	_, err := io.WriteString(w, body) //nolint:gosec // raw test fixtures verify byte-for-byte response fidelity
```
yduartep: Re: Copilot review — inconsistent gosec suppression directive (`export_test.go:68`):

Switched from `#nosec G705` to `//nolint:gosec` to match the dominant test convention (`main_test.go`, `usageevent_process_test.go`). Suppression stays scoped to this `writeRawJSON` helper (all mock response bodies flow through here). Suite-wide G705 exclusion in `.golangci.yaml` remains reverted.

`mise run lint` passes (0 issues).

Addressed in d0e1587c.

annanay25:
I wonder if this is a good candidate for GCX extensions @dafydd-t. I'm not sure if the design on that has reached consensus though.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,184 @@
...
+		{"appends LIMIT when UNION is only a literal", "SELECT 'UNION' FROM t", 100, "SELECT 'UNION' FROM t LIMIT 100", false},
+		{"appends LIMIT when OPTION is only a literal", "SELECT * FROM t WHERE note = 'OPTION(timeoutMs=1)'", 100, "SELECT * FROM t WHERE note = 'OPTION(timeoutMs=1)' LIMIT 100", false},
+		{"appends LIMIT when LIMIT offset,count is only a literal", "SELECT * FROM t WHERE hint = 'LIMIT 10, 20'", 100, "SELECT * FROM t WHERE hint = 'LIMIT 10, 20' LIMIT 100", false},
+		{"appends LIMIT when UNION is only a quoted identifier", `SELECT "UNION" FROM t`, 100, `SELECT "UNION" FROM t LIMIT 100`, false},
+		{"appends LIMIT when UNION is only in a line comment", "SELECT 1 -- UNION\nFROM t", 100, "SELECT 1 -- UNION\nFROM t LIMIT 100", false},
+		{"appends LIMIT when block open is only in a line comment", "SELECT 1 -- /* note\nFROM t", 100, "SELECT 1 -- /* note\nFROM t LIMIT 100", false},
+		{"appends LIMIT when UNION is only in a block comment", "SELECT 1 FROM t /* UNION */", 100, "SELECT 1 FROM t /* UNION */ LIMIT 100", false},
+		{"bail on unclosed block comment", "SELECT 1 FROM t /* keep", 100, "SELECT 1 FROM t /* keep", false},
```
dafydd-t, line 112: **required** — this table covers a *keyword* inside a comment (`UNION`, `/*`) but never a quote character or a `--`. `TestExtractTableName` has both shapes at lines 41-42, added with 583ecc48 when `maskNonTableFrom` was fixed. `limitAppendUnsafe` (`types.go:268`) did not get the same fix: it runs the trailing-comment tests on `stripQuoted(sql)`, which masks literals and quoted identifiers but not comments, so a `'` or a `--` inside a comment corrupts the scan and the statement bails. `--limit` is dropped and the user is told the query "uses UNION, OFFSET, OPTION, or a trailing comment" when it uses none of them.

Four cases that mirror the two `TestExtractTableName` ones. All four fail on the head of this branch, and no existing case in the table changes:

```suggestion
		{"bail on unclosed block comment", "SELECT 1 FROM t /* keep", 100, "SELECT 1 FROM t /* keep", false},
		{"appends LIMIT when an apostrophe is only in a line comment", "SELECT a -- don't stop\nFROM events", 100, "SELECT a -- don't stop\nFROM events LIMIT 100", false},
		{"appends LIMIT when an apostrophe is only in a block comment", "SELECT a /* don't stop */ FROM events", 100, "SELECT a /* don't stop */ FROM events LIMIT 100", false},
		{"appends LIMIT when -- is only inside a block comment", "SELECT /* note -- x */ a FROM events", 100, "SELECT /* note -- x */ a FROM events LIMIT 100", false},
		{"appends LIMIT when a block comment holding -- is not at the end", "SELECT a FROM events /* a -- b */ WHERE x = 1", 100, "SELECT a FROM events /* a -- b */ WHERE x = 1 LIMIT 100", false},
```

dafydd-t (commented):
This `--limit` handling logic is complicated isnt it. There's one gap here. 

`TestExtractTableName` covers a `--` and an apostrophe inside a comment, but `TestEnforceLimit` and `TestLimitNotEnforced` do not.

internal/query/pinot/types_test.go
```diff
@@ -0,0 +1,184 @@
...
+		{"leading block comment before SET still enforces limit", "/* note */ SET useMultistageEngine = true;\nSELECT 1", false},
+		{"dml word in literal is not union/offset", "SELECT * FROM t WHERE action = 'delete'", false},
+		{"union word in literal is not a union", "SELECT 'UNION' FROM t", false},
+		{"option word in literal is not an option clause", "SELECT * FROM t WHERE note = 'OPTION(timeoutMs=1)'", false},
+		{"limit offset,count in literal is not that form", "SELECT * FROM t WHERE hint = 'LIMIT 10, 20'", false},
+		{"union in quoted identifier is not a union", `SELECT "UNION" FROM t`, false},
+		{"union in line comment is not a union", "SELECT 1 -- UNION\nFROM t", false},
+		{"block open in line comment is not unclosed block", "SELECT 1 -- /* note\nFROM t", false},
+		{"union in block comment is not a union", "SELECT 1 FROM t /* UNION */", false},
+		{"unclosed block comment skips enforcement", "SELECT 1 FROM t /* keep", true},
```
dafydd-t: **required** — the other half of the comment on `TestEnforceLimit` above. This table is what pins the second symptom: the stderr warning firing on a query that has none of the four named shapes.

```suggestion
		{"unclosed block comment skips enforcement", "SELECT 1 FROM t /* keep", true},
		{"apostrophe in a line comment is not a trailing comment", "SELECT a -- don't stop\nFROM events", false},
		{"apostrophe in a block comment is not a trailing comment", "SELECT a /* don't stop */ FROM events", false},
		{"-- inside a block comment is not a trailing comment", "SELECT /* note -- x */ a FROM events", false},
		{"block comment holding -- mid-statement is not a trailing comment", "SELECT a FROM events /* a -- b */ WHERE x = 1", false},
```

internal/query/pinot/client.go
```diff
@@ -0,0 +1,82 @@
+package pinot
```
dafydd-t, line 1: **recommended** — `docs/architecture/project-structure.md` enumerates every `internal/datasources/<kind>` and `internal/query/<kind>` package, and neither `pinot/` row was added. This package and `internal/datasources/pinot/` are both missing. `CLAUDE.md`'s PR checklist step 4 (doc maintenance gate) names the package map specifically, and T11 in `integrate-with-gcx/references/self-review.md` repeats it for new packages.

The `internal/datasources/` list is alphabetical, so this goes immediately above `postgres/` at line 106:

```
│   │   ├── pinot/            # StarTree Pinot datasource commands (query)
```

The `internal/query/` list is grouped rather than alphabetical, with the SQL clients last, so this goes between `mysql/` and `postgres/` (line 124) and leaves `postgres/` as the `└──` entry:

```
│   │   ├── pinot/            # StarTree Pinot query client (PinotQL via unified query API)
```

Not blocking, and no `suggestion` block because the file is not in this diff.

One judgement call worth naming: I checked the Reference Implementations table in `.claude/skills/add-datasource/SKILL.md` too, and left it alone. That table already omits bigquery, mysql, elasticsearch and opensearch, so it reads as a curated set of examples rather than a register, and adding pinot there is not obviously right.

dafydd-t (commented):
One doc gate from the PR checklist. Comment below is prefixed **recommended** (1).

internal/query/pinot/types_test.go
yduartep: Added. `TestEnforceLimit` now appends LIMIT when an apostrophe is only in a line or block comment, and when `--` is only inside a block comment (including mid-statement). Those queries no longer bail. The same shapes are covered in `TestLimitEnforcementNotice` and `bail_internal_test.go`.

internal/query/pinot/types_test.go
yduartep: `TestLimitNotEnforced` is gone; notices are asserted in `TestLimitEnforcementNotice` instead. The apostrophe / `--`-in-comment cases no longer emit the skip wording. When `--limit` is omitted they get the “appended default LIMIT 100” notice. Unclosed `/*` still skips. Same fixtures are in `bail_internal_test.go`.

internal/query/pinot/client.go
yduartep: Added both package-map rows in `docs/architecture/project-structure.md` as suggested (`internal/datasources/pinot` and `internal/query/pinot`). Left the add-datasource skill reference table alone — agreed it is a curated example set, not a register.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,515 @@
...
+
+// hasTrailingLineComment reports a -- line comment on the last line of the statement.
+func hasTrailingLineComment(sql string) bool {
+	trimmed := strings.TrimRight(sql, "; \t\n\r")
+	line := trimmed
+	if idx := strings.LastIndexByte(trimmed, '\n'); idx >= 0 {
+		line = trimmed[idx+1:]
+	}
+	line = strings.TrimRight(
+		maskLexical(line, lexicalMask{strings: true, quotedIdents: true, blockComments: true}),
```
awsome-o: Could we avoid reporting that a row limit was applied when it lands inside a comment? In a local helper test, `EnforceLimit("SELECT * FROM events /*\n' */ -- tail", 100, 1000)` appends `LIMIT 100` inside `-- tail`, while the notice says it applied the limit.

Starting the scan at the last line loses the surrounding block comment, so the apostrophe hides the real `--`. I haven't checked this against live Pinot. This is an unusual input, and I'm happy to leave it as an optional followup.

awsome-o (dismissed):
The earlier commented LIMIT case is fixed.
I'm approving with one optional followup below.
Focused tests pass; I haven't checked live Pinot.

internal/query/pinot/types.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+	return string(out)
+}
+
+func stripQuotedLiterals(sql string) string {
+	return maskLexical(sql, lexicalMask{strings: true, quotedIdents: true})
+}
+
+// fromFuncRe matches Calcite functions whose FROM argument is not a table.
+var fromFuncRe = regexp.MustCompile(`(?is)\b(?:EXTRACT|SUBSTRING|SUBSTR|TRIM|OVERLAY)\s*\([^)]*\)`)
```
aranhave: Non-blocking: `SELECT TRIM(BOTH LOWER('X') FROM city) FROM events` sends `tableName: "city"`, because this regex stops at the inner `)`. Could we handle nested parentheses here? I reproduced it through both CLI commands with a local HTTP test. `--table events` works around it, so I'm happy to leave this as a follow-up.

internal/query/pinot/types.go
yduartep: Addressed in e5cdc625.

`hasTrailingLineComment` now walks the full statement (same lexical order as the other scanners) instead of starting at the last line. That keeps the surrounding `/* */`, so the apostrophe cannot hide a real trailing `--`. We bail instead of appending `LIMIT` into the comment, and the existing skip notice is the one that fires.

Verified against a live StarTree datasource:

```
GCX_AGENT_MODE=false ./bin/gcx datasources pinot query \
  -d UID --since 1h --limit 50 -o json \
  --expr $'SELECT "timestamp" FROM faro_pinot_measurements_v1 /*\n\' */ -- tail'
```

That returned 10 rows (Pinot's default when no live `LIMIT` is present) with:

`You asked for --limit 50, but a row limit was not appended (this query shape is not safe to modify). Add LIMIT in the SQL or use --limit 0.`

aranhave (approved):
LGTM! 🏅

awsome-o (approved):
LGTM

PR #1257, opened 2026-08-27
feat(frontend): add command to fetch frontend o11y session

yduartep:
## Summary

Adds `gcx frontend sessions get` so humans and agents can fetch **one Frontend Observability session** and write it as a plain-text dump: session metadata once, then the user-journey events.

This is stacked on `feat/datasources-pinot-query`. Pinot is the analytics backend for Faro session detail (the same tables Explore uses). Loki remains available on stacks that have not moved to Pinot.

- `-d` / `--datasource` is **required** and is a **Grafana datasource UID** (not `loki` or `pinot`). gcx loads that datasource and **infers Loki vs Pinot from its type**.
- `--app` is the Faro app (slug-id or numeric id). `--since` or `--from` / `--to` is required (mutually exclusive).
- `--app-type web|mobile` is optional. Faro apps do not store web vs mobile; when omitted, gcx infers it from `sdkName` / `osName` so mobile journeys drop `app_memory` / `app_cpu_usage`.
- Without `--save`, Pinot prints metadata as tables and the journey as TSV; Loki prints named envelope fields once, then timestamp + stripped log lines.
- `--save` writes that dump to a file and prints a small artifact receipt on stdout. Agent mode requires `--save` so telemetry is not dumped into the conversation.
- There is no JSON/YAML encoding of the dump.

Pinot `appId` is an unquoted numeric SQL literal. `substPinot` now rejects anything that is not a base-10 integer (`pinot.FormatSQLInt`) so values like `66; DROP TABLE events` never reach the query. Session IDs stay quote-escaped.

## How to test

Unit tests:

```bash
mise exec -- go test ./internal/query/pinot/ ./internal/providers/faro/
```

Against a stack that has Frontend Observability data and a logged-in gcx context. List UIDs with `gcx datasources list` if you need them:

**Note**: Use [this session](https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/96/sessions/76rGv2rmGK) for testing against real datasource.

```bash
# Pinot on stdout (metadata tables, journey TSV)
gcx frontend sessions get <session-id> --app <app-id> -d <pinot-datasource-uid> --since 7d

# Pinot dump to a file (what agents should use)
gcx frontend sessions get <session-id> --app <app-id> -d <pinot-datasource-uid> --since 7d \
  --save /tmp/session-<session-id>.txt

# Loki dump
gcx frontend sessions get <session-id> --app <app-id> -d <loki-datasource-uid> --since 7d \
  --save /tmp/session-<session-id>.txt

# Force mobile SQL (app_memory / app_cpu_usage excluded)
gcx frontend sessions get <session-id> --app <app-id> --app-type mobile \
  -d <pinot-datasource-uid> --since 7d --save /tmp/session-<session-id>.txt
```

Expect two labeled blocks (`=== session metadata ===` and `=== events ===`). With `--save`, stdout should be an artifact receipt pointing at the file, not the dump itself.

**LOKI Web Session Output**:

Loki Web Session Get stdout

**LOKI Mobile Session Output**:

Loki Mobile Session Get stdout

**Pinot Mobile Session Output**:

Pinot Mobile Session Get stdout

**Pinot Web Session Output**:

Pinot Web Session Get stdout


internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,201 @@
...
+	flags.StringVar(&o.App, "app", "", "Frontend Observability app slug-id or numeric id (required)")
+	flags.StringVar(&o.AppType, "app-type", "", "web or mobile. Optional: inferred from sdkName/osName when omitted")
+	flags.StringVar(&o.Datasource, "datasource", datasourceLoki, "Telemetry backend: loki or pinot (default loki)")
+	flags.StringVar(&o.Save, "save", "", "Write the session dump to this path instead of stdout")
+}
+
+func (o *sessionsGetOpts) Validate() error {
+	if strings.TrimSpace(o.App) == "" {
+		return errors.New("--app is required")
+	}
```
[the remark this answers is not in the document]
yduartep: Fixed in 8ed17be. `Validate()` now trims `--app`, `--app-type`, `--datasource`, and `--save` (and lowercases the enums) before checks, so padded values like `--app " 66 "` reach `resolveAppID` in canonical form. Covered by `TestSessionsGetOptsValidateTrimsInputs`.

internal/providers/faro/session_format.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+func tsvCell(v any) string {
+	if v == nil {
+		return ""
+	}
+	switch val := v.(type) {
+	case string:
+		return val
+	case float64:
```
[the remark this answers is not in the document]
yduartep: Fixed in 8ed17be. `tsvCell` now runs string values through `escapeTSV`, which encodes `\`, tab, CR, and LF so a message with embedded newlines/tabs stays one TSV row. Covered by `TestFormatPinotTSVEscapesControlChars`.

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,204 @@
...
+	Datasource string
+	Save       string
+}
+
+func (o *sessionsGetOpts) setup(flags *pflag.FlagSet) {
+	o.SetupTimeFlags(flags)
+	flags.StringVar(&o.App, "app", "", "Frontend Observability app slug-id or numeric id (required)")
+	flags.StringVar(&o.AppType, "app-type", "", "web or mobile. Optional: inferred from sdkName/osName when omitted")
+	flags.StringVar(&o.Datasource, "datasource", datasourceLoki, "Telemetry backend: loki or pinot (default loki)")
+	flags.StringVar(&o.Save, "save", "", "Write the session dump to this path instead of stdout")
```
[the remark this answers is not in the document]
yduartep: Fixed. Dropped the manual `(default loki)` so Cobra only prints `(default "loki")`. `--datasource` and `--app-type` help now say `case-insensitive`, matching `Validate()` which trims and lowercases both. Regenerated `docs/reference/cli/gcx_frontend_sessions_get.md`.

yduartep:
@claude review

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,377 @@
...
+  FIRSTWITHTIME(appVersion, "timestamp", 'STRING') FILTER (WHERE appVersion <> '' AND appVersion <> 'null') AS app_version,
+  FIRSTWITHTIME(browserName, "timestamp", 'STRING') FILTER (WHERE browserName <> '' AND browserName <> 'null') AS browser_name,
+  FIRSTWITHTIME(browserVersion, "timestamp", 'STRING') FILTER (WHERE browserVersion <> '' AND browserVersion <> 'null') AS browser_version,
+  FIRSTWITHTIME(browserOs, "timestamp", 'STRING') FILTER (WHERE browserOs <> '' AND browserOs <> 'null') AS browser_os,
+  FIRSTWITHTIME(geoCountryCode, "timestamp", 'STRING') FILTER (WHERE geoCountryCode <> '' AND geoCountryCode <> 'null') AS geo_country_iso,
+  FIRSTWITHTIME(geoCity, "timestamp", 'STRING') FILTER (WHERE geoCity <> '' AND geoCity <> 'null') AS geo_city,
+  min("timestamp") AS session_start,
+  max("timestamp") AS session_end,
+  min("timestamp") FILTER (WHERE eventName = 'faro.session_recording.started') AS session_replay_start
+FROM faro_pinot_events_v2
```
aranhave: This hardcodes the ops events table, but the Frontend Observability app selects `faro_pinot_events_v1` on grafana-dev.net and v2 on grafana-ops.net. Can we resolve the table from the current server or make it configurable so the Pinot path works on dev too?

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,377 @@
...
+const pinotEventsMetadataSQL = `SET useMultistageEngine = true;
+SELECT
+  FIRSTWITHTIME(appName, "timestamp", 'STRING') FILTER (WHERE appName <> '' AND appName <> 'null') AS app_name,
+  FIRSTWITHTIME(appVersion, "timestamp", 'STRING') FILTER (WHERE appVersion <> '' AND appVersion <> 'null') AS app_version,
+  FIRSTWITHTIME(browserName, "timestamp", 'STRING') FILTER (WHERE browserName <> '' AND browserName <> 'null') AS browser_name,
+  FIRSTWITHTIME(browserVersion, "timestamp", 'STRING') FILTER (WHERE browserVersion <> '' AND browserVersion <> 'null') AS browser_version,
+  FIRSTWITHTIME(browserOs, "timestamp", 'STRING') FILTER (WHERE browserOs <> '' AND browserOs <> 'null') AS browser_os,
+  FIRSTWITHTIME(geoCountryCode, "timestamp", 'STRING') FILTER (WHERE geoCountryCode <> '' AND geoCountryCode <> 'null') AS geo_country_iso,
+  FIRSTWITHTIME(geoCity, "timestamp", 'STRING') FILTER (WHERE geoCity <> '' AND geoCity <> 'null') AS geo_city,
+  min("timestamp") AS session_start,
```
aranhave: `session_start` should come from the actual `session_start` event, while `session_end` should be the last row in the full journey. With `min`/`max` on the events table, the start can be too early and later exceptions or measurements can be missed. Can we derive both from the journey instead?

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,377 @@
...
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''name'']', 'STRING', '') AS nav_name,
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''duration'']', 'STRING', '') AS nav_duration,
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''ttfb'']', 'STRING', '') AS nav_ttfb
+  FROM faro_pinot_events_v2
+  WHERE appId = {{APP_ID}}
+    AND sessionId = '{{SESSION_ID}}'
+    AND eventName NOT IN ('faro.performance.resource', 'faro.performanceEntry')
+    AND $__timeFilter("timestamp")
+) journey
+ORDER BY "timestamp" ASC`
```
aranhave: [Pinot defaults a select without `LIMIT` to 10 rows](https://docs.pinot.apache.org/build-with-pinot/querying-and-sql/sql-syntax/sql-reference). That means longer sessions will be silently cut off here. Should this paginate, or at least use an explicit limit and mark the dump as truncated?

internal/providers/faro/session_format.go
```diff
@@ -0,0 +1,378 @@
...
+	writeLogfmtField(b, key, value)
+	b.WriteByte('\n')
+}
+
+func writeLogfmtField(b *strings.Builder, key, value string) {
+	b.WriteString(key)
+	b.WriteByte('=')
+	if strings.ContainsAny(value, " \t\"") {
+		b.WriteByte('"')
+		b.WriteString(strings.ReplaceAll(value, `"`, `\"`))
```
aranhave: Decoding and re-emitting a quoted value turns escaped `\n` into a literal newline because this only escapes quotes. A multiline exception or message then breaks the one-record-per-line dump. Can we use the logfmt encoder or escape backslashes and control characters here?

dafydd-t:
@claude review

internal/providers/faro/session_queries.go
yduartep: Addressed. The events table is no longer hardcoded to v2.

`gcx frontend sessions get` reads `grafana.server` from the current context (the stack URL from login) and picks the same table Frontend Observability uses: `faro_pinot_events_v2` on `*.grafana-ops.net`, `faro_pinot_events_v1` everywhere else (including grafana-dev). Measurements and exceptions stay `*_v1` on both clusters.

That name is substituted into the metadata query and the journey events leg. The journey UNION also sets StarTree `tableName` to the same events table, because the outer `FROM (` has no extractable table and the plugin requires the field.

internal/providers/faro/session_queries.go
yduartep: In commit [010297be](https://github.com/grafana/gcx/commit/010297be4d50eec499cd8acbb542a9be5c808773) I fixed `session_start` to use the Faro `session_start` event (`eventName = 'session_start'`) instead of `min("timestamp")` across all events. `session_last_event` stays `max("timestamp")`.

internal/providers/faro/session_queries.go
yduartep: Addressed in 33f57d85.

Pinot defaults a `SELECT` with no `LIMIT` to 10 rows, so a long session was silently truncated. `EnforceLimit` cannot append `LIMIT` onto a `UNION` (it would bind only the last leg), and `gcx datasources pinot query --limit 0` means “do not rewrite SQL”, not “return every row”.

**What I did.** Keep the Explore-style `UNION ALL` (measurements + exceptions + events). Put `LIMIT 1000` on the *outer* `SELECT` so it bounds the merged journey, not one table. Page with a timestamp cursor:

- first page: no extra `WHERE`
- after a full page: refetch `WHERE "timestamp" = last_ms` so same-millisecond rows that missed that `LIMIT` are kept
- next page: `WHERE "timestamp" > last_ms`

Stop when a page is shorter than 1000. Cap at 100 pages. Fingerprint-dedup so the same-ms refetch does not double-count, and so identical projected rows (the same exception or fetch again in the same millisecond) stay one analysis row. `OFFSET` is not used: it would re-scan all three UNION legs on every page.

**Evidence (ops Pinot, one real web session, fixed `--from`/`--to`).** Compared the paged UNION dump to three separate `SELECT`s (same `WHERE` as each UNION leg) and to `COUNT(*)` per table:

| Source | measurements | exceptions | events | sum |
|---|---:|---:|---:|---:|
| `COUNT(*)` / separate `SELECT` | 380 | 565 | 8808 | 9753 |
| Unique after journey projection | 380 | 494 | 8802 | 9676 |
| Paged UNION dump | 380 | 494 | 8802 | 9676 |

The 77-row gap is not a missing time range. Those rows are the same exception/fetch again in the same millisecond (same type + hash, or the same tracing event). For analysis we keep one copy. Every distinct projected journey row is present.

Pagination is actually running: 9676 ≫ 10 and ≫ 1000. Verbose HTTP showed 21 Pinot POSTs = 2 metadata + 10 full pages + 9 same-ms bucket refetches. A shorter `--since 3h` window produced 2774 rows and 7 POSTs (2 meta + 3 pages + 2 buckets).

Unit coverage: `TestFetchPinotSessionPagesUntilComplete` (first page + `=` bucket + `>` page, no `OFFSET`) and `TestAppendPinotRowsSkipsEqualJourneyRows`.

internal/providers/faro/session_format.go
yduartep: Addressed in b4df2020.

`writeLogfmtField` now re-emits through `logfmt.NewEncoder`. The decoder still turns quoted `\n` into a real newline; the encoder writes it back as escaped `\n` (and handles `\"`, tabs, and backslashes the same way). A multiline exception or message stays on one dump record.

If the encoder rejects a key, we fall back to `escapeTSV` so the rest of the line is not dropped.

**How to verify**

```bash
go test ./internal/providers/faro/ -count=1 -run TestWriteLogfmtFieldEscapesNewlines -v
```

The fixture is a Loki line `exception_value="line1\nline2"`. After decode + re-encode the dump is still one TSV/logfmt row:

```
1\tkind=exception exception_value="line1\nline2" message="say \"hi\""
```

Not `line1` then a raw newline then `line2`.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,377 @@
...
+// measurements as separate indexed Loki streams.
+func lokiSessionEventKinds() []string {
+	return []string{lokiKindEvent, lokiKindException, lokiKindLog, lokiKindMeasurement}
+}
+
+func lokiEventsQueryForKind(p sessionQueryParams, kind string) string {
+	app := escapeLogQLString(p.AppID)
+	session := escapeLogQLString(p.SessionID)
+	q := fmt.Sprintf(
+		`{app_id="%s", kind="%s"} |= "session_id=%s" !~ "performanceEntry|faro.performanceEntry|faro.performance.resource" | logfmt`,
```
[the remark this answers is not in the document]
yduartep: Fixed in 94d52691.

`lokiEventsQueryForKind` now finishes with `| logfmt | session_id="%s"`, same exact match as the metadata and replay queries. `|= "session_id=%s"` stays as the cheap stream filter. Covered by `TestLokiQueries` and `TestFetchLokiSession`.

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,217 @@
...
+				}
+				result, err = fetchPinotSession(ctx, client, opts.Datasource, params, start, end)
+			case datasourceLoki:
+				client, clientErr := loki.NewClient(cfg)
+				if clientErr != nil {
+					return fmt.Errorf("failed to create loki client: %w", clientErr)
+				}
+				result, err = fetchLokiSession(ctx, client, opts.Datasource, params, start, end)
+			default:
+				return fmt.Errorf("unsupported datasource kind %s", kind)
```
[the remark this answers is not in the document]
yduartep: Fixed in 94d52691.

Removed the unreachable `default`. `sessionKindFromDatasourceType` still rejects anything that is not Loki or Pinot before the switch.

internal/query/pinot/client.go
```diff
@@ -0,0 +1,77 @@
...
+// open; Pinot executes pinotQlCode, and a bad query is returned as a plugin error.
+func (c *Client) Query(ctx context.Context, datasourceUID string, req QueryRequest) (*querysql.QueryResponse, error) {
+	intervalMs := req.IntervalMs
+	if intervalMs == 0 {
+		intervalMs = 60000
+	}
+
+	from := strconv.FormatInt(req.Start.UnixMilli(), 10)
+	to := strconv.FormatInt(req.End.UnixMilli(), 10)
+	if req.Start.IsZero() || req.End.IsZero() {
```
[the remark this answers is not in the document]
yduartep: Fixed in 94d52691.

The 1h default now applies only when both Start and End are unset (`&&`). A lone Start is no longer overwritten. Covered by `TestQuery_DefaultsTimeRangeOnlyWhenBothUnset`.

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+			if err != nil {
+				return err
+			}
+
+			if opts.Save == "" {
+				return result.writeTables(cmd.OutOrStdout())
```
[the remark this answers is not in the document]
yduartep: Addressed in 3c633d41.

Human stdout and `--save` now go through `cmdio.Encode` with a `text` codec (`sessionDumpCodec`). The dump bytes are unchanged. `-o json` / `-o yaml`, `--json`, and `--jq` fail in `Validate()` (`session dump is text-only; use --save for the file`) so nothing is printed. Default `-o` is `text`. Agent mode still requires `--save` and emits the artifact receipt.

yduartep:
@claude review

cmd/gcx/root/testdata/output_classes.json
```diff
@@ -275,6 +275,7 @@
  "gcx frontend apps list": "finite",
  "gcx frontend apps list-sourcemaps": "finite",
  "gcx frontend apps update": "finite",
+ "gcx frontend sessions get": "artifact",
```
dafydd-t, line 279: every other command that creates a file uses `pull`  not `get`, could we do that here too?

dafydd-t: OH, i see this writes to stdout by default. so `get` is fine, but the command description is wrong (it says it writes to file).

docs/reference/cli/gcx_frontend_sessions_get.md
```diff
@@ -0,0 +1,81 @@
...
+### Options
+
+```
+      --app string          Frontend Observability app slug-id or numeric id (required)
+      --app-type string     web or mobile (case-insensitive). Optional: inferred from sdkName/osName when omitted
+  -d, --datasource string   Grafana datasource UID (required). Type is inferred (loki or pinot)
+      --from string         Start time (RFC3339, Unix timestamp, or relative like 'now-1h')
+  -h, --help                help for get
+  -o, --output string       Output format. One of: text (default "text")
+      --save string         Write the session dump to this path instead of stdout
```
dafydd-t, line 61: Can we get rid of `--save` and instead make an example with using normal bash operations e.g. `> /tmp/session-dggfdgfdg.txt` ?

internal/providers/faro/session_fetch.go
```diff
@@ -0,0 +1,407 @@
...
+		n := lokiEntryCount(resp)
+		appendLokiEvents(merged, resp)
+		if n < lokiEventsPageSize {
+			break
+		}
+		earliest, ok := minLokiTime(resp)
+		if !ok {
+			break
+		}
+		next := earliest.Add(-time.Millisecond)
```
aranhave: Moving back a millisecond skips any rows left at the page boundary. My test with 1,001 distinct events and only the last two sharing a timestamp returned 1,000 with no error. Could we overlap/refetch that boundary and deduplicate or report truncation when we can't recover it?

internal/providers/faro/session_fetch.go
```diff
@@ -0,0 +1,407 @@
...
+			return merged, nil
+		}
+		// Refetch the last millisecond so rows that shared that timestamp
+		// but missed this LIMIT are kept, then continue after that instant.
+		bucket, err := queryPinotJourney(ctx, client, uid, table, p, start, end, lastMS, true)
+		if err != nil {
+			return nil, err
+		}
+		appendPinotRows(merged, bucket, seen)
+		cursorMS = lastMS
```
aranhave, line 169: What happens if one millisecond contains more than 1,000 distinct rows? The `=` refetch also has `LIMIT 1000`, then this advances past that timestamp. my 1,001-row test returned 1,000 without an error. Please page that bucket or fail clearly when it fills.

cmd/gcx/root/testdata/output_classes.json
yduartep: > every other command that creates a file uses `pull` not `get`, could we do that here too?

Hey @dafydd-t I am not sure about the pull. We are fetching data as primary function with the option to save it also in a file but you could get that data printed out in the terminal as well. Example for fetching app data we use **get**:  `gcx frontend apps get`.

docs/reference/cli/gcx_frontend_sessions_get.md
yduartep: Hi @dafydd-t so, `--save` is the agent contract, not a convenience alias for `>`.

This command is class `artifact`: in agent mode stdout is a `gcx.artifact_receipt` with `files[0].path`, and the dump stays on disk. The [frontend-session-rca](https://github.com/grafana/skills/pull/94) skill always runs `gcx frontend sessions get --save …`, then reads that path. It must never paste the dump into the chat.

`>` does the opposite for an agent: the file is written, **stdout is empty**, and there is no receipt. The model only sees “exit 0” and often does not follow the path it typed. gcx can require `--save` in agent mode; it cannot require `>`. Without the flag, a long session dump lands in the conversation (or the agent thinks the command produced nothing).

Humans can still print to the terminal or redirect themselves. Agents need `--save` so the skill has a stable pointer to the file.

internal/providers/faro/session_fetch.go
yduartep: Addressed in 49b4fcb0.

After a full Loki page I refetch `[earliest, earliest+1ms)` and keep leftover rows at that timestamp (deduped), then continue with `End = earliest` instead of `earliest-1ms`. If that 1ms window is also full we error instead of truncating.

Covered by `TestFetchLokiEventPagesKeepsBoundaryRowsThatShareTimestamp`.

internal/providers/faro/session_fetch.go
yduartep: Addressed in 49b4fcb0.

A full `=` bucket no longer advances past that millisecond. We keep paging it with `OFFSET` until a short page, then continue with `timestamp > lastMS`.

Covered by `TestFetchPinotJourneyKeepsMoreThanPageInOneMillisecond`.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,455 @@
...
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''name'']', 'STRING', '') AS nav_name,
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''duration'']', 'STRING', '') AS nav_duration,
+    JSON_EXTRACT_SCALAR(attributesJson, '$[''ttfb'']', 'STRING', '') AS nav_ttfb
+  FROM {{EVENTS_TABLE}}
+  WHERE appId = {{APP_ID}}
+    AND sessionId = '{{SESSION_ID}}'
+    AND eventName NOT IN ('faro.performance.resource', 'faro.performanceEntry')
+    AND $__timeFilter("timestamp")
+) journey
+{{JOURNEY_CURSOR}}ORDER BY "timestamp" ASC
```
awsome-o: Could this still silently lose rows when more than 1,000 share a timestamp? The bucket now uses `OFFSET`, but `ORDER BY timestamp` leaves their relative order unspecified. A local stub returning 1,001 distinct rows in different valid tie orders produced 1,000 rows with no error.

Could we give those rows a stable ordering across requests and cover changing tie order in the test? I haven’t checked against live Pinot, so an existing guarantee that keeps these pages stable would change this finding.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,455 @@
...
+// measurements as separate indexed Loki streams.
+func lokiSessionEventKinds() []string {
+	return []string{lokiKindEvent, lokiKindException, lokiKindLog, lokiKindMeasurement}
+}
+
+func lokiEventsQueryForKind(p sessionQueryParams, kind string) string {
+	app := escapeLogQLString(p.AppID)
+	session := escapeLogQLString(p.SessionID)
+	q := fmt.Sprintf(
+		`{app_id="%s", kind="%s"} |= "session_id=%s" !~ "performanceEntry|faro.performanceEntry|faro.performance.resource" | logfmt | session_id="%s"`,
```
awsome-o: Could we keep exceptions and logs whose message mentions `performanceEntry`? This exclusion runs against the whole line for every kind, so an exception such as `exception_value="performanceEntry is undefined"` disappears from the session dump. Escaping the dots wouldn’t fix that case.

Could we filter the parsed `event_name` only for event records, matching the Pinot path? I checked the generated query and regex locally, but haven’t run it against live Loki.

awsome-o (commented):
Could we address the Pinot paging order and Loki filter before merging?
The existing session tests pass, but focused local probes reproduce both omissions.
I haven’t tested against live Pinot or Loki.

internal/providers/faro/session_commands.go
```diff
@@ -0,0 +1,242 @@
...
+				return opts.IO.Encode(cmd.OutOrStdout(), result)
+			}
+
+			var buf strings.Builder
+			if err = opts.IO.Encode(&buf, result); err != nil {
+				return err
+			}
+			if err = os.WriteFile(opts.Save, []byte(buf.String()), 0o600); err != nil {
+				return fmt.Errorf("writing session dump: %w", err)
+			}
```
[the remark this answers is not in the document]
yduartep: Fixed in 96f2b959. The `--save` path now opens the file and streams through `opts.IO.Encode(f, result)` instead of building a strings.Builder and copying into []byte for WriteFile.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,455 @@
...
+		app, kind, session, session,
+	)
+	if kind == lokiKindMeasurement && p.mobile() {
```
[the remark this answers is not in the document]
yduartep: Fixed in 96f2b959. Removed the line-level `!~` filter. Performance events are excluded only on `kind=event` streams, after `logfmt`, via `event_name!~"faro\\.performanceEntry|faro\\.performance\\.resource"` — matching the Pinot events leg. Exceptions and logs that mention `performanceEntry` in message/exception_value are kept.

internal/query/pinot/client.go
```diff
@@ -45,7 +45,7 @@ func (c *Client) Query(ctx context.Context, datasourceUID string, req QueryReque
 
 	from := strconv.FormatInt(req.Start.UnixMilli(), 10)
 	to := strconv.FormatInt(req.End.UnixMilli(), 10)
-	if req.Start.IsZero() || req.End.IsZero() {
+	if req.Start.IsZero() && req.End.IsZero() {
 		now := time.Now()
 		from = strconv.FormatInt(now.Add(-1*time.Hour).UnixMilli(), 10)
```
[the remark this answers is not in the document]
yduartep: Fixed in 96f2b959. `Client.Query` now returns `ErrPartialTimeRange` when exactly one of Start/End is set, instead of sending a year-0001 UnixMilli for the missing boundary. Default last-hour range still applies only when both are zero.

internal/query/pinot/client_test.go
```diff
@@ -157,6 +159,32 @@ func TestQuery_ExplicitTableNameOverridesExtract(t *testing.T) {
...
+	client := newTestClient(t, server.URL)
+	_, err := client.Query(context.Background(), "pinot-uid", pinot.QueryRequest{
+		RawSQL:    "SELECT 1",
+		TableName: "events",
+		Start:     start,
+	})
+	require.NoError(t, err)
+	assert.Equal(t, wantFrom, captured["from"])
+	assert.Equal(t, wantTo, captured["to"])
+}
```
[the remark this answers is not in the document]
yduartep: Fixed in 96f2b959. Replaced the Start-only/year-0001 test with `TestQuery_SendsExplicitTimeRange` (both boundaries set) and `TestQuery_RejectsPartialTimeRange` (Start XOR End errors). `TestQuery_DefaultsTimeRangeWhenBothUnset` covers the both-zero default.

internal/providers/faro/session_queries.go
yduartep: Fixed in 96f2b959. Added stable tie-break columns to the outer journey `ORDER BY` (`kind`, `event_name`, `hash`, `attribute_hash`, `message`, `exception_type`, `measurement_type`, `traceID`) so same-ms `OFFSET` paging is deterministic. `TestFetchPinotJourneySameMSStableWhenShuffled` shuffles 1,001 distinct same-ms rows (including per-request shuffle in the stub) and asserts all are kept.

internal/providers/faro/session_queries.go
yduartep: Fixed in 96f2b959. The line-level regex is gone. Performance exclusion now applies only to `kind=event` after `logfmt`, filtering parsed `event_name` — same semantics as Pinot's `eventName NOT IN (...)` on the events leg. Exceptions such as `exception_value="performanceEntry is undefined"` stay in the dump.

internal/providers/faro/session_queries.go
yduartep: I am testing on[ this real app](https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/96/sessions/76rGv2rmGK?tracesQuery=%7B%22queryType%22:%22traceql%22,%22refId%22:%22A%22,%22query%22:%22%7Bspan.session_id%3D%5C%2276rGv2rmGK%5C%22%20%7C%7C%20span.session.id%3D%5C%2276rGv2rmGK%5C%22%7D%22,%22limit%22:20%7D) in case you want to test against real DS

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+// pinotEventsTable matches Frontend Observability: ops Pinot serves
+// faro_pinot_events_v2; grafana-dev and every other host still serve v1.
+func pinotEventsTable(serverURL string) string {
+	host := hostnameFromServerURL(serverURL)
+	if hostHasSuffix(host, grafanaOpsHost) && !hostHasSuffix(host, grafanaDevHost) {
+		return pinotEventsTableOps
+	}
+	return pinotEventsTableDev
```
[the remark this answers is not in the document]
yduartep: Done in 2ea1de85 — dropped the redundant `!hostHasSuffix(host, grafanaDevHost)` guard.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+func (p sessionQueryParams) mobile() bool {
+	return p.AppType == appTypeMobile
+}
+
+func substPinot(sql string, p sessionQueryParams) (string, error) {
+	// appId is an unquoted numeric literal in the templates. EscapeSQLString
+	// only doubles quotes, so a value like `66; DROP TABLE events` would
+	// still inject. Canonicalize to a base-10 int64 or refuse to build SQL.
```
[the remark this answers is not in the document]
yduartep: Done in 2ea1de85 — comment now says EscapeSQLString only escapes single quotes.

internal/providers/faro/session_queries.go
```diff
@@ -0,0 +1,466 @@
...
+	// Pinot defaults a SELECT with no LIMIT to 10 rows. Page the journey
+	// UNION at this outer LIMIT and continue while a page is full. OFFSET
+	// pages a same-millisecond bucket when it exceeds LIMIT; tie-breakers
+	// keep that bucket stable across requests.
+	pinotJourneyPageSize = 1000
+	pinotJourneyMaxPages = 100
+
+	// Outer journey sort: timestamp alone is not stable when OFFSET pages a
+	// dense same-ms bucket (Pinot may return ties in arbitrary order).
+	pinotJourneyOrderBy = `ORDER BY "timestamp" ASC, kind ASC, event_name ASC, hash ASC, attribute_hash ASC, message ASC, exception_type ASC, measurement_type ASC, traceID ASC`
```
aranhave: The paging fix still leaves ties when rows differ only in `page_id`, `http_url`, or other fields outside this sort. A local variant of the shuffled test with 2,001 same-ms events differing by page ID returned 1,742 without an error. Could the ordering cover the full projected row so these pages can't silently skip events?

awsome-o, line 34: Could we fail clearly instead of returning an incomplete session here? My mocked test still returned 1,000 of 1,001 distinct events without an error when all had the same timestamp and differed only in `http_url`.

Those rows tie on every field in this `ORDER BY`, so they can change places across the `OFFSET` requests in `session_fetch.go:158`. I haven't reproduced that ordering against live Pinot. A guard that rejects a full timestamp bucket, like the Loki path, would be enough for me before merge.

awsome-o (commented):
Could we handle the remaining Pinot pagination case before merging?
The mocked test still loses one of 1,001 events without an error.
A clear error for a full timestamp bucket would be enough; I haven't checked live Pinot.

internal/providers/faro/session_queries.go
yduartep: Addressed in 7bfe1d4f.

A same-millisecond `=` bucket that fills `LIMIT 1000` now errors instead of being paged with `OFFSET`:

`pinot journey: more than 1000 rows share timestamp <ms>; dump would be truncated`

That is the Loki path: we still refetch `WHERE "timestamp" = last_ms` so leftover rows at a page boundary are kept when that bucket is short, but we do not `OFFSET` a full bucket. Ties on `ORDER BY` (including events that differ only in `http_url`) can reshuffle across requests, so paging them was able to return 1,000 of 1,001 with no error.

Covered by `TestFetchPinotJourneyErrorsWhenSameMSBucketIsFull` (1,001 same-ms rows that differ only in `http_url`). The usual first-page + short `=` bucket + `>` page path is unchanged (`TestFetchPinotSessionPagesUntilComplete`).

aranhave (dismissed):
LGTM

awsome-o (dismissed):
LGTM

dafydd-t (approved):

PR #1258, opened 2026-08-28
feat(alert): add notification history commands

fayzal-g:
## Summary

Adds a read-only `gcx alert notification-history` command group backed by the alerting historian API (`historian.alerting.grafana.app/v0alpha1`).

Notification History is a historical delivery log with no create/update semantics, so it is exposed as a **status reader under the existing `gcx alert` provider** (mirroring `gcx alert instances list`) rather than as a resource adapter — consistent with the provider's existing `TypedRegistrations() nil` design note. The historian API is only reachable via custom POST subresource routes (no CRUD Kind), so `gcx resources` cannot serve it.

- **`notification-history list`** — `POST notification/query` with time-range (`--from`/`--to`/`--since`) and `--status`/`--outcome`/`--receiver`/`--rule-uid` filters. Default `table` + `wide` codecs, plus `json`/`yaml`. The table surfaces `OUTCOME`, a formatted `DURATION` (from nanoseconds), and delivery `ERROR`.
- **`notification-history alerts`** — `POST notifications/queryalerts` to fetch the individual alerts for a notification `--uuid` (entries never carry their own alerts, so this is a required second call).
- A `422` from the API (feature disabled) maps to a clear "notification history is not enabled on this stack" error.

The namespace is derived from the existing `NamespacedRESTConfig.Namespace` — no new config/provider/blank-import wiring.

### Files
- `internal/providers/alert/notification_history_{types,client,commands}.go` — wire types, client methods, command group.
- `internal/providers/alert/client.go` — `Client` now carries the namespace.
- `internal/providers/alert/provider.go` — mounts the group.
- `internal/agent/command_annotations.go` + `cmd/gcx/root/testdata/output_classes.json` — agent `token_cost`/output-class metadata for the two new leaves.
- `docs/reference/cli/` — regenerated CLI reference.

## Test plan

- [x] `go test ./...` (full suite green, incl. agent consistency + output-class gates)
- [x] `golangci-lint` clean on the alert package
- [x] httptest contract tests: request path/body, agent single-JSON output, human-table byte-identity, `alerts` pivot, `--uuid` required, `422` disabled-state error
- [x] CLI reference regenerated (`GCX_AGENT_MODE=false ... cmd-reference`), no drift
- [x] Live smoke test against a stack with notification history enabled: `list` (table + json) and `alerts` pivot both return correct data
- [ ] CI `mkdocs build` (not run locally — mise/mkdocs unavailable)

## Out of scope (v1)
- `type=counts`/`range_counts` time-series mode — easy follow-up on the same client method.
- Any adapter / `resources push/pull` integration — not applicable to a historical log.


titolins (dismissed):

titolins (dismissed):

titolins (approved):

dafydd-t (approved):
Looks good from a UX point of view. thanks!

PR #1261, opened 2026-08-31
feat(fleet): move Fleet Management to the collector app plugin proxy

wardbekker:
## Why

Fleet Management was the only gcx product that needed a grafana.com Cloud Access Policy token. `internal/fleet/config.go` read `AgentManagementInstanceURL` and `AgentManagementInstanceID` from the grafana.com stack record, then called the Fleet Management API directly with Basic authentication.

That cost a user three things:

- A `gcx login` to the stack was not enough. A user also needed `gcx cloud login`, or a Cloud Access Policy token.
- gcx requested the `fleet-management:read` and `fleet-management:write` scopes for every user, whether or not they used Fleet.
- Every fleet command paid for a grafana.com stack lookup, even the commands that never read the stack record.

The `grafana-collector-app` plugin proxies the same API for its own user interface, and adds the credentials server-side. This pull request moves gcx onto that path, with no fallback.

**Outcome:** a Grafana stack login alone drives `gcx fleet` and `gcx instrumentation`.

## What changed

- `internal/fleet/client.go` carries no credentials. It takes an `*http.Client` built with `rest.HTTPClientFor` from the stack REST config.
- `internal/fleet/config.go` composes the prefix `/api/plugin-proxy/grafana-collector-app/fleet-management-api` and takes a `LoadGrafanaConfig` loader. `LoadClientWithStack` reads `cloud.StackInfo` through the bounded `grafanacom-api/instances/` proxy route. Each call performs one lookup, so a transient failure cannot affect a later call. `LoadClient` performs no stack lookup.
- `internal/fleet/preflight.go` owns collector app state and route-action checks. `cmd/gcx/setup` only renders the result.
- `internal/providers/fleet/client.go` returns the typed `fleet.HTTPError` from every method. Before this change `convertFleetHTTPErrors` never fired for pipeline or collector errors.
- `GetPipeline` and `GetCollector` test the body for `plugin route match not found` before they report a missing resource. Behind the proxy a 404 has two meanings.
- `cmd/gcx/fail` reports an absent plugin with the approved `Endpoint not available` summary. A 403 explains named read routes and wildcard admin routes.
- `internal/auth/gcom.go` and `internal/login/login.go` no longer name the Fleet scopes.
- `gcx setup status` reports the plugin state and both route actions. A missing or disabled plugin emits the complete status document and exits with code 1.

`providers.CloudRESTConfig`, `LoadCloudConfig`, and `cloud.GCOMClient` stay. k6, Synthetic Monitoring, the Faro sourcemap upload, Adaptive telemetry, `gcx cloud stacks`, and login validation still use them.

## Breaking change

A Cloud Access Policy token with `fleet-management` scopes no longer authenticates Fleet commands. Users need a Grafana stack login and the enabled collector app. Older stack tokens can require a new `gcx login` to obtain `grafana-api:write`.

Permission follows the matched plugin route, not the command intent. Named routes need `grafana-collector-app:read`. Wildcard routes need the Admin role or `grafana-collector-app:admin`. Some read-only commands, including `tenant get-limits` and `RunK8sMonitoring`, use wildcard routes.

## Compliance

- **CONSTITUTION** — the invariant named Fleet as an external-domain API that must use `httputils.NewDefaultClient`. Fleet now calls `cfg.Host`, so the rule inverts. This pull request amends the invariant and adds a Fleet paragraph next to the synth carve-out. **This is an invariant change and needs a reviewer's explicit approval.**
- **ADR-023** records the decision, the measurements, the route-based permission model, and the rejected fallback option.
- **DESIGN** — no new command or grammar change. The `gcx setup status` document gains one product row. Error summaries use the closed vocabulary.
- **ARCHITECTURE** — Fleet-specific preflight logic lives in `internal/fleet`. The setup command owns presentation only.

## Verification

`GCX_AGENT_MODE=false mise run all` passes: lint reports 0 issues, the full test suite passes, and the build, reference checks, skill checks, and documentation build succeed.

Focused tests cover:

- Retry after a transient stack lookup failure.
- The 1 MiB response limit.
- Missing, disabled, and unknown plugin states.
- All route-action combinations.
- The missing-plugin exit code and single-document output contract.
- The approved error summary and route-based 403 suggestions.

Live verification against a Grafana Cloud stack with plugin version 4.27.0 before the review fixes:

- `gcx fleet pipelines list`, `collectors list`, `tenant get-limits`, `gcx instrumentation status`, and `gcx resources get pipelines` returned data.
- A create, get, update, and delete round trip through `gcx fleet pipelines` succeeded.
- `gcx fleet tenant get-limits --context <stack-only-context>` worked without a cloud entry.

## What a reviewer must know

1. Route permissions are part of `plugin.json`. They do not follow read-only or write command intent.
2. `plugin.json` is not a stable contract. A plugin release can change the route set. `gcx setup status` reports effective actions, and the real RPC remains authoritative.
3. Fleet Management is a Grafana Cloud product. A self-hosted Grafana never had it, and still does not.
4. A stack token that predates `grafana-api:write` fails with HTTP 403 because Fleet RPCs use POST. Run `gcx login` again to replace that token.

### DIFF
--- a/cmd/gcx/fail/convert.go
+++ b/cmd/gcx/fail/convert.go
@@ -1171,8 +1171,8 @@ func convertFleetHTTPErrors(err error) (*gcxerrors.DetailedError, bool) {
 	if httpErr.Status == http.StatusNotFound && fleet.IsPluginMissingBody(httpErr.Body) {
 		return &gcxerrors.DetailedError{
 			Parent:  err,
-			Summary: "Fleet Management plugin not available",
-			Details: "The grafana-collector-app plugin is not installed or not enabled on this stack",
+			Summary: "Endpoint not available",
+			Details: "The " + fleet.CollectorAppID + " plugin is not installed or not enabled on this stack",
 			Suggestions: []string{
 				"Check the plugin and your permissions: gcx setup status",
 				"Install or enable the Collector app in Grafana: Administration > Plugins",
@@ -1201,8 +1201,8 @@ func convertFleetHTTPErrors(err error) (*gcxerrors.DetailedError, bool) {
 			Summary: "Authorization failed",
 			Details: "HTTP 403 from " + httpErr.Path,
 			Suggestions: []string{
-				"Read commands need the grafana-collector-app:read action on this stack",
-				"Write commands need the Admin role, or the grafana-collector-app:admin action",
+				"Named read routes need the " + fleet.CollectorAppReadAction + " action on this stack",
+				"Wildcard routes need the Admin role, or the " + fleet.CollectorAppAdminAction + " action; some read-only commands use these routes",
 				"Check what your login holds: gcx setup status",
 			},
 			DocsLink: docs.RolesAndPermissions,
--- a/cmd/gcx/fail/convert.go
+++ b/cmd/gcx/fail/convert.go
@@ -1168,7 +1168,7 @@ func convertFleetHTTPErrors(err error) (*gcxerrors.DetailedError, bool) {
 	// Grafana returns this when the collector app plugin is absent or disabled.
 	// It arrives as a 404, the same status Fleet Management uses for an absent
 	// resource, so the body decides.
-	if strings.Contains(httpErr.Body, fleet.PluginRouteMissingMarker) {
+	if httpErr.Status == http.StatusNotFound && fleet.IsPluginMissingBody(httpErr.Body) {
 		return &gcxerrors.DetailedError{
 			Parent:  err,
 			Summary: "Fleet Management plugin not available",
--- a/cmd/gcx/fail/convert.go
+++ b/cmd/gcx/fail/convert.go
@@ -1105,34 +1105,6 @@ func convertCloudConfigErrors(err error) (*gcxerrors.DetailedError, bool) {
 		}, true
 	}
 
-	// Fleet API scope error on read operations.
-	if strings.Contains(msg, "fleet:") && strings.Contains(msg, "invalid scope") &&
-		(strings.Contains(msg, "list ") || strings.Contains(msg, "get ")) {
-		return &gcxerrors.DetailedError{
-			Parent:  err,
-			Summary: "Fleet Management: permission denied",
-			Suggestions: []string{
-				"Ensure your cloud token's access policy includes the fleet-management:read scope",
-			},
-			DocsLink: docs.AccessPolicies,
-			ExitCode: new(gcxerrors.ExitAuthFailure),
-		}, true
-	}
-
-	// Fleet API scope error on write operations.
-	if strings.Contains(msg, "fleet:") && strings.Contains(msg, "invalid scope") &&
-		(strings.Contains(msg, "create ") || strings.Contains(msg, "update ") || strings.Contains(msg, "delete ")) {
-		return &gcxerrors.DetailedError{
-			Parent:  err,
-			Summary: "Fleet Management: permission denied",
-			Suggestions: []string{
-				"Ensure your cloud token's access policy includes the fleet-management:write scope",
-			},
-			DocsLink: docs.AccessPolicies,
-			ExitCode: new(gcxerrors.ExitAuthFailure),
-		}, true
-	}
-
 	// Adaptive Traces scope errors.
 	if strings.Contains(msg, "adaptive-traces:") && strings.Contains(msg, "invalid scope") {
 		return &gcxerrors.DetailedError{
@@ -1162,21 +1134,6 @@ func convertCloudConfigErrors(err error) (*gcxerrors.DetailedError, bool) {
 		}, true
 	}
 
-	// Fleet management not available.
-	if strings.Contains(msg, "fleet management endpoint is not available") ||
-		strings.Contains(msg, "fleet management instance ID is not available") {
-		return &gcxerrors.DetailedError{
-			Summary: "Fleet Management not available",
-			Details: msg,
-			Parent:  err,
-			Suggestions: []string{
-				"Fleet Management may not be enabled for this stack",
-				"Contact Grafana Cloud support to enable Fleet Management",
-			},
-			DocsLink: docs.FleetManagement,
-		}, true
-	}
-
 	// Stack info lookup forbidden — access policy missing stacks:read scope.
 	if strings.Contains(msg, "failed to get stack info for") && strings.Contains(msg, "status 403") {
 		suggestions := []string{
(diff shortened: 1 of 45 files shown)


internal/fleet/config.go
```diff
@@ -2,67 +2,142 @@ package fleet
...
+// fetchStackInfo reads the grafana.com instance record for the stack through the
+// collector app plugin proxy. The result is memoized per host, because several
+// commands load the client more than once in a run.
+func fetchStackInfo(ctx context.Context, host string, httpClient *http.Client) (cloud.StackInfo, error) {
+	v, _ := stackInfoCache.LoadOrStore(host, &stackInfoEntry{})
+	entry, ok := v.(*stackInfoEntry)
+	if !ok {
+		return cloud.StackInfo{}, fmt.Errorf("fleet: unexpected stack info cache entry for %q", host)
+	}
+	entry.once.Do(func() {
```
radiohead: issue – this will cause any transient error (e.g. a network timeout) to permanently "poison" an otherwise healthy context.

Consider reusing the logic from `internal/config/stack_id.go` – it caches only successes for exactly this reason.

Also, it looks like caching isn't even necessary, since every call site I could find only load the client once per `RunE`.

cmd/gcx/setup/preflight.go
```diff
@@ -0,0 +1,179 @@
...
+	collectorAppID = "grafana-collector-app"
+
+	pluginSettingsPath = "/api/plugins/" + collectorAppID + "/settings"
+	userActionsPath    = "/api/access-control/user/actions"
+
+	// actionRead is the action a read command needs. actionAdmin is the action
+	// every write command needs, because a write matches the wildcard route of
+	// the plugin.
+	actionRead  = collectorAppID + ":read"
+	actionAdmin = collectorAppID + ":admin"
```
radiohead: issue – collector app actually mixes up different actions per-route, i.e. some read-only actions need `admin` permissions:
```
$ gcx setup status --context dev
PRODUCT           ENABLED  HEALTH    DETAILS
fleet-management  yes      degraded  read only; write commands need the grafana-collector-app:admin action
instrumentation   no       unknown   the check failed: RunK8sMonitoring: fleet: HTTP 403 ...{"message":"plugin proxy route access denied"}
```

And also:
```
$ gcx fleet tenant get-limits --context dev
Error: Authorization failed
├─ Suggestions:
│ • Read commands need the grafana-collector-app:read action on this stack   ← login already HAS :read
│ • Write commands need the Admin role, or the grafana-collector-app:admin action  ← this is the actual fix
```

So we need to put together more elaborate checks here.

cmd/gcx/setup/command.go
```diff
@@ -98,33 +101,86 @@ func newStatusCommand(loader *providers.ConfigLoader) *cobra.Command {
...
+			state, err := checkCollectorApp(ctx, cfg, httpClient)
+			if err != nil {
+				return fmt.Errorf("setup: %w", err)
+			}
+
+			products := []setupProductStatus{state.row()}
 
-			status := newSetupStatus([]setupProductStatus{
-				{
+			if !state.mayServe() {
```
radiohead: question – should we treat disabled collector app plugin as a failure and return an error here instead?

docs/adrs/fleet-plugin-proxy/001-fleet-via-collector-app-proxy.md
```diff
@@ -0,0 +1,91 @@
...
+- `LoadClient` performs no stack lookup at all. It used to pay for one and
+  discard the result.
+- The proxy holds the Fleet Management credential, so gcx never handles it.
+
+Harder:
+
+- gcx now depends on the plugin. A stack without the `grafana-collector-app`
+  plugin cannot run these commands. Self-hosted Grafana never could, because
+  Fleet Management is a Grafana Cloud product.
+- Every write needs the Admin role or the `grafana-collector-app:admin` action.
```
radiohead: question – do we need a migration guide here? What's our policy on breaking changes?

internal/fleet/config.go
```diff
@@ -2,67 +2,142 @@ package fleet
...
 	"github.com/grafana/gcx/internal/cloud"
-	"github.com/grafana/gcx/internal/providers"
+	"github.com/grafana/gcx/internal/config"
+	"k8s.io/client-go/rest"
 )
 
-// ConfigLoader can load Grafana Cloud configuration from the active context.
+const (
+	// pluginID is the Grafana app plugin that proxies the Fleet Management API.
+	pluginID = "grafana-collector-app"
```
radiohead: suggestion – `pluginID` is redeclared in `cmd/gcx/setup/preflight.go` and `cmd/gcx/fail/convert.go`, consider making it exported here and re-using in those other two places.

cmd/gcx/setup/preflight.go
radiohead, line 1: suggestion – this file is quite heave on domain-specific logic, consider moving this to `internal/fleet` and leaving only the presentation logic here.

radiohead (commented):
LGTM overall with a few minor issues.

internal/fleet/config.go
```diff
@@ -2,67 +2,142 @@ package fleet
...
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return cloud.StackInfo{}, &HTTPError{
+			Status: resp.StatusCode,
+			Path:   stackInfoPath,
+			Body:   ReadErrorBody(resp),
+		}
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
wardbekker: Fixed in 7139590f. Stack information and preflight responses now use a 1 MiB limit. Fleet diagnostic bodies are also capped.

internal/fleet/config.go
wardbekker: Fixed in 7139590f. I removed the cache. Each `LoadClientWithStack` call performs one lookup. A regression test proves that a transient failure does not affect the next call.

docs/adrs/fleet-plugin-proxy/001-fleet-via-collector-app-proxy.md
wardbekker: Addressed in 7139590f. The Unreleased changelog now marks this as a breaking authentication change and gives the required login and plugin actions. We chose a changelog notice instead of a separate migration page.

cmd/gcx/setup/command.go
wardbekker: Fixed in 7139590f. A missing or disabled plugin emits the complete status document and returns general error code 1. Code 4 remains for genuine partial aggregate failures.

cmd/gcx/setup/preflight.go
wardbekker: Fixed in 7139590f. Setup now reports named read routes and wildcard admin routes instead of read and write intent. The 403 guidance also states that some read-only commands need the admin action.

internal/fleet/config.go
wardbekker: Fixed in 7139590f. `internal/fleet` now exports the collector app ID and both action names. Setup and error conversion reuse them.

cmd/gcx/fail/convert.go
```diff
@@ -1198,26 +1155,44 @@ func convertCloudConfigErrors(err error) (*gcxerrors.DetailedError, bool) {
...
 		return nil, false
 	}
 
+	// Grafana returns this when the collector app plugin is absent or disabled.
+	// It arrives as a 404, the same status Fleet Management uses for an absent
+	// resource, so the body decides.
+	if httpErr.Status == http.StatusNotFound && fleet.IsPluginMissingBody(httpErr.Body) {
+		return &gcxerrors.DetailedError{
+			Parent:  err,
+			Summary: "Fleet Management plugin not available",
```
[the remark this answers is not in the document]
wardbekker: Fixed in 7139590f. The converter now uses the approved `Endpoint not available` summary.

cmd/gcx/setup/preflight.go
wardbekker: Fixed in 7139590f. Plugin detection, action checks, response limits, and domain state now live in `internal/fleet/preflight.go`. `cmd/gcx/setup` contains presentation logic only.

radiohead (dismissed):

dafydd-t (approved):

PR #1266, opened 2026-09-04
feat(config): add trusted credentials.keychain policy

radiohead:
## What

Adds `credentials.keychain: on|off`, a configuration field that turns off OS keychain credential storage and persists credentials in the mode-`0600` config file instead. The default is `on`. `GCX_KEYCHAIN` keeps highest precedence.

The field is honoured only from trusted layers: system config, user config, or a config file selected deliberately through `--config` or `GCX_CONFIG`. An automatically discovered repository-local config cannot set it. Using it there is ignored and warns once per invocation, naming the trusted places to set it instead.

`gcx config set credentials.keychain` and `gcx config unset credentials.keychain` run as one locked transaction through `MutateKeychainPolicy`. The field is self-referential — it decides which credential store the write itself uses — so the load, the mutation, and the write all run under the *intended* post-mutation policy inside a single lock. Disabling storage never probes the previous store; enabling it stages any plaintext credentials and the policy update in the same atomic write.

Three supporting fixes, each in its own commit:

- The resolved policy is carried through owner reloads and OAuth refresh persistence, so an asynchronous refresh cannot re-derive a different answer from its own bytes.
- The legacy-migration reread is bound to the identity of the config source whose lock is held.
- `loadForWrite` resolves the policy across all discovered sources rather than only the `--file` target, so `--file system` honours a `keychain: off` declared in the user layer instead of pushing secrets into a store a trusted layer opted out of.

Adds no context keys. The policy travels as a `loadOptions` field. The exported entry point is `LoadUnderResolvedPolicy`, which takes a whole `Config` rather than a policy value, so a caller cannot manufacture a plaintext opt-out without first loading trusted configuration.

Based on the `refactor/config-load-write-options` branch and must merge after it.

## Why

Some environments have no OS credential store at all. CI runners and headless machines cannot rely on one, and a user disabling it deliberately needs that choice to be explicit and durable rather than an accidental side effect of a probe failure.

The trust rule exists because a checked-in repository config must not be able to downgrade another developer's credential storage to plaintext just by being discovered. That would turn a per-project convenience file into a security control nobody asked for.

Resolution has to happen before anything is loaded, because the credential store must be constructed before a config file can be read. That ordering constraint, not the field itself, is what the machinery here is for.

An earlier version of this change added four context keys and an exported function with one caller to carry the resolved policy across reloads. Rebasing onto the explicit-options refactor removed all of that: the four keys became one option field, and the single-caller export was replaced by one with two callers. The policy-propagation work dropped from roughly 1,460 lines to under 200.

## How it was tested

- [x] `GCX_AGENT_MODE=false mise run gate` passed — lint 0 issues, full suite green, build succeeded
- [x] `GCX_AGENT_MODE=false mise run reference` regenerated with no drift; `git status` clean afterwards
- [x] `GCX_AGENT_MODE=false go test ./internal/config/... ./internal/credentials/... ./internal/login/... ./cmd/gcx/config/... -count=1` passed, plus `-count=2` and `-shuffle=on` on the two config packages
- [x] Every new test verified to fail with its production change reverted
- [x] The `--file local` guardrail verified end to end against a built binary: the refusal now recommends `GCX_CONFIG`, exporting it and re-running writes the value, and a later invocation with no flag reads `{"credentials":{"keychain":"off"}}` with no warning, confirming the advice persists
- [x] `TestLoadForWriteHonoursHigherLayerPolicyOverTargetLayer` covers the `--file system` case honouring a user-layer opt-out
- [x] Confirmed the context-key set in `internal/config` is byte-identical to the base branch, with no new `context.WithValue` in the package diff
- [x] Confirmed `ContextWithResolvedKeychainPolicy` and the four context keys the earlier version added are all absent
- [x] Exported `Load`, `Write`, `LoadLayered`, and `LoadForWrite` signatures confirmed byte-identical to `main`
- [x] Confirmed the fail-closed behaviour change contributes zero lines to this branch — it ships separately
- [ ] The plaintext-fallback prose here matches current `main`. Once the fail-closed change lands, its "No automatic plaintext fallback" and "Keychain unavailable" documentation sections need reinstating; no repository check covers hand-written prose, so this is tracked rather than gated

## Related

- Follow-up to #1242

---
**Generated by Claude Code**


dafydd-t:
> The setting is honoured only from the user, system, or an explicit (GCX_CONFIG) config layer. An auto-discovered repository-local config cannot downgrade credential storage to plaintext; using the field there is ignored and produces a warning naming the ignored value and the trusted places to set it instead.


This was the reason i skipped adding it to config: the fact that it couldn't be set at all layers.

Does it warn on every invocation in this case? I'd be OK with it if it makes a lot of noise - i dont want it to be a warning we see once and is easily ignored.

radiohead:
> > The setting is honoured only from the user, system, or an explicit (GCX_CONFIG) config layer. An auto-discovered repository-local config cannot downgrade credential storage to plaintext; using the field there is ignored and produces a warning naming the ignored value and the trusted places to set it instead.
> 
> This was the reason i skipped adding it to config: the fact that it couldn't be set at all layers.
> 
> Does it warn on every invocation in this case? I'd be OK with it if it makes a lot of noise - i dont want it to be a warning we see once and is easily ignored.

It should warn on every invocation, yes. FWIW we could update the logic to allow disabling keychain at the local level too, but I believe this is a security risk (i.e. repo-local, checked-in config could disable keychain usage if user chooses to update contexts defined outside of the local config).

Also, I'm working on cleaning this PR up – I would like to merge your fix first, since it's small, targeted and addresses the issue, with this PR being a follow-up, adding support for disabling keychain via config files.

internal/config/path.go
```diff
@@ -40,6 +40,18 @@ func ValidateConfigPath(cfg Config, path string) (string, error) {
...
+		case "keychain":
+			// cmd/gcx/config/command.go intercepts "credentials.keychain"
+			// before validation runs (set and unset both need the locked
+			// keychain-policy transaction, not the generic path), so this
+			// arm never fires from the CLI today. Kept as defense in depth
+			// so ValidateConfigPath is correct for any other caller.
+			return path, nil
+		case "":
+			return "", fmt.Errorf("invalid path %q: credentials has no directly settable field; use credentials.keychain", path)
+		}
```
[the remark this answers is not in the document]
radiohead: Applied in 990e820d, and it was worth slightly more than a nit.

The fall-through did not just omit `credentials.keychain` — it produced `unknown path "credentials.keychan": paths start from a top-level section (stacks.<name>., cloud.<entry>., contexts.<name>., resources., current-context)`. That tells the user their top-level section is wrong and lists sections excluding `credentials`, when `credentials` is valid and the leaf is the typo. Actively misdirecting rather than merely unhelpful.

Added the `default:` arm as suggested. It also catches the over-deep case (`credentials.keychain.extra`), which fell through the same way.

Two existing tests pinned the old message (`{"credentials.unknown", "top-level section"}` and the same for `.keychain.extra`); both now pin the new one. Mutation-checked: removing the arm fails both with the old string.

---
**Generated by Claude Code**

awsome-o (commented):
I'd want the storage issue below fixed before merge.
The existing tests pass, but default config mutations reproduce the failure.
The one line fix passed the reproductions and the config suite.

internal/config/loader.go
```diff
@@ -1142,6 +1209,15 @@ func loadLayered(ctx context.Context, explicitFile string, opts loadOptions, ove
 	if err := preflightLayeredSources(sources, &hasLegacyLayer); err != nil {
 		return Config{}, err
 	}
+	// Resolve the credential-storage policy once, from every trusted layer,
+	// and bind it to the parent options so each per-layer copy below inherits
+	// it. A layer that re-derived the policy from its own bytes would obey a
+	// different decision than the load it is part of.
+	policy, err := resolveKeychainPolicy(ctx, sources)
```
awsome-o: Enabling the keychain can report success while leaving the token in plaintext. With a single discovered user config containing `credentials.keychain: off`, I reproduced this with `gcx config set credentials.keychain on` and `gcx config unset credentials.keychain`. A command test with a fake store also shows the reverse: setting `off` moves the token into the keychain, where the new policy prevents reading it. I would want this fixed before merge.

In [loader.go:1216](https://github.com/grafana/gcx/blob/990e820d3ec76433801572549ed9cd2c8b37a81e/internal/config/loader.go#L1216), `loadLayered` resolves the old file policy again and replaces the intended policy in `opts`. The later assignment updates `cfg.keychainPolicy` but leaves the old store attached. Could this use `resolveKeychainPolicyForSources(ctx, opts, sources)` and add set/unset cases without `--config` or `--file`? That one line change passed all four reproductions locally, including enabling with an unavailable store.

```suggestion
	policy, err := resolveKeychainPolicyForSources(ctx, opts, sources)
```

internal/config/loader.go
```diff
@@ -100,6 +100,16 @@ func keychainStoreForMode(mode keychainMode) credentials.Store {
 	return openedStore
 }
 
+// keychainStoreForPolicy is the sole selector for which credential store
+// backs a resolved policy: the already-resolved policy decides disabled vs.
+// enabled, so nothing here re-reads GCX_KEYCHAIN.
+func keychainStoreForPolicy(policy keychainPolicy) credentials.Store {
+	if policy.mode == keychainModeDisabled {
+		return disabledStore{}
```
awsome-o, line 108: Someone who disables storage through `credentials.keychain: off` gets recovery advice for an environment variable they never set. I reproduced `gcx config check` reporting `keychain use is disabled by GCX_KEYCHAIN` with that variable unset. The plaintext warning and deletion error also tell them to unset it, which leaves the configured policy unchanged.

Returning `disabledStore` here now covers both sources, but the messages in [keychain.go:1006](https://github.com/grafana/gcx/blob/990e820d3ec76433801572549ed9cd2c8b37a81e/internal/config/keychain.go#L1006) and lines 1484 and 1522 still assume the environment variable was responsible. Could those messages describe the effective policy and give a repair that works when the value comes from a config file? Fine to handle as a followup.

docs/sources/keychain.md
```diff
@@ -28,64 +28,91 @@ copied file separately.
...
-`off` is the only value that disables the credential store. gcx keeps using the
-store for every other value, so a typo cannot write credentials in plaintext
-without your intent. gcx warns once per command when it ignores a value:
+The precedence, highest first, is:
 
-```
-warn: GCX_KEYCHAIN="disabled" is not a recognized value and was ignored; the OS credential store is still in use. Set GCX_KEYCHAIN=off to disable it.
-```
+1. `GCX_KEYCHAIN`
+2. A deliberately selected file (`--config` or `GCX_CONFIG`)
```
awsome-o: A user with `credentials.keychain: off` in their user config could read this list as saying that `--config other.yaml` inherits that value when the selected file omits the field. It actually resolves to the default `on`, because selecting an explicit file bypasses the user and system layers entirely ([loader.go:1177](https://github.com/grafana/gcx/blob/990e820d3ec76433801572549ed9cd2c8b37a81e/internal/config/loader.go#L1177)).

Could we spell out the two cases here: with an explicit file, environment then selected file then default; otherwise environment then user then system then default? The same wording in `docs/architecture/config-system.md:189` could use that clarification too. Fine to handle as a documentation followup.

awsome-o (commented):
Two UX followups from the review. Both can be handled as followups; the storage issue already posted remains the merge blocker.

internal/config/loader.go
radiohead: Good catch, addressed!

internal/config/loader.go
radiohead: Updated the error messages, let me know if they make more sense now.

docs/sources/keychain.md
radiohead: Fixed!

dafydd-t (approved):

PR #1243, opened 2026-08-25
test(cli): enforce the command naming conventions in CI

dafydd-t:
Follows the verb taxonomy work in #387.

A test which fails if someone adds a new command operation that is not in the "canonical" list. authors can add a new non-standard command to the golden file (a new addition in this PR too), but they will need gcx team's approval to do so. 

We want to delegate more code ownership to product teams, so tools like this help us defend against sprawl.

PR #1267, opened 2026-09-04
feat(agento11y): pull experiment data

portertech:
## Summary

- add the experimental `gcx agento11y experiments pull <run-id>` artifact command
- pull experiment metadata, the aggregate report (including per-trial scores and artifact metadata), every trial page, and indexed conversation IDs by default
- optionally download full conversation payloads with `--include-conversations`
- preserve every successful API response body byte-for-byte and inventory it in a checksummed manifest
- generate an `AGENTS.md` with privacy, prompt-injection, integrity, redaction, and cleanup guidance
- generate a deny-all `.gitignore` so bundle contents are excluded from Git by default
- fetch deduplicated conversations with bounded concurrency while keeping memory bounded to active requests
- publish through a private staging directory with atomic no-replace semantics

## Usage

```bash
# Experiment, aggregate report, trial pages, and conversation IDs
gcx agento11y experiments pull <run-id> -d ./exports/<run-id>

# Also download full conversation payloads
gcx agento11y experiments pull <run-id> -d ./exports/<run-id> --include-conversations
```

## Output layout

```text
<output-dir>/
  AGENTS.md
  .gitignore
  manifest.json
  indexes/trials.jsonl
  raw/experiment.json
  raw/report.json
  raw/trial-pages/*.json
  raw/conversations/*.json  # only populated with --include-conversations
```

The manifest's `includes.conversations` field makes the requested scope explicit. Conversation filenames are derived from SHA-256 IDs; the manifest and trial index retain the ID-to-path mapping. Generated safety files are also inventoried and checksummed.

## Safety and failure behavior

- requires a destination that does not already exist
- verifies atomic no-replace directory publication support before downloading; if an unexpected final publication failure occurs while the destination remains absent, the completed private staging bundle is retained and its path is reported
- creates directories with `0700` permissions and files with `0600` permissions
- treats all pulled data—not only conversation text—as sensitive and potentially instruction-bearing
- tells agents to use only approved runtimes/providers, treat data as inert, avoid external services/subagents, redact derived output, validate manifest sizes/checksums, and clean up afterward
- documents that `complete: true` records success for the requested scope rather than current file integrity, and that checksums detect changes but do not authenticate the bundle
- ignores the complete bundle directory from Git by default
- records per-conversation failures in both the manifest and artifact receipt when conversations are included
- relies on gcx's shared bounded retry transport for 429 and transient 502/503/504 responses, including `Retry-After`
- returns exit code 4 for mixed conversation-fetch results and exit code 1 when every requested conversation fails
- preserves cancellation as exit code 5 and removes canceled staging data
- leaves fine-tuning schema conversion out of scope so raw provider-specific generation fields remain available for later curation

`AGENTS.md` and `.gitignore` provide defense in depth; they are not substitutes for an approved agent runtime/model provider or other host security controls.

## Validation

- `GCX_AGENT_MODE=false go test -race ./internal/providers/agento11y/eval/experiments -count=1`
- cross-compiled the experiments tests for Linux, Windows, and FreeBSD

## Review note

The command is intentionally marked experimental until Agent Observability owners attest backend pagination/rate limits and large-run behavior.


internal/providers/agento11y/eval/experiments/publish_other.go
```diff
@@ -0,0 +1,11 @@
+//go:build !darwin && !linux && !windows
+
+package experiments
+
+import "os"
+
+// os.Rename rejects an existing directory on Go's supported platforms. Linux,
+// macOS, and Windows use stronger kernel-level no-replace operations above.
```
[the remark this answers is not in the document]
portertech: Fixed in 849150e9. The fallback comment now states that the caller performs the absence check and that os.Rename cannot prevent replacement of a concurrently created empty directory.

jgordley (dismissed):

internal/providers/agento11y/eval/experiments/export.go
```diff
@@ -0,0 +1,660 @@
...
+	if o.Concurrency < 1 {
+		return fmt.Errorf("invalid --concurrency value %d: must be at least 1", o.Concurrency)
+	}
+	return nil
+}
+
+func newExportCommand(loader *providers.ConfigLoader) *cobra.Command {
+	opts := &exportOpts{}
+	cmd := &cobra.Command{
+		Use:   "export <run-id>",
```
awsome-o: Picking up the naming thread: `export` trips a naming gate that landed after your merge base. `TestConsistency_NoNewCommandOperationsAdded` fails with `leaf command "gcx agento11y experiments export" does not end in a canonical operation`, and the frozen v1 verb list in cmd/gcx/root/commandoperations_test.go:17 doesn't have `export` in it.

I think `pull` is what you want, and it also answers Daf's question about whether `export` on its own is too vague. docs/design/command-naming.md calls `pull` "the manifest read/export counterpart: it writes remote resources to local files", which is exactly this. `gcx resources pull`, `gcx slo definitions pull` and `gcx slo reports pull` all do the same job, and the two SLO ones already take `-d/--output-dir` the way you have it. Keeping `export` is possible but it means adding the path to cmd/gcx/root/testdata/non_canonical_command_operations.json, which is an explicit exception to a surface CONSTITUTION froze. Does `gcx agento11y experiments pull <run-id> -d <dir> --include-conversations` read alright to you?

internal/providers/agento11y/eval/experiments/pull.go
```diff
@@ -0,0 +1,660 @@
...
+instructions and a .gitignore that excludes the entire bundle from Git by
+default.`,
+		Example: `  # Export experiment metadata, aggregate report, trials, and conversation IDs.
+  gcx agento11y experiments export <run-id> -d ./exports/run-1
+
+  # Also download every referenced conversation with reduced request pressure.
+  gcx agento11y experiments export <run-id> -d ./exports/run-1 --include-conversations --concurrency 4`,
+		Args: exactArgsWithSuggestion(1, "gcx agento11y experiments export <run-id> -d <directory>"),
+		Annotations: map[string]string{
+			agent.AnnotationStability: agent.StabilityExperimental,
```
awsome-o, line 150: Heads up that a rebase turns CI red here, and not for anything you did. Two gate files landed on main on 7 September (7952f70c for #1220, bdd7c4cd for #1243) and your merge base is from 4 September, so your green run never saw either of them. I applied the branch onto current main to check.

Since the command carries `agent.StabilityExperimental`, `TestExperimental_AnnotatedCommandsAreAdvertised` wants `[experimental] ` on the front of Short, and `TestExperimental_LongDescriptionCarriesPreamble` wants Long to open with "This command is experimental. It may be removed, or its subcommands, flags and responses may change without following the normal semantic versioning conventions." internal/datasources/tempo/baseline.go:91 has the shape. Both change the generated help, so the files under docs/reference/cli need a regen or `reference-drift` catches it.

awsome-o (commented):
Really nice piece of work, the manifest and the generated AGENTS.md especially. Two things I'd want sorted before merge: a rebase turns CI red on two gate tests that landed after your merge base, and `export` isn't in the frozen verb set so the naming gate rejects it outright.

On Daf's question I think `pull` is the answer and it's the one the repo already uses for this. Everything else below is take or leave, though the publish step and the missing scores are the two I'd actually think about.

internal/providers/agento11y/eval/experiments/publish_linux.go
```diff
@@ -0,0 +1,15 @@
+//go:build linux
+
+package experiments
+
+import "golang.org/x/sys/unix"
+
+func publishDirectoryNoReplace(stagingDir, outputDir string) error {
+	return unix.Renameat2(
```
awsome-o, line 8: I think someone writing the output to NFS, or to /mnt/c under WSL2, or to an SMB share loses the entire export at the very last step. `RENAME_NOREPLACE` isn't universally supported: the Linux NFS client rejects renameat2 flags with EINVAL, kernels before 3.15 give ENOSYS, and on darwin `RENAME_EXCL` comes back ENOTSUP on SMB and exFAT. export.go:399 wraps that as `publish output directory`, and the deferred RemoveAll then discards everything that was just fetched, which on a big run with `--include-conversations` is a lot of downloading.

publish_other.go already concedes that plain os.Rename is an acceptable degradation, so letting EINVAL, ENOSYS and ENOTSUP fall through to the same path seems consistent with what you already decided. Separately, and take or leave this one: would `os.Mkdir(outputDir, 0o700)` up front do the job these four files are doing? It's an atomic exclusive create everywhere, so it closes the same window requireMissingDirectory leaves open at export.go:159 and the two syscall dependencies go away. The tradeoff is that the destination is briefly visible while incomplete, which manifest.json already covers, since it's written last and the generated AGENTS.md tells readers to check it.

internal/providers/agento11y/eval/experiments/export.go
```diff
@@ -0,0 +1,660 @@
...
+
+	receipt := cmdio.NewArtifactReceipt("exported", "json")
+	receipt.Dir = outputDir
+	receipt.Files = append(receipt.Files, cmdio.ArtifactFile{
+		Path:  outputDir,
+		Kind:  "AgentO11yExperimentExport",
+		Count: len(trials),
+	})
+	receipt.Summary = cmdio.MutationSummary{
+		Succeeded: len(pathByConversation),
```
awsome-o: An agent that runs this without `--include-conversations` gets `"summary":{"succeeded":0,"failed":0}` back from a completely successful export, which reads as nothing happened. `Succeeded: len(pathByConversation)` only counts conversation payloads, so the default mode counts nothing even though the experiment, the report, the trial pages and the index all got written.

internal/output/mutation.go:41 puts the invariant as "every matched target is counted exactly once", so counting trials or written files would sit better. Same thought on the Files entry just above: cmd/gcx/resources/pull.go:252 appends one ArtifactFile per directory written with a real count for that directory, whereas here Path repeats receipt.Dir and Count is a trial count, which isn't what ArtifactFile.Count documents itself as.

internal/providers/agento11y/eval/experiments/pull.go
```diff
@@ -0,0 +1,660 @@
...
+	g.SetLimit(concurrency)
+	for i, id := range conversationIDs {
+		if ctx.Err() != nil {
+			break
+		}
+		g.Go(func() error {
+			if err := ctx.Err(); err != nil {
+				return err
+			}
+			body, fetchErr := fetchRawJSON(ctx, base, "/query/conversations/"+url.PathEscape(id))
```
awsome-o, line 444: This is the rate limit question from your description, and I think the honest answer is that a single 429 permanently drops that conversation from the bundle. fetchRawJSON returns the error, line 437 stores it as a terminal fetchedConversation.err, and the bundle publishes with `complete: false`. At the default `--concurrency 10` against a plugin API that throttles, I'd expect to hit that on any large run.

What makes it expensive is that there's no way back into it. requireMissingDirectory refuses an existing destination, so recovering means deleting the directory and refetching every conversation, into the same throttle. Would a bounded retry honoring `Retry-After` on 429 and on 502/503/504 be enough here, or would you want a resume path too?

internal/providers/agento11y/eval/experiments/pull.go
```diff
@@ -0,0 +1,660 @@
...
+
+type fetchedConversation struct {
+	id        string
+	path      string
+	hash      string
+	sizeBytes int64
+	err       error
+}
+
+func exportExperimentBundle(ctx context.Context, base *agento11yhttp.Client, runID, outputDir string, includeConversations bool, concurrency int) (*experimentExportResult, error) {
```
awsome-o, line 289: For the training data use case you mentioned in the thread, I think the piece that's missing is scores. The bundle carries the experiment, the aggregate report, the raw trial pages and the conversations, but not `/eval/experiments/{id}/scores` (ListScores, agento11y client.go:243), not `/eval/test-case-trials/{id}/scores` (ListTrialScores, client.go:199) and not `/artifacts` (ListTrialArtifacts, client.go:205). The aggregate report has a row per test case but not the evaluator scores for an individual trial.

So curating "every trial the judge scored above 0.8" means a second `gcx agento11y experiments list-scores` run whose output isn't checksummed, isn't in the manifest and isn't joined to the trial index, which undercuts what the bundle is for. The manifest already declares `schema_version: "1"` and `includes` is an extensible shape, so is it worth putting `raw/scores/` and `includes.scores` in now rather than bumping the schema later?

internal/agent/command_annotations.go
```diff
@@ -577,6 +577,10 @@ var commandAnnotations = map[string]annotation{
 		Cost: "small",
 		Hint: "<run-id> -o json",
 	},
+	"gcx agento11y experiments export": {
+		Cost: "large",
+		Hint: "<run-id> --output-dir ./exports/<run-id> --include-conversations --concurrency 10",
+	},
 
 	"gcx agento11y experiments test-suites list":             {Cost: "small"},
```
awsome-o, line 588: Small thing: this hint tells an agent to pass `--include-conversations`, which works against the default Jack asked for in the thread. An agent that follows it pulls every transcript on the first call, which is both the expensive path and the one carrying the sensitive payloads the generated AGENTS.md warns about. Worth dropping it from the hint so it gets added only when the transcripts are actually wanted?

internal/providers/agento11y/eval/experiments/pull.go
portertech: I did not add redundant score/artifact requests because the premise is not true for the report contract: `raw/report.json` already contains each trial’s `final_score`, complete `scores` array, and `artifacts` metadata, and the body is preserved byte-for-byte. The captured API fixture in `client_test.go` confirms that shape. I made this explicit in help/skill docs and strengthened the raw-fidelity export fixture with score/artifact fields in f6990a6a.

internal/providers/agento11y/eval/experiments/export.go
portertech: Valid. Renamed the public command to the canonical `gcx agento11y experiments pull`, changed the receipt action to `pulled`, renamed the implementation file, and updated annotations, fixtures, examples, skills, and generated reference docs in f6990a6a.

internal/providers/agento11y/eval/experiments/export.go
portertech: Valid. The receipt now counts the experiment bundle as one successful target plus each successfully requested conversation; conversation failures remain failed targets, so default mode reports one success instead of zero. Human conversation counts use the manifest explicitly. The receipt file entry now points to the actual `manifest.json` and its count is the number of inventoried files. Fixed and tested in f6990a6a.

internal/agent/command_annotations.go
portertech: Valid. Removed `--include-conversations` (and the concurrency flag) from both the leaf command annotation hint and the experiments parent LLM hint. Agents now default to the ID-only pull and add conversation payloads only when explicitly needed. Fixed in f6990a6a.

internal/providers/agento11y/eval/experiments/pull.go
portertech: Valid. After rebasing onto current main, the command now has the required `[experimental]` Short prefix and exact Long preamble. The generated CLI reference was regenerated and the new naming/experimental gates pass in f6990a6a.

internal/providers/agento11y/eval/experiments/publish_linux.go
portertech: Valid concern. I kept the atomic no-replace guarantee rather than degrading to plain `os.Rename`: gcx now probes the destination filesystem before any API download and fails early when atomic publication is unsupported. If an unexpected final publication error occurs while the destination is still absent, the completed private staging bundle is retained and its path is reported instead of deleting the download. Destination-race cleanup is unchanged. Added focused tests in f6990a6a.

internal/providers/agento11y/eval/experiments/pull.go
portertech: The retry concern is already covered by gcx’s shared transport: ConfigLoader installs `retry.Transport` in the REST config, `rest.HTTPClientFor` applies that wrapper for AgentO11y, and GETs are retried for 429/502/503/504 with bounded backoff and `Retry-After` support. I added an export-path test proving a 429 conversation request is retried successfully. I did not add a second retry loop or resume mode; partial failures remain explicit and auditable.

dafydd-t (approved):

PR #1270, opened 2026-09-06
fix(auth): use an OS-assigned ephemeral port for the OAuth callback

DarkWanderer:
## Summary
- The OAuth callback server picked its local port by scanning a hardcoded `54321-54399` range one `bind()` at a time. Under WSL2 with mirrored networking, that range is shared with Windows' own ephemeral port pool (`49152-65535`), so any Windows process transiently holding a port in that band can exhaust the whole range and fail login with `no available port in range 54321-54399`, even though nothing is actually listening.
- Replaced the scan with a single `net.Listen` call on port `0`, letting the kernel hand back any free port — it can never collide the way scanning a fixed range can. Applies to both the instance login flow (`internal/auth/flow.go`) and the Grafana Cloud OAuth flow (`internal/auth/gcom.go`), which share the same helper.
- `--oauth-callback-port` (fixed port override) is unchanged.
- Updated the flag help text, CLI reference doc, and `docs/architecture/auth-system.md` (which still described the old fixed range).

## Test plan
- [x] `go test ./...` (full suite passes; `GCX_AGENT_MODE=false` needed locally due to `CLAUDECODE` env var auto-detection, same as `mise run tests` pins)
- [x] `go build ./...`, `go vet ./...`, `gofmt -l` all clean
- [x] Manually verified `gcx login --oauth` picks a real OS ephemeral port (e.g. `callback_port=45089`) and completes login end-to-end
- [x] Manually verified `--oauth-callback-port <port>` still forces the exact port
- [x] Regenerated `docs/reference/cli/gcx_login.md` via `scripts/cmd-reference`

https://claude.ai/code/session_01WwXwNwz6ccRKJt3kFSXeg8


DarkWanderer:
@wardbekker greetings from ex-Grafanista. Would appreciate a review if you have a moment. This fixes an issue of `gcx` refusing to run in WSL, which already has this port range busy. Also, "get free port through OS" approach is much more robust than a cycle.

PR #1274, opened 2026-09-07
refactor(output): collapse provider table codecs behind a generic

dafydd-t:
Part of https://github.com/grafana/gcx/issues/1275

## Summary

- Define an interface for the table codec: `output.Table[T]`
- Migrates two providers onto it as a pilot: `faro` apps, `synth/checks` checks
- Adds a golden-file helper so each migration pins its rendered bytes

This is an example use: `Table[T]` takes an ordered list of columns. it contains the header name whether the column should only appear in `-o wide`, and then a function to present the correct data: 

```go
func CheckTable() cmdio.Table[Check] {
	wide := []string{cmdio.FormatWide}

	return cmdio.Table[Check]{
		Columns: []cmdio.Column[Check]{
			{Header: "NAME", Content: checkDisplayName},
			{Header: "JOB", Content: func(c Check) string { return c.Job }},
			{Header: "ENABLED", Visibility: cmdio.WideOnly, Content: func(c Check) string { return strconv.FormatBool(c.Enabled) }},
		},
	}
}
```

I've also implemented the change to two providers in this PR: faro and synth/checks. There are "golden files" added for each, so we can ensure that table output is unchanged in these and future changes. The golden files are captured from the old codec.

`GCX_UPDATE_GOLDEN=1 go test ./...` rewrites the golden files, but is useful for adding new golden files for providers before migrating them to the interface

## How to review this PR

- the table codec interface is defined [here](https://github.com/grafana/gcx/pull/1274/changes#diff-e7db48870f37deac32208f8e0216ac90addb48069398608c346305845a601dc4R1)
- Then, check the two pilot provider migrations:
  - synth/checks: [golden file example](https://github.com/grafana/gcx/pull/1274/changes#diff-fafba0363d0eb7cb20051767d92001609c25fba642b0f0b87b8be8a9b70bf18d), [provider change](https://github.com/grafana/gcx/pull/1274/changes#diff-d1240eefb8a1b1ec0110a7327e7e33e09c29c5ea5c2bc73dd2972b1298bc63ba)
  - faro: [golden file exampel](https://github.com/grafana/gcx/pull/1274/changes#diff-6540a6e0025ae935d5b078d0b17045712bc64e084e861ba605f426619f3b1422), [provider diff](https://github.com/grafana/gcx/pull/1274/changes#diff-20b7d12c680b365aed4588953e3592480aae733a884a9be75a944512cf8ae73d)

## Scope

This PR adds the table codec interface only. we'll tackle the other codecz in their own PRs

I'll follow up this PR with PRs moving providers over to using this interface rather than their own implementation.

Once we've settled on the interface, I'll write the devision into an ADR.

`internal/providers/instrumentation/output` is next — it is the one package with cross-package codec consumers, so it will surface the visibility questions before the long tail.


dafydd-t:
@claude review

radiohead (approved):
Nice work, thank you!

Issue #1275, opened 2026-09-07
🚀 simplify the codec providers


dafydd-t:
### Your idea

Currently, each provider has its own implementation of each codec type, which means we have lots of repetition of codec implementations across all the providers. There is loads of repetition particularly across the table and text codecs. I'm talking about implementations like `successLineCodec` in  `internal/providers/faro/mutation_output.go`. There are ~150 Decode functions which are identical, and we have ~180 codec implementations!

I think we can simplify this by adding an interface for each of table and text codecs, and then get each implementation to use the interface instead.

As part of this, we can add some golden file tests for table output codecs, because we currently dont have those covered in tests at all.

### table codec implementations

There are 117 table codec implementations across 33 packages. Each one has a type with its own `Encode` calling `style.NewTable`, plus its own type assertion, `Format` and `Decode` methofs.

| Package | Table codecs | Files |
|---|---|---|
| `internal/providers/agento11y/agents` | 2 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/agents/commands.go) |
| `internal/providers/agento11y/conversations` | 3 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/conversations/commands.go) |
| `internal/providers/agento11y/eval/collections` | 1 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/collections/commands.go) |
| `internal/providers/agento11y/eval/evaluators` | 2 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/evaluators/commands.go), [test_command.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/evaluators/test_command.go) |
| `internal/providers/agento11y/eval/experiments` | 6 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/experiments/commands.go) |
| `internal/providers/agento11y/eval/guards` | 1 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/guards/commands.go) |
| `internal/providers/agento11y/eval/judge` | 2 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/judge/commands.go) |
| `internal/providers/agento11y/eval/rules` | 1 | [commands.go](https://github.com/grafana/gcx/blob/main/internal/providers/agento11y/eval/rules/commands.go) |
[...]

PR #1277, opened 2026-09-07
refactor(instrumentation): render tables through the generic codec

dafydd-t:
Part of #1275. Follows #1274, which is now merged, so this targets `main`.

We're moving more providers to using the `output.Table[T]` interface for table codecs.  This PR tackles the `gcx instrumentation` related providers.

## Summary

- `ClusterTableCodec`, `AppTableCodec` and `ServiceTableCodec` become `ClusterTable`, `AppTable` and `ServiceTable`, using column lists instead of separate implementations of `Encode`.
- Add `EncodeList` in the instrumentation provider, so we can present the un-enveloped data when its presenting to a table codec.
- Golden files for all three tables, captured from the old codecs
- Collapse the three list envelopes into one generic `ListEnvelope[T]`

## Notes

- `STATUS` renders differently depending on table width. In normal tables,  `STATUS` is normalised to `OK`/`FAILING`/`NODATA`, but in the wide tables it is left with the raw value. To maintain this, you can define the same "column" twice, once for each setting (see the `Visible: cmdio.TableOnly|WideOnly` field). Only 4 codecs do this, 3 of which are in this PR.

    ```go
    {Header: "STATUS", Visible: cmdio.TableOnly, Content: func(c ClusterView) string {
        return NormalizeStatus(c.InstrumentationStatus)
    }},
    {Header: "STATUS", Visible: cmdio.WideOnly, Content: func(c ClusterView) string {
        return string(c.InstrumentationStatus)
    }},
    ```
- this provider's output content is often an envelope, not a slice. JSON and YAML codecs want the `{"items": [...]}` structure. the provider now has common logic to either present the envelope, or the envelope's contents:
    ```go
    switch string(codec.Format()) {
    case cmdio.FormatTable, cmdio.FormatWide, cmdio.FormatText:
	    return codec.Encode(w, rows)
    default:
	    return opts.Encode(w, ListEnvelope[T]{Items: rows})
    }
    ```
- some providers dont call the narrow table output (`-o`) "table". So we make a slight change to the interface to add `RegisterTableAs`, which allows table codecs to register with a different name. 

    ```go
  cmdio.RegisterTable(&o.IO, instroutput.ServiceTable())                        // status:   "table"
  cmdio.RegisterTableAs(outOpts, instrumout.ServiceTable(), cmdio.FormatText)   // services: "text"
  ```
  We also change the visibility enum to be based on when something should be visible, rather than its name:

    ```go
    const (
	    VisibleAll Visibility = ""       // both
	    NarrowOnly Visibility = "narrow" // not wide, whatever the command calls it
	    WideOnly   Visibility = FormatWide
    )
    ```

- Output is unchanged. Like the example providers in the last PR, we add golden files of output from before the change, and encode them as testdata. This way we know the final table output is unchanged.

## How to review this PR

Four commits. Suggested order:

1. `internal/providers/instrumentation/output/codec.go` — the three column lists and `EncodeList`. The `STATUS` pairs are the interesting part.
2. `internal/providers/instrumentation/output/testdata/*.golden` — seven files pinning the current output; `clusters_table.golden` vs `clusters_wide.golden` shows the `STATUS` difference directly.
3. The `cmd/gcx/instrumentation` call sites — registration changes, plus `EncodeList` at the six list sites.
4. `internal/output/table.go` — the `Visibility` values, `RegisterTableAs`, and the `FormatText` constant.


dafydd-t:
@claude review

dafydd-t:
@claude review

dafydd-t:
@claude review

radiohead (approved):

PR #1279, opened 2026-09-07
feat(skills): improve the Grafana investigation workflow

javiermolinar:
Current skill does the job but it can improved. Previously it was a checklist, inspect logs, then dashboards then... Some of the guidance was also wrong like the error-ratio example.

The new version: 
 - Start with the user’s question and reuse supplied queries
 - Choose the next query based on what it would distinguish—not which signal comes next in a checklist.
 - Preserve scope, time windows, and measurement semantics.
 - Stop when the question is answered, or explain the specific evidence gap.


javiermolinar:
@claude review

dafydd-t (commented):
aware that this might make the agent more interactive, and therefore the skill wont work so well in evals, but we have other skills that encourage interaction, and that seems more useful to me. Thank you!

dafydd-t (approved):

PR #1280, opened 2026-09-07
docs: add guidelines for engineering teams in grafana

dafydd-t:
Closes https://github.com/grafana/gcx-internal/issues/29

We want to encourage product engineering teams in grafana to take ownership of their command areas in gcx, and to do that they need to be able to contribute independently. To compliment the existing .github/CODEOWNERS file, this addition to CONTRIBUTING.md exists to encode that practice somewhere we can reference to. We might need to iterate on the CODEOWNERS content, but this section should make it clear to those other engineering teams that they are enouraged to contribute independently, and without our explicit approval.

This follows similar patterns we see in e.g. grafana/grafana, grafana/terraform-provider-grafana, and so on.


awsome-o (approved):
this is great!!

PR #1283, opened 2026-09-08
feat(config): require explicit context with GCX_REQUIRE_CONTEXT

julvo712:
## Summary
- Adds opt-in `GCX_REQUIRE_CONTEXT` so gcx refuses commands that would silently inherit `current-context` from the config file. That field is shared across shells and coding agents; another session's `gcx config use-context` can otherwise retarget later invocations.
- An invocation satisfies the requirement with `--context <name>` (root or subtree-bound) or `GRAFANA_SERVER`. Local/meta commands stay exempt; `config check` and `instrumentation check --fix-plan=assistant` still require an explicit target.
- The guard runs pre-dispatch in `main()` rather than root `PersistentPreRun`, because Cobra only runs the closest hook and provider groups that define their own would otherwise skip it.

## Test plan
- [ ] `export GCX_REQUIRE_CONTEXT=true` then run `gcx slo definitions list` without `--context` — expect exit 2 and a message to pass `--context`
- [ ] Same with `gcx --context <name> slo definitions list` — command proceeds
- [ ] `GRAFANA_SERVER=https://example.grafana.net gcx resources list-types` satisfies the guard without `--context`
- [ ] `gcx version`, `gcx config view`, `gcx login`, `gcx agent skills list` still work without `--context`
- [ ] `gcx config check` without `--context` is refused; with `--context` it runs
- [ ] Unset `GCX_REQUIRE_CONTEXT` — existing `current-context` fallback is unchanged

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -160,6 +160,14 @@ credential-bearing stack/Cloud entries atomically, select `--context` or
 `GRAFANA_TOKEN`, `GRAFANA_PROVIDER_{NAME}_{KEY}`). Overrides take runtime
 precedence but remain ephemeral.
 
+**Strict context mode:** `GCX_REQUIRE_CONTEXT` drops the `current-context`
+fallback from that chain, so an invocation must name its target with `--context`
+or `GRAFANA_SERVER` or be refused with exit code 2. It protects workstations
+holding many contexts, where one session's `config use-context` would otherwise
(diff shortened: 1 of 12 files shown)


radiohead:
@claude review this PR

radiohead:
Hey @julvo712, thank you for submitting this! Let me know if you need help getting this to pass Claude's code review! 🙂

PR #1287, opened 2026-09-08
fix(output): Make --jq compatible with -o agents

javiermolinar:
Previously, gcx rejected this combination even though agents output is JSON. This change lets jq transform the full result before agents applies compact formatting and automatic spilling.

```bash
   gcx config list-contexts -o agents --jq '.contexts[0].name'
```

This now returns the first context’s name as compact JSON instead of an incompatible-output error.

Bare --jq and explicit -o json behavior remain unchanged.


javiermolinar:
@claude review

dafydd-t (approved):

PR #1292, opened 2026-09-08
feat(skills): refine the Grafana investigation workflow

javiermolinar:
This PR tries to refines the existing investigation skill to improve some pathological behaviors.  We evaluated the skill by asking an agent to investigate a historical production alert repetitively. 

Previously GCX used 4 turns by repeating this pattern:

1. fetch JSON into a file
2. read part of the file
3. write a python script to summarize it
4. interpret the summary

To solve this all the examples now use jq and agent mode by default. The other main change is to delete the scenarios that were mostly noise.

During refinement, the main investigation instructions were reduced from approximately 1,800 to 900 words, with detailed guidance kept in references loaded when needed.

These observations support stronger evidence gathering and more careful control selection with a smaller entry point


awsome-o:
@claude review

javiermolinar:
@claude review

dafydd-t (approved):

PR #1293, opened 2026-09-08
refactor(providers): migrate SLO and OnCall registration

radiohead:
## What

Migrates SLO definitions to `adapter.Resource[T]` and replaces OnCall's private registration builder with the shared builder from #931. Moves SLO's create/update-and-refetch behavior into its client and constructs fresh command trees for each CLI root.

Keeps the existing command surface, fail-safe dry-run behavior, and resource output. Provider tests stay with the migration, including coverage for equivalent adapter output and isolated Cobra command state. Updates the provider-authoring guidance to use the migrated examples.

## Why

This is the adoption step after #931's reusable machinery. Its diff contains only provider migrations, their tests, and the associated guidance, so reviewers can assess the behavior change separately from the new API.

## How it was tested

- [x] `GOFLAGS=-mod=readonly GCX_AGENT_MODE=false mise run all` passed independently on this branch.
- [x] Race-enabled adapter, SLO, and OnCall tests passed.
- [x] Generated references remained unchanged.

## Related

- Depends on #931; merge the machinery PR first.

<hr/>

#### Stack

* **PR #931**
  * **PR #1293** 👈

---
**Generated by Codex**

### DIFF
--- a/.claude/skills/add-provider/SKILL.md
+++ b/.claude/skills/add-provider/SKILL.md
@@ -287,8 +287,8 @@ Summary of the key steps:
 1. Provider interface + `init()` with a single `providers.Register()` call + `providers.ConfigLoader` (mirror the SLO reference)
 2. Config keys + validation
 3. Commands with UX compliance
-4. Types + client; adapter only for resources placed in the `resources` pipeline (returned from `TypedRegistrations()`, non-nil `Schema`)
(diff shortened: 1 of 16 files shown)


awsome-o (dismissed):
LGTM but shall 931 goes first？

awsome-o (dismissed):

dafydd-t (approved):

PR #1295, opened 2026-09-08
ci(review): move the review rules into the skill

dafydd-t:
Follows #1290. claude review CI jobs were still running silently, but for #1292 we have a trranscript, so I try and use that to fix things. #1292 had a "clean" review, which is my theory for it passing silently. I want it to comment "no issues" in this scenario.

- We were repeating instructions in the prompt and in the review-pr skill. move the majority of prompting into the skill. the prompt in the CI is now much shorter. Hopefully this will make the instructions on how to present results more clear to the agent
- Also, allow `git fetch` since the skill needs that.

other denials from the transcript were `gh pr checkout`, `go build` and `bin/gcx --help`. I left those disallowed as we dont want it to build the program.

**Verifying**

Like #1290, this cannot be checked from the PR, because the workflow only takes effect once merged.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -7,6 +7,7 @@
       "Skill",
       "Task",
       "Bash(git diff:*)",
+      "Bash(git fetch:*)",
       "Bash(git log:*)",
       "Bash(git show:*)",
       "Bash(gh issue list:*)",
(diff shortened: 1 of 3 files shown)


awsome-o:
let me try to merge it and see if it fixes

awsome-o (approved):
will follow up and test

PR #1299, opened 2026-09-08
fix(synth): preserve browser check channels

### LINKED ISSUE #1298: Synthetic Monitoring browser check creation omits required k6 channel
## Summary

`gcx synthetic-monitoring checks create` cannot create a browser check when the Synthetic Monitoring API requires a k6 execution channel.

The API returns HTTP 400 with `Invalid incoming check`.

## Cause

The remote `Check` model includes `channels`.

The user-facing `CheckSpec` model does not include `channels`.

The check adapter also treats `channels` as server-managed data and removes it from manifests.

Current browser check requests require a value such as:

```yaml
channels:
  k6:
    id: v2
```

As a result, the Synthetic Monitoring command cannot send the required field.

## Reproduction

Use a browser check manifest with a valid base64-encoded k6 browser script:

```yaml
apiVersion: syntheticmonitoring.ext.grafana.app/v1alpha1
kind: Check
metadata:
  name: browser-example
spec:
  job: browser-example
  target: https://example.com
  frequency: 300000
  timeout: 60000
  enabled: true
  probes:
    - Paris
  alertSensitivity: none
  basicMetricsOnly: false
  settings:
    browser:
      script: <base64-script>
```

Run:

```console
gcx synthetic-monitoring checks create -f browser.yaml
```

The command fails:

```text
failed to create check: request failed with status 400: Invalid incoming check
```

Send the same request through `gcx api`. Add `channels.k6.id: v2`. The request succeeds.

## Expected result

The Synthetic Monitoring create and update commands must support browser check channels.

The command can select the default compatible channel, or it can accept `spec.channels` in the manifest.

The check export and update flow must preserve the selected channel.

## Additional improvement

Return the API `err` field when it exists. The direct API response can contain a useful cause, but the Synthetic Monitoring command currently shows only the generic `msg` field.

wardbekker:
## Summary

- Preserve `spec.channels` through Synthetic Monitoring check create, update, and export flows.
- Send the selected k6 execution channel through both Synthetic Monitoring transport paths.
- Use a string `err` field from API error responses when it gives a more useful cause.

## Verification

- Live CRUD test on a Grafana Cloud test context:
  - Created a disabled browser check with `channels.k6.id: v2`.
  - Read the check and confirmed that the API returned the selected channel.
  - Updated the check and confirmed that the channel remained present.
  - Deleted the check and confirmed that no test resource remained.

## Scope

This change extends the existing check manifest model. It adds no command, flag, package, dependency, or architecture pattern.

Closes #1298.


mem (approved):

javiermolinar:
@claude review

javiermolinar (approved):
LGTM

Issue #1098, opened 2026-07-29
gcx kg notifications upsert: accept get/list output shapes + strict YAML parse


kblaschke512:
## Summary
Follow-up hardening for `gcx kg notifications upsert`, deferred from the review on #1077 (see [this comment](https://github.com/grafana/gcx/pull/1077#issuecomment-5119384071)). Two independent, low-cost items.

> **Update:** the non-empty `name` check (originally bundled below) landed in #1077 directly — it has clear precedent in the other kg upsert paths. The strict-parse and round-trip items remain here.

## 1. `get`/`list` output can't be fed back into `upsert -f`
`get` encodes a bare `AlertConfig` and `list` encodes a bare `[]AlertConfig`, but `upsert` only unmarshals the wrapped shape:

```yaml
alertConfigs:
  - name: ...
```

So the natural read → edit → apply flow fails:

```bash
gcx kg notifications get foo -o yaml > foo.yaml
gcx kg notifications upsert -f foo.yaml   # -> "no notification configs found in file"
```

...which reads like the file is empty rather than like it needs a wrapper key.

**Fix:** accept all three input shapes on `upsert` (the `alertConfigs:` wrapper, a single bare config, and a bare list), or at minimum change the error to say `expected top-level "alertConfigs:"`.

## 2. Strict YAML unmarshal on `upsert`
`upsert` uses non-strict `yaml.Unmarshal`, so a mistyped key is silently dropped — e.g. `matchlabels:` (lowercase L) or `duration: 5m` instead of `for:`. The result is a POST of a config with no matchers, i.e. a match-everything rule, and with no validate endpoint / no `--dry-run` there is no step where a person would notice. Paired with `silenced: true` that's especially bad.

**Note on scope:** nothing in the codebase currently uses strict YAML decoding (`KnownFields`/`UnmarshalStrict`) — every kg parse path is non-strict. Adding it to notifications alone would be a one-off. Worth deciding whether to apply it just to this backstop-less endpoint or repo-wide (suppressions/prom-rules/model-rules/entities all share the non-strict behavior).

**Fix:** `yaml.NewDecoder(...).KnownFields(true)` (or `UnmarshalStrict`) on `upsert`, so mistyped keys fail locally at parse time.

## Acceptance criteria
- [ ] `upsert -f` accepts the `alertConfigs:` wrapper, a single bare config, and a bare list (or a clearly worded error naming the expected wrapper key)
[...]

PR #1301, opened 2026-09-08
fix(kg): notifications list zero-result shape, NoArgs, category help

kblaschke512:
## What

Small follow-up to #1077 (merged 2026-08-20), catching three gaps found in a review pass against the newly-landed `integrate-with-gcx` contributor skill:

1. `Args: cobra.NoArgs` on the flags-only `list` and `upsert` leaves — without it they silently swallow stray positionals instead of rejecting them.
2. `gcx kg notifications list -o json|yaml` on zero results now serializes `[]` instead of `null`, with a regression test (`TestNotifications_List_ZeroResults`).
3. `--category` flag help now states its empty/omitted behavior and that matching is server-side and exact.

## Notes

- Unrelated to the two items still open in #1098 (accepting `get`/`list` output shapes on `upsert`, strict YAML parse) — no overlap.
- `gcx kg suppressions list` has the same null-vs-`[]` zero-result issue; out of scope here, flagged separately as family-wide debt.
- Full gate green: `GCX_AGENT_MODE=false mise run all` (lint 0 issues, all tests, reference docs regenerated with no drift).

### DIFF
--- a/docs/reference/cli/gcx_kg_notifications_list.md
+++ b/docs/reference/cli/gcx_kg_notifications_list.md
@@ -9,7 +9,7 @@ gcx kg notifications list [flags]
 ### Options
 
 ```
-      --category string   Filter by category: request, resource, health, or slo.
+      --category string   Filter by category: request, resource, health, or slo (server-side, exact match). Empty or omitted lists all categories.
   -h, --help              help for list
       --jq string         jq expression to apply to JSON output. Mutually exclusive with --json.
       --json string       Comma-separated list of fields to include in JSON output, or 'list' (or '?') to discover available fields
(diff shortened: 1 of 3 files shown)


kblaschke512:
Closing without merging — on closer inspection, none of the three items here are actually notifications-specific gaps. Checked repo-wide: only 6/66 `list` leaves declare `Args: cobra.NoArgs` (scattered, not an adopted convention), zero of 429 bare-slice `Encode` call sites normalize a nil slice to `[]` before encoding, and none of the ~10 comparable enum-filter flags (`--state`, `--status`, `--type` elsewhere) document match-semantics or empty-value behavior in their help text. So all three "fixes" would have been one-off deviations from how the rest of the codebase already works, not corrections to notifications-specific defects. Not worth the inconsistency for the marginal gain.

PR #1306, opened 2026-09-10
fix(ci): publish the Claude review from the workflow

awsome-o:
The Claude reviewer can finish its analysis without posting a review. This PR uses one Claude session to review the code with the gcx skill, then lets the GitHub workflow publish the structured findings as one `COMMENT` review.

## Before and after

| Area | Before | Fix | Expected behavior |
| --- | --- | --- | --- |
| Review process | On main, the gcx skill invokes an upstream plugin whose stop instruction can end the session before delivery. The original version of this PR added a second session. | One session performs correctness and gcx checks; remove the plugin and second session. | The reviewer returns a complete result without a plugin handoff. |
| Posting | Claude must choose and execute a posting command that passes tool permissions and shell checks. | The workflow posts the structured output using its own token. | One review with a summary and inline comments, authored by `github-actions[bot]`. |
| Missing results | A successful model step does not establish that a review was completed or published. | Require structured output; reject `review: null` and incomplete payloads; validate GitHub's submitted review response. | An incomplete review or unsuccessful publication fails visibly. A clean review has a summary and an empty comments array. |
| Invalid inline locations | There is no workflow-owned recovery from rejected inline anchors. | Retry once, body-only, for an HTTP 422 identifying an inline-location error. Preserve every finding, file/line range, and reviewed SHA. | Findings remain visible when inline anchoring fails. Other API errors fail normally. |
| PR changes during review | Main marks the run failed even if a review was published correctly before the head moved. | Check the head before publication and after confirmation. | A changed head before posting fails without posting. A changed head after confirmed publication produces a warning; the review remains attached to the reviewed commit. |
| Review model | The workflow has used `claude-opus-4-6` since #1217 (August 25); the initial version of this PR retained it. | Pin `claude-opus-5`, using the same gcx skill, output schema, and permissions. | Trial Opus 5 within the existing single-session flow. The model change adds no review stages. |
| Status and retries | Main adds `claude-reviewed`; removing it requests another review. | Remove the label logic and retain `@claude review`. | The posted review identifies the reviewed commit; the comment command requests a fresh review. |
| Transcript upload failure | In the earlier PR revision `c6ec90a5`, publication still inherited GitHub's implicit `success()` check, despite checking the review step's outcome. | Add an explicit `!cancelled()` status check alongside the successful-review condition. | A completed review can publish after an upload failure. Cancelled runs and failed reviews do not publish. The upload failure still remains visible in the job result. |

## Latest correction

The publish condition is now:

```yaml
if: ${{ !cancelled() && steps.claude-review.outcome == 'success' }}
```

GitHub adds an implicit `success()` condition when an `if` expression contains no status function. Checking only `steps.claude-review.outcome` therefore did not remove the dependency on a successful transcript upload. The regression check now requires the condition that also preserves cancellation and review-failure handling. [GitHub documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions)

## Evidence and validation

- [Run 34571589662](https://github.com/grafana/gcx/actions/runs/34571589662) ended on the upstream plugin's instruction to stop without posting. [Run 34574827847](https://github.com/grafana/gcx/actions/runs/34574827847) encountered posting-command rejections before eventually publishing a summary. These motivate separating review generation from publication.
- `GCX_AGENT_MODE=false mise run all`: lint, tests, build, skill validation, generated references, and docs pass. Documentation builds with existing link warnings; generated references have no diff.
- `actionlint` and `git diff --check` pass. The strengthened publication-condition test failed against the old condition before the workflow fix.
- Existing publication tests execute the real workflow shell against a fake GitHub API, covering clean and inline reviews, incomplete output, inline-location fallback, unrelated API failures, head movement, response validation, and literal shell syntax in findings.
- A live local run against `c6ec90a5` used the workflow's prompt and arguments with project settings: 16 turns, about 6 minutes, $1.68, complete structured output, and no permission denials. The transcript confirms that `/review-pr` read the PR diff and files, invoked no upstream plugin, and attempted no publication. It returned zero inline findings and missed the condition bug corrected above; this establishes that the review protocol can complete, not that all defects are detected.

## Model comparison

Two fresh local sessions used Claude Code 2.1.269, the same frozen pre-fix change (`5f6cfafb..c6ec90a5`), the same workflow instructions and schema, project settings, and a $5 limit per run. The local prompt selected the frozen git diff instead of the live PR, which already contained the fix. Neither session consulted the live PR or attempted publication.

| Result | Opus 4.6 | Opus 5 |
| --- | --- | --- |
| Caught the implicit `success()` publication bug | No | Yes |
| Returned complete structured review output | Yes | Yes |
| Duration | 6m48s | 7m01s |
| Reported cost | $1.67 | $2.09 |

Opus 5 also raised a permissions blocker that this test did not substantiate: its own transcript successfully ran the `git show ... | grep` pipeline it claimed would be refused. It recovered from one unrelated compound-command denial by using the allowed file reader. The result supports a model trial; one bug is not a general review-quality benchmark, and findings still need human verification. Its proposed `always()` fix was not applied: this PR retains the cancellation-aware `!cancelled()` guard.

The model-only update passes the repository's full quality gates and actionlint. This comparison was a local analysis test on the pre-fix snapshot; it did not exercise the current revision's full Actions publication path.

## Remaining validation and scope

The full Actions publication path has not been exercised with this revision. Pushing to the PR branch does not trigger this reviewer because `synchronize` is not configured. Separately, the GitHub App has rejected runs whose workflow differs from the default branch, and the pinned action restores trusted `.claude/` configuration from the PR base branch.

After merge, verify one clean PR and one known-defect PR: each should produce a submitted review on the reviewed commit, with the defect anchored inline in the latter. Inspect the transcript and the published review, rather than relying only on a green check.

The workflow now pins `claude-opus-5`. Existing gcx review criteria remain in place; the local posting-command rewrite and broader skill-policy changes are deferred. This PR also removes the plugin from local `/review-pr` use, so finding quality and cost remain things to measure on real reviews.


CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ Grafana engineering teams are welcome to contribute to and maintain their areas
...
 
-- [The claude code review GH action, with prompt & references](.github/workflows/claude-code-review.yml). This should encourage authors to adhere to the guidelines linked above.
+- [The Claude review workflow](.github/workflows/claude-code-review.yml) reviews every non-draft PR opened by a person (bot PRs are skipped) with the repository's own [`review-pr`](.claude/skills/review-pr/SKILL.md) checks, so a PR gets the same treatment in CI as it does when a developer runs the review locally. The session returns its findings and the workflow publishes them as one comment review against the commit it reviewed, including a summary when there are no findings; missing or incomplete findings fail the run instead of publishing a clean review. Comment `@claude review` for a fresh review, and read the `claude-review-transcript-<PR>` artifact on the run to see what the session did. The review reads code: the normal CI jobs own builds, tests and generated docs.
```
dafydd-t: I think we can revert this,we dont need this kind of detail.

dafydd-t (approved):
This looks a bit overengineered, but we'll give it a go.

PR #1307, opened 2026-09-10
Add named-query backend datasource client for SM

Pokom:
## Summary

Adds `gcx synthetic-monitoring query <name>` — a command that executes a **named query** against the SM datasource's backend query API (`queryType: <name>`, via Grafana's unified datasource query API) instead of gcx hand-building PromQL/LogQL.

- New `BackendDatasourceClient` (`internal/query/synth/backend_datasource_client.go`) wraps `grafanaquery.Client` to call a backend-owned named query by name + params, returning raw frames or a client-side reduction (`Mean`, `HasNumericField`).
- New `namedquery` package/command wires this into `gcx synthetic-monitoring`, with `-p/--param`, `--from`/`--to`, and table/JSON output.
- Renamed the existing SM REST-API proxy client `client.go` → `proxy_client.go` to disambiguate it from the new backend-query path — these are two distinct transports (proxy vs. backend query API) and will stay separate.

This keeps the query definition in exactly one place — the SM plugin's own query registry — so gcx and the SM app report the same values, rather than gcx re-deriving PromQL that can drift from what the plugin computes.

## Why

Demonstrates the shape of calling backend-owned named queries from gcx rather than every consumer hand-building PromQL. Landing this now gets alignment on the pattern with a real, useful command (`checks_uptime`, `probe_execution_rate`), even though only two named queries exist today.

## Not in scope (future improvements)

1. **Live discovery + caching** — `synthetic-monitoring-app` already generates `src/datasource/schema/v0alpha1/query.types.json`, a `QueryTypeDefinitionList` describing every valid query name, description, and JSON-Schema param shape (source:
`pkg/plugin/namedqueries.go` registry + `pkg/plugin/schema_test.go` `querySchemas`). Grafana core serves this statically at the plugin's schema path. A follow-up could have gcx fetch and cache it for validation/autocomplete — no backend changes required.
2. **Migrate existing hand-built PromQL/LogQL** (e.g. `internal/providers/synth/checks/status.go`'s `probe_success` avg-over-time) to call the backend datasource client instead, once enough named queries exist to cover them.
3. **Agent discovery mechanism** — surface named query names/params through the `gcx commands` catalog
(`internal/agent/command_annotations.go`), likely backed by (1) above, so an agent can enumerate valid queries without already knowing their names.

- relates https://github.com/grafana/synthetic-monitoring/issues/634


Pokom:
@claude review

internal/providers/synth/namedquery/commands.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// points counts the samples behind the reduced value, so an unexpectedly round
+// number can be traced to an empty or single-point series.
+func points(res *synth.NamedResult) int {
+	if res == nil || len(res.Frames) == 0 {
+		return 0
+	}
+
+	values := res.Frames[0].Data.Values
```
The-9880: All the results referencing `Frames[0]` assume that the named query is only returning one frame. Each `Frame` corresponds to a series.

We have two currently:
* `checks_uptime`: uses a `max by ()` query, so it is 1 frame (1 timeseries).
* `probe_execution_rate`: is ` sum(rate(...)) by (probe)` - you should expect one frame (TS) per probe.

This is in the following methods as well:
* `Mean`
* `HasNumericField`

In addition to here, in `points`.

internal/providers/synth/namedquery/commands.go
```diff
@@ -0,0 +1,254 @@
...
+// points counts the samples behind the reduced value, so an unexpectedly round
+// number can be traced to an empty or single-point series. It reports 0 for a
+// query with no numeric field, since there is no reduced value for it to count
+// samples behind.
+func points(res *synth.NamedResult) int {
+	if res == nil || len(res.Frames) == 0 || !synth.HasNumericField(res) {
+		return 0
+	}
+
+	values := res.Frames[0].Data.Values
```
The-9880: All the results referencing `Frames[0]` assume that the named query is only returning one frame. Each `Frame` corresponds to a series.

We have two currently:
* `checks_uptime`: uses a `max by ()` query, so it is 1 frame (1 timeseries). ✅ 
* `probe_execution_rate`: is ` sum(rate(...)) by (probe)` - you should expect one frame (TS) _per probe_. ❌ 

This is in the following methods as well:
* `Mean`
* `HasNumericField`

In addition to here, in `points`.

docs/reference/cli/gcx_synthetic-monitoring_query.md
```diff
@@ -0,0 +1,57 @@
...
+### Examples
+
+```
+
+  # Uptime for a check, as the app computes it
+  gcx synthetic-monitoring query checks_uptime \
+    -p job=my-check -p instance=https://example.com -p frequency=60000
+
+  # Reachability over the last day
+  gcx synthetic-monitoring query reachability \
```
The-9880: Is `reachability` actually a named query we have? I encountered 2:
* `checks_uptime`
* `probe_execution_rate`

Running this gets me:
```
 unknown query "reachability"   (Source: plugin), exit 1
```

internal/providers/synth/namedquery/commands.go
```diff
@@ -0,0 +1,254 @@
...
+			params[key] = true
+		case "false":
+			params[key] = false
+		default:
+			if n, err := strconv.ParseFloat(value, 64); err == nil {
+				params[key] = n
+				continue
+			}
+			params[key] = value
+		}
```
The-9880, line 213: Small catch here - if the key is `job` and the value happens to be numeric for whatever reason (`1234`), it will parse as a number and be rejected by our API, which expects `job` to be `string`.

internal/providers/synth/namedquery/commands.go
Pokom: That's a good catch, thanks. Let me figure out how to properly handle when multiple frames are returned.

docs/reference/cli/gcx_synthetic-monitoring_query.md
Pokom: Ah, that's slop, sorry about that. When I initially started this PR, I had claude one shot all of the named queries just to stress test it and this got left behind. Let me review the docs a bit better and make sure there are no other cases.

internal/providers/synth/namedquery/commands.go
Pokom: Very good catch! Let me fix that up

Pokom: Okay, I added a bit of a hack to better handle this that I intend to use as a short term solution. Long term, I'd like to look at the updated the backend datasource to provide an affordance on what the type of the parameter is so we can marshal it properly. Let me know if it's passible

Pokom: https://github.com/grafana/gcx/pull/1307/changes/6d304c7a207667f1988f5de41ba1494de7da7bc5#diff-08871dfef71538f7e758bf61ab0901dd6b20bb8daeba7909b4a65b8c9983e2c3R187-R191

internal/providers/synth/namedquery/commands.go
Pokom: Okay, the latest commit handles multiple frames now. I left a comment that calling `synth.Mean(frame)` is fine for the first two named queries, but anything that's a range may not necessarily be appropriate to transform the frame with a mean. I'll need to consider in the next change to SM's backend datasource providing some affordance on which transform to apply.

dafydd-t (approved):

The-9880 (approved):

PR #1308, opened 2026-09-10
fix(agento11y): rename resource group and preserve sigil aliases

### LINKED ISSUE #1344: fix(agento11y): rename sigil.ext.grafana.app API group with backwards compatibility
Agent Observability resource types are still registered under the pre-rename `sigil.ext.grafana.app` API group:

```
$ gcx resources list-types --no-schema -o text | grep sigil
sigil.ext.grafana.app  v1alpha1  collections
sigil.ext.grafana.app  v1alpha1  evalrules
sigil.ext.grafana.app  v1alpha1  evaluators
sigil.ext.grafana.app  v1alpha1  hookrules
```

Every other provider names its group after the provider (`appo11y.ext.grafana.app`, `slo.ext.grafana.app`, `adaptive-metrics.ext.grafana.app`, …), so `sigil` is the only group that doesn't match the command surface (`gcx agento11y`). It was missed when the provider was renamed. Anyone writing GitOps manifests today has to know an internal codename that appears nowhere else in the CLI, and the short form of the group resolves as `sigil` rather than `agento11y`.

The group is baked into user-authored YAML (`apiVersion: sigil.ext.grafana.app/v1alpha1`) and into `gcx resources pull` output, so the rename needs a compatibility path rather than a straight swap.

Affected descriptors:

| Kind | Plural | File |
|------|--------|------|
| `Evaluator` | `evaluators` | `internal/providers/agento11y/eval/evaluators/resource_adapter.go` |
| `EvalRule` | `evalrules` | `internal/providers/agento11y/eval/rules/resource_adapter.go` |
| `HookRule` | `hookrules` | `internal/providers/agento11y/eval/guards/resource_adapter.go` |
| `Collection` | `collections` | `internal/providers/agento11y/eval/collections/resource_adapter.go` |

## Fix options

1. **Rename to `agento11y.ext.grafana.app`, keep `sigil.ext.grafana.app` accepted on input** (preferred). Change the four `StaticDescriptor()` groups, then:
   - register a `resources.RegisterGVKNormalizer` that collapses the old group onto the new one, which covers manifests going through `push` and `delete` (`internal/resources/remote/pusher.go`, `deleter.go`) - same mechanism datasources already uses for its per-plugin groups
   - teach the discovery index about a deprecated group alias so selectors still resolve (`gcx resources get evalrules.v1alpha1.sigil.ext.grafana.app`, and the `gcx resources get evalrules.sigil` short form). `RegistryIndex.RegisterStatic` only takes name aliases today, so this needs a small addition - group aliases in `adapter.Registration`, or a `RegisterGroupAlias` on the index
   - emit the new group only on `pull` / `get -o yaml`, and warn on stderr when the old group is used
2. **Register both groups as real descriptors.** Rejected: duplicates all four types in `list-types`, and makes preferred-version resolution per group ambiguous.
3. **Hard rename, no compatibility.** Cheapest, but silently breaks existing manifests with a "no converter found"-style error rather than telling the user what to change.

## Acceptance criteria

- [ ] `gcx resources list-types --no-schema -o text | grep agento11y` lists all four types under `agento11y.ext.grafana.app`, and no type is listed under `sigil.ext.grafana.app`
- [ ] `gcx resources push` and `gcx resources delete` accept a manifest with `apiVersion: sigil.ext.grafana.app/v1alpha1` and route it to the same adapter as the new group
- [ ] Using the old group prints a single deprecation warning on stderr naming the replacement, and does not change the exit code
- [ ] `gcx resources get evalrules.v1alpha1.sigil.ext.grafana.app` and `gcx resources get evalrules.v1alpha1.agento11y.ext.grafana.app` return the same thing
- [ ] `gcx resources pull` writes `apiVersion: agento11y.ext.grafana.app/v1alpha1`
- [ ] Round trip works against a real stack: `pull` with the new group, then `push` the result unchanged
- [ ] CLI reference docs regenerated (`GCX_AGENT_MODE=false mise run reference`)

## Out of scope

The Cloud access-policy scope `sigil:write` keeps its name - that's a Cloud-side resource name, unrelated to the gcx API group. Don't touch the scope references in `claude-plugin/skills/agento11y-*`.

alexander-akhmetov:
Renames the Agent Observability resource group from `sigil.ext.grafana.app` to `agento11y.ext.grafana.app`.

Old manifests and selectors keep working. New output uses the new group.

fixes https://github.com/grafana/gcx/issues/1344


alexander-akhmetov:
@claude review

dafydd-t:
Can we add:

- deprecation warning on stderr when using the old resource name?
- evidence that the full pull/push workflow works with these group names (and the old aliased ones)?

alexander-akhmetov:
> deprecation warning on stderr when using the old resource name?

Added https://github.com/grafana/gcx/pull/1308/changes/1ccafb8ba7fb2c5fd9add7352ad1aa859918dae8

> evidence that the full pull/push workflow works with these group names (and the old aliased ones)?

Added a test in https://github.com/grafana/gcx/pull/1308/changes/73da7579ff33e03ecd611b0f32cab67e43471b07

PR #1309, opened 2026-09-11
feat: Decode Check Scripts

Pokom:
Synthetic Monitoring API returns the contents of scripts for browser and scripted checks as base64 encoded representations. Harnesses are certainly capable of figuring out how to decode them, but this wastes tokens for something we know must be done to view/edit scripts. This change introduces a new flag `--decode-script` that can be passed to the check subcommand that will perform the decoding in place.

The expected workflow would be
1. list out checks that are scripted/browser
2. get the script of a check that needs to be edited and store it
3. let the harness edit the script on disk
4. validate it works with k6(if available locally)
5. update the script

Future improvements can be made around the validation part. SM exposes an adhoc endpoint where you can provide a check to execute, and the check will flow through the full system. GCX isn't wired up today to make these calls, so this change focuses purely on the decoding side.

I validated this works against a live stack by going through the expected workflow above. The neat part was having GCX query mimir/loki to gather metrics generated by the script and find wasteful logs.

### DIFF
--- a/docs/reference/cli/gcx_synthetic-monitoring_checks_get.md
+++ b/docs/reference/cli/gcx_synthetic-monitoring_checks_get.md
@@ -17,11 +17,15 @@ gcx synthetic-monitoring checks get NAME [flags]
 
   # Get check with current execution status.
   gcx synthetic-monitoring checks get grafana-instance-health-5594 --show-status
+
+  # Get a scripted/browser check with a readable script, ready to edit and 'checks update'.
+  gcx synthetic-monitoring checks get grafana-instance-health-5594 --decode-script -o yaml > check.yaml
 ```
 
 ### Options
 
 ```
+      --decode-script   Decode a scripted/browser check's base64 script to plaintext (yaml/json output only, for editing and 'checks update')
   -h, --help            help for get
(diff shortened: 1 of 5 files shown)


digitalcrab (changes requested):
Looks good. One note that I am not sure about 100%, but looks like we can update checks via `gcx resources push` and it leads to `internal/providers/synth/checks/resource_adapter.go` with `NewTypedCRUD` and `UpdateFn` defined and it's not part of that PR. 

As I am writing the review i decided to simply try ...

I tried building gcx from your branch and was able to get the check with decode - all good. And then i run `gcx resources push` and it failed with `illegal base64 data at input byte 6`

```bash
  $ gcx synthetic-monitoring checks get local-tools-scripted-89333 --decode-script -o yaml > decoded/decoded.yaml                                                                                                               
  $ gcx resources push checks -p ./decoded/                                                                                                                                                                                     
  ✘ 0 resources pushed, 1 errors                                                                                                                                                                                                
                                                                                                                                                                                                                                
    "error": "failed to update check 89333: request failed with status 400:                                                                                                                                                     
              illegal base64 data at input byte 6"
```

digitalcrab (approved):

PR #1311, opened 2026-09-11
feat: implement adhoc checks for SM

Pokom:
Synthetic Monitoring API provides a way to execute an ad hoc execution of check for testing purposes. The results are stored in Loki/Mimir, but the content of the check does not get permenantly created/updated. The app provides a "test" mechanism that customers use today to validate changes.

This change wires up a `test` command for `gcx synthetic-monitoring` that will
1. Make an HTTP request to the SM API and encode a check to test
2. Gather the correlation ID that can be used to query Loki
3. Periodically polls Loki for up to 30 seconds to discover results of the test
4. Outputs the result of the test

Ideally the polling of loki would be passed off to our backend datasource to handle, but that path hasn't been implemented yet.


docs/reference/cli/gcx_synthetic-monitoring_checks_test.md
```diff
@@ -0,0 +1,58 @@
...
+Run a Synthetic Monitoring check once against its probes without persisting
+it — the check is never saved, and no schedule is created.
+
+Results are recovered by polling Loki for the log lines the probes emit for
+this execution (tagged type="adhoc"), the same mechanism the Synthetic
+Monitoring app's "Test" button uses. Requires a Loki datasource containing SM
+ad-hoc logs, and read access to it.
+
+Note: ad-hoc test executions are billed the same as scheduled check
+executions. See https://grafana.com/docs/grafana-cloud/cost-management-and-billing/manage-invoices/understand-your-invoice/synthetic-monitoring-invoice.md.
```
The-9880, line 16: Is this a fact or is it being conservative for if that happens in the future?

internal/providers/synth/checks/status.go
```diff
@@ -879,21 +911,20 @@ func smMetricsDatasourceName(ctx context.Context, restCfg config.NamespacedRESTC
...
 
 	var body struct {
-		JSONData struct {
-			Metrics struct {
-				GrafanaName string `json:"grafanaName"`
-			} `json:"metrics"`
+		JSONData map[string]struct {
+			GrafanaName string `json:"grafanaName"`
 		} `json:"jsonData"`
 	}
```
The-9880, line 922: Tier 4 of `resolveDataSourceUID` is when the agent fetches the the DS via the SM plugin settings - this is typical for a fresh session/context.

I hit this error: `json: cannot unmarshal string into Go struct field .jsonData of type struct { GrafanaName string }`.

Because the type:
```
JSONData map[string]struct {
			GrafanaName string `json:"grafanaName"`
		} `json:"jsonData"`
```

Suggests that `jsonData` has the following format:
```
"jsonData": {
	"value1": { "grafanaName": "name-1"},
	"value2": { "grafanaName": "name-2"},
	"value3": { "grafanaName": "name-3"}
}
```

But `jsonData` actually has:

```
"jsonData": {
	"apiHost": "api-url", // fails to decode
	"stackId": 9880, // fails to decode
	"metrics": { "grafanaName": "name-3"},
	"logs": { "grafanaName": "name-3"}
}
```
-----
If `smPluginDatasourceName` only cares about `logs` and `metrics`, just add the missing field for `logs`:
```
var body struct {
		JSONData struct {
			Metrics struct {
				GrafanaName string `json:"grafanaName"`
			} `json:"metrics"`
			Logs struct {
				GrafanaName string `json:"grafanaName"`
			} `json:"logs"`
		} `json:"jsonData"`
	}
```

And then select the one you want:
```
var name string
switch section {
case "metrics":
    name = body.JSONData.Metrics.GrafanaName
case "logs":
    name = body.JSONData.Logs.GrafanaName
default:
    return "", fmt.Errorf("unknown SM plugin datasource section %q", section)
}
```

internal/providers/synth/checks/adhoc_results.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	reported := make(map[string]AdHocProbeResult, len(probeNames))
+	expr := BuildAdHocLogsQuery(id)
+
+	pollCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	poll := func() error {
+		resp, err := lokiClient.Query(pollCtx, dsUID, loki.QueryRequest{Query: expr})
```
The-9880, line 94: Using `pollCtx` in the `lokiClient.Query` call means you could cancel an in-flight Query if the `pollCtx` has expired. That's fine for strictly respecting the timeout, but if a `poll` is already in flight the results may potentially come back with it.

The-9880: ^ That suggestion loosens the time-bounding in favour of potentially getting the results back.

If you want to strictly respect the `timeout`, however, then I believe the `select` needs to be corrected:
```
for len(pendingByName) > 0 {
		select {
		case <-pollCtx.Done():
			for name, probeID := range pendingByName {
				reported[name] = AdHocProbeResult{ProbeID: probeID, ProbeName: name, Status: AdHocTimeout}
			}
			return orderedAdHocResults(reported, probeNames), nil
		case <-ticker.C:
			if err := poll(); err != nil {
				return nil, fmt.Errorf("querying ad-hoc results: %w", err)
			}
		}
	}
```

^^^ You are erroring with `fmt.Errorf("querying ad-hoc results: %w", err)` when the timeout is hit. What you want to do, I think, is return the same reporting as the `pollCtx.Done()` branch which shows whatever partial results you _have_ acquired - so change the `case <-ticker.C` branch like so:

```
case <-ticker.C:
    if err := poll(); err != nil {
        if  errors.Is(pollCtx.Err(), context.DeadlineExceeded) {
            for name, probeID := range pendingByName {
				reported[name] = AdHocProbeResult{ProbeID: probeID, ProbeName: name, Status: AdHocTimeout}
			}
            return orderedAdHocResults(reported, probeNames), nil
        }
        return nil, fmt.Errorf("querying ad-hoc results: %w", err)
    }
```

docs/reference/cli/gcx_synthetic-monitoring_checks_test.md
Pokom: > Is this a fact or is it being conservative for if that happens in the future?

I was being conservative about it being a fact in the future.

The-9880: I think that's OK for now - it _could_ cause billing concerns, but the statement currently is that we do not bill for these. Easier to reverse than the other way around.

cmd/gcx/root/testdata/non_canonical_command_operations.json
```diff
@@ -102,6 +102,7 @@
   "gcx slo reports status",
   "gcx slo reports timeline",
   "gcx synthetic-monitoring checks status",
+  "gcx synthetic-monitoring checks test",
```
The-9880, line 105: Nit: is `test` the right word, or should we call it `adhoc`? Or `test-check`?

internal/providers/synth/checks/commands.go
```diff
@@ -674,6 +676,187 @@ func existingSensitivity(ctx context.Context, loader smcfg.Loader, checkID int64
...
+	return o.IO.Validate()
+}
+
+// checkTestResult is the finite result document for `checks test`.
+type checkTestResult struct {
+	Type          string             `json:"type" yaml:"type"`
+	SchemaVersion string             `json:"schema_version" yaml:"schema_version"`
+	Job           string             `json:"job" yaml:"job"`
+	Target        string             `json:"target" yaml:"target"`
+	AdHocID       string             `json:"adhoc_id" yaml:"adhoc_id"`
```
The-9880, line 715: Nit: `schema_version` and `adhoc_id` are snake-case, meanwhile we've got `probeId`/camelCasing elsewhere.

The-9880 (approved):
LGTM, I left a couple of comments - let me know if you need a re-approve if you choose to address them.

cmd/gcx/root/testdata/non_canonical_command_operations.json
Pokom: Test aligns with what our app calls it today, which is why I landed on `test`. I personally prefer `adhoc`, but I'd rather a consistent naming experience where we can.

dafydd-t (approved):
LGTM from gcx, just looking an new command names 👍

PR #1313, opened 2026-09-11
Stop OAuth when sandbox blocks keychain writes

### LINKED ISSUE #1312: OAuth login does not explain keychain persistence failures
## Description

A user must repeat `gcx login --oauth` several times each day for one non-production Grafana context. The same user previously had to sign in only every one or two weeks. A separate production context continues to work as expected.

A possible cause is that gcx cannot write the refreshed OAuth token to the operating system keychain. This can occur when an agent sandbox blocks keychain access. The current behavior does not make this cause clear. It can look like an OAuth expiry or server configuration problem.

## Expected behavior

gcx must report when it cannot persist an OAuth token or refreshed token. The message must identify the keychain access failure and give an actionable next step. Agent-mode output must include enough structured detail for the agent harness to explain that keychain access is required.

## How can we reproduce this?

1. Configure a context that uses OAuth.
2. Run gcx in an environment that blocks access to the operating system keychain.
3. Complete the OAuth login flow.
4. Use the context after the active token expires or requires a refresh.
5. Observe that gcx starts the OAuth flow again without a clear keychain persistence error.

## Environment details

- macOS
- OAuth login
- An execution environment can restrict keychain access

## Additional context

The exact cause in the original report is not yet confirmed. The repeated prompt is specific to one environment, so a server-side token policy change is another possible cause. Better diagnostics must distinguish token lifetime or server rejection from local token persistence failure.

wardbekker:
## Summary

- Add a non-secret write, read, and delete probe for the OS credential store.
- Run the probe before OAuth login starts the browser flow.
- Run the probe before OAuth refresh sends the refresh token.
- Keep pending rotated tokens on the persistence-only retry path.
- Classify macOS authorization failures as restricted execution sessions.
- Tell agent users to retry the same command outside the sandbox.
- Keep plaintext storage only when the user explicitly disables the credential store.

## Rationale

Some sandboxes allow keychain reads and block keychain writes. A read-only check reports a false success in this state. The capability probe detects the real failure without an agent-specific environment variable.

The refresh path stops before network use. The stored OAuth session stays unchanged.

## Validation

- Focused credential, auth, config, login, and error tests
- Signed commit verification

Fixes #1312

### DIFF
--- a/docs/architecture/auth-system.md
+++ b/docs/architecture/auth-system.md
@@ -195,8 +195,8 @@ backend (which issues and refreshes tokens), and a short-lived callback
 server that gcx starts on a loopback port.
 
 Before gcx starts the browser flow, it confirms that the credential store can
-write, read, and remove a random non-secret value. It stops before browser or
-network use when a sandbox or another process policy blocks the write.
+write, read, and remove a random non-secret value. If the check fails, gcx stops
+the OAuth flow.
 
 ```mermaid
 sequenceDiagram
@@ -370,10 +370,7 @@ the server-side refresh token. Cancellation is still honored before the lock
 and network request begin.
 
 After the lock and reload, the transport checks that the credential store is
-writable. This check occurs only when a network refresh is still required. A
-failure stops the operation before gcx sends the refresh token. A retry of an
-already rotated pending generation skips this check and retries only the
-persistence callback.
+writable. If the check fails, gcx stops the token refresh flow.
 
 A successful refresh response is not exposed to protected requests until
 persistence succeeds. Rotation-capable issuers may return a new refresh token;
--- a/docs/architecture/auth-system.md
+++ b/docs/architecture/auth-system.md
@@ -194,6 +194,10 @@ spans three remote actors: the Grafana instance (which hosts the
 backend (which issues and refreshes tokens), and a short-lived callback
 server that gcx starts on a loopback port.
 
+Before gcx starts the browser flow, it confirms that the credential store can
+write, read, and remove a random non-secret value. It stops before browser or
+network use when a sandbox or another process policy blocks the write.
+
 ```mermaid
 sequenceDiagram
     participant User
@@ -364,6 +368,12 @@ uses a context detached from the caller's request context so that a caller
 cancellation cannot abandon a refresh that has already consumed and rotated
 the server-side refresh token. Cancellation is still honored before the lock
 and network request begin.
+
+After the lock and reload, the transport checks that the credential store is
+writable. This check occurs only when a network refresh is still required. A
+failure stops the operation before gcx sends the refresh token. A retry of an
+already rotated pending generation skips this check and retries only the
+persistence callback.
 
 A successful refresh response is not exposed to protected requests until
 persistence succeeds. Rotation-capable issuers may return a new refresh token;
--- a/docs/architecture/config-system.md
+++ b/docs/architecture/config-system.md
@@ -352,12 +352,10 @@ backend, replacing or deleting an existing generation, replacing a missing or
 rejected sentinel, and
 value-size, policy, cancellation, or unknown backend failures all fail closed.
 
-Some sandboxes allow credential reads and block credential writes. OAuth login
-and refresh use a random non-secret write, read, and delete probe to detect this
-condition before they send OAuth credentials. Known macOS authorization exit
-codes use `credentials.ErrRestrictedSession`. Other probe failures stay fatal.
-The check uses credential-store behavior instead of an agent-specific
-environment variable.
+Some sandboxes allow credential reads and block credential writes. The OAuth
+login and refresh flows check for this condition before they start. They stop
+if the condition is true. gcx returns `credentials.ErrRestrictedSession` for
+known macOS authorization exit codes. Other probe failures stay fatal.
 
 Silently continuing in those cases could orphan the only resolvable credential,
 leave an old credential active, downgrade a credential for an unrelated backend
--- a/docs/architecture/config-system.md
+++ b/docs/architecture/config-system.md
@@ -351,6 +351,14 @@ same cause while keeping the rotated generation pending for retry. A locked
 backend, replacing or deleting an existing generation, replacing a missing or
 rejected sentinel, and
 value-size, policy, cancellation, or unknown backend failures all fail closed.
+
+Some sandboxes allow credential reads and block credential writes. OAuth login
+and refresh use a random non-secret write, read, and delete probe to detect this
+condition before they send OAuth credentials. Known macOS authorization exit
+codes use `credentials.ErrRestrictedSession`. Other probe failures stay fatal.
+The check uses credential-store behavior instead of an agent-specific
+environment variable.
+
 Silently continuing in those cases could orphan the only resolvable credential,
 leave an old credential active, downgrade a credential for an unrelated backend
 error, or write a secret in plaintext while a real secret backend exists.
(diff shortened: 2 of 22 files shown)


dafydd-t:
@claude review

wardbekker:
Smoke test passed on commit `89d3126e`.

- The macOS Keychain probe succeeded outside the sandbox.
- The probe returned `ErrRestrictedSession` with exit status 152 inside the sandbox.
- OAuth login stopped before the browser flow.
- OAuth refresh stopped before any HTTP request.
- Focused tests passed for credentials, auth, config, login, and error handling.
- The PR binary built successfully.
- All GitHub CI checks passed.

I found no smoke-test blockers.

docs/design/errors.md
```diff
@@ -165,6 +165,9 @@ Adding a new summary requires a PR amending this list.
 | `Authentication failed` | Token expired or missing |
 | `Keychain locked` | The OS keychain answers, but it is locked or the current session cannot unlock it, so gcx cannot store or use the credential |
 | `Keychain unavailable` | The OS keychain cannot be reached, so gcx cannot store or use the credential without an explicit plaintext-storage opt-out |
+| `OS credential store access is restricted` | The credential store is available, but the current execution session cannot write to it |
+| `OAuth login cannot persist credentials` | gcx cannot confirm that it can persist a new OAuth credential before the browser flow starts |
+| `OAuth refresh cannot persist credentials` | gcx cannot confirm that it can persist a rotated OAuth credential before it sends the refresh token |
```
dafydd-t: do we need all three of these? Isn't the first one the underlying cause for the second two? I would vote to just keep the first one

docs/sources/keychain.md
```diff
@@ -119,6 +119,23 @@ This applies only to credentials that are in the credential store. A credential
...
 the store, so `gcx config unset` and entry deletion both work as usual.
 
+## Restricted execution sessions
+
+Some agent tools run commands in a sandbox. The sandbox can allow credential
+reads and block credential writes. A read-only check cannot detect this state.
+
+Before an OAuth login or token refresh, gcx writes, reads, and removes a random
+non-secret test value. If this test fails, gcx stops before it starts the browser
+flow or sends the refresh token. The stored OAuth session stays unchanged.
```
dafydd-t:
```suggestion
Before an OAuth login or token refresh, gcx writes, reads, and removes a random
non-secret test value. If this test fails, gcx aborts the OAuth flow and the stored OAuth session stays unchanged.
```

internal/config/rest.go
```diff
@@ -284,6 +298,30 @@ func tokenPersistenceIdentity(sources []ConfigSource, path string) (string, erro
...
+func (cfg *Config) CheckOAuthCredentialPersistence() error {
+	if cfg == nil || cfg.keychainStore == nil {
+		return CheckOAuthCredentialPersistence()
+	}
+	return checkOAuthCredentialPersistence(cfg.keychainStore)
+}
+
+func checkOAuthCredentialPersistence(store credentials.Store) error {
+	err := credentials.CheckWritable(store)
+	if errors.Is(err, credentials.ErrUnavailable) || errors.Is(err, credentials.ErrDisabled) {
```
javiermolinar: ErrUnavailable and ErrDisabled are two different things. This means that readable credentials with unavailable writes can pass preflight, consume a refresh token, and then fail to save it

docs/sources/keychain.md
```diff
@@ -119,6 +119,23 @@ This applies only to credentials that are in the credential store. A credential
...
+
+Run the same gcx command outside the sandbox. If an agent ran the command, use
+the agent approval flow to run gcx with access to the operating system
+credential store. Do not run `gcx login` when a refresh command reports this
+error. Retry the original command.
+
+gcx detects this condition from credential-store behavior. It does not depend
+on an environment variable from one agent tool.
```
dafydd-t:
```suggestion
The remedy is to run the same command from outside the sandbox. If an agent ran the command, use
the agent approval flow to run gcx with access to the operating system
credential store. 
```

docs/architecture/auth-system.md
```diff
@@ -194,6 +194,10 @@ spans three remote actors: the Grafana instance (which hosts the
 backend (which issues and refreshes tokens), and a short-lived callback
 server that gcx starts on a loopback port.
 
+Before gcx starts the browser flow, it confirms that the credential store can
+write, read, and remove a random non-secret value. It stops before browser or
+network use when a sandbox or another process policy blocks the write.
```
dafydd-t:
```suggestion
Before gcx starts the browser flow, it confirms that the credential store can
write, read, and remove a random non-secret value. If it fails, the OAuth flow is aborted.
```

docs/architecture/auth-system.md
```diff
@@ -365,6 +369,12 @@ cancellation cannot abandon a refresh that has already consumed and rotated
 the server-side refresh token. Cancellation is still honored before the lock
 and network request begin.
 
+After the lock and reload, the transport checks that the credential store is
+writable. This check occurs only when a network refresh is still required. A
+failure stops the operation before gcx sends the refresh token. A retry of an
+already rotated pending generation skips this check and retries only the
+persistence callback.
```
dafydd-t:
```suggestion
After the lock and reload, the transport checks that the credential store is
writable. If it fails, gcx aborts the token refresh flow.
```

docs/architecture/config-system.md
```diff
@@ -351,6 +351,14 @@ same cause while keeping the rotated generation pending for retry. A locked
 backend, replacing or deleting an existing generation, replacing a missing or
 rejected sentinel, and
 value-size, policy, cancellation, or unknown backend failures all fail closed.
+
+Some sandboxes allow credential reads and block credential writes. OAuth login
+and refresh use a random non-secret write, read, and delete probe to detect this
+condition before they send OAuth credentials. Known macOS authorization exit
+codes use `credentials.ErrRestrictedSession`. Other probe failures stay fatal.
+The check uses credential-store behavior instead of an agent-specific
+environment variable.
```
dafydd-t:
```suggestion
Some sandboxes allow credential reads and block credential writes. The OAuth login
and refresh flows check for this condition before they start, and abort if it is true. We return  `credentials.ErrRestrictedSession` for known macOS authz exit
codes. Other probe failures stay fatal.
```

docs/design/errors.md
wardbekker: Updated in 54256708. The converter now uses only the underlying restricted-session summary for direct, login, and refresh failures. I removed the two OAuth-specific summaries and their converter.

docs/sources/keychain.md
wardbekker: Updated in 54256708. The keychain source and generated reference now state that gcx stops the OAuth flow when the check fails.

internal/config/rest.go
wardbekker: Fixed in 54256708. Both OAuth preflight paths now permit plaintext storage only for ErrDisabled. ErrUnavailable fails closed. I added tests for the direct login preflight and the wired refresh preflight.

docs/sources/keychain.md
wardbekker: Updated in 54256708. The keychain source and generated reference now use this shorter remedy.

docs/architecture/auth-system.md
wardbekker: Updated in 54256708. The authentication architecture text now states that gcx stops the OAuth flow when the check fails.

docs/architecture/auth-system.md
wardbekker: Updated in 54256708. The refresh section now states that gcx stops the token refresh flow when the check fails.

docs/architecture/config-system.md
wardbekker: Updated in 54256708. The configuration architecture text now uses the shorter explanation and keeps the macOS error mapping.

wardbekker: Updated in 54256708. The configuration architecture text now uses the shorter explanation and keeps the macOS error mapping.

dafydd-t (approved):
Approved, but this handling logic is OS specific and so we'll likely see a sprawl of other instances of this for other OSes which we'll have to follow up with

Issue #1078, opened 2026-07-27
gcx kg thresholds — CRUD over v1 threshold config


xujiaxj:
## Summary
Add `gcx kg thresholds` over the **v1** threshold config API (`/v1/config/threshold-rule(s)`, `ThresholdRulesConfigController`). Target **v1, not v2** — see "Why v1" below.

## Why v1 (not v2)
The Asserts app UI and every threshold users have actually configured run on **v1** — `ManageAssertions.service.ts` calls only `/v1/config/threshold-rule(s)`. The **v2** API (`ThresholdsV2ConfigController`, added 2025-09-11) has **zero usage in the plugin UI** (only present in the proxy authz allow-list), so a v2 client would diverge from what users and the product actually use. Build v1; revisit v2 if/when the UI migrates.

## Backend endpoint map (verified: asserts-adi + asserts-app-plugin)

| gcx verb | Endpoint | Body / notes |
|---|---|---|
| `list --category request\|resource` | `GET /v1/config/threshold-rules/{request\|resource}` → `ThresholdRulesDto` | structured; **only request/resource** (no health in v1) |
| `get` | `GET /v1/config/threshold-rules` (Accept: x-yaml) | whole config as YAML |
| `set -f` | `POST /v1/config/threshold-rules/` (x-yaml body) | **replaces whole config** |
| `upsert` | `POST /v1/config/threshold-rule` `{record, expr, labels}` | create-or-update a single threshold |
| `delete` | `POST /v1/config/threshold-rule/delete` `{record, expr, labels}` | delete a single threshold (POST-with-body, **not** a DELETE verb) |

## Command shape
```bash
gcx kg thresholds list --category request|resource      # structured per-category view
gcx kg thresholds get [-o yaml]                          # dump whole config (YAML)
gcx kg thresholds set -f thresholds.yaml                 # replace whole config (destructive → confirm)
gcx kg thresholds upsert --record <name> --expr '<promql>' [--label k=v ...]
gcx kg thresholds delete --record <name> --expr '<promql>' [--label k=v ...]
```

## Structural facts driving the design
- **Three access patterns**: per-category structured reads (request/resource), whole-file YAML get/set, and single-threshold upsert/delete.
- **Single-threshold identity is the full `{record, expr, labels}` triple** — there is no delete-by-name; `delete` requires record + expr + labels (mirror the UI's `deleteThreshold`).
[...]

PR #1316, opened 2026-09-11
feat(kg): add thresholds read commands over v1 threshold config

kblaschke512:
## Summary

Adds read-only `gcx kg thresholds` over the Asserts **v1** threshold config API. Refs #1078 — writes (`set -f`, `upsert`, `delete`) follow in a second PR, so this one carries no destructive paths.

```bash
gcx kg thresholds get                                  # whole config; table by default
gcx kg thresholds get -o yaml                         # PrometheusRules resource envelope
gcx kg thresholds list --category request|resource    # custom/global split; table by default
```

## Why v1, not v2

The Asserts app UI and every threshold users have actually configured run on v1 — `ManageAssertions.service.ts` calls only `/v1/config/threshold-rule(s)`, across six call sites. `ThresholdsV2ConfigController` has zero call sites in the plugin UI; it appears only in the proxy authz allow-list (`pkg/plugin/authz/authz.go`) and its own tests. So a v2 client would diverge from what the product actually uses.

Re-checked at `asserts-app-plugin@dc0b960b` and `asserts-adi@4f4d8a8880` (2026-09-11): both controllers still exist, neither is `@Deprecated`, and the v1 endpoint map is unchanged.

The two read endpoints in this PR were verified live against a dev stack — response shapes match the DTOs exactly (`{active, expr, labels, record}`). The write endpoints are from source reading only; they land with the follow-up PR.

## Notes for review

- **`get` supports human and machine output.** Table is the default and reuses the existing prom-rules table codecs. JSON/YAML preserve the `PrometheusRulesDto` resource envelope through the existing `RuleToResource` path.
- **`list` tables follow ADR-002.** One column list defines narrow and wide output; `LABELS` is the only `WideOnly` column.
- **Machine formats preserve backend shapes.** A KG-local helper sends flattened rows to human-readable formats and the DTO/resource envelope to machine formats. It remains local because this PR owns KG, not Instrumentation or the shared `internal/output` contract. ADR-002 and those packages are unchanged.
- **Expressions remain complete.** Multi-line PromQL is folded onto one table row without truncation, while whitespace inside quoted literals is preserved. Special label values are quoted so they cannot break or ambiguate table cells.
- **Zero results serialize as `[]`, never `null`.**
- **`--category` is validated before any request** and path-escaped before being added to the URL.

## Testing

- `GCX_AGENT_MODE=false mise run all` — lint 0 issues, full suite green, docs/reference generation and build green
- Unit tests cover both endpoints, default and machine output, golden narrow/wide/empty tables, untruncated multi-line expressions, quoted literal whitespace, special labels, zero-result JSON, category validation, and URL path escaping
- Both commits carry verified SSH signatures

## Out of scope

- Writes (`set -f` whole-config replace, `upsert`, `delete`) — follow-up PR; `set` replaces everything, so it gets a confirmation prompt plus `--yes`
- v2 API — unadopted by the UI
- Shared output abstraction or Instrumentation changes
- The remaining hand-rolled KG codecs — those belong to the #1275 sweep


dafydd-t:
@claude review

internal/providers/kg/thresholds.go
```diff
@@ -0,0 +1,290 @@
...
+	cmd.AddCommand(getCmd, listCmd)
+	return cmd
+}
+
+type thresholdsGetOpts struct {
+	IO cmdio.Options
+}
+
+func (o *thresholdsGetOpts) setup(flags *pflag.FlagSet) {
+	o.IO.RegisterCustomCodec(cmdio.FormatTable, &RuleTableCodec{})
```
radiohead, line 117: **required** — current `main` removed `RuleTableCodec` and `RuleWideTableCodec` in favor of the shared `RuleTable()` declaration. A virtual merge with current `main` fails to compile at these two registrations, even though the stale PR-head CI and live PR build pass. Please restack and replace these registrations with `cmdio.RegisterTable(&o.IO, RuleTable())`; is there a current-base build that would show these deleted symbols remain available?

---
**Generated by Codex**

internal/providers/kg/thresholds.go
```diff
@@ -0,0 +1,290 @@
...
+	var b strings.Builder
+	b.Grow(len(expr))
+
+	var quote rune
+	escaped := false
+	pendingSpace := false
+	for _, r := range expr {
+		if quote != 0 {
+			switch r {
+			case '\n':
```
radiohead, line 233: **required** — this changes the meaning of valid PromQL. Inside a backtick literal, replacing a newline with the characters `\n` changes the literal because PromQL does not process escapes there; outside literals, folding the newline after `#` extends the line comment across the rest of the expression. Please make compaction PromQL-aware and add cases for multiline backtick strings and comments, or point to a backend validation guarantee that excludes both constructs.

---
**Generated by Codex**

radiohead (changes requested):
**Intent.** Adds read-only `gcx kg thresholds get` and `list` commands over the Asserts v1 threshold API, with human tables and structured output for agents.

**Blocking.** The PR no longer compiles when merged with current `main`, and the table formatter can change the meaning of valid PromQL containing comments or multiline backtick strings.

**Validation.** The PR-head binary passed live read-only testing against `dev`: both endpoints, output formats, field selection, agent output, and validation errors behaved correctly. The live data did not contain PromQL comments or multiline backtick strings, so that formatter boundary was verified from the implementation and PromQL syntax contract rather than the smoke test.

**Fix summary.**

1. Restack onto current `main` and register `RuleTable()` through `cmdio.RegisterTable`.
2. Make expression compaction preserve PromQL comment and raw-string semantics, with regression tests for both cases.

**Verdict.** Request changes for the current-base compile failure and lossy PromQL rendering.

Comments below are prefixed **required** (2). **required** means fix before merge.

---
**Generated by Codex**

PR #1321, opened 2026-09-12
feat(agento11y): pull experiment artifact content

portertech:
## Summary

- add explicit `--include-artifacts` support to AgentO11y experiment pulls
- download trial artifact content through the plugin proxy into private,
  digest-addressed files
- inventory artifact metadata, payload paths, sizes, and SHA-256 digests
- preserve atomic no-replace publication and partial-failure receipts
- bound artifact, JSON response, and paginated trial data sizes
- verify no-replace behavior during publication preflight

## Validation

- `go test ./...`
- `go test ./internal/providers/agento11y/eval/experiments`
- `mise run reference:cli`
- `git diff --check`
- live dev pull of one experiment with conversation and artifact inclusion at
  concurrency 1; all 9 manifest entries verified by size and SHA-256, with
  private file permissions

## Security

Artifact downloads are opt-in. Response bodies are bounded, payload filenames
are derived from artifact-ID SHA-256 values, files are mode `0600`, bundle
directories are mode `0700`, and any fetch failure produces an incomplete
manifest rather than silently omitting content.

### DIFF
--- a/docs/reference/cli/gcx_agento11y_experiments_pull.md
+++ b/docs/reference/cli/gcx_agento11y_experiments_pull.md
@@ -9,16 +9,18 @@ responses may change without following the normal semantic versioning convention
 
 Pull the experiment record, aggregate report, and paginated trial responses to
 a new directory. The report includes the evaluator scores and artifact metadata
-returned for each trial. The trial index contains referenced conversation IDs,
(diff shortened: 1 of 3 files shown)


portertech:
@claude review

dafydd-t (approved):

PR #1329, opened 2026-09-15
feat(appo11y): gate services commands on App Observability activation

cedricziel:
## What

Adds a reusable `IsActivated` pre-flight (`internal/providers/appo11y/activation`) that checks the plugin-proxy settings endpoint for a 404, mirroring the "not installed or not enabled" detection already used by `appo11y/settings` and `appo11y/overrides`. Wires it into all five `gcx appo11y services` subcommands so they fail fast with a clear error instead of running PromQL/Tempo queries against a stack where App Observability isn't enabled.

## Why

Today `gcx appo11y services *` runs its queries regardless of whether the App Observability plugin is actually enabled for the stack, which produces confusing "no data" output instead of a clear error. This also lays the foundation for gating the Knowledge Graph work later in this stack, which needs the same activation-check shape.

## Stack

| # | PR | Status |
|---|---|---|
| 1 | #1329 gate services on App O11y activation ← **this PR** | 👀 |
| 2 | #1330 service metadata fields + Explore links | 👀 |
| 3 | #1331 extract internal/query/kg | 👀 |
| 4 | #1332 refactor kg provider onto internal/query/kg | 👀 |
| 5 | #1333 fold Knowledge Graph consumption into services | 👀 |
| 6 | #1334 add operations list/get | 👀 |

> Diff this PR against its base (`main`), not the full stack.


jereenv (approved):

jereenv:
LGTM

PR #1330, opened 2026-09-15
feat(appo11y): first-class service metadata fields + Explore links

cedricziel:
## What

Adds `Environment`, `Cluster`, `Version`, and `Kind` as named fields on `Service` (previously only reachable, if at all, via the untyped `Labels` map), and an Explore deep-link field on `services get`'s RED snapshot, reusing the existing `prometheus.QueryExploreURL` builder. All additive via `omitempty` so JSON on unaffected paths is unchanged.

`Version` deliberately stays empty when a service's matched series disagree on `service_version`, rather than picking one and silently misreporting a mid-rollout service.

## Why

The Grafana Assistant's app-o11y tool surfaces this metadata (language, service type, environment, cluster, version) plus deep links as part of its unified `app_observability_handler`. gcx's `services list/get` were missing everything but `language` as a first-class field. This closes that gap on the structured (JSON/agent-mode) output — no new prose formatting, since agent-mode consumers already get to choose their own presentation.

## Stack

| # | PR | Status |
|---|---|---|
| 1 | #1329 gate services on App O11y activation | 👀 |
| 2 | #1330 service metadata fields + Explore links ← **this PR** | 👀 |
| 3 | #1331 extract internal/query/kg | 👀 |
| 4 | #1332 refactor kg provider onto internal/query/kg | 👀 |
| 5 | #1333 fold Knowledge Graph consumption into services | 👀 |
| 6 | #1334 add operations list/get | 👀 |

> Diff this PR against its base (#1329), not the full stack.


internal/providers/appo11y/services/query.go
```diff
@@ -323,11 +336,16 @@ func parseServicesResponse(resp *prometheus.QueryResponse) ([]Service, error) {
...
 			byKey[k] = svc
 		}
 		for lk, lv := range sample.Metric {
-			if lk == "job" || lk == "telemetry_sdk_language" || lk == "__name__" || lv == "" {
+			// service_version is deliberately excluded here: it feeds the
+			// ambiguity-tracked Version field below via versionsSeen, and a
+			// flat first-wins copy into Labels would leak an arbitrary
+			// single value for a service mid-rollout across two versions —
+			// exactly the misreport Version's "" case exists to avoid.
+			if lk == "job" || lk == "telemetry_sdk_language" || lk == "__name__" || lk == "service_version" || lv == "" {
```
jereenv, line 348: By dropping `service_version` from labels, we are causing `--columns service_version` to return empty as it read directly from Labels https://github.com/grafana/gcx/blob/aef2b5c24a0f1a13ce8696f9091b3202a97762dc/internal/providers/appo11y/services/commands.go#L451-L453

Since` svc.Version` is already populated with the version, could we just repopulate Labels["service_version"] from svc.Version once that's set?

jereenv (dismissed):

jereenv (approved):

internal/datasources/query/explore.go
```diff
@@ -43,6 +46,27 @@ func ExploreRange(from, to string, instant bool) (string, string) {
...
+func DatemathFrom(window string) string {
+	d, err := model.ParseDuration(window)
+	if err != nil {
+		return "now-" + window
+	}
+	seconds := int64(time.Duration(d).Round(time.Second) / time.Second)
+	if seconds <= 0 {
+		seconds = 1
+	}
+	return fmt.Sprintf("now-%ds", seconds)
```
radiohead: Functionally this is OK, but it does look weird in the UI. I wonder if we could optimise this a bit to use seconds / minutes / hours / days where appropriate:

Image

radiohead (commented):
LGTM

radiohead (approved):

PR #1339, opened 2026-09-16
feat(skills): add lifecycle catalog and local reconciliation

javiermolinar:
## What

Add a release lifecycle catalog at `claude-plugin/skills-catalog.yaml` and shared read-only reconciliation against the selected local `.agents` installation.

## Why

When removing a skill from GCX we were leaving orphan skills from the user installations. The catalog can keep current and deleted skills


javiermolinar:
@claude review

wardbekker (approved):

PR #1356, opened 2026-09-18
fix(datasources): retain plugin types in list responses

dafydd-t:
Grafana's datasource list responses can omit both `spec.type` and `apiVersion` from individual items. Preserve the plugin type from the queried endpoint when neither field supplies it, fixing empty types in `gcx datasources list` and invalid `.datasource.grafana.app` API groups in `gcx resources get datasources`.

Extend the multi-plugin transport test to cover omitted item metadata and assert the resulting manifest API groups. Live verification changed 11/11 missing types to 0/11, with valid API groups for all 11 resources.

Validation: `GCX_AGENT_MODE=false mise run all` passed with a fresh lint cache, along with targeted datasource and resource package tests. Generated reference docs have no drift.


awsome-o (approved):
lgtm

